PR #18, opened 2016-09-27
updated required pip version and added setuptools as an install_requirements entry

mvfc:
https://github.com/pybee/briefcase/issues/8

pip requirement updated to pip 8.1 and setuptools >= 27.0 added as an install_requires entry, as said in the issue


freakboy3742:
Thanks for the contribution!

PR #28, opened 2017-03-12
Initial cut for django briefcase target

glasnt:
To be used with https://github.com/pybee/Python-Django-template. The how-to is in the 3.5 branch of the template

It's probably not 100%, there's still some Django-specifics that could be nicer, but it's a start. 

Feel free to djangoify it, @freakboy3742

### DIFF
--- a/briefcase/app.py
+++ b/briefcase/app.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 
 import os
 import json
+import random
 import re
 import sys
 import uuid
@@ -40,6 +41,8 @@ class app(Command):
          "Name of the icon file."),
         ('guid=', None,
          "GUID identifying the app."),
+        ('secret-key=', None,
+         "Secret key for the app."),
         ('splash=', None,
          "Name of the splash screen file."),
         ('app-requires', None,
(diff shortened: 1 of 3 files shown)


freakboy3742:
I've had a bit of a swing at this - it's cleaned up a few things. There's still a problem generating a `package.json` file that has content that doesn't raise warnings, but it's getting there.

briefcase/django.py
```diff
@@ -43,13 +43,7 @@ def install_support_package(self):
     def install_extras(self):
         # Install additional elements required for Django
         print(" * Installing extras...")
-        print(" * Installing NPM requirements...")
-        package_url = "https://raw.githubusercontent.com/pybee/toga/master/src/django/package.json"
```
glasnt: I wasn't sure if we should be using the toga-django install requirements, or a package.json like the source I was using from `togatest`. But it looks like you've added the package.json into the template, which works 👍

briefcase/app.py
```diff
@@ -120,28 +134,41 @@ def find_support_pkg(self):
...
+        return os.path.join(os.getcwd(), self.resource_dir, 'app_packages')
+
     def generate_app_template(self):
         print(" * Writing application template...")
 
         if self.template is None:
             self.template = 'https://github.com/pybee/Python-%s-template.git' % self.platform
         print("Project template: %s" % self.template)
 
+        # import pdb; pdb.set_trace()
```
glasnt: stray debugging

glasnt:
🛰

PR #33, opened 2017-04-18
Update tutorials to include briefcase-template

glasnt:
Still rough, but it should be right. Needs someone to run them in an isolated test to confirm

@freakboy3742


glasnt:
I'd appreciate some pointers here

glasnt:
Defaults updated

PR #40, opened 2017-05-23
Add `--build` and `--execute` options to the setup script

nikmolnar:
This PR adds a `--build` option, which will automatically build the Xcode project, and an `--execute` option, which will launch the app in the iOS simulator. Additional `--os` and `--device` options allow the user to build/run the app on different OS/device combinations.

Currently, these options are ignored for all targets other than iOS. ~~I plan to add an implementation for macOS next.~~ (I see this isn't necessary for macOS)

### DIFF
--- a/briefcase/app.py
+++ b/briefcase/app.py
@@ -52,6 +52,8 @@ class app(Command):
          'URL for the support package to use'),
         ('download-dir=', None,
          "Directory where the project support packages will be cached"),
+        ('build', None,
+         "Build the project after generating")
     ]
 
     def initialize_options(self):
@@ -70,6 +72,7 @@ class app(Command):
         self.version_code = None
         self.guid = None
         self.secret_key = None
+        self.build = False
 
     def finalize_options(self):
         if self.formal_name is None:
@@ -257,6 +260,9 @@ class app(Command):
     def install_extras(self):
         pass
 
+    def build_app(self):
+        pass
+
     def post_run(self):
         print()
         print("Installation complete.")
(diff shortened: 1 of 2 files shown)


phildini:
This is flipping amazing. I have a couple comments incoming.

briefcase/app.py
```diff
@@ -52,6 +52,14 @@ class app(Command):
          'URL for the support package to use'),
         ('download-dir=', None,
          "Directory where the project support packages will be cached"),
+        ('build', None,
+         "Build the project after generating"),
+        ('execute', None,
+         "Run the application after building"),
+        ('os-version=', None,
+         "For iOS, the OS version to run (e.g., iOS 10.2)"),
```
phildini: Let's not make the docstring iOS-specific. What if we wanted to target a specific android version some day?

Maybe make the docstring something like "Sets the OS version. Currently only iOS supported."

briefcase/app.py
```diff
@@ -52,6 +52,14 @@ class app(Command):
...
+        ('device=', None,
+         "For iOS, the device to run (e.g., iPhone 7 Plus)")
```
phildini: See above, re: device. Let's generalize, and leave ourselves open for the future.

briefcase/app.py
```diff
@@ -70,6 +78,10 @@ def initialize_options(self):
         self.version_code = None
         self.guid = None
         self.secret_key = None
+        self.build = False
+        self.execute = False
+        self.os_version = 'iOS 10.2'
```
phildini: Related to the above, let's set these props to `None` or something in the general `app.py`, and set the specific values in ios.py

briefcase/app.py
nikmolnar: Agreed.

briefcase/app.py
nikmolnar: This is an excellent point; will do.

PR #45, opened 2017-05-25
Fixes #47: Detection of NodeJS and Install process

### LINKED ISSUE #47: Django: Executing migrations for the users
It may be useful to execute the migrations for the user once the setup has been successful. There might be some inconveniences so I will like to discuss the pros and cons of this option.

elyak123:
When you run `python setup.py django` it will detect the version of Node is it exists, check if its major version is 6.x, raise a Runtime Error if its below.

If no version of NodeJS is found, proceeds to install it, as well all the npm dependencies.

Hope it helps, open to feedback

### DIFF
--- a/briefcase/django.py
+++ b/briefcase/django.py
@@ -1,6 +1,10 @@
 import os
+import stat
 import shutil
-import subprocess
+import subprocess, shlex
+import sys
+import requests
+import tempfile
 
 try:
     from urllib.request import urlopen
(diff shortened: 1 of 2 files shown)


briefcase/django.py
```diff
@@ -78,10 +82,48 @@ def install_extras(self):
...
+        system_platform = sys.platform
+        node = shutil.which('node')
+        npm  = shutil.which("npm")
+        try:
+            node_version_unicode = subprocess.check_output([node, '--version'])
+            node_version = node_version_unicode.decode('utf-8')
+            print('Node version %s detected' % node_version)
+        except:
+            node_version = ''
+            print('No version of Node detected\n Installing NodeJS... \nPlease Follow the instructions.')
```
phildini: This is good, maybe should include a link to the docs?

briefcase/django.py
```diff
@@ -78,10 +82,48 @@ def install_extras(self):
...
+        if node and node_version[1] == '6':
+            subprocess.Popen(['npm', 'install'], cwd=os.path.abspath(self.dir)).wait()
+        elif node and node_version[1] != '6':
+            err_message = ('ERROR: Cannot run with the current version of Node and Npm, please \n'
+                           'unnistall the current version and install Node version 6.x\n'
+                           'Installation cancelled.'
+            )
+            raise RuntimeError(err_message)
+        else:
+            ses = requests.Session()
```
phildini: I'm not a fan of trying to download node for the user... We should link to docs, but the user should install their own packages, I think.

Part of this is I have a lot of FUD around making sure this works on every platform, and being able to test it.

elyak123: I can take out the installation process, but, should I raise an error of no version of node is found?? or just an warning, print a link that may correspond to their platform and continue with the process??

phildini: I think raising an error or warning saying they should install Node and try again (with a link, definitely) is the right approach here.

elyak123: Great!! working on it!!

elyak123:
Made some changes, hope it helps.

Also, I heard we might want to `python manage.py migrate` for the user so, I will like to confirm is that is something we actually want to do so I can get into it.

briefcase/django.py
```diff
@@ -78,13 +82,45 @@ def install_extras(self):
...
+            if not node or node_version[1] != '6':
+                err_message = ('ERROR: Cannot run with the current version of Node and Npm, please \n'
+                               'unnistall the current version and install Node version 6.x\n'
+                               'Installation cancelled.'
+                )
+            raise RuntimeError(err_message)
         print("   - Building Webpack assets...")
         subprocess.Popen([npm, "run", "build"], cwd=os.path.abspath(self.dir)).wait()
 
+    def django_migrate(self):
```
phildini: There's an argument that we should migrate for the user automatically, can you open an issue on briefcase to discuss it, and remove this code? Thanks!

briefcase/django.py
```diff
@@ -78,13 +82,45 @@ def install_extras(self):
         # Install additional elements required for Django
         print(" * Installing extras...")
         print("   - Installing NPM requirements...")
-
-        npm = shutil.which("npm")
-        subprocess.Popen([npm, "install"], cwd=os.path.abspath(self.dir)).wait()
-
+        node_attributes = {
```
phildini: Since we're no longer trying to install node and npm, I think this code is no longer necessary?

phildini:
Overall I think this is getting really close, a couple of deletions and we'll be there.

briefcase/django.py
elyak123: well, I use it to get the download links, but I can delete the `'command_prefix'` key from the dictionary

briefcase/django.py
elyak123: Done and done!! Thanks!!

briefcase/django.py
```diff
@@ -78,10 +82,35 @@ def install_extras(self):
...
-        npm = shutil.which("npm")
-        subprocess.Popen([npm, "install"], cwd=os.path.abspath(self.dir)).wait()
-
+        node_download_link = {
+            'darwin': 'https://nodejs.org/dist/v6.10.3/node-v6.10.3.pkg',
+            'linux' : '',
+            'win32' :'https://nodejs.org/dist/v6.10.3/node-v6.10.3-x86.msi',
+        }
+        system_platform = sys.platform
+        required_node_version = '6'
```
freakboy3742: There should also be a required_npm_version; the default version installed by node 6 is npm 3, which has problems.

briefcase/django.py
```diff
@@ -78,10 +82,35 @@ def install_extras(self):
...
+            subprocess.Popen(['npm', 'install'], cwd=os.path.abspath(self.dir)).wait()
+        else:
+            if node_version[1] != required_node_version:
+                err_message = ('ERROR: Cannot run with the current version of Node and Npm, please \n'
+                               'unnistall the current version and install Node version 6.x\n'
+                               'Installation cancelled.'
+                )
+            else:
+                err_message=('Could not finish installation because NodeJs is not installed\n'
+                    'Please install NodeJs at:\n %s' % node_download_link[system_platform])
```
freakboy3742: This message won't be useful if you're on Linux. Might be better to just send the user to the npm install page, rather than linking to a specific installer.

briefcase/django.py
elyak123: If I understand, npm 3 has problems? if so, which version of npm should be required???

elyak123: Any thoughts regarding this??

PR #49, opened 2017-07-04
Update tutorial-0.rst

Ry10p:
I'm new at contributing.
UPDATED FOR LINUX UBUNTU 16.04
If I can do anything better please let me know I just want to be helpful!

Cheers,
-Ryan

### DIFF
--- a/docs/tutorials/tutorial-0.rst
+++ b/docs/tutorials/tutorial-0.rst
@@ -10,9 +10,9 @@ Setup
 This tutorial assumes you've read and followed the instructions in
 :doc:`/intro/getting-started`. If you've done this, you should have:
 
-* XCode installed on your Mac,
+* XCode installed on your Mac, (if linux no Xcode required)
 * A ``tutorial`` directory,
-* A activated Python 3.4 virtual environment,
+* A activated Python 3.4 virtual environment, (linux used 3.5)
 * Briefcase installed in that virtual environment,
 
 Start a new project
@@ -45,6 +45,8 @@ Check out what the provided ``tutorial_0/app.py`` file contains:
 
 This won't do much as it is, but we can make it useful. 
 
+IF LINUX import __main__ script
+
 Add this into the ``app.py`` to make it useful:
 
 .. code-block:: python
@@ -64,7 +66,7 @@ It is all ready for using ``briefcase``. You can invoke it, using:
 
 .. code-block:: bash
 
-    $ python setup.py ios
+    $ python setup.py ios OR linux
 
 to create the iOS app.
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/tutorial-0.rst
```diff
@@ -10,9 +10,9 @@ Setup
 This tutorial assumes you've read and followed the instructions in
 :doc:`/intro/getting-started`. If you've done this, you should have:
 
-* XCode installed on your Mac,
+* XCode installed on your Mac, (if linux no Xcode required)
 * A ``tutorial`` directory,
-* A activated Python 3.4 virtual environment,
+* A activated Python 3.4 virtual environment, (linux used 3.5)
```
freakboy3742: 3.4 or 3.5 should generally work - it's not a mac/linux specific feature.

docs/tutorials/tutorial-0.rst
```diff
@@ -45,6 +45,8 @@ Check out what the provided ``tutorial_0/app.py`` file contains:
 
 This won't do much as it is, but we can make it useful. 
 
+IF LINUX import __main__ script
```
freakboy3742: It's not clear what you're trying to say here - there shouldn't be any "linux specific" startup instructions...

docs/tutorials/tutorial-0.rst
```diff
@@ -64,7 +66,7 @@ It is all ready for using ``briefcase``. You can invoke it, using:
 
 .. code-block:: bash
 
-    $ python setup.py ios
+    $ python setup.py ios OR linux
```
freakboy3742: This is potentially confusing - do I type `python setup.py ios OR linux`, or do I type `python setup.py ios` or `python setup.py linux`? It would be better to break this out into 2 separate commands, describing which one you run depending on which platform you're on.

docs/tutorials/tutorial-0.rst
```diff
@@ -75,7 +77,7 @@ There is a new folder in your project called 'iOS', which contains the Xcode pro
 
 .. code-block:: bash
 
-    open iOS/Tutorial\ 0.xcodeproj
+    open iOS/Tutorial\ 0.xcodeproj OR ./linux/tutorial_0
 
 You can test the app by running it in Xcode. As our application only shows a message, the iOS application will show only a blank screen. You can check if it is working in the console logs, which should contain something like this:
```
freakboy3742: These instructions will need to be updated as well - no point talking about XCode if you're on linux.

In fact, it might be better to write a completely separate "getting started on Linux" tutorial, and give the user the option of which tutorial they run based on the platform they're using.

docs/tutorials/tutorial-0.rst
Ry10p: My intent was to tell the user to import the main script into the app.py
or for example in my code at the top of the file I have:

import __main__

docs/tutorials/tutorial-0.rst
Ry10p: :+1: I see hat you mean; command used was:
python setup.py linux

docs/tutorials/tutorial-0.rst
Ry10p: I will rewrite a getting started on Linux tutorial.

THANK YOU SO MUCH!!!

Ry10p:
Created a Pull request for the added tutorial

PR #69, opened 2017-09-08
Windows: use numeric version numbers.

andrewleech:
Windows installer fails when non-integer elements are present in the version string.

I ran into this trying to use briefcase with a package using setuptools_scm which gave me a version number like "1.2.dev7+g3126144.d20170908".

This isn't a particularly elegant solution, but it's a suggestion of how I've worked around the issue.

### DIFF
--- a/briefcase/app.py
+++ b/briefcase/app.py
@@ -57,7 +57,9 @@ class app(Command):
         ('os-version=', None,
          "Set the device OS version. (e.g., iOS 10.2)"),
         ('device-name=', None,
-         "Set the device to run. (e.g., iPhone 7 Plus)")
+         "Set the device to run. (e.g., iPhone 7 Plus)"),
+        ('sanitize-version', None,
+         "Forces installer version to only contain numbers.")
     ]
 
     def initialize_options(self):
@@ -80,6 +82,7 @@ class app(Command):
         self.start = False
         self.os_version = None
         self.device_name = None
+        self.sanitize_version = None
 
     def finalize_options(self):
         if self.formal_name is None:
@@ -167,6 +170,13 @@ class app(Command):
 
     def generate_app_template(self, extra_context=None):
         print(" * Writing application template...")
+
+        if self.sanitize_version and self.version_numeric != self.version:
+            print(" ! Version currently contains characters: %s" % self.version)
+            print(" ! Installer version sanitized to: %s" % self.version_numeric)
+
+            extra_context = extra_context or {}
+            extra_context['version'] = self.version_numeric
 
         if self.template is None:
             template_path = os.path.expanduser('~/.cookiecutters/Python-%s-template' % self.platform)
--- a/briefcase/app.py
+++ b/briefcase/app.py
@@ -111,10 +111,7 @@ class app(Command):
             int(match.groups()[4]) if match.groups()[4] else 0,
         )
         self.version_code = '%02d%02d%02d' % self._numeric_version_parts
-        self.numeric_version = '%d.%d.%d' % self._numeric_version_parts
-        if self.version != self.distribution.get_version():
-            print("Installer needs version in format x.x.x where x is integer")
-            print("Using version: %s" % self.version)
+        self.version_numeric = '%d.%d.%d' % self._numeric_version_parts
 
         # The app's GUID (if not manually specified) is a namespace UUID
         # based on the URL for the app.
@@ -168,7 +165,7 @@ class app(Command):
     def version(self):
         return self.distribution.get_version()
 
-    def generate_app_template(self):
+    def generate_app_template(self, extra_context=None):
         print(" * Writing application template...")
 
         if self.template is None:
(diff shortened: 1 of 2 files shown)


briefcase/app.py
```diff
@@ -159,10 +164,6 @@ def app_dir(self):
     def app_packages_dir(self):
         return os.path.join(os.getcwd(), self.resource_dir, 'app_packages')
 
-    @property
-    def version(self):
```
phildini: this is definitely used by some of the templates, at the very least the Django one. I don't think this is safe to remove?

phildini:
Thanks for the PR! Could you talk a little bit more about the error you were having? I'm curious what Windows is doing, and I'm also thinking if there is a fix needed for Windows it should go in the Windows subclass. 

Overall, I'm leery about a change where a user could end up with a version different from the version specified.

andrewleech:
Yeah to be honest I would reject this PR as is too. I've been a bit unwell and should have lodged this as a well documented issue rather than just throwing this up here but wasn't thinking very clearly. 

I'll follow up soon with more appropriate details.

briefcase/app.py
andrewleech: It was replaced with the self.version variable set above which would/should behave the same, other than technically being writable and not necessarily created early enough.

andrewleech:
As mentioned earlier I use `setuptools_scm` for versioning. When I'm on a semantic versioned tag, the build will work fine. In between tags/releases however the version is more descriptive, including git hash, date and/or distance from tag. I would still like to be able to build and potentially share installers of these in-between builds (beta release, etc)

On windows, the installer / application version cannot have anything other than numbers and periods, eg. 1.2.3.0

With strings in the version, I get a failure on windows during installer build.
`python.exe setup.py windows --build`
```
<snip>
Installation complete.

 * Looking for WiX Toolset...
   - Using C:\Program Files (x86)\WiX Toolset v3.11\
 * Compiling application installer...
Windows Installer XML Toolset Compiler version 3.11.0.1701
Copyright (c) .NET Foundation and contributors. All rights reserved.

briefcase.wxs
C:\Users\corona\src\feralsync\windows\briefcase.wxs(5) : error CNDL0108 : The Product/@Version attribute's value, '1.2.dev7+g3126144.d20170910', is not a valid version.  Legal version values should look like 'x.x.x.x' where x is an integer from 0 to 65534.

C:\Users\corona\src\feralsync\windows\briefcase.wxs(5) : error CNDL0010 : The Product/@Version attribute was not found; it is required.
 * Linking application installer...
light.exe : error LGHT0103 : The system cannot find the file 'briefcase.wixobj' with type 'Source'.

Build complete.
``` 
On a related note, it reports `Build complete.` but there is no installer actually built. 
It appears the failures isn't being detected / reported.

The head of `briefcase.wxs` contains:
```
<?xml version="1.0"?>
<?define ProductVersion = "1.2.dev7+g3126144.d20170910" ?>
<?define ProductUpgradeCode = "2b314b12-07d5-362b-aef7-7d6f119aa067" ?>
<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
   <Product
            Id="*"
            UpgradeCode="$(var.ProductUpgradeCode)"
            Name="feralsync"
            Version="$(var.ProductVersion)"
```

This is a limitation of only certain properties in windows apps so filtering / changing of the version string should only be done in the bare minimum of places with appropriate warnings that this has been done.

phildini:
These changes look good, but there's something I'm still not clear about.

I think I can sum this up as: Will the Windows installer _never_ take non-number versions, or is it `setuptools_scm` that means it can't take non-number versions?

I want to be a little careful about adding complexity for a package unless we're 100% certain we want to support the package.

If it's truly _all_ Windows installs that won't take non-number versions, then this is totally ready to accept (although that's weird). If it's specific to `setuptools_scm`, we'll have to think about this more.

andrewleech:
As far as I can find it's a fixed restriction on native windows apps using the standard mechanism of an application manifest, including windows installers.  

http://oncoding.blogspot.com.au/2012/04/net-assembly-version-numbers.html

There are a few different version numbers defined for windows apps. 
The main fields (AssemblyVersion, FileVersion) are enforced to the number.number.number scheme. 
According to the above article there's another field that is free text in most circumstances, "ProductVersion", however this is the name of the field name in WIX project file that is already used and throws the failure. 

Either way it definitely appears the main version number for the installer must be numerical. Perhaps there might be another property field that the real version could be stored in for purely information use?

Other References:
https://docs.microsoft.com/en-us/dotnet/framework/app-domains/assembly-versioning
https://msdn.microsoft.com/en-us/library/windows/desktop/aa370859(v=vs.85).aspx
https://stackoverflow.com/a/9313229

phildini:
Ok, so I think what the best thing to do here is complain loudly when using non-numeric version numbers and _hard fail_. I'm leery about changing the versions people are using out from under them, because that could introduce all kinds of subtle bugs that are going to be hard to detect if you miss the warning message.

I know that make some of your work unnecessary, sorry about that, but I'm trying to think through a concept of "no surprises" for the user.

andrewleech:
I definitely agree with an ethos of zero surprises.
I also want a way to be able to still build an installer on every CI run, regardless of whether it's on a tag.

How about hard fail on non-numeric versions, but add a command line option to `--sanitise-version` which will then run my code from here?

This way, with my patch, the application and the installer filename still has the full version number in it as defined in the project, it's just in the installer properties that has the sanitized version. 

On that note, do you prefer sanitise or sanitize? As an Australian I prefer the former, but I know not everyone does ;-)

phildini:
I'm pretty ok with failing by default but adding a flag like `--fix-version-for-windows` (no spelling issues in that case 😉 )

emansije:
How about a slightly shorter flag like `--fix-windows-version`? :smiley:

phildini:
I think this is fine for now, and we can munge the flag later if we feel the need. 

Thanks for your patience, and for walking this through to completion!

PR #70, opened 2017-09-10
Detect failures in build_app()

andrewleech:
Currently, if anything fails in the various build_app() subprocesses it goes undetected, leaving briefcase trying to continue regardless. It eventually reports "Build complete." regardless of whether it has succeeded or not.
This PR monitors the returncode from the subprocess calls and reports failures if detected.

Now, post_build() will only be run on success of build_app()

### DIFF
--- a/briefcase/android.py
+++ b/briefcase/android.py
@@ -107,11 +107,13 @@ class android(app):
     def build_app(self):
         if not self.start:
             print(" * Building %s..." % (self.formal_name))
-            subprocess.Popen([
+            proc = subprocess.Popen([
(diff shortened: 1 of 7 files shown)


andrewleech:
On a related note, is there an intention to fully support python 3.4?
briefcase setup.py does list 3.4 as supported however there is no briefcase-template for 3.4 and attempting to build fails completely on 3.4

If 3.4 support is not needed, I'd prefer to use the newer, cleaner `subprocess.run()` command rather than `Popen` and `wait`

PR #75, opened 2017-10-01
Allow use of Setuptools launchers

andrewleech:
Currently the application startup script is hard coded in the per-platform template.  
This is quite restrictive in that it only works if you have a module named 'app' with a function 'main' which returns an object with a function `main_loop` (on windows at least, other platforms can be different).

Also, on windows this means that in task manager you see `pythonw.exe` rather than the name of the distributed application.

To make briefcase more flexible I propose using the built in mechanism for defining entry points / applications. 
On windows this already creates exe launchers that have the desired application names. 
This also allows having multiple gui and console applications in the one installer package.

In this PR the windows launcher creation has been modified slightly to create relocatable exe applications in the contents/app folder.
Start menu shortcuts will be created for each of these rather than the static launcher currently defined in the template.
These executables will show  in task manager with the user specified names.

On linux, executable launch scripts will be created for each entrypoint defined.

On MacOS, if an entrypoint is defined with the same name as the `formal_name` specified it will be the main application. There are launch scripts for any others also defined in the bundle/Contents/MacOS the an end user could run from the command line etc. 

Mobile platforms have not been adapted to this mechanism as there's no way to produce multiple executables. 
If desired however they could also be updated in the same way to allow defining the custom entry point which would be used in preference to the templated method.

If no entry_points are defined it falls back to the launching mechanism as defined in the platform templates.

### DIFF
--- a/briefcase/app.py
+++ b/briefcase/app.py
@@ -8,12 +8,16 @@ import re
 import subprocess
 import shutil
 import sys
+import textwrap
 import uuid
 
 from datetime import date
 from distutils.core import Command
 
 import pip
+import pkg_resources
+from setuptools.command import easy_install
+from setuptools.dist import Distribution
 
 from botocore.handlers import disable_signing
 import boto3
@@ -239,7 +243,7 @@ class app(Command):
             print("No requirements.")
 
     def install_platform_requirements(self):
-        print(" * Installing plaform requirements...")
+        print(" * Installing platform requirements...")
         if self.app_requires:
             pip.main([
                     'install',
@@ -261,6 +265,51 @@ class app(Command):
                 '--target=%s' % self.app_dir,
                 '.'
             ])
+
+    @property
+    def launcher_header(self):
+        """
+        Override the shebang line for launcher scripts
+        This should return a suitable relative path which will find the
+        bundled python for the relevant platform
+        """
+        return None
+
+    def install_launch_scripts(self):
+        print(" * Creating launchers...")
+        pip.main([
+                     'install',
+                     '--upgrade',
+                     '--force-reinstall',
+                     '--target=%s' % self.app_packages_dir,
+                     'setuptools'
+                 ])
+
+        easy_install.ScriptWriter.template = textwrap.dedent("""
+            # EASY-INSTALL-ENTRY-SCRIPT: %(spec)r,%(group)r,%(name)r
+            __requires__ = %(spec)r
+            import re
+            import sys
+            from os.path import dirname, abspath, join
+            sys.path.insert(0, join(dirname(__file__), '..', 'app_packages'))
+            sys.path.insert(0, dirname(__file__))
+            from pkg_resources import load_entry_point
+    
+            if __name__ == '__main__':
+                sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
+                sys.exit(
+                    load_entry_point(%(spec)r, %(group)r, %(name)r)()
+                )
+        """).lstrip()
+
+        ei = easy_install.easy_install(self.distribution)
+        for dist in pkg_resources.find_distributions('.'):
+            ei.args = True  # To allow finalize_options to run
+            ei.finalize_options()
+            ei.script_dir = self.app_dir
+
+            for args in easy_install.ScriptWriter.best().get_args(dist, header=self.launcher_header):
+                ei.write_script(*args)
 
     def install_resources(self):
         if self.icon:
@@ -351,6 +400,7 @@ class app(Command):
         self.install_app_requirements()
         self.install_platform_requirements()
         self.install_code()
+        self.install_launch_scripts()
         self.install_resources()
         self.install_extras()
         self.post_install()
--- a/briefcase/app.py
+++ b/briefcase/app.py
@@ -269,9 +269,10 @@ class app(Command):
     @property
     def launcher_header(self):
         """
-        Override the shebang line for launcher scripts
+        Optionally override the shebang line for launcher scripts
         This should return a suitable relative path which will find the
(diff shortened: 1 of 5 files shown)


andrewleech:
Hi, could I get some feedback on this feature? 

I feel it should be ready to go and won't impact existing users who don't change their setup.py to use the feature.

docs/background/quickstart.rst
```diff
@@ -75,3 +75,29 @@ Then, you can invoke ``briefcase``, using:
...
+                'utility = example.main:main',
+            ]
+        }
+        ...
+
+For more details on the format see:
+http://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins
+
+On windows and linux this allows for multiple executables to be defined.
+MacOS will use the entry point with the same name as your `formal_name` as the
```
freakboy3742: Minor capitalisation issues here - Windows, Linux and (ironically), macOS (even though it's starting the sentence).

briefcase/macos.py
```diff
@@ -35,6 +36,30 @@ def install_icon(self):
...
+        pyexe = '../Resources/python/bin/python%s' % (3 if sys.version_info.major == 3 else '')
+        return '#!/bin/sh\n'\
+               '"exec" "`dirname $0`/%s" "$0" "$@"\n' % pyexe
+
+    @property
+    def launcher_script_location(self):
+        macos_dir = os.path.abspath(os.path.join(self.resource_dir, '..', 'MacOS'))
+        return macos_dir
+
+    # def install_launch_scripts(self):
```
freakboy3742: Is there any reason to keep this code around?

docs/background/quickstart.rst
```diff
@@ -75,3 +75,29 @@ Then, you can invoke ``briefcase``, using:
 You can also use the ``-b`` (or ``--build``) argument to automatically
 perform any compilation step required; or use ``-s`` (``--start``) to
 start the application.
+
+For desktop os's (macOS, Windows, Linux) the entry point(s) to your program can
```
freakboy3742: Minor capitaliation fix: OS.

briefcase/app.py
```diff
@@ -8,12 +8,16 @@
...
 import sys
+import textwrap
 import uuid
 
 from datetime import date
 from distutils.core import Command
 
 import pip
+import pkg_resources
+from setuptools.command import easy_install
```
freakboy3742: /me has a Pavlovian shudder at the utterance of these two magic words... :-)

briefcase/app.py
```diff
@@ -239,7 +243,7 @@ def install_app_requirements(self):
             print("No requirements.")
 
     def install_platform_requirements(self):
-        print(" * Installing plaform requirements...")
+        print(" * Installing platform requirements...")
```
freakboy3742: Oops :-)

briefcase/app.py
```diff
@@ -262,6 +266,75 @@ def install_code(self):
...
             ])
 
+    @property
+    def launcher_header(self):
+        """
+        Override the shebang line for launcher scripts
+        This should return a suitable relative path which will find the
+        bundled python for the relevant platform
+        """
+        return None
```
freakboy3742: Should this be a `return None`, or a `NotImplementedException`?

briefcase/app.py
```diff
@@ -262,6 +266,75 @@ def install_code(self):
...
+
+    @property
+    def launcher_script_location(self):
+        return self.app_dir
+
+    def install_launch_scripts(self):
+        exe_names = []
+        if self.distribution.entry_points:
```
freakboy3742: What's the behavior here if the user has an app with multiple entry points, which *will* make sense for desktop deploys, but *won't* make sense for mobile? 

Would it make sense to roll out a full mobile template for each app? (and, for that matter - for macOS too, where the alternate entry points will be buried)

briefcase/app.py
andrewleech: I know, but this is the version that setuptools itself ships to use, so I felt ok about the functional reuse rather than reimplement :-)

briefcase/app.py
andrewleech: Actually no, `None` means the setuptools default will be used, which may be suitable depending on the platform.
I'll update the comment to make this more clean.

briefcase/macos.py
andrewleech: Nope, that was test code I overlooked before committing.

briefcase/app.py
andrewleech: I've gone back and forth on this one myself. It's both a pro and con of the packaged app formats that you can't share any bulk between them.

Initially I baulked at the idea of producing isolated copies of the entire app just for the sake of a modified startup script, however now I think it's probably the way to go. 
It ultimately gives the end user the option of which of the produced apps they want to install, and not worry about any library dependencies, for the sake of a few more MB of disk space. 

This would be a significant extra piece of work that I'd prefer to do in a follow up PR if that's ok. It'll mean a bit of restructuring of the main workflow in app, checking for entry points early on and if so probably looping over the rest of the workflow with some optiomisations to minimise duplication of common calculations like building the main apps bdist.

This will also be needed on linux if it moves to snappy/flatpack.

andrewleech: As far as current behaviour is concerned.... on mobile releases a defined entry point with the same name as the package formal_name will override the hardcoded entrypoint in the template, otherwise it will essentially be ignored and the template method will be used.

freakboy3742: Sounds good - tackling this as a follow up PR works for me.

briefcase/app.py
freakboy3742: True. But old wounds... :-)

PR #84, opened 2017-10-09
Allow building android output on windows

andrewleech:
As well as (existing) linux/mac/*nix support.
Simply checks the current runtime platform and runs gradlew.bat on windows, gradlew (sh script) on others.


briefcase/android.py
```diff
@@ -107,11 +108,18 @@ def post_install(self):
...
+                proc = subprocess.Popen([
+                    'cmd.exe', '/c', 'gradlew.bat', 'build'
                 ],
-                cwd=os.path.abspath(self.dir)
-            )
+                    cwd=os.path.abspath(self.dir)
+                )
+            else:
+                proc = subprocess.Popen([
+                        '/usr/bin/env', 'sh', './gradlew', 'build'
```
eliasdorneles: Is the `/usr/bin/env sh` bit really needed?

eliasdorneles:
@phildini I believe those instructions could be updated instructing people to run `python setup.py android --start` and `python setup.py android --build` instead of `cd android; ./gradlew run`

eliasdorneles:
Sorry for getting in the way -- looks like we were looking into the PR at the same time :)

briefcase/android.py
andrewleech: I felt it was safer than assuming the shell script had executable permissions, and also matched the windows cmdline better by declaring the shell it's running in.

andrewleech:
Yep I agree the instructions could definitely do with a cleanup, I'll do that when I get a chance at home.

PR #87, opened 2017-10-11
Windows: Limit directory id length in wix project

andrewleech:
On Windows, WIX has a max directory id length of 72

As the id is generated from path names this can easily exceed this, at which point a warning is shown during build:
```
C:\build\PC-App-2\windows\briefcase.wxs(10552) : warning CNDL1026 : The Directory/@Id attribute's 
value, 'DIR_app_packages.pip._vendor.requests.packages.urllib3.contrib.__pycache__', 
is too long for an identifier.  Standard identifiers are 72 characters long or less.
```

This limits the id (generated from file path) to this restriction

### DIFF
--- a/briefcase/windows.py
+++ b/briefcase/windows.py
@@ -70,6 +70,7 @@ class windows(app):
         content = []
         contentrefs = []
         shortcuts = []
+        dir_ids = []
 
         def walk_dir(path, depth=0):
             files = []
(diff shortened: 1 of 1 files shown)


phildini:
The code here looks fine, but could you talk a bit about what this results in? It seems strange to me that shortening the path is the right thing to do here.

andrewleech:
The id's aren't particularly important, they just have to be unique. As far as I know they're just used internally during installation and have no impact on the installed files. 
The file id's by comparison are simply uuids and have no relation to the file path.
I considered changing dir id's to uuids also however I suspect there may be times when some relationship to the path might be useful during debugging installers.

phildini:
Hey @andrewleech, sorry this has taken so long to get back to. If you can resolve the conflict, I'm happy to merge!

andrewleech:
@phildini It's all rebased now, should be good to go thanks

phildini:
@andrewleech it occurs to me after the fact that ideally these magic numbers would be encoded as constants at the top of the file. I'm going to merge this as-is for now, but a follow-up PR making these better constants would be 💯

andrewleech:
@phildini valid point. 
At least the 72 should be labelled as a constant, the rest probably should be used as offsets  from this.

PR #89, opened 2017-10-13
Add developer mode.

ramiroluz:


ramiroluz:
@eliasdorneles @freakboy3742 can you help me?

The smoke-test is broken, using python3.5 on my machine it runs well.

ramiroluz:
@berinhard can you review my tests? Can we pair someday to refactor the code design?

phildini:
I think the code looks ok overall, can you talk a bit about how you see this being used? 

ie, why would you use this method instead of `pip install`ing something that was dev-specific?

eliasdorneles:
Hey @phildini , the context of this is in issue https://github.com/pybee/briefcase/issues/73: the idea is to have a "developer mode" where we can override app dependencies with a development version (useful if you're working in a toga backend, for example).

ramiroluz:
@phildini as it is implemented by now it is possible to create a developer requirement file similar to:

toga-gtk /home/user/projects/python/toga-gtk

It may be necessary to use the full string dependency like:

toga-core>=0.2.15  /home/user/projects/python/toga-gtk

To use it:
python setup.py linux --dev=requirements.dev

eliasdorneles:
Hey @ramiroluz, thank you so much for this, I'm sorry for taking long to review this!

I've been trying this out and I'm having some trouble trying to use it, which isn't entirely due to the code in here.
I don't know exactly what the problem is just yet, but it seems to be related to toga packaging.
I want to sort these things out before merging this PR, please bare with me a bit longer. :blush: 

In any case, I'm going to document here the steps I'm doing to try this out:

1) Check out the code from this PR
2) Create a temporary virtualenv (using virtualenvwrapper I just hit `mktmpenv`), and install on it `cookiecutter` and `briefcase`
3) Go to /tmp and run `cookiecutter https://github.com/pybee/briefcase-template app_name=myapp formal_name=MyApp --no-input` -- this will create a simple toga-enabled app in a directory called `myapp`
4) Enter the app directory (`cd myapp`) and create a `requirements.dev` file looking like this:
```
toga /home/elias/src/toga
toga-core /home/elias/src/toga/src/core/
toga-gtk /home/elias/src/toga/src/gtk/
```
One can replace `toga-gtk` by their toga backend of choise.

**NOTE:** btw, it would be nice to be able to use `~/src` instead of `/home/elias/src` -- @ramiroluz you can use `os.path.expanduser` for that.

5) Then I run `python setup.py linux --dev requirements.dev`. After running this, the expected would be that it generates the `linux` directory the same as with vanilla briefcase, except it would replace the packages from PyPI for the ones I overrode in the `requirements.dev`.

So, it seems to be correctly *trying* to install the toga-gtk from the places I indicated (which is pointing to a checkout of toga which is in sync with master), but it fails (apparently because of some toga versioning issue) and proceeds even so.

Here is an example output, note how the "Install platform requirements step" doesn't finish cleanly, but briefcase moves forward anyway:

```
$ python setup.py linux --dev requirements.dev 
running linux
 * Writing application template...
Template Python-Linux-template succesfully updated.
Project template: /home/elias/.cookiecutters/Python-Linux-template
 * Installing requirements...
No requirements.
 * Installing plaform requirements...
Processing /home/elias/src/toga/src/gtk
Processing /home/elias/src/toga/src/core
Processing /home/elias/src/toga
Collecting toga-core>=0.3.0.dev1 (from toga-gtk==0.3.0.dev1)
  Could not find a version that satisfies the requirement toga-core>=0.3.0.dev1 (from toga-gtk==0.3.0.dev1) (from versions: 0.2.0, 0.2.1, 0.2.2, 0.2.3, 0.2.4, 0.2.5, 0.2.6, 0.2.7, 0.2.8, 0.2.9, 0.2.10, 0.2.11, 0.2.12, 0.2.13, 0.2.14, 0.2.15)
No matching distribution found for toga-core>=0.3.0.dev1 (from toga-gtk==0.3.0.dev1)
 * Installing project code...
Processing /tmp/myapp
Installing collected packages: myapp
  Running setup.py install for myapp ... done
Successfully installed myapp-0.0.1
 * No icons defined - using default...
 * No splash screen defined...

Installation complete.
```

The end result is that app_packages is empty:

```
$ ls linux/app_packages/
README
```

PS: there is probably more than one issue in here, I will check out this with more patience soon and create proper issues in other repos if confirmed.

ramiroluz:
I was afraid of something like that.

The way the code is, it search for the full name of the requirement, toga-gtk for example.

If there is a toga-gtk requirement in the list it will replace, but if it came from an external requirement file, it will not. Also, if the name on the list is not exactly the same, it will not replace.

This is one of the lists that are replaced:
https://github.com/pybee/briefcase/pull/89/files#diff-06a54fb54932e606a440a08fb6908619R161

Can you try to use a requirements.dev file like this?
```
toga /home/elias/src/toga
toga-core>=0.3.0.dev1 /home/elias/src/toga/src/core/
toga-gtk==0.3.0.dev1 /home/elias/src/toga/src/gtk/

```

Don't know if it will work.

hiteshramani:
Hi,

I was just going through the issue and the PR. I found some basic optimizations in the code, so just want to ask can I review the code?

eliasdorneles:
Hello @hiteshramani, thanks for offering, but before optimizing anything, it's more important to verify (and specify) how this feature is actually going to work.

If you're interested, the best thing you can do to help this to move forward is to 1) understand the problem it's trying to solve 2) test it locally. Then, maybe you can propose a better way of implementing the feature (it currently doesn't really work) -- ideally, by sending a follow-up PR.

freakboy3742:
A developer mode has been added in the v0.3 branch. Thanks for the contribution - sorry we weren't able to get this merged!

PR #93, opened 2017-12-19
Windows: download correct amd64/win32 support package

andrewleech:
Currently the support package on windows downloaded is always the amd64 python.

This patch auto detects whether briefcase is being run as 32 or 64 bit and uses the correct package to suit.

While I do use amd64 by default for my windows work 32 bit is still required at times to support Win7 32bit users.

### DIFF
--- a/briefcase/windows.py
+++ b/briefcase/windows.py
@@ -1,5 +1,7 @@
 import distutils.command.install_scripts as orig
 import os
+import struct
+
 from pkg_resources import Distribution, PathMetadata
 import re
 import shutil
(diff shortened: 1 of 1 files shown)


andrewleech:
In general it wouldn't be possible to create an installer for a different arch to the one running as pip will download/install wheels for the running arch as well.

This function already takes the running python version without an option to change so I think keeping arch consistent would be best. 

If needed, the whole support package URL can be overridden at runtime to pick a different version/arch by the end user._

freakboy3742:
Oh, I know about the pip/wheel architecture issue - that's why I raised this. It's currently an issue with building iOS/Android packages that have binary components.

However, the complexity of handling cross-targeted builds right probably means we need to tackle it with a more comprehensive approach; so lets just merge this and deal with the edge cases later.

PR #105, opened 2018-02-25
Add more information to the setup.py for briefcase

phildini:
Per https://packaging.python.org/tutorials/distributing-packages/#project-urls
we can have even more helpful information in our package and on our PyPI page by
adding urls to the setup.py


glasnt:
This looks great; this will take effect on the next package update. ping @freakboy3742

PR #118, opened 2018-04-15
merge commits from freakboy3742

jedie:
see also: https://github.com/pybee/briefcase/issues/117#issuecomment-381393379


jedie:
>  NameError: name 'errno' is not defined

fixed with: https://github.com/pybee/briefcase/pull/119

PR #119, opened 2018-04-15
merge commits from freakboy3742 + Bugfix NameError: name 'errno' is not defined

jedie:
see: https://github.com/pybee/briefcase/issues/117#issuecomment-381393379

and if fixed `NameError: name 'errno' is not defined`

See: https://travis-ci.org/jedie/pybee-docker/builds/366737315#L2250-L2252


freakboy3742:
Thanks for this; but it turns out there was even more changes needed. I was testing on a very old branch... turns out it's a bad idea to make releases early on a Sunday morning :-)

PR #132, opened 2018-05-26
Add option to create DMG file for MacOS

emmagordon:
As requested in https://github.com/pybee/briefcase/issues/130

### DIFF
--- a/briefcase/macos.py
+++ b/briefcase/macos.py
@@ -3,6 +3,8 @@ import sys
 import shutil
 import subprocess
 
+import dmgbuild
+
 from .app import app
 
 
@@ -12,7 +14,7 @@ class macos(app):
     def finalize_options(self):
         # Copy over all the options from the base 'app' command
         finalized = self.get_finalized_command('app')
-        for attr in ('formal_name', 'organization_name', 'bundle', 'icon', 'download_dir'):
+        for attr in ('formal_name', 'organization_name', 'bundle', 'icon', 'download_dir', 'dmg'):
             if getattr(self, attr) is None:
                 setattr(self, attr, getattr(finalized, attr))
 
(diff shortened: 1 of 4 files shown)


briefcase/macos.py
```diff
@@ -52,6 +54,12 @@ def launcher_script_location(self):
         return macos_dir
 
     def build_app(self):
+        if self.dmg:
+            print("Building DMG file...")
+            dmg_name = self.formal_name.replace(' ', '_') + '.dmg'
```
freakboy3742: I don't think this replace is necessary; macOS filenames don't have a problem with spaces.

briefcase/macos.py
```diff
@@ -52,6 +54,12 @@ def launcher_script_location(self):
...
+            dmg_path = os.path.join(os.path.abspath(self.dir), dmg_name)
+            dmgbuild.build_dmg(filename=dmg_path, volume_name=self.formal_name)
```
freakboy3742: This *creates* a dmg, but AFAICT, it doesn't put anything *into* the DMG. From a quick scan of the docs, we also need a `settings` kwarg, specifying at least a `files` key (pointing to the .app file). For bonus points, we could also include:
* `symlinks` (to provide the fast installation link to Applications)
* `icon` (to provide an icon for the volume)
* `background` (to provide a custom background image)
* `window_rect` (to set the initial size of the mounted window)

The DMG file should also be created in the macOS directory, so that all the macOS products end up in the same place.

docs/background/quickstart.rst
```diff
@@ -22,6 +22,7 @@ specific settings can be specified using a platform key::
                     'toga-cocoa'
                 ],
                 'icon': 'icons/macos',
+                'dmg': True,
```
freakboy3742: This won't be needed.

Tadaboody:
Is this still being worked on? should I fork from this and try to address the review or start from scratch?

freakboy3742:
@Tadaboody There hasn't been any activity on this for a while; so you're welcome to take this on. It's probably going to be easier to start from scratch on your own PR; but if you end up using any of @emmagordon's code, please give credit where it's due.

danyeaw:
Finished by #162

PR #134, opened 2018-06-04
Enforce python version 3.5 and above

danyeaw:
Signed-off-by: Dan Yeaw <dan@yeaw.me>


danyeaw:
Beekeeper already has python_requires, closing PR.

PR #141, opened 2018-06-13
Add Mu success story.

freakboy3742:


docs/background/success.rst
```diff
@@ -5,3 +5,4 @@ Want to see examples of Briefcase in use? Here's some:
 
 * `Travel Tips <https://itunes.apple.com/au/app/travel-tips/id1336372310>`_ is an app in the iOS App Store that was packaged for distribution using Briefcase.
 
+* `Mu <https://codewith.mu>` is a simple code editor for beginner programmers. It uses Briefcase to prepare installers for Windows and MacOS.
```
ntoll: We only use Briefcase for MacOS packaging. (We were not aware of Briefcase when we packaged for Windows... we already have an automated process that uses pynsist).

PR #148, opened 2018-07-09
Fix #146 find_support_pkg for > b9 builds

### LINKED ISSUE #146: Support package version comparison breaks with versions > b9
When determining the support package to use (in `briefcase/app.py`, function `find_support_pkg()`), the code looks to S3, finds all matching support packages, reverse sorts them, and then returns the first item.

This is an alphabetical sort, based on filename.

A support package has a name matching the pattern `Python-3.7-macOS-support.b1.tar.gz` (for Python 3.7, build 1)

Alphabetical sorts will work fine for builds 1-9... but build 10 will appear alphabetically between b1 and b2.

We need to split the filename to extract the build number, and then sort numerically.

rayrrr:
This addresses https://github.com/pybee/briefcase/issues/146 by using a combination of Python string methods.

### DIFF
--- a/briefcase/app.py
+++ b/briefcase/app.py
@@ -162,7 +162,7 @@ class app(Command):
                 if build_number > top_build_number:
                     top_build_number, top_build = build_number, item['Key']
         try:
-            return S3_URL + latest_build
+            return S3_URL + top_build
         except IndexError:
             return None
 
--- a/briefcase/app.py
+++ b/briefcase/app.py
@@ -164,11 +164,9 @@ class app(Command):
                     top_build_number = build_number
                     top_build = item['Key']
         if top_build:
-            try:
-                return S3_URL + top_build
-            except IndexError:
-                pass
-        return None
+            return S3_URL + top_build
+        else:
+            return None
 
     @property
     def app_dir(self):
--- a/briefcase/app.py
+++ b/briefcase/app.py
@@ -146,7 +146,8 @@ class app(Command):
         s3 = boto3.client('s3', region_name=S3_REGION)
         s3.meta.events.register('choose-signer.s3.*', disable_signing)
 
-        top_build_number, top_build = 0, None
+        top_build_number = 0
+        top_build = None
         paginator = s3.get_paginator('list_objects')
         for page in paginator.paginate(
                         Bucket=S3_BUCKET,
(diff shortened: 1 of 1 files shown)


briefcase/app.py
```diff
@@ -157,11 +157,16 @@ def find_support_pkg(self):
...
-        except IndexError:
-            return None
+                build_number = int(
+                    item['Key'].rstrip('.tar.gz').split('.')[-1].lstrip('b'))
+                if build_number > top_build_number:
+                    top_build_number, top_build = build_number, item['Key']
+        if top_build:
+            try:
+                return S3_URL + top_build
+            except IndexError:
```
freakboy3742: Under what conditions will this raise an IndexError?

briefcase/app.py
```diff
@@ -146,7 +146,7 @@ def find_support_pkg(self):
         s3 = boto3.client('s3', region_name=S3_REGION)
         s3.meta.events.register('choose-signer.s3.*', disable_signing)
 
-        candidates = []
+        top_build_number, top_build = 0, None
```
freakboy3742: Stylistically, there's no reason for this to be a multi-assignment; could you split this into 2 statements?

briefcase/app.py
```diff
@@ -157,11 +157,16 @@ def find_support_pkg(self):
...
             for item in page.get('Contents', []):
-                candidates.append(item['Key'])
-        try:
-            return S3_URL + sorted(candidates, reverse=True)[0]
-        except IndexError:
-            return None
+                build_number = int(
+                    item['Key'].rstrip('.tar.gz').split('.')[-1].lstrip('b'))
+                if build_number > top_build_number:
+                    top_build_number, top_build = build_number, item['Key']
```
freakboy3742: Similarly - no reason for this to be a multi-assign.

briefcase/app.py
rayrrr: This exception handling was already in there, returning `None` just as it still does now (eventually); I was just preserving the existing behavior. `IndexError` results when the requested S3 path does not exist; I suppose there could be an edge case that this catches where a build is released to S3 but then deleted in the midst of this running.

briefcase/app.py
rayrrr: Sure.

briefcase/app.py
rayrrr: Sure sure.

rayrrr:
@freakboy3742 thanks for the quick review! Logic question answered, style changes made. Looking forward to my first contribution being merged!

PR #149, opened 2018-07-10
Updated windows.py message when Wix is not found.

AltoRetrato:
Added recommendation to set environment variable when Wix Toolset is installed but not found.


danyeaw:
@AltoRetrato Thanks for the contribution. Sorry for the long delay in this getting reviewed. Instead of adding more details to the error message itself, what if we point users to the documentation?
https://briefcase.readthedocs.io/en/latest/background/getting-started.html#windows

The way I got it to work, I had to install WiX, and then install Visual Studio Build Tools to get msbuild to be available on the command line using x64 Native Tools Command Prompt for VS 2017. There might be an easier way, but we should figure out what the optimum install instructions are and add those to the docs, since I don't think "Install WiX" is sufficiently clear.

AltoRetrato:
@danyeaw I'm not sure that coupling error messages with URLs for online documentation is a good idea. URLs can change, documentation can change, and to keep things up-to-date you'd need to track those changes. Also, you put users through hoops to find a solution.

In this case, I changed the error message to be specifically what I needed to know to get through: a simple instruction that can be performed in almost no time at all (and that other would need to know as well).

I don't remember if I needed to do anything else to get it working, though.

briefcase/windows.py
```diff
@@ -171,11 +171,11 @@ def build_app(self):
         print(" * Looking for WiX Toolset...")
         wix_path = os.getenv('WIX')
         if not wix_path:
-            print("Couldn't find WiX Toolset. Please visit:")
+            print("Couldn't find WiX Toolset. Please install the latest stable release from:")
```
brutusthebee: At column 80: (E501) line too long (94 > 79 characters)

briefcase/windows.py
```diff
@@ -171,11 +171,11 @@ def build_app(self):
...
             print()
             print("    http://wixtoolset.org/releases/")
             print()
-            print("and install the latest stable release.")
+            print("If already installed, set the WIX environment variable to the installation path.")
```
brutusthebee: At column 80: (E501) line too long (101 > 79 characters)

danyeaw:
Thanks for the contribution, I think updating this error message will improve clarity ❤️

PR #150, opened 2018-07-10
Fixed constants, missing directory

AltoRetrato:
Removed repeated / redundant path creations and used previously created path "constants".

Directory "\windows\content\python" was not created, causing app execution to fail with "error: [WinError 267] The directory name is invalid".


src/briefcase/windows.py
```diff
@@ -209,10 +209,13 @@ def build_app(self):
     def start_app(self):
         print()
         print(" * Starting {}...".format(self.formal_name))
+        cwd = os.path.abspath(self.support_dir)
+        if not os.path.exists(cwd):
+           os.mkdir(cwd)
```
brutusthebee: At column 12: (E111) indentation is not a multiple of four

danyeaw:
@AltoRetrato Thanks for the contribution ❤️ . Sorry this took so long to be reviewed. This is some great refactoring to simplify and cleanup some of the path code 👍

PR #169, opened 2018-10-18
Added urllib3 version to install_requires

kayrahbear:
Assigned the package `urllib3`  to use version `1.22`. This should eliminate the VersionConflict errors being raised by `botocore` for the time being.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ setup(
         'requests < 3.0',
         'boto3 >= 1.4.4',
         # 10/18/18 - Pinning the urllib3 version to solve Boto dependency issues
-        'urllib3 = 1.22'
+        'urllib3 == 1.22'
     ],
     license='New BSD',
     classifiers=[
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -48,6 +48,8 @@
         'setuptools >= 40.0',
         'requests < 3.0',
         'boto3 >= 1.4.4',
+        # 10/18/18 - Pinning the urllib3 version is a temporary fix to solve conflicting dependency issues with Boto3 & Requests
```
brutusthebee: At column 120: (E501) line too long (128 > 119 characters)

setup.py
```diff
@@ -48,6 +48,8 @@
...
+        'urllib3 = 1.22'
```
brutusthebee: At column 25: (W291) trailing whitespace

setup.py
kayrahbear: 👍

setup.py
kayrahbear: 👍

PR #185, opened 2019-01-11
add pycairo to pip install

Aventine1:
I had the same issue as https://github.com/pybee/briefcase/issues/176
but for me
pip install pycairo
solved it.  Also I think this is a preferable solve in a virtual environment, instead of installing a linux package, which was the solve in the linked issue.

### DIFF
--- a/docs/tutorial/tutorial-0.rst
+++ b/docs/tutorial/tutorial-0.rst
@@ -24,7 +24,7 @@ Let's get started by using
 
 .. code-block:: bash
 
-    $ pip install cookiecutter briefcase
+    $ pip install pycairo cookiecutter briefcase
     $ cookiecutter https://github.com/pybee/briefcase-template
 
 This will ask a bunch of questions of you. We'll use an `app_name` of


danyeaw:
@Aventine1 Thanks for the problem report and suggesting a solution ♥️. Briefcase has Toga as a dependency, and toga-gtk depends on pycairo, so it *should* install automatically when you `pip install briefcase`. Your solution to manually `pip install pycairo` may have worked but it isn't the right solution. Do you have the log that shows why it failed?

Aventine1:
Although the log ends in "App started.", no window appears.
It does work if I ``pip install pycairo``

[log.txt](https://github.com/pybee/briefcase/files/2769976/log.txt)

danyeaw:
I have seen this error before as well, I think it is some minor problems with PyGObject / pycairo pip install support. Adding it manually as a dependency doesn't fix the issue, since if you look at the log, pycairo said it successfully pip installed prior to trying to install PyGObject.

PR #186, opened 2019-01-17
Remove redundant pip install briefcase

DannyCork:
Briefcase instructions already included on getting-started page. No need  to pip it again.


danyeaw:
Hi @DannyCork Thanks for your contribution :heart:.

PR #189, opened 2019-01-23
Windows shortcuts updates

andrewleech:
While adding briefcase support to a new project at work I found my `entry_points` functionality (#75) was no longer working, eventually tracking it down to #170

Looking at that, I realised there was a significant bug in my original implementation, hence the need for those lines to be removed.

This PR re-instates the functionality properly, only removing the hardcoded shortcut if there are `entry_points` defined in `setup.py`

Further to that, I then found that gui_scripts were still using python.exe rather than pythonw.exe
Turns out the function that builds the launcher script automatically fixes that in the shebang header only if it funds the described pythonw.exe on the path. This has been fixed in the second commit.

Lastly, on a separate project where we've heavily customised the wix script, I've found it really useful to have a second copy of the app shortcuts in the program folder, in addition the the one in start menu. Generally if users navigate to a program folder in explorer, it's nice if it's obvious how to run the app. Having a shortcut there provides this. The third commit is needed for this in conjunction with an update to the template: https://github.com/pybee/Python-Windows-template/pull/3

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 #/usr/bin/env python
 import io
 import re
+import platform
 
 from setuptools import setup, find_packages
 
@@ -55,7 +56,8 @@ setup(
         # this will break either requests or boto3 (or possibly both), at which point
         # the requirement can probably be dropped.
         'urllib3<1.24',
-        'dmgbuild >= 1.3.2',
+        *(['dmgbuild >= 1.3.2'] if platform.system() == 'Darwin' else
+          ['pywin32-ctypes'] if platform.system() == 'Windows' else []),
     ],
     license='New BSD',
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ setup(
         # the requirement can probably be dropped.
         'urllib3<1.24',
         *(['dmgbuild >= 1.3.2'] if platform.system() == 'Darwin' else
-          ['pywin32-ctypes'] if platform.system() == 'Windows' else []),
+          ['pywin32-ctypes', 'windows-entry-exe'] if platform.system() == 'Windows' else []),
     ],
     license='New BSD',
     classifiers=[
@@ -80,4 +80,8 @@ setup(
         'Tracker': 'https://github.com/pybee/briefcase/issues',
         'Source': 'https://github.com/pybee/briefcase',
     },
+    **(
+        dict(use_windows_entry_exe=True) if platform.system() == 'Windows' else {}
+    )
 )
+
--- a/src/briefcase/windows.py
+++ b/src/briefcase/windows.py
@@ -133,8 +133,8 @@ class windows(app):
                                 Name="{exe_name}"
                                 Icon="ProductIcon"
                                 Description="{description}"
-                                Target="[AppDir]\\app\\{exe_name}.exe"
-                                WorkingDirectory="[AppDir]\\app" />""".format(**locals()))
+                                Target="[WorkingDir]\\{exe_name}.exe"
+                                WorkingDirectory="WorkingDir" />""".format(**locals()))
 
         # Generate the full briefcase.wxs file
         briefcase_wxs = os.path.join(self.dir, 'briefcase.wxs')
--- a/src/briefcase/windows.py
+++ b/src/briefcase/windows.py
@@ -7,6 +7,7 @@ import sys
 import uuid
 
 from .app import app
+from .windows_icon import apply_icon
 
 
 class windows(app):
@@ -27,6 +28,7 @@ class windows(app):
         if self.dir is None:
             self.dir = 'windows'
 
+        self.iconfile = None
         self.resource_dir = os.path.join(self.dir, 'content')
         self.support_dir = os.path.join(self.resource_dir, 'python')
 
@@ -39,10 +41,8 @@ class windows(app):
         super(windows, self).generate_app_template(extra_context=extra_context)
 
     def install_icon(self):
-        shutil.copyfile(
-            '%s.ico' % self.icon,
-            os.path.join(self.app_dir, '%s.ico' % self.distribution.get_name())
-        )
+        self.iconfile = os.path.join(self.app_dir, '%s.ico' % self.distribution.get_name())
+        shutil.copyfile('%s.ico' % self.icon, self.iconfile)
 
     def install_splash(self):
         raise RuntimeError("Windows doesn't support splash screens.")
@@ -172,6 +172,16 @@ class windows(app):
         with open(briefcase_wxs, 'w') as template:
             for line in lines:
                 template.write('{}\n'.format(line))
+
+    def post_install(self):
+        if self.iconfile:
+            py_exes = [os.path.join(self.support_dir, f)
+                       for f in os.listdir(self.support_dir) if f.lower().endswith('.exe')]
+            py_exes.extend([os.path.join(self.app_dir, f)
+                            for f in os.listdir(self.app_dir) if f.lower().endswith('.exe')])
+
+            for exe in py_exes:
+                apply_icon(icon=self.iconfile, dest=exe)
 
     def build_app(self):
         print()
--- a/src/briefcase/windows.py
+++ b/src/briefcase/windows.py
@@ -153,7 +153,11 @@ class windows(app):
                     lines.extend(content)
                 elif line.strip() == '<!-- CONTENTREFS -->':
                     lines.extend(contentrefs)
-                elif line.strip() == '<!-- SHORTCUTS -->':
+                elif len(shortcuts) and line.strip() == '<!-- SHORTCUTS_PROVIDED -->':
+                    # Comment out existing shortcut in template if setup provides own
+                    lines.append('                        <!--')
+                elif len(shortcuts) and line.strip() == '<!-- SHORTCUTS -->':
+                    lines.append('                        -->')
                     lines.extend(shortcuts)
                 else:
                     lines.append(line.rstrip())
--- a/src/briefcase/windows.py
+++ b/src/briefcase/windows.py
@@ -71,6 +71,7 @@ class windows(app):
         content = []
         contentrefs = []
         shortcuts = []
+        shortcuts_instance = 0
         dir_ids = []
 
         def walk_dir(path, depth=0):
@@ -125,10 +126,10 @@ class windows(app):
                 for entry in entries:
                     exe_name = entry.split('=')[0].strip()
                     description = self.distribution.get_description()
-                    shortcutid = uuid.uuid4().hex
+                    # This is used multiple times below so shortcut_id will be filled later
                     shortcuts.append("""\
                             <Shortcut
-                                Id="AppShortcut_{shortcutid}"
+                                Id="Shortcut_{{shortcut_id}}"
                                 Name="{exe_name}"
                                 Icon="ProductIcon"
                                 Description="{description}"
@@ -160,7 +161,11 @@ class windows(app):
                     lines.append('                        <!--')
                 elif len(shortcuts) and line.strip() == '<!-- SHORTCUTS -->':
                     lines.append('                        -->')
-                    lines.extend(shortcuts)
+
+                    shortcuts_instance += 1
+                    for shortcut in shortcuts:
+                        shortcut_id = uuid.uuid4().hex
+                        lines.append(shortcut.format(**locals()))
                 else:
                     lines.append(line.rstrip())
 
--- a/src/briefcase/windows.py
+++ b/src/briefcase/windows.py
@@ -134,7 +134,7 @@ class windows(app):
                                 Icon="ProductIcon"
                                 Description="{description}"
                                 Target="[AppDir]\\app\\{exe_name}.exe"
-                                WorkingDirectory="AppDir\\app" />""".format(**locals()))
+                                WorkingDirectory="[AppDir]\\app" />""".format(**locals()))
 
         # Generate the full briefcase.wxs file
         briefcase_wxs = os.path.join(self.dir, 'briefcase.wxs')
--- a/src/briefcase/windows.py
+++ b/src/briefcase/windows.py
@@ -59,7 +59,9 @@ class windows(app):
         This should return a suitable relative path which will find the
         bundled python for the relevant platform
         """
-        return "#!.\\python\\python.exe\n"
+        # we need the specified python to be findable relative to a path entry
+        os.environ['PATH'] += os.pathsep + os.path.join(self.dir, 'content')
+        return "#!..\\python\\python.exe\n"
 
     def install_extras(self):
         print(" * Finalizing application installer script...")
@@ -131,7 +133,7 @@ class windows(app):
                                 Icon="ProductIcon"
                                 Description="{description}"
                                 Target="[AppDir]\\app\\{exe_name}.exe"
-                                WorkingDirectory="AppDir" />""".format(**locals()))
+                                WorkingDirectory="AppDir\\app" />""".format(**locals()))
 
         # Generate the full briefcase.wxs file
         briefcase_wxs = os.path.join(self.dir, 'briefcase.wxs')
--- a/src/briefcase/windows_icon.py
+++ b/src/briefcase/windows_icon.py
@@ -1,0 +1,96 @@
+import os
+from ctypes import Structure, sizeof, c_byte, c_ushort, c_int
+from win32ctypes.pywin32 import win32api
+
+
+# Structures in ico file
+class IconHeader(Structure):
+    _fields_ = [
+        ("Width", c_byte),
+        ("Height", c_byte),
+        ("Colors", c_byte),
+        ("Reserved", c_byte),
+        ("Planes", c_ushort),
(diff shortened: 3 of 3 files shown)


freakboy3742:
@andrewleech FYI - I needed to do a "minor emergency" release to get a project unstuck; I took a look at this patch to see if I could get it into the release, but I couldn't find an obvious fix for the issues @danyeaw raised in his review. 

However, since you've been doing such good work on the Windows installer, I'll commit to getting a release of this patch in the wild ASAP if you can resolve these issues.

src/briefcase/windows.py
```diff
@@ -123,10 +126,9 @@ def walk_dir(path, depth=0):
                 for entry in entries:
                     exe_name = entry.split('=')[0].strip()
                     description = self.distribution.get_description()
-                    shortcutid = uuid.uuid4().hex
                     shortcuts.append("""\
                             <Shortcut
-                                Id="AppShortcut_{shortcutid}"
+                                Id="Shortcut_%s"
```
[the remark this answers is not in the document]
andrewleech: It is mixing old and new, that's true. IIRC I did this so it could be parsed separately, the initial template fill handles the `.format()` args, but that shortcut name needs to be provided later once the id is generated.
Looking over the code again now, I suspect this is a holdover from an earlier implementation of it where the id was matching something that wasn't known until later.... currently it's a uuid3 which could in all likelihood be replaces by any old uuid up on this line instead

src/briefcase/windows.py
```diff
@@ -59,6 +59,8 @@ def launcher_header(self):
         This should return a suitable relative path which will find the
         bundled python for the relevant platform
         """
+        # we need the specified python to be findable relative to a path entry
+        os.environ['PATH'] += os.pathsep + os.path.join(self.dir, 'content')
```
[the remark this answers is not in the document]
andrewleech: It's only added to the path within the python executable for the duration of the build. It doesn't affect the rest of the pc or any other apps.

It's neccesary because setuptools/pip reads the launcher header when it builds the launcher exe's and tries to find the python.exe declared there. If it can't find it, the provided header is thrown out.

src/briefcase/windows.py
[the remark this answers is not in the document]
danyeaw: @andrewleech If this is a holdover from an earlier implementation, is this something we could fix to just directly populate a uuid3?

src/briefcase/windows.py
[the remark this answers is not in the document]
danyeaw: Thanks, that makes sense to me now.

src/briefcase/windows.py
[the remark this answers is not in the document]
andrewleech: I had a bit more of a look yesterday, my last commit on this branch was explicitly replacing a uuid setting at the top with what's the now, I can't remember exactly why it was needed though. I'll need to test it without that change again.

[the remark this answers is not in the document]
andrewleech: Ah yes, the AppShortcut template generated is later used twice for both the application folder and the start menu shortcuts, this results in a duplicated Shortcut Id which Wix fails at.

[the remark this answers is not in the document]
freakboy3742: The shortcuts are being handled by re-writing the Wix config file, looking for those magic comment markers - would it be possible to roll that into the template itself? If we pass in extra context to the template, that block could be handled as a conditional (and/or loop) in the template, rather than needing a rewrite

setup.py
```diff
@@ -1,6 +1,7 @@
 #/usr/bin/env python
```
brutusthebee: At column 1: (E265) block comment should start with '# '

src/briefcase/windows.py
```diff
@@ -153,15 +156,31 @@ def walk_dir(path, depth=0):
...
                 else:
                     lines.append(line.rstrip())
 
         with open(briefcase_wxs, 'w') as template:
             for line in lines:
                 template.write('{}\n'.format(line))
 
+    def post_install(self):
+        if self.iconfile:
+            py_exes = [os.path.join(self.support_dir, f) for f in os.listdir(self.support_dir) if f.lower().endswith('.exe')]
```
brutusthebee: At column 120: (E501) line too long (125 > 119 characters)

src/briefcase/windows.py
```diff
@@ -153,15 +156,31 @@ def walk_dir(path, depth=0):
...
+            py_exes.extend([os.path.join(self.app_dir, f) for f in os.listdir(self.app_dir) if f.lower().endswith('.exe')])
```
brutusthebee: At column 120: (E501) line too long (123 > 119 characters)

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
+from ctypes import *
+from win32ctypes.pywin32 import win32api
```
brutusthebee: At column 1: (F403) 'from ctypes import *' used; unable to detect undefined names

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+# Structures in ico file
+class IconHeader(Structure):
+    _fields_ = [
```
brutusthebee: At column 18: (F405) 'Structure' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("Width", c_byte),
+        ("Height", c_byte),
```
brutusthebee: At column 19: (F405) 'c_byte' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("Colors", c_byte),
```
brutusthebee: At column 20: (F405) 'c_byte' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("Reserved", c_byte),
```
brutusthebee: At column 20: (F405) 'c_byte' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("Planes", c_ushort),
```
brutusthebee: At column 22: (F405) 'c_byte' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("BitsPerPixel", c_ushort),
```
brutusthebee: At column 20: (F405) 'c_ushort' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("ImageSize", c_int),
```
brutusthebee: At column 26: (F405) 'c_ushort' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("ImageOffset", c_int),
```
brutusthebee: At column 23: (F405) 'c_int' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+    ]
```
brutusthebee: At column 25: (F405) 'c_int' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class GroupIcon(Structure):
+    _fields_ = [
```
brutusthebee: At column 17: (F405) 'Structure' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("Reserved", c_ushort),
+        ("ResourceType", c_ushort),
```
brutusthebee: At column 22: (F405) 'c_ushort' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("ImageCount", c_ushort),
```
brutusthebee: At column 26: (F405) 'c_ushort' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        # ("Enries", PIconDirResEntry),
```
brutusthebee: At column 24: (F405) 'c_ushort' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+    ]
+
+
+# Structure for updating resource
+class IconDirResEntry(Structure):
+    _fields_ = [
```
brutusthebee: At column 23: (F405) 'Structure' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("Width", c_byte),
+        ("Height", c_byte),
```
brutusthebee: At column 19: (F405) 'c_byte' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("Colors", c_byte),
```
brutusthebee: At column 20: (F405) 'c_byte' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("Reserved", c_byte),
```
brutusthebee: At column 20: (F405) 'c_byte' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("Planes", c_ushort),
```
brutusthebee: At column 22: (F405) 'c_byte' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("BitsPerPixel", c_ushort),
```
brutusthebee: At column 20: (F405) 'c_ushort' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("ImageSize", c_int),
```
brutusthebee: At column 26: (F405) 'c_ushort' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+        ("ResourceID", c_ushort),
```
brutusthebee: At column 23: (F405) 'c_int' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+    ]
```
brutusthebee: At column 24: (F405) 'c_ushort' may be undefined, or defined from star imports: ctypes

src/briefcase/windows_icon.py
```diff
@@ -0,0 +1,92 @@
...
+    :param str icon: .ico file path
+    :param str dest: path of the .exe to update.
+    """
+
+    print("Applying icons from %s to %s", (icon, dest))
+
+    with open(icon, 'rb') as iconfile:
+        # header = GroupIcon()
+        group_header = GroupIcon.from_buffer_copy(iconfile.read(sizeof(GroupIcon)))
+        if group_header.ResourceType != 1:
```
brutusthebee: At column 65: (F405) 'sizeof' may be undefined, or defined from star imports: ctypes

src/briefcase/windows.py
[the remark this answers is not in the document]
andrewleech: The problem with inclusion in the template is it needs to be duplicated for each entry-point defined in setup.py
All the other <File> and <Compoenent> sections are generated in a loop and injected into the markers in the same way, the shortcut block is generated as multiline for readability but it's really not that different.

Does cookiecutter support looping sections of a template? If so arguably all the content, contentrefs and shortcuts could be pushed to that perhaps?

setup.py
brutusthebee: At column 1: (E265) block comment should start with '# '

src/briefcase/windows.py
brutusthebee: At column 120: (E501) line too long (125 > 119 characters)

src/briefcase/windows.py
brutusthebee: At column 120: (E501) line too long (123 > 119 characters)

setup.py
brutusthebee: At column 1: (E265) block comment should start with '# '

andrewleech:
so @brutusthebee is very verbose.... could he be adjusted to only comment on things changed in the patchset? 
Gitlabs' code style checker is nice in this way, just reporting the difference in code smell, rather than all of it every push.

andrewleech:
Accidentally hit close button...

andrewleech:
FYI I found another issue for windows use, irrespective of whether entry_point launchers or the original shortcuts are in use: 
both of them are effectively just means of launching the embedded `python.exe` with the included start script.

This means that in the windows taskbar, if you right-click on the app, the application that is shown is python.exe, not your desired app name.  
Similarly, if you "Pin to taskbar" it's python being pinned, without any startup args. that means the pinned app never works.

The standard entry_point launchers don't work either because they run python.exe as a child process, which again is what's shown in the taskbar.

I've solved this issue by building new entry_point launcher exe's that use the python c api to directly load and run the script, avoid the child process issue. 
https://gitlab.com/alelec/windows-entry-exe

I could make my launchers get used by default here too if you're interested, though they havne't had a lot of testing yet. I might reach out to the setuptools team too, see if they're interested.

freakboy3742:
Definitely want to ensure that there's a clean user experience for starting an application - having programs appear as "python.exe" isn't a good look.

I will admit to not being a Windows user, so I'm not in the best position to judge the best solution here; but I thought from my original testing that the lnk file approach let you set and icon and control the text that appeared in menus, etc. Is that not correct?

I'm mildly hesitant to introduce an executable - or, more specifically, anything that would introduce the need for end-users to have a compiler configured to be able to create a Windows installer. I'm currently on the road for a conference, so I'm not in a position to test anything at the moment; would the approach you're proposing break that requirement?

Another (somewhat hacky) approach - would it be possible to simply rename the python.exe to "<yourapp>.exe"? If there are multiple entry points, creating multiple copies of python.exe which are all the same file, but with different execution points? Or use .cmd files to represent entry points?

andrewleech:
With the current shortcut system, the app that shows up does generally have the correct name and icon in the window, and when hovering on the app in the taskbar. It's when getting the right-click context menu from the taskbar it shows python.exe, as well as looking for the process in task manager.

Renaming python.exe would probably fix the visible name in context menu/task manager however it will not resolve the pinning to taskbar issue, as the app is dependent on the script arg being passed to the exe by the shortcut which is lost on pin. 
You can pin the shortcut itself to the taskbar/start menu, but then the app starts up as a new icon on the taskbar because it's seen as a different app.

I haven't tried cmd files, I'm not sure if they're handled particularly differently to bat files?

My module does not need compiling, it simply provides new pre-built exe's to replace the exe's currently bundled in setuptools. My ones don't rely on running a child process like the original ones, it dynamically loads the correct pointed-to pythons dll instead and runs the script directly without needing args passed to it.

setup.py
brutusthebee: At column 1: (E265) block comment should start with '# '

setup.py
```diff
@@ -78,4 +80,8 @@
         'Tracker': 'https://github.com/pybee/briefcase/issues',
         'Source': 'https://github.com/pybee/briefcase',
     },
+    **(
+        dict(use_windows_entry_exe=True) if platform.system() == 'Windows' else {}
+    )
 )
+
```
brutusthebee: At column 1: (W391) blank line at end of file

freakboy3742:
I've just landed the v0.3 branch, which is a complete ground-up rewrite of Briefcase. It takes a complete different approach to packaging metadata, so this PR is, unfortunately, no longer applicable. 

Some of the broader issues (in particular, executable naming) may still apply; however, the patch will need to be radically updated to be compatible with the new codebase.

PR #195, opened 2019-02-24
Support environment markers in setuptools.setup install_requires arg.

tmontes:
Hello PyBee,

I've been contributing to [Mu editor](https://github.com/mu-editor/mu) recently and, in particular, started exploring [this issue there](https://github.com/mu-editor/mu/issues/776), where we'll be looking at improving the way dependencies are managed/specified, among other development setup and packaging related issues.

Mu uses briefcase to build macOS packaged versions.

One idea that will be useful in Mu's packaging context is using [environment markers, as defined in PEP 508](https://www.python.org/dev/peps/pep-0508/#id23), to turn platform dependency variations into declarative expressions.

In my exploration I found out that neither the most recent briefcase release, 0.2.8, nor the current master branch, install Mu's dependencies that were "tagged" with such environment markers. For example, if `setup.py` includes something like:

```
setup(
    ...
    install_requires=[
        'pycodestyle==2.4.0',
        'black>=18.9b0;python_version > "3.5"',
    ],
    ...
)
```

Running `python setup.py macos` does not add `black` to the `app_packages` when packaging with a Python version > 3.5.

I explored this problem for a while, reading on `setuptools`, `pkg_resources`, `pip`, and more, trying to understand "what's going on here" and looking for an authoritative doc on "how such dependencies should be handled". I do confess that while I learned quite a bit, I did not find such an authoritative doc (I may have missed it, of course!).

Nevertheless, I did come up with what seems like an effective solution, using a single `pkg_resources` public API. Here's what I found out, and what this PR implements, in order to support such environment markers in a project's dependencies:

(in the context of the `app` / `distutil.core.Command` instances)

* `install_requires` entries with environment markers are not present in `self.distribution.install_requires`.
* Instead, they are:
  * Grouped per environment marker.
  * Added to the `self.distribution.extras_require` dict where:
    * Keys are the marker expression strings, starting with a literal `:`.
    * Values are the lists of requirements stripped of their environment markers.

With this in mind, the solution this PR implements builds the `requirement_list`:

* By copying `self.distribution.install_requires`.
* Extending it with the entries in `self.distribution.extras_require` such that the key is an environment marker (`.startswith(':')`) and that marker evaluates to `True` on the packaging platform: for this, I used `pkg_resources.evaluate_marker`.

A few parting thoughts:
* Did not find any tests in the source tree to run, but with this PR I can produce a working Mu application on macOS. Am I missing something? :)
* Merging this in will be very useful for the work we're doing with Mu (as well as, for any other projects with environment markers in their `setup.py` files!), but I'd love to hear other people's perspective on the solution I'm proposing. Maybe there are better ways?
* Thanks for BeeWare, briefcase, and taking the time to review this proposal.

:)

### DIFF
--- a/src/briefcase/app.py
+++ b/src/briefcase/app.py
@@ -298,6 +298,28 @@ class app(Command):
 
     def install_app_requirements(self):
         print(" * Installing requirements...")
+        # setuptools install_requires setup() argument list entries end up in
+        # two different self.distribution arguments:
+        # - Entries with no environment markers
+        #   Will be found in self.distribution.install_requires.
+        # - Entries with environment markers
+        #   Will be found under new keys in self.distribution.extras_require;
+        #   each such key is a string starting with ':' followed by the original
+        #   marker expression text itself (eg: ':python_version>"3.5"'); the
+        #   values will be lists of all the entries in the install_requires
+        #   setup() argument containing that particular marker.
+        # With this, the full requirements are:
+        # - The list in self.distribution.install_requires.
+        # - Extended with all True-evaluating marker entries in
+        #   self.distribution.extras_require.
+        requirement_list = list(self.distribution.install_requires)
+        for extra, extra_list in self.distribution.extras_require.items():
+            if not extra.startswith(":"):
+                # Discard non-environment marker entries.
+                continue
+            if pkg_resources.evaluate_marker(extra[1:]):
+                # Marker evaluates to True, bring it in.
+                requirement_list.extend(extra_list)
         if self.distribution.install_requires:
             subprocess.Popen(
                 [
@@ -305,7 +327,7 @@ class app(Command):
                     "--upgrade",
                     "--force-reinstall",
                     '--target={}'.format(self.app_packages_dir)
-                ] + self.distribution.install_requires,
+                ] + requirement_list,
             ).wait()
         else:
             print("No requirements.")


tmontes:
@freakboy3742, thanks for welcoming the change and for the quick response. You can certainly count on feedback and future contributions we find valuable not only for Mu, but for the general BeeWare audience.

Question: which kinds of plans would you have WRT cutting a release? At some point, I'll need to review Mu's packaging automation which I expect to depend on a briefcase release containing this change.

PS: Much like you, I also feel uncomfortable in changing code with no tests... But if Python packaging is non-trivial, testing it is even less obvious! :)

freakboy3742:
@tmontes I don't have any concrete plans to cut a release - I mostly make releases whenever I add a new feature or bugfix that has a significant impact. This change would seem to meet that criterion; so it's just a matter of me finding some spare time to cut a release.

What's your timeline for cutting the next Mu release?

tmontes:
@freakboy3742 thanks for your feedback

I'm not aware of a strict timeline for the next Mu release: the first alpha is in the works, now. I'd say a final release is still several weeks away. No hurries WRT to cutting an updated briefcase release, from my standpoint, thus! :)

I've been reviewing Mu's project dependency and packaging related aspects which led me to this PR. I think my work around it will be completed within a week or so. In that timeframe I may find the need for other improvements to briefcase (odds are low, I fathom) so I'd suggest let's hold on for a bit.

Once I'm ready to merge those improvements to Mu, I'll ping you to see when an updated briefcase release can be cut... Would that be ok? Like I said, no hurries there. :)

PS: In the meantime I guess I can pin Mu's briefcase dependency to a particular commit/branch here so, again, there is no pressure to cut a new briefcase release.

Thanks!

ntoll:
Hi Folks, Thank you for all the efforts with this.

I just ran into this issue when preparing for the 1.1-alpha.1 release. The long and short of it is we're blocked from releasing until this lands in briefcase. :-(

Is there **anything** I can do to help test/check/whatever so this fix can be released..?

Thank you!

ntoll:
OK... so the obvious solution is to pin briefcase to a commit/branch in GitHub -- but this feels "hacky" if you see what I mean... ;-)

ntoll:
I can confirm that Mu builds properly for OSX when we pip install from master rather than PyPI.

freakboy3742:
@ntoll Sorry this caused you problems. I'll find some time after work this afternoon to do some release testing and get this out there.

ntoll:
@freakboy3742 you're a :star: and IOU a beer or something at PyCon. Thank you!

freakboy3742:
@ntoll @tmontes FYI I've just released 0.2.9, which should include this patch. Let me know if you have any problems.

ntoll:
Stonking stuff... I've just kicked of a new build of Mu and will test / confirm once my meeting is over... ;-)

ntoll:
....and THANK YOU!!!! :cow: :heart: :bee: :beers:

ntoll:
Just to say, I've just tested the new build and it all works as expected. Thanks again! :-)

tmontes:
@freakboy3742, things are looking good. Thanks again. :)

PR #201, opened 2019-03-16
Fix all DeprecationWarning: invalid escape sequence

BoboTiG:
Hello,

This is a little patch to fix `DeprecationWarning`s about invalid escape sequence.

Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>

### DIFF
--- a/src/briefcase/app.py
+++ b/src/briefcase/app.py
@@ -139,7 +139,7 @@ class app(Command):
 
         # The Version Code is a pure-string, numerically sortable
         # version number.
-        match = re.match('(?P<major>\d+)(\.(?P<minor>\d+)(\.(?P<revision>\d+))?)?', self.distribution.get_version())
+        match = re.match(r'(?P<major>\d+)(\.(?P<minor>\d+)(\.(?P<revision>\d+))?)?', self.distribution.get_version())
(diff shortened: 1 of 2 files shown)


src/briefcase/app.py
```diff
@@ -406,7 +406,7 @@ def install_launch_scripts(self):
                 from pkg_resources import load_entry_point
 
                 if __name__ == '__main__':
-                    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
+                    sys.argv[0] = re.sub(r'(-script\\.pyw?|\\.exe)?$', '', sys.argv[0])
```
danyeaw: This string is already a "raw" string with the `r''` added. Do we need to double escape the two regex special characters for this line?

BoboTiG: Yes, because the entire string is inside another string itself. It starts at line 393, this is not really readable just within this snippet :)

danyeaw: @BoboTiG Thanks for the explanation, I missed that. I needed to look a bit further up and down :smile:

danyeaw:
Congrats on your first contribution :tada:

PR #209, opened 2019-04-13
Suggesting next possible step when the iOS setup/build exits

ahter:
This PR suggests a minor improvement to the ''No iOS device simulators found" exit message of the iOS build.

It proposes the next possible step for the build when simulator is not found (might simply be due to permission settings in macOS Mojave 10.14.4) so the first time users, especially when doing the tutorials get this error, they know how to proceed.

### DIFF
--- a/src/briefcase/ios.py
+++ b/src/briefcase/ios.py
@@ -132,7 +132,7 @@ class ios(app):
             if self.os_version is None:
                 os_list = [label for label in data['devices'] if label.startswith('iOS')]
                 if len(os_list) == 0:
-                    print('No iOS device simulators found', file=sys.stderr)
+                    print('No iOS device simulators found. Build your xcodeproj file manually using XCode.', file=sys.stderr)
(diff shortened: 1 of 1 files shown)


src/briefcase/ios.py
```diff
@@ -132,7 +132,7 @@ def set_device_target(self):
             if self.os_version is None:
                 os_list = [label for label in data['devices'] if label.startswith('iOS')]
                 if len(os_list) == 0:
-                    print('No iOS device simulators found', file=sys.stderr)
+                    print('No iOS device simulators found. Build your xcodeproj file manually using XCode.', file=sys.stderr)
```
brutusthebee: At column 120: (E501) line too long (125 > 119 characters)

freakboy3742:
The messaging around detection of iOS simulators has been updated in the v0.3 development version; if you think this problem still exists, please open a new PR.

PR #210, opened 2019-04-13
Fixing available iOS versions

ahter:
For some unknown reason, iOS versions (on macOS 10.14.4 and Xcode 10.2) do not all seem to start with iOS.

This fixes that issue.

### DIFF
--- a/src/briefcase/ios.py
+++ b/src/briefcase/ios.py
@@ -133,27 +133,29 @@ class ios(app):
             data = json.loads(output)
 
             if self.os_version is None:
-                os_list = []
+                os_dict = collections.OrderedDict()
                 for label in data['devices']:
                     if label.startswith('iOS'):
-                        os_list.append(label)
+                        os_dict[label] = label
                     elif label.startswith('com.apple.CoreSimulator.SimRuntime.iOS'):
-                        os_list.append(label[-8:].replace("-", ".").replace(".", " ", 1))
+                        os_dict[self._get_human_readable_label_name(label)] = label
 
-                if len(os_list) == 0:
+                os_list = list(os_dict.values())
+
+                if len(os_dict) == 0:
                     print('No iOS device simulators found', file=sys.stderr)
                     sys.exit(1)
-                elif len(os_list) == 1:
+                elif len(os_dict) == 1:
                     print('Building for {}...'.format(os_list[0]))
                     self.os_version = os_list[0]
                 else:
                     print()
                     while self.os_version is None:
                         print('Available iOS versions:')
-                        for i, label in enumerate(os_list):
+                        for i, label in enumerate(os_dict):
                             print('  [{}] {}'.format(i+1, label))
                         try:
-                            index = int(input('Which iOS version do you want to target: '))
+                            index = input('Which iOS version do you want to target: ')
                             self.os_version = os_list[int(index) - 1]
                         except:
                             print("Invalid selection.")
--- a/src/briefcase/ios.py
+++ b/src/briefcase/ios.py
@@ -119,6 +119,9 @@ class ios(app):
         # Complete bypass launch scripts for iOS
         print(" * Skipping creation of launch scripts.")
 
+    def _get_human_readable_label_name(self, label):
+        return label[-8:].replace("-", ".").replace(".", " ", 1)
+
     def set_device_target(self):
         if self.os_version is None or self.device_name is None or self.device is None:
             # Find an appropriate device
--- a/src/briefcase/ios.py
+++ b/src/briefcase/ios.py
@@ -129,8 +129,14 @@ class ios(app):
 
             data = json.loads(output)
 
+            def clean_label_output(label):
+                if label.startswith('com.apple.CoreSimulator.SimRuntime.iOS'):
+                    return label[-8:-2]
+                else:
+                    return label
+
             if self.os_version is None:
-                os_list = [label for label in data['devices'] if label.startswith('iOS') or label.startswith('com.apple.CoreSimulator.SimRuntime.iOS')]
+                os_list = [clean_label_output(label) for label in data['devices'] if label.startswith('iOS') or label.startswith('com.apple.CoreSimulator.SimRuntime.iOS')]
                 if len(os_list) == 0:
                     print('No iOS device simulators found', file=sys.stderr)
                     sys.exit(1)
--- a/src/briefcase/ios.py
+++ b/src/briefcase/ios.py
@@ -130,7 +130,7 @@ class ios(app):
             data = json.loads(output)
 
             if self.os_version is None:
-                os_list = [label for label in data['devices'] if label.startswith('iOS')]
+                os_list = [label for label in data['devices'] if label.startswith('iOS') or label.startswith('com.apple.CoreSimulator.SimRuntime.iOS')]
                 if len(os_list) == 0:
                     print('No iOS device simulators found', file=sys.stderr)
                     sys.exit(1)
--- a/src/briefcase/ios.py
+++ b/src/briefcase/ios.py
@@ -136,7 +136,11 @@ class ios(app):
                     return label
 
             if self.os_version is None:
-                os_list = [clean_label_output(label) for label in data['devices'] if label.startswith('iOS') or label.startswith('com.apple.CoreSimulator.SimRuntime.iOS')]
+                os_list = []
+                for label in data['devices']:
+                    if label.startswith('iOS') or label.startswith('com.apple.CoreSimulator.SimRuntime.iOS'):
+                        os_list.append(clean_label_output(label))
+
                 if len(os_list) == 0:
                     print('No iOS device simulators found', file=sys.stderr)
                     sys.exit(1)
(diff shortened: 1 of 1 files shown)


ahter:
That's a very good idea. I will be doing the post-processing.

src/briefcase/ios.py
```diff
@@ -130,7 +130,7 @@ def set_device_target(self):
             data = json.loads(output)
 
             if self.os_version is None:
-                os_list = [label for label in data['devices'] if label.startswith('iOS')]
+                os_list = [label for label in data['devices'] if label.startswith('iOS') or label.startswith('com.apple.CoreSimulator.SimRuntime.iOS')]
```
brutusthebee: At column 120: (E501) line too long (151 > 119 characters)

ahter: Yep, good idea. I will respect E501 too.

ahter: To respect E501, I decomposed the one-liner. I hope that's OK.

src/briefcase/ios.py
```diff
@@ -129,8 +129,18 @@ def set_device_target(self):
...
+                    return label[-8:-2]
+                else:
+                    return label
+
             if self.os_version is None:
-                os_list = [label for label in data['devices'] if label.startswith('iOS')]
+                os_list = []
+                for label in data['devices']:
+                    if label.startswith('iOS') or label.startswith('com.apple.CoreSimulator.SimRuntime.iOS'):
+                        os_list.append(clean_label_output(label))
```
freakboy3742: Rather than try to catch both 'startswith' cases in a single if clause, and then have an inner if - why not handle this with a single if/elif pair?
```
if startswith iOS
    append label
elif startswith com.apple...
    append trimmed label
```
That also removes the need for an inline function.

src/briefcase/ios.py
```diff
@@ -130,21 +134,29 @@ def set_device_target(self):
...
                     while self.os_version is None:
                         print('Available iOS versions:')
-                        for i, label in enumerate(os_list):
+                        for i, label in enumerate(os_dict):
                             print('  [{}] {}'.format(i+1, label))
                         try:
-                            index = int(input('Which iOS version do you want to target: '))
+                            index = input('Which iOS version do you want to target: ')
                             self.os_version = os_list[int(index) - 1]
                         except:
```
brutusthebee: At column 25: (E722) do not use bare 'except'

src/briefcase/ios.py
ahter: @magic-lantern @freakboy3742 
Apologies with late reply. Seems I have some Python and 'zsh' related issues at my home computer. I realized this when I've just tried doing some of the patching now. 

So, I wrote the patch but couldn't do the testing. Is it possible to get some help testing this out with some different configuration (macOS versions etc)?

magic-lantern: I'm running macOS 10.14.4 with Xcode 10.2.1. The proposed changes to ios.py do resolve the problem of "No iOS device simulators found" and correctly start an iOS simulator with the selected iOS version and hardware.

I am getting some new/additional errors though. My errors I'm receiving appear to be the same as mentioned in #245, so I believe they are separate from this pull request.

ahter:
@magic-lantern 
Great to know it works for you.

magic-lantern:
@ahter and others: So while this patch does fix the problem of "No iOS device simulators found" there are some other related problems not fixed. I've been investigating and haven't completed my analysis, but I know for example that the line:

```
'-sdk', 'iphonesimulator%s' % (self.os_version.split(' ')[-1],),
```

no longer works as well due to the change in the format of the output from `xcrun simctl list -j`

Would is the preference for getting the iOS portion of briefcase completely working? For example, would it be preferred to have this file completely working as part of this patch, have multiple patches, or something else?

Also, seeing how there are many parts of this file that are broken due to Xcode, it seems better unit testing is needed to help with troubleshooting.

freakboy3742:
@magic-lantern I don't have any firm opinions on how to tackle the problem. Smaller, self-contained patches are easier to review; but they're also harder to test if they only solve "part" of the problem at a time. Ultimately, I'm willing to be flexible towards whatever approach gets the work done. For me, results matter more than any hard process requirements. The only caveat I have is that whatever approach we take, each 'merged to master' state can't be any *less* stable than the one before it. 

I also don't know if @ahter is motivated to continue working on this PR; he's done some great work getting the patch to where it is, but I don't know if he's got the time or inclination to continue working to resolve the other problems you've described. @ahter - let us know if you want to keep working on this, or if you're happy to hand off to @magic-lantern (or, if you want to work in tandem on it).

Oh - and absolutely no argument that there's a *huge* need for unit tests. The bad excuse I have for this at the moment is that testing setuptools extensions is a bit painful; testing integration with external tools is even worse; and the bulk of the work on briefcase was originally done in a "get the demo working for a conference presentation" sprint. If a test suite was to magically appear, I wouldn't object in the slightest :-) 

The other big need is error checking - the bulk of the calls to external tools are implemented in a "assume it's going to work" form. Which is fine when they do... but when they don't... 😆

ahter:
I would love to continue working on Briefcase. I am going though a very busy month so it would be best if we worked in tandem. 

Since the other issue is a different one and seems independent of this patch,  perhaps it could be addressed in a different patch/PR. That’s my personal view.

magic-lantern:
Thanks @ahter - I'll do as you and @freakboy3742 suggest and work on the next phase of fixing how briefcase parses the output from Xcode to automate the iOS build process. So far, my changes I'm working on need to build on this patch, so hopefully this change can be finalized soon.

As far as I know, these changes and my planned changes only work with the current version of Xcode. What is the plan for supporting old versions of Xcode? Also, I would guess that it is likely that the recently announced Xcode 11 could break the iOS portions of briefcase again. Is there some recommendations on how to deal with that?

As for unit testing - I doubt I'll have time to complete a full set of tests, but I should at least be able to start creating stubs for testing (sort of the opposite of TDD I'd say) separate from any Xcode related fixes. While I've built unit tests for Python code, I've never unit tested a module. I think testing would be a good candidate for long term goals.

freakboy3742:
@magic-lantern So, my general opinion is that Apple *really* wants you to keep updating XCode, so we don't need to support historically ancient XCode behaviors if it's onerous. However, if we're able to continue supporting old versions (and maintain testing of those versions), then we might as well keep that support. 

If we *do* deprecate support for an old XCode version, then we should make sure we report to the user that this is the source of the problem. If, for example, I have an old box with XCode 9, I should get an error if I try to deploy to iOS. (I should also get warnings if I don't have XCode installed *at all*, or if I don't have the command line tools installed).

In terms of testing - an incremental approach sounds fine. 

I'm also contemplating doing a rework of Briefcase to decouple it from setuptools - so it would become a standalone command line tool. This would hopefully improve the testing story (and also provide a convenient point at which to start adding tests).

magic-lantern:
@ahter - Not sure what your availability is, but turns out my changes won't work at all without your changes first. Due to the change in Xcode output, my fixes can't be tested without yours first in place. I did send a pull request against your fork - not sure if that's the right way to collaborate with you on this.

Seems like the blocker for this pull request is just some pycodestyle checks. I'm happy to help resolve those failed checks to move this forward.

Let me know!

ahter:
@magic-lantern Hello Seth, I will only be able to have a look this weekend.

If you'd have time, I'd be more than happy if you can have a look at the those issues and move them forward.

src/briefcase/ios.py
```diff
@@ -119,79 +120,99 @@ def install_launch_scripts(self):
...
+                            except Exception:
+                                print("Invalid selection.")
+                                print
+
+                if self.device is None:
+                    device_list = data['devices'].get(self.os_version, [])
+                    self.device = [x for x in device_list if x['name'].lower() == self.device_name.lower()][0]
+
+            except subprocess.CalledProcessError as cpe:
+                print('xcrun simctl list resulted in an error. Return code was', cpe.returncode, 'with output:', cpe.output)
```
brutusthebee: At column 120: (E501) line too long (124 > 119 characters)

src/briefcase/ios.py
```diff
@@ -119,79 +120,99 @@ def install_launch_scripts(self):
...
+                int(v)
+                for v in output.split('\n')[0].split(' ')[1].split('.')[:2]
+            )
+            if version < (8, 0):
+                print('\nAutomated builds require XCode 8.0 or later', file=sys.stderr)
+                return False
+            else:
+                return True
+        except subprocess.CalledProcessError as cpe:
+            print('xcrun xcodebuild -version resulted in an error. Return code was', cpe.returncode, 'with output:', cpe.output)
```
brutusthebee: At column 120: (E501) line too long (128 > 119 characters)

tanabi:
Any update on this?  This is going to impact a lot of people and it looks like its probalby ready to merge?

magic-lantern:
Actually I think you are right! @ahter and I were working on this together. There were some style issues that I resolved but didn't realize they had been merged. I think everything now is ready.

@freakboy3742 - Thanks for your help along the way as well. Not sure who besides you has write access - when you have time please review and let us know if this can be merged in!

ahter:
Yes, seems good to go?

PR #217, opened 2019-04-27
Docs fixes

rixx:
I'm also very tempted to put a big fat warning on top of the Android section, telling first-timers that they'll be expected to download at least 2-2.5GB of data, but that's my frustration :robot:  speaking, and probably not something that needs extra mentioning.

(Though a `.. note: Setting up an Android development setup like this will require downloads of at least 2GB. If you just want to try out briefcase quickly, consider starting off with a different target like …` probably wouldn't hurt.)


freakboy3742:
That's a good point - desktop platforms are definitely a better first trial.

And, for the record - the huge download problem exsits for iOS too - you need XCode for IOS to work. A developer on macOS is *marginally* more likely to have XCode installed (it's used by a lot of other tools) but if they don't it's a couple of gig first time out.

PR #223, opened 2019-05-06
use the correct python executable when calling pip

MiriamSexton:
I believe this should address #208. `pip` should now be called using the correct version of python and not the system default.


danyeaw:
Thanks @MiriamSexton for the contribution! :tada:

PR #266, opened 2019-09-07
Convert Briefcase to be a PEP518 tool, rather than a setuptools extension

freakboy3742:
The 0.2 branch of Briefcase has served us well, but certain aspects of it's user experience is are showing strains.

Various conversations at PyCons have lead me to believe that part of the problem may be that we are trying to cram "application packaging" into setuptools, when application packaging is really a standalone task. 

PEP518 gives a way for us to provide a face for this task. By making briefcase it's own standalone PEP518 tool, we can set our own configuration and UI direction.

This PR is a WIP branch describing a possible future for Briefcase 0.3. Some sample usage of the new interface:

* Show usage 
```
briefcase
```
* Create a new stub app for the default platform, in the default format
```
briefcase create
```
* Create a new stub app for Linux, using the default format for Linux
```
briefcase create linux
```
* Create a new stub linux app in the `snap` format (which isn't the default)
```
briefcase create linux snap
```
* Update the code in the existing stub app (default platform and format)
```
briefcase update
```
* Update the code in the existing stub macOS DMG project
```
briefcase update macOS dmg
```
* Build the code in the existing stub macOS DMG project. This step is only required when there's a compilation step of some kind; macOS apps, for example, would be a no-op on build (since they're just a directory convention)
```
briefcase build macOS dmg
```
* Run the app in the default platform and format
```
briefcase run
```
* Run the app in the default platform and format, but update the code and rebuild the app first.
```
briefcase run -u
```
* Run the iOS project in a specific simulator 
```
briefcase run iOS --device <device UUID>
```
* Publish the code for the default platform and format to an s3 bucket
```
briefcase publish -c s3 
```

All configuration would be contained in `pyproject.toml`. A simple sample project would look something like:

````
[build-system]
requires = ["briefcase"]

[tool.briefcase]
version = "0.1"

[tool.briefcase.app.myapp]
formal_name = "My App"
description = "My first Briefcase App"
bundle = "com.example.myapp"
sources = ['src/myapp']

[tool.briefcase.app.myapp.macOS]
requires = ['toga-cocoa==0.3.0.dev15']
```


setup.py
```diff
@@ -1,81 +1,4 @@
 #/usr/bin/env python
```
brutusthebee: At column 1: (E265) block comment should start with '# '

src/briefcase/__main__.py
```diff
@@ -0,0 +1,239 @@
+import argparse
+import importlib
+import pkg_resources
```
brutusthebee: At column 1: (F401) 'importlib' imported but unused

glasnt:
When I specify the `-f` flag for format options without a platform, I would expect the same functionality as when I didn't provide a platform, i.e. my local platform. 

that is

`briefcase create -f`
would be the same as
`briefcase create macos -f`

`-f` should hard exit with the options, in either case. 

The current error message has the reason behind the error  hidden in the message, the last time is nonsensical. 


<details>

```
$ briefcase create -f
Formats are platform specific. Specify a platform to see availble formats.
  File "/Users/glasnt/git/beeware/briefcase/src/briefcase/__main__.py", line 246, in main
    command = parse_cmdline(sys.argv[1:])
  File "/Users/glasnt/git/beeware/briefcase/src/briefcase/__main__.py", line 185, in parse_cmdline
    raise UnknownFormatsError()
briefcase.__main__.UnknownFormatsErrorFormats are platform specific. Specify a platform to see availble formats.
: <exception str() failed>

During handling of the above exception, another exception occurred:

  File "/Users/glasnt/git/beeware/briefcase/venv/bin/briefcase", line 11, in <module>
    load_entry_point('briefcase', 'console_scripts', 'briefcase')()
  File "/Users/glasnt/git/beeware/briefcase/src/briefcase/__main__.py", line 250, in main
    print(e, file=sys.stdout if e.error_code == 0 else sys.stderr)
TypeError: __str__ returned non-string (type NoneType)
```

</details>

freakboy3742:
@glasnt Thanks for that report - the `-f` issue should be resolved in 75eab02.

PR #271, opened 2019-10-01
macOS | Include project version in output DMG file name

tmontes:
In the context of [this Mu packaging issue](https://github.com/mu-editor/mu/issues/925), where I found out that `briefcase`-produced *application bundles* do not include the underlying project version, I created [this PR](https://github.com/beeware/Python-macOS-template/pull/3) in the `Python-macOS-template` project.

Along those lines, I figured that `briefcase`-produced DMG files should (could?) also include the project version in their filename. This PR implements precisely that change.

It is nearly trivial and works nicely on my dev system.

Thanks in advance for reviewing it: I'll be glad to address any improvements you may point me to.

### DIFF
--- a/src/briefcase/macos.py
+++ b/src/briefcase/macos.py
@@ -103,7 +103,7 @@ class macos(app):
 
     def build_app(self):
         print("Building DMG file...")
-        dmg_name = self.formal_name + '.dmg'
+        dmg_name = self.formal_name + '-' + self.version + '.dmg'
         dmg_path = os.path.join(os.path.abspath(self.dir), dmg_name)
 
         dmgbuild.build_dmg(


tmontes:
Thanks for the quick response @freakboy3742. I'm tempted to propose a few other DMG-related improvements. :)

freakboy3742:
@tmontes As a heads up, I'm in the process of doing a ground-up rebuild of Briefcase - you can see the work in progress on #266. This is to address a number of problems we've hit along the way, including differentiating between app and dmg builds.

To that end - if you've got ideas for big features in DMG support (or anything else, for that matter), it might be more productive to document them on a tickets, rather than dig into large scale rewrites of the existing DMG handling code.

tmontes:
@freakboy3742, thanks for the heads up - I had noticed #266, indeed, but haven't looked at it yet. I do agree with you that it will be more productive to share ideas within issues if such a change is in progress -- I will do just that.

PS: I tried to give the code in #266 a quick run but could not understand how to "make it go", from the builtin help/written notes in the PR -- I gather I'll need to configure "things" in `pyproject.toml` which, I suppose, will be `briefcase`-specific. Then again, not being very familiar with PEP-517, maybe there's some homework for me to do there... I'd love to track its progress, testing it out in the Mu packaging process... Am I missing something obvious? Can you share any hints? Thanks for that.

freakboy3742:
@tmontes You're not missing anything obvious; at this point, it's a bare bones framework, and can't actually package anything yet. A lot of the rebuild has been focusing on testing (which is an area where we've historically been lax) so what's there is very well tested, but doesn't actually *do* anything much (yet).

The end goal is that `pyproject.toml` will replace setup.py as a source of configuration information, and it will be possible to override configuration on a per-platform basis. It also allows for a single codebase to generate multiple output executables, and allows for different output formats on each platform (e.g., differentiating between NSIS and MSI for Windows, or raw `.app` and `.dmg` for macOS).

I'll be sure to loop you in once we get to a point where we have feature compatibility with the current working branch.

tmontes:
> The end goal is that `pyproject.toml` will replace setup.py as a source of configuration information, and it will be possible to override configuration on a per-platform basis. It also allows for a single codebase to generate multiple output executables, and allows for different output formats on each platform (e.g., differentiating between NSIS and MSI for Windows, or raw `.app` and `.dmg` for macOS).

(going a bit out of context -- in this PR -- but following up on your words: maybe an issue can be created for such discussion, later)

While I think I understand the general `pyproject.toml` philosophy, I'm left wondering how some things will work. In particular, I'm concerned about "single sources of truth" in the general Mu codebase, including what pertains to packaging -- some [non-trivial effort](https://github.com/mu-editor/mu/pull/779) was put in place to have things like dependencies and version declared in a single place: version sourced from `__init__.py:__version__`, and dependencies sourced from `setup.py`. Maybe it is too early, but do you foresee any challenges in the general "having a single source of truth" for such meta-info?

freakboy3742:
So, from my perspective: the fact that a "single source of truth" is hard to define at present is essentially the problem we're trying to solve. setuptools/distutils is a useful collection of utilities, but they're very much targeted at distributing python packages to PyPI. Application distribution requires different metadata, handled in different ways, and my hope is that by building a tool that is focused on solving the application deployment problem, we'll be able to provide solid answers to any source-of-truth questions. And since we won't be trying to cram the solution into a setup.py shaped hole, we'll be able to pick solutions that work.

Right now, what I'm trying to build is a base framework that:
1. Recognizes that an application has a base configuration that, but that configuration may need to be amended or altered on a per-platform basis;
2. Recognizes that there are multiple deployment "options" on any given platform, and that it may be necessary to support multiple options for a single platform;
3. Is user-extensible; and
4. Is inherently testable.

Once that framework is in place, we'll be in a much stronger position to address specifics of individual platform configuration needs.

Looking at Mu's use case specifically - I can imagine that *some* of the single-source-of-truth problem is caused by the fact that you have (by necessity) used a range of different tools for packaging, each of which answers configuration questions in a slightly different way. From my perspective as Briefcase's author - if Briefcase isn't in a position to replace all those tools and present a unified picture for application packaging, then we've failed as a project. I recognize that the practicality of Mu actually adopting Briefcase as a unified packaging solution is another matter entirely; but as a goal, reaching a point where Briefcase is good enough that Mu can use it for all app packaging is a reasonable measure of success for me, personally.

Issue #281, opened 2019-12-21
RfC: Add GitHub Actions Workflow to Publish Briefcase to PyPI


danyeaw:
As part of #266, @freakboy3742 swapped to using GitHub Actions for CI. One thing that we were chatting about that would be nice is to add a CD component in order to publish a new version to PyPI.

This issue to collect feedback on this update. I think there are four main components to this discussion:

1. What events we want it to push to PyPI on?
2. Do we want to use setuptools, or one of the more recent tools like Twine or Poetry?
3. Do we want to upload any artifacts to the Release on GitHub?
4. Do we want to do some checks after it is uploaded like try to pip install the library on all the platforms?

For 1: I think we have these main options:
1. Any time a new tag in the format of `v#.#.#` is pushed to the repo
2. Any time a new Release is created, this is the default that the Publish Python Package GitHub Actions template uses
3. Any time a new Release is published
This library isn't currently making use of adding release notes or uploading assets to the releases. I think going with the 2nd option would be a straightforward approach in this case. You would push your tag, draft a release which kicks off the workflow, and then once the workflow is done manually publish the release.

For 2:
I see the options are leave things alone and use setuptools, go all-in and use Poetry (with `poetry config virtualenvs.create false` or `poetry config virtualenvs.in-project = true` so that it doesn't take over virtualenv creation), or use Twine just for building and uploading to PyPI.

For 3: since artifacts aren't currently part of the Releases, I would assume we don't start now.

For 4: this is more of a nice to have, but could give us more confidence in a release after it is pushed to PyPI.

freakboy3742:
Thanks for putting this together, @danyeaw. 

There's a part of me that likes the simplicity of (1) and cutting a release by just pushing a tag. However, given that it's almost impossible to undo mistakes made on PyPI, I'd feel better if there was a slightly more deliberate, human-reviewable action in place. 

I tend to agree that (2) is the best of the options you present. However, an even better option might be a hybrid: create a *draft* release whenever vX.X.X is tagged, but formally publish to PyPI when the release is published. That automates the process setting up and formally publishing the release, but makes the *trigger* of the actual publication a manual, human-approved process. If adopting a tool like [towncrier](https://github.com/hawkowl/towncrier) for release notes will help automating the drafting process, I'd be open to doing that.

I'm not sure what you're driving at with your discussion of setuptools, twine and poetry; as far as I'm aware, those are three independent tools, solving different problems:
[...]

PR #282, opened 2019-12-22
Create Workflows to Upload new Releases

danyeaw:
Initial PR for #281.

This PR creates two new GitHub Actions workflows:

1. Drafts a release when a new version tag is pushed.
2. Uploads a release to PyPI once the release is published.

In order to take advantage of this, PyPI credentials need to be stored as [secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets). The preferred way to do this would be to create a [PyPI API token](https://pypi.org/help/#apitoken), but your username and password could also be used.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install setuptools wheel twine
+        pip install --upgrade setuptools wheel twine
     - name: Build and publish
       env:
         TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,4 +23,5 @@ jobs:
         TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
       run: |
         python setup.py sdist bdist_wheel
+        twine check dist/*
         twine upload dist/*
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,26 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: published
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
+        twine upload dist/*
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,24 @@
+name: Create Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  build:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@master
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          draft: true
+          prerelease: false


.github/workflows/publish.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
```
freakboy3742: These probably need to be `--upgrade` too; setuptools *should* be installed in a default environment.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
```
freakboy3742: TIL 3.x was possible in Github Actions...

.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          draft: true
+          prerelease: false
```
freakboy3742: Ideally, we'd set this dependent on whether the version tag has 'a', 'b', 'rc', 'dev', or 'pre' in it... is that possible? Or do we just have to keep our wits about us and make sure we uncheck this if we tag a dev release?

.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
+name: Create Release
+
+on:
+  push:
+    tags:
+      - 'v*'
```
freakboy3742: It's an unlikely edge case, but if we tagged "very_important_milestone", this would trigger this workflow.  If I'm reading the docs right, a pattern of `'v[0-9]+.[0-9]+*'` would give us at least some superficial protection against this.

danyeaw: I think the risk here and for the next comment is low. If an extra tag was pushed, it would just create a draft release. We would just need to go to the release tab and click on the delete draft button to clean it up. All the tags in the repo right now are for releases. I could add in the regular expression, but might be overkill to prevent an extra draft release for some one off event.

freakboy3742: Fair enough. As long as it's easy to clean up drafts that get mistakenly created, I agree that the extra complexity probably isn't worth it.

danyeaw:
@freakboy3742 Thanks for the feedback.

> We might want to set this configuration up on a test repository (and publish to testPyPI) to validate the processes will work as expected before we try to land this into a live project.

I could test this on my fork before we merge. Do you think I could get test-pypi write access?

I add the `twine check` command :+1:.
 
> Lastly - what happens if a tag is updated?

By updated, I think you mean for example the v1.0 tag is deleted and a new v1.0 tag is created and pushed? A new draft release would be created, and we would just have to delete the previous draft release on the Releases tab.

.github/workflows/release.yml
danyeaw: Yup, it is very easy to clean them up, a single button click. :+1:

freakboy3742:
Yeah - that's what I meant about deleting or changing a version tag.

I've just given you maintainer access on test.PyPI.org for briefcase.

Assuming the `twine check` fails... what's the recovery procedure? We can't retag the same hash, and the release will have already been published. Would it be better to do the `twine check` as a step on the tag event? That way, at least if it fails, we'll have to create an updated tag anyway. 

Alternatively, we could add `twine check` as a normal build check (i.e., as another pycodestyle check). That way, we hopefully won't ever end up accepting merging a PR that has a potential publication problem baked into the markup.

danyeaw:
@freakboy3742 I have completed my testing, and found one issue with the `twine check` command that I have fixed.

You can see the successful result here:
https://test.pypi.org/project/briefcase/0.3.0.dev2/
The upload is the result of my PR:
https://github.com/danyeaw/briefcase/pull/1
and this release:
https://github.com/danyeaw/briefcase/releases/tag/v0.3.0.dev2
I hope it is OK that I burned the 0.3.0.dev2 version on the test server.

I also added an error to the README and tried to publish a 0.3.0.dev3 version. As soon as the `twine check` command failed, the workflow immediately cancelled and the version was not uploaded to test.pypi.org. You can then delete the tag (locally and on the repo) and this sets the Release back to draft automatically, then push a new tag to fix the issue.

I also made use of the PyPI API Token, and that worked well with using `__token__` as the username.

danyeaw:
Yup, that is a good summary of the process 👍

freakboy3742:
Righto then - Lets 🚢!

PR #309, opened 2020-03-06
dev command can now handle windows separator

### LINKED ISSUE #308: Cannot find package when run 'briefcase dev'
I have the following pyproject.toml file:
```
[build-system]
requires = ["briefcase"]

[tool.briefcase]
project_name = "Fit GUI"
bundle = "com.fitgui"
version = "0.0.1"
url = "https://fitgui.com/FitGui"
license = "Apache Software License"
author = 'Sagi Shadur'
author_email = "saroad2@gmail.com"
sources = ['src/lab_util']

[tool.briefcase.app.fitgui]
formal_name = "Fit GUI"
description = "Fitting GUI for laboratory experiments"
icon = "src/fitgui/resources/FitGui"
sources = ['src/fitgui']
requires = [
    'numpy>=1.18.1',
    'scipy>=1.4.1',
    'matplotlib>=3.1.3',
    'pandas>=1.0.1',
    'prettytable>=0.7.2',
    'xlrd>=1.2.0',
]


[tool.briefcase.app.fitgui.macOS]
requires = [
    'toga-cocoa>=0.3.0.dev18',
]

[tool.briefcase.app.fitgui.linux]
requires = [
    'toga-gtk>=0.3.0.dev18',
]

[tool.briefcase.app.fitgui.windows]
requires = [
    'toga-winforms>=0.3.0.dev18',
]

# Mobile deployments
[tool.briefcase.app.fitgui.iOS]
requires = [
    'toga-iOS>=0.3.0.dev18',
]

[tool.briefcase.app.fitgui.android]
requires = [
    'toga-android>=0.3.0.dev18',
]
```

When I try to run "briefcase dev" I get the following error:

```
c:\users\saroa\repositories\lab_util\venv\scripts\python.exe: No module named fitgui

Unable to start application 'fitgui'
```

Do you have any idea why?
I am working on a windows computer.

saroad2:
fixes #308

### DIFF
--- a/tests/commands/dev/conftest.py
+++ b/tests/commands/dev/conftest.py
@@ -70,5 +70,5 @@ def third_app(tmp_path):
         bundle="com.example",
         version="0.0.2",
         description="The third simple app",
-        sources=["src/third", "src2/b"],
+        sources=["src/third", "src/common", "other"],
     )
--- a/tests/commands/dev/conftest.py
+++ b/tests/commands/dev/conftest.py
@@ -6,7 +6,7 @@ from briefcase.config import AppConfig
 
 
 @pytest.fixture
-def dev_command_with_dummy_subprocess(tmp_path):
+def dev_command(tmp_path):
     command = DevCommand(base_path=tmp_path)
     command.subprocess = mock.MagicMock()
     return command
--- a/tests/commands/dev/conftest.py
+++ b/tests/commands/dev/conftest.py
@@ -32,8 +32,13 @@ class DummyDevCommand(DevCommand):
 
 
 @pytest.fixture
-def dev_command(tmp_path):
+def dummy_dev_command(tmp_path):
     return DummyDevCommand(base_path=tmp_path)
+
+
+@pytest.fixture
+def dev_command(tmp_path):
+    return DevCommand(base_path=tmp_path)
 
 
 @pytest.fixture
--- a/tests/commands/dev/conftest.py
+++ b/tests/commands/dev/conftest.py
@@ -1,4 +1,5 @@
 import pytest
+from unittest import mock
 
 from briefcase.commands import DevCommand
 from briefcase.commands.base import full_kwargs
@@ -17,18 +18,7 @@ class DummyDevCommand(DevCommand):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, apps=[], **kwargs)
 
-        self.actions = []
-        self.env = dict()
-
-    def install_dev_dependencies(self, app, **kwargs):
-        self.actions.append(('dev_dependencies', app.app_name, kwargs))
-
-    def run_dev_app(self, app, env, **kwargs):
-        self.actions.append(('run_dev', app.app_name, kwargs))
-        self.env.update(env)
-        return full_kwargs({
-            'run_dev_state': app.app_name
-        }, kwargs)
+        self.subprocess = mock.MagicMock()
 
 
 @pytest.fixture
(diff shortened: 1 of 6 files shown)


saroad2:
As you can see in my commit, I have fully tested 1 and 2.
I think 3 is a byproduct of 1 and 2 and doesn't need to be tested directly. eventually, you call the __call__ method and see that the subprocess command was called with the environment dictionary.

saroad2:
I think I understood what you requested, but I respectfully disagree on the way you suggested doing that.

I think that we should test the *behavior* of the __call__ function, and not the inner implementation of it, invoking `get_environment` and `run_dev_app`. In that way, you could change the inner implementation tomorrow or next week and the tests will stay the same because the behavior hasn't changed.
This is your project, and therefore I will do exactly as you requested, but I must say that I do not think its the right thing to do.

Working on the fix now.

saroad2:
@freakboy3742 Ready for your review :)

tests/commands/dev/conftest.py
```diff
@@ -18,22 +19,35 @@ def __init__(self, *args, **kwargs):
...
         self.actions.append(('dev_dependencies', app.app_name, kwargs))
 
-    def run_dev_app(self, app, **kwargs):
-        self.actions.append(('run_dev', app.app_name, kwargs))
+    def get_environment(self, app):
+        return self.env
+
+    def run_dev_app(self, app, env, **kwargs):
+        self.actions.append(('run_dev', app.app_name, kwargs, env))
+        self.env.update(env)
```
freakboy3742: The "update" step (and the command-level environment state) isn't needed here - we only need to track that the app is invoked with whatever environment is provided to `run_dev_app`.

tests/commands/dev/test_run_dev_app.py
```diff
@@ -0,0 +1,27 @@
+import sys
+from subprocess import CalledProcessError
+import pytest
+
+from briefcase.exceptions import BriefcaseCommandError
+
+
+def raise_called_process_error(*args, **kwargs):
+    raise CalledProcessError(returncode=2, cmd="cmd")
```
freakboy3742: The extra callable isn't needed here; if the side effect is an instance of an exception, mock will cause it to be raised.

tests/commands/dev/test_run_dev_app.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def test_subprocess_running_successfully(dev_command, first_app):
+    env = dict(a=1, b=2, c=3)
+    dev_command.run_dev_app(first_app, env)
+    dev_command.subprocess.run.assert_called_once_with([sys.executable, '-m', first_app.app_name],
+                                                       env=env,
+                                                       check=True)
```
freakboy3742: We are aiming to follow "black" style indentation, not pycharm "hanging" indents. The eventual goal of actually applying black consistently across the codebase.

tests/commands/dev/conftest.py
```diff
@@ -18,22 +19,35 @@ def __init__(self, *args, **kwargs):
...
         return full_kwargs({
-            'run_dev_state': app.app_name
+            'run_dev_state': app.app_name,
+            'env': env
         }, kwargs)
 
 
 @pytest.fixture
-def dev_command(tmp_path):
+def dummy_dev_command(tmp_path):
```
freakboy3742: This rename causes a lot of code churn, but doesn't significantly increase readability. Since the 'replacement' `dev_command` is only used in `test_run_dev_app`, I'd suggest moving the "new" fixture into that specific test.

saroad2:
@freakboy3742 Fixed :)

tests/commands/dev/test_call.py
```diff
@@ -1,184 +1,217 @@
...
+@pytest.fixture
+def dummy_dev_command(tmp_path):
+    return DummyDevCommand(base_path=tmp_path)
+
+
+def test_no_git(dummy_dev_command, first_app):
     "If Git is not installed, an error is raised"
     # Mock a non-existent git
-    dev_command.git = None
+    dummy_dev_command.git = None
```
freakboy3742: This is the code churn issue I was referring to. Patches should, generally, be as small as possible. Unless there's a good reason, arbitrary renames and restructures should be avoided. The problem I was flagging was the mass-rename of "dev_command" to "dummy_dev_command". That rename is a significant portion of the final diff, but doesn't actually improve any aspect of the patch.

PR #311, opened 2020-03-07
WIP: Add initial Android APK commands

paulproteus:
Right now, `briefcase create android` and `briefcase build android` seem to work in my testing. I haven't actually written `briefcase run android` yet, so I haven't validated the resulting APK, but I will do that. :)

Requesting your feedback @freakboy3742 .

I'll leave some comments on the pull request, too. Expect more work on this in the coming days, but wanted to share what I have now.


src/briefcase/commands/create.py
```diff
@@ -20,7 +20,7 @@ class InvalidTemplateRepository(BriefcaseCommandError):
     def __init__(self, template):
         self.template = template
         super().__init__(
-            'Unable to clone application template; is the template path {template!r} correct?'.format(
+            "Unable to clone application template; is the template path {template!r} correct?".format(
```
paulproteus: I accidentally-ish ran `black` on this file. I can separate that into a follow-up commit. Is that the right code formatter to use? I like black, but I am happy to defer to whatever you advise.

src/briefcase/commands/create.py
```diff
@@ -368,34 +355,39 @@ def install_app_support_package(self, app: BaseConfig):
...
+            raise NetworkFailure("downloading support package")
 
         try:
             print("Unpacking support package...")
             support_path = self.support_path(app)
             support_path.mkdir(parents=True, exist_ok=True)
             self.shutil.unpack_archive(
-                str(support_filename),
-                extract_dir=str(support_path)
+                str(support_filename), extract_dir=str(support_path), format="zip"
```
paulproteus: If I didn't specify `format="zip"` here, then `unpack_archive` would raise `ValueError` with invalid format.

BTW, what's up with the UUID-style filename that gets created for this file?

Basically... how does this ever work? :)

src/briefcase/commands/create.py
```diff
@@ -410,11 +402,14 @@ def install_app_dependencies(self, app: BaseConfig):
             try:
                 self.subprocess.run(
                     [
-                        sys.executable, "-m",
```
paulproteus: Skip to the next file if you're reading this pull request to leave comments :)

src/briefcase/platforms/android/apk.py
```diff
@@ -1 +1,205 @@
...
+
+
+class ApkUpdateCommand(ApkMixin, UpdateCommand):
+    description = "Update an existing Android APK."
+
+
+class ApkBuildCommand(ApkMixin, BuildCommand):
+    description = "Build an Android APK."
+
+    # e.g.
```
paulproteus: I haven't written these tests yet, but I remember our discussion.

src/briefcase/platforms/android/apk.py
```diff
@@ -1 +1,205 @@
...
+
+
+# The way to split this up is that you might start with more
+# stuff externally specified.
+
+
+class ApkCreateCommand(ApkMixin, CreateCommand):
+    description = "Create and populate an Android APK."
+
+    # TODO: Move these two into my app, and out of this pull request!
```
paulproteus: I intend to follow this TODO, but I haven't yet.

src/briefcase/platforms/android/apk.py
```diff
@@ -1 +1,205 @@
...
+        return Path.home() / ".briefcase" / "tools" / "android_sdk_tmp"
+
+    @property
+    def android_sdk_path(self):
+        return Path.home() / ".briefcase" / "tools" / "android_sdk"
+
+    def binary_path(self, app):
+        return (
+            self.platform_path
+            / app.formal_name
```
paulproteus: BTW, why does "formal_name" exist? What's the real difference between an app name and its formal name? Anyway, here I'm hoping it's safe for the filesystem. I'm OK continuing with that assumption of filesystem-safe formal name, but it's a little odd.

src/briefcase/platforms/android/apk.py
```diff
@@ -1 +1,205 @@
...
+            raise BriefcaseCommandError(
+                """
+Found Python version {self.python_version_tag}. Android packaging currently requires Python 3.7.
+        """.format(
+                    self=self
+                )
+            )
+
+
+# The way to split this up is that you might start with more
```
paulproteus: This was a note-to-self about how to split up the pull request. Ignore it.

src/briefcase/platforms/android/apk.py
```diff
@@ -1 +1,205 @@
...
+            print()
+            raise BriefcaseCommandError(
+                "Error while building app {app.app_name}.".format(app=app)
+            )
+
+
+class ApkRunCommand(ApkMixin, RunCommand):
+    description = "Run a Linux AppImage."
+
+    def run_app(self, app: BaseConfig, **kwargs):
```
paulproteus: Not implemented yet :)

src/briefcase/platforms/android/apk.py
```diff
@@ -1 +1,205 @@
...
+        #
+        # Show all possible targets
+        #
+        # If you picked an emulator device but it's not started, start it
+        #
+        # If you picked the briefcase auto-generated device, then create it :)
+
+
+class ApkPackageCommand(ApkMixin, PackageCommand):
+    description = "Package an Android APK."
```
paulproteus: Is it okay if `package` and `build` are the same for Android?

I'll keep this a no-op like this, per https://github.com/beeware/briefcase/blob/master/src/briefcase/platforms/iOS/xcode.py , but figured I'd check.

src/briefcase/platforms/android/apk.py
```diff
@@ -1 +1,205 @@
...
+
+
+class ApkPublishCommand(ApkMixin, PublishCommand):
+    description = "Publish an Android APK."
+
```
paulproteus: Not implemented yet.

src/briefcase/commands/create.py
freakboy3742: We haven't *formally* adopted black, but we're definitely leaning towards black style choices. At some point (probably in the not too distant future) we'll almost certainly formally adopt black into the toolchain.

src/briefcase/commands/create.py
freakboy3742: Interesting - omitting the format flag should cause the archive format to be autodetected. That's useful because some of the support packages are tarballs, and some are zip files.

Not sure what you're referring to with the "uuid-style filename", though - the support package should be the name as downloaded from the support folder (i.e. ,"Python-Apple-support.b3.tar.gz"). Are you seeing different behaviour?

src/briefcase/platforms/android/apk.py
freakboy3742: This is covered in the docs, but "Formal name" is a human readable name, and may include capitalization, punctuation, start with a number, or more. App name must be a PEP508 compliant name. Think "Hello World!" vs "hello_world".

We don't currently do any filesystem-safety checks on formal name - but you're right, we probably should. We *do* validate app_name is PEP508 compliant, though.

src/briefcase/platforms/android/apk.py
```diff
@@ -1 +1,205 @@
...
+        if not self.android_sdk_path.exists():
+            try:
+                android_sdk_zip_path = self.download_url(
+                    url=self.android_sdk_url,
+                    download_path=Path.home() / ".briefcase" / "tools",
+                )
+            except requests_exceptions.ConnectionError:
+                raise NetworkFailure("downloading Android SDK")
+            with ZipFile(android_sdk_zip_path) as android_sdk_zip:
+                android_sdk_zip.extractall(path=self.android_sdk_path_tmp)
```
freakboy3742: Any reason you're using an explicitly specified, constant tmp path, rather than a genuine tmpfile generated, unique-every-time directory? If we're going to go to the trouble of caring about the edge case of a "partially unrolled zipfile" being a problem, then using the same temp dir every time would seem to be a risk.

src/briefcase/platforms/android/apk.py
```diff
@@ -1 +1,205 @@
...
+            for binpath in (self.android_sdk_path_tmp / "tools" / "bin").glob("*"):
+                binpath.chmod(0o755)
+            self.android_sdk_path_tmp.rename(self.android_sdk_path)
+
+        print("Ensuring all Android SDK licenses are accepted...")
+        self.subprocess.run(
+            [self.android_sdk_path / "tools" / "bin" / "sdkmanager", "--licenses"],
+            check=True,
+            cwd=self.android_sdk_path,
+        )
```
freakboy3742: Error handling required - any call to subprocess needs to catch the non-zero return code and report that as a problem.

src/briefcase/platforms/android/apk.py
```diff
@@ -1 +1,205 @@
...
+            except requests_exceptions.ConnectionError:
+                raise NetworkFailure("downloading Android SDK")
+            with ZipFile(android_sdk_zip_path) as android_sdk_zip:
+                android_sdk_zip.extractall(path=self.android_sdk_path_tmp)
+            # Remove the ZIP file; it has no purpose now that it is extracted.
+            android_sdk_zip_path.unlink()
+            # Set executable permissions; Python ZipFile ignores these, but we need them.
+            for binpath in (self.android_sdk_path_tmp / "tools" / "bin").glob("*"):
+                binpath.chmod(0o755)
+            self.android_sdk_path_tmp.rename(self.android_sdk_path)
```
freakboy3742: I can see a bunch of possible error cases here that aren't caught (bad zipfile; no permissions to unpack zipfile; disk space problems when unpacking...).

src/briefcase/platforms/android/apk.py
```diff
@@ -1 +1,205 @@
...
+            appdir_path = self.bundle_path(app)
+            self.subprocess.run(
+                [appdir_path / "gradlew", "assembleDebug",],
+                env=env,
+                check=True,
+                cwd=str(appdir_path),
+            )
+
+            # Make the binary executable.
+            self.os.chmod(str(self.binary_path(app)), 0o755)
```
freakboy3742: Really? The APK file needs to be `+x`? I haven't noticed this requirement in manual testing...

src/briefcase/platforms/android/apk.py
freakboy3742: At least for first pass, it's fine for package to be a no-op. Longer term, package would be where APK signing would occur.

src/briefcase/commands/create.py
paulproteus: +1

I'll keep this pull request more minimal.

I'm interested in volunteering to `black`-ify BeeWare projects, but we can discuss that later.

src/briefcase/commands/create.py
paulproteus: re: UUID-style filename: It probably has to do with how GitHub chooses URLs for assets. I'll take a look at that.

src/briefcase/platforms/android/apk.py
paulproteus: +1

I'll continue writing code happily assuming it's filesystem-safe for now, then, and we can make a project-wide decision about that later/never. :)

src/briefcase/platforms/android/apk.py
paulproteus: Gotcha -- makes sense that `package` would involve signing.

src/briefcase/platforms/android/apk.py
paulproteus: +1

src/briefcase/platforms/android/apk.py
paulproteus: Just a quick initial implementation. For the next revision, I am planning on doing `rmtree` on the `_tmp` filename, rather than creating temporary paths; that way, I avoid creating files in the filesystem that we never clean up. I can leave a comment explaining that rationale.

I'm also okay using a unique filename, but I think that it's a substantial virtue if we're able to clean up after ourselves, in which case there's no real upside to the random name IMHO.

src/briefcase/platforms/android/apk.py
paulproteus: I think this is unneeded copy-pasta. I'll check.

paulproteus:
I have one remaining big-picture question about this:

I'm using `./gradlew assembleDebug` which creates an APK built in debug mode. There's a different gradle command to build a non-debug APK.

For now, I plan to continue creating debug APKs unless you insist otherwise. Debug APKs have the upside that if there's a crash in C code (for example), the quality of stack traces is higher.

However, in the long run, what's our plan to enable creating non-debug APKs? Does it make sense to create a different `format` (`apk` and `production_apk`?). Or does it make sense to rebuild in production mode when doing the beeware `package` command? I lean toward the latter, but am curious for your input.

I'll keep working on this code, and either submit a mergeable version of this work on this PR, or on a separate PR, probably the latter, but TBD.

src/briefcase/platforms/android/apk.py
freakboy3742: I suppose the bigger question is why the tmp version is needed at all. If the zip fails to unpack, then I don't see that having a temporary directory actually saves us from anything - we're going to need to do a cleanup. If it unpacks successfully, an in-situ unpack works; if ti doesn't, we should be purging the "bad" directory as part of the "it failed" cleanup.

freakboy3742:
I'm inclined to agree with your "debug during testing, release for packaging" approach; sticking with debug builds for now mirrors what happens with iOS (i.e., simulator builds are all Debug builds). That mirrors what I'd eventually like to see on iOS, too - although that's currently being hampered by working out how to automate the signing process.

src/briefcase/platforms/android/apk.py
paulproteus: I guess I'm scared of the case where we can't run our own cleanup for some reason. I'm OK ignoring that case for now, and focusing on writing good cleanup. :)

paulproteus: I tend to love the pattern where "if it exists, it works", that I've learned from some well-written shell scripts and Debian tools; that's where this comes from for me.

paulproteus:
I'm going to close this in favor of #321 , which is fewer lines of diff.

PR #312, opened 2020-03-09
Parse Content-Disposition header if provided

paulproteus:
GitHub assets redirect to strange URLs in Amazon S3, and those URLs
provide Content-Disposition headers to indicate the appropriate filename
to use. This commit adjusts filename calculation to use
Content-Disposition accordingly.

### DIFF
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -1,5 +1,6 @@
 
 import argparse
+from cgi import parse_header
 import importlib
 import inspect
 import os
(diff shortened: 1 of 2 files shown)


paulproteus:
@freakboy3742 Requesting your review.

I would click "Request review" in the interface, but I don't see that under "reviewers," perhaps since I don't have some permission within this repo.

tests/commands/base/test_download_url.py
```diff
@@ -39,6 +40,35 @@ def test_new_download_oneshot(base_command):
...
+def test_new_download_content_disposition(base_command):
+    base_command.requests = mock.MagicMock()
+    response = mock.MagicMock()
+    response.url = 'https://example.com/path/to/irrelevant.zip'
+    response.content = b'content'
+    response.iter_content.return_value = iter([response.content])
+    # We use strange case & whitespace in this sample header to ensure we're lenient in what we accept.
+    # https://tools.ietf.org/html/rfc6266 permits spaces, quotes, and strange case in these places.
+    response.headers = HTTPHeaderDict({
+        'content-disposition': 'Content-Disposition: ATTACHment; filename=    "something.zip"',
```
freakboy3742: This test seems to be trying to validate a bunch of things in one. Could we parameterize the test for a bunch of possible cases (including the RFC6266 case you flag in the implementation)?

src/briefcase/commands/base.py
```diff
@@ -374,9 +375,18 @@ def download_url(self, url, download_path):
...
-        cache_name = urlparse(response.url).path.split('/')[-1]
+        # we need the *final* response. We look at either the `Content-Disposition`
+        # header, or the final URL, to extract the cache filename.
+        cache_full_name = urlparse(response.url).path
+        if 'Content-Disposition' in response.headers:
+            # See also https://tools.ietf.org/html/rfc6266
+            header_value = response.headers['Content-Disposition']
+            value, parameters = parse_header(header_value)
+            if value.split(':', 1)[-1].strip().lower() == 'attachment':
+                # We only the `filename=` parameter. RFC 6266 defines a more complicated
```
freakboy3742: We only.... what... the filename= parameters?

paulproteus: +1 will fix :)

tests/commands/base/test_download_url.py
paulproteus: +1

PR #316, opened 2020-03-09
Add initial adb `get_devices()` integration

paulproteus:
Hi @freakboy3742 !

I wrote this initial `adb devices` output parser as part of my Android briefcase integration work. The Android briefcase integration work isn't done yet, but I wanted to get your review on this to make sure I'm following the right direction.

If it passes review, please merge!

### DIFF
--- a/src/briefcase/integrations/adb.py
+++ b/src/briefcase/integrations/adb.py
@@ -1,0 +1,27 @@
+import subprocess
+import os
+
+from briefcase.exceptions import BriefcaseCommandError
+
+
+def get_devices(sdk_path, sub=subprocess):
+    # Add our `adb` to start of the path.
+    env = dict(os.environ)
+    env['PATH'] = str(sdk_path / 'platform-tools') + ':' + env.get('PATH', '')
+    try:
+        output = sub.check_output(['adb', 'devices'], env=env)
+    except subprocess.CalledProcessError:
+        raise BriefcaseCommandError("Unable to run `adb devices`")
+
+    results = []
+    for line in output.split('\n'):
+        line = line.strip()
+        if not line:
+            continue
+        if line == 'List of devices attached':
+            continue
+        if '\t' in line:
+            first, rest = line.split('\t', 1)
+            if rest.strip() == 'device':
+                results.append(first)
+    return results
--- a/tests/integrations/adb/test_get_devices.py
+++ b/tests/integrations/adb/test_get_devices.py
@@ -1,0 +1,47 @@
+import subprocess
+from pathlib import Path
+from unittest import mock
+
+import pytest
+
+from briefcase.exceptions import BriefcaseCommandError
+from briefcase.integrations.adb import get_devices
+
+DUMMY_SDK_PATH = Path('.')
+
+
+def devices_result(name):
+    """Load `adb devices` output file from the sample directory, and return the content"""
+    filename = Path(__file__).parent / 'devices' / '{name}.txt'.format(name=name)
+    with filename.open() as f:
+        return f.read()
+
+
+def test_adb_missing():
+    "If adb is missing or fails to start, an exception is raised."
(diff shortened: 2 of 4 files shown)


src/briefcase/integrations/adb.py
```diff
@@ -0,0 +1,27 @@
+import subprocess
+import os
+
+from briefcase.exceptions import BriefcaseCommandError
+
+
+def get_devices(sdk_path, sub=subprocess):
+    # Add our `adb` to start of the path.
+    env = dict(os.environ)
+    env['PATH'] = str(sdk_path / 'platform-tools') + ':' + env.get('PATH', '')
```
freakboy3742: Any reason to do this by adding to PATH, rather than invoking adb with the explicit path? (I guess they're both equally valid; just interested in the decision process if there was one)

src/briefcase/integrations/adb.py
```diff
@@ -0,0 +1,27 @@
...
+    try:
+        output = sub.check_output(['adb', 'devices'], env=env)
```
freakboy3742: Do we want to invoke `adb devices -l` and parse more details? From my testing, `adb devices` tells you:
```
List of devices attached
04ea55ee8292009a	device
KAPHN4G9AA1513	device
emulator-5554	device
```
whereas `adb devices -l` gives you:
```
04ea55ee8292009a       device usb:340869888X product:hammerhead model:Nexus_5 device:hammerhead transport_id:4
KAPHN4G9AA1513         device usb:340870144X product:Kogan_Agora_9 model:Kogan_Agora_9 device:Kogan_Agora_9 transport_id:3
emulator-5554          device product:sdk_phone_x86 model:Android_SDK_built_for_x86 device:generic_x86 transport_id:2
```
That would also allow the UI to present a more helpful list of choices when there are multiple devices available.

tests/integrations/adb/test_get_devices.py
```diff
@@ -0,0 +1,47 @@
...
+    assert devices == []
+
+
+def test_one_device():
+    "If there are no devices available, expect the empty list"
+    sub = mock.MagicMock()
+    sub.check_output.return_value = devices_result('one-device')
+    devices = get_devices(sub=sub, sdk_path=DUMMY_SDK_PATH)
+
+    assert devices == ['emulator-5554']
```
freakboy3742: Returning ID *and* a label would be preferable here.

src/briefcase/integrations/adb.py
freakboy3742: Another useful data point: if a device is *not* authorized for debugging mode, but *is* plugged in, you get:
```
List of devices attached
04ea55ee8292009a       unauthorized usb:340869888X transport_id:4
KAPHN4G9AA1513         device usb:340870144X product:Kogan_Agora_9 model:Kogan_Agora_9 device:Kogan_Agora_9 transport_id:3
emulator-5554          device product:sdk_phone_x86 model:Android_SDK_built_for_x86 device:generic_x86 transport_id:2
```

paulproteus: Oh, huh. This is a good idea. +1

tests/integrations/adb/test_get_devices.py
paulproteus: Interesting idea. +1.

src/briefcase/integrations/adb.py
paulproteus: My thinking was to run `adb` with its `argv[0]` staying `adb` not `/path/to/adb`. Obsessing over `argv[0]` might be a little silly. Now that I think about it, it's simplest to run it directly rather than monkey with `PATH`, so I'll switch how it works.

paulproteus:
I'm closing this for now, since I'm going to try to push a much more limited `briefcase run` out the door. I'll re-open it, or make a new PR, when ready.

PR #324, opened 2020-03-15
Add BaseCommand.dot_briefcase_path

paulproteus:
This adds a customizable `dot_briefcase_path` to the `BaseCommand`.

The main reason to add this, for now, is to ensure the tests (1) don't interfere with other tests and (2) don't create invalid state for my laptop's real `~/.briefcase`. (2) is something that actually happened to me -- a testing `.zip` file ended up in `~/.briefcase`, resulting in weird behavior from real briefcase.

One day, we could use something like [user-config](https://pypi.org/project/user-config/) or [config-path](https://pypi.org/project/config-path/) or [appdirs](https://pypi.org/project/appdirs/) or [codi](https://pypi.org/project/codi/) to be more flexible about the location of Briefcase configuration data. That's a task for another pull request IMHO.

### DIFF
--- a/tests/commands/create/test_install_app_support_package.py
+++ b/tests/commands/create/test_install_app_support_package.py
@@ -24,7 +24,7 @@ def test_install_app_support_package(create_command, myapp, tmp_path, support_pa
 
     # Confirm the right URL was used
     create_command.download_url.assert_called_with(
-        download_path=Path.home() / '.briefcase' / 'support',
+        download_path=create_command.dot_briefcase_path / 'support',
         url='https://briefcase-support.org/python?platform=tester&version=3.X',
     )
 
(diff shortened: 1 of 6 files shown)


tests/commands/create/test_install_app_support_package.py
```diff
@@ -24,7 +24,7 @@ def test_install_app_support_package(create_command, myapp, tmp_path, support_pa
 
     # Confirm the right URL was used
     create_command.download_url.assert_called_with(
-        download_path=Path.home() / '.briefcase' / 'support',
+        download_path=create_command.dot_briefcase_path / 'support',
```
paulproteus: This could be `download_path=tmp_path / 'dot-briefcase' / 'support'` instead. Would you prefer that?

Asking based on discussion at https://github.com/beeware/briefcase/pull/309 , which I skimmed.

freakboy3742: If we need to make a change for Windows, it's likely going to be a change of the ".briefcase" part, not the "support" part; so sticking with "dot_briefcase_path" for now will be fine.

freakboy3742:
Of course, there's a PEP8 problem that I didn't see in my code review... :-)

PR #325, opened 2020-03-17
Add `line_length` so `black` operates properly

paulproteus:
`black` is a code formatter that I enjoy using. Its default configuration
meets briefcase's style needs except for the max line length. Therefore,
this commit adds the line-length configuration.

I know briefcase hasn't standardized on using `black` yet. I figure this
is OK to add if using `black` is optional.


paulproteus:
Per discussion with @freakboy3742 , `black`'s output _should_ already be permitted by our code style, so I'll close this for now, and discuss more in the case that it doesn't.

PR #342, opened 2020-04-08
Add publishing & signing docs; generate Android App Bundle

### LINKED ISSUE #438: Switch to using AAB format for Android builds.
This allows the Play Store to dynamically build the smallest APK appropriate to a device installing an app.

paulproteus:
Fixes #438.

I'd love your feedback on this. I know there's more changes I need to make, but I wanted to get a draft out ASAP.

Some questions for you:

- Is it organized in the way you'd expect?

- Did I miss any TODOs? :)

- Where do we ask the user to store their Android app signing keys? We could use `~/.briefcase/keys/` by default in these instructions. I get the impression there is no true consistent default keystore path in Android; the official docs use an arbitrary-seeming path: https://developer.android.com/studio/publish/app-signing#generate-key . Whatever path we decide on, I'm going to have to make sure these docs use it consistently, so I left it a TODO until we chat.

I'll proofread it at least once before resubmitting to you.


paulproteus:
filename:

```
~/.briefcase/android-app-keys.jks
```

The `alias` should be the reverse dotted name of the app.

paulproteus:
I'll try to migrate this to use Google's hosted signing thing, which means our key would become an "uploading" key in their terminology.

docs/how-to/publishing/android.rst
```diff
@@ -0,0 +1,132 @@
...
+same as your app's Formal Name in the BeeWare ecosystem.
+
+This will take you to **Store Listing** section of your app. You will need to
+answer a variety of questions. Take your time to answer them well.
+
+You will need to upload an icon. For briefcase apps, an icon is typically specified in
+your app's ``pyproject.toml`` file. Look for a line starting with ``icon = ``.
+The Google Play Store will require a variety of other information, including at least
+one screenshot of your app and your assessment of likelihood that children will want
+to use the app.
```
freakboy3742, line 85: Icon needs to be 512x512, which isn't a size required during setup.

Minium 2 screenshots are required; they must be at least 320px, no more than 3480px, and can't have an aspect ratio > 2:1.

Feature graphic (1024x500px) is also required.

Other details like a promo graphic, tv banner, promo vidio, and daydream 360 degree stereoscopic image can be provided, but aren't required.

docs/how-to/publishing/android.rst
```diff
@@ -0,0 +1,132 @@
...
+
+Configure app signing
++++++++++++++++++++++
+
+Click **App Signing** in the left navigation. The Google Play Store offers an app signing
+feature where they maintain the signing key for your app. For simplicity, you will
+click **Opt Out** to opt out of that feature.
```
freakboy3742, line 92: This appears out of order; I couldn't manage keys until uploading an APK.

docs/how-to/publishing/android.rst
```diff
@@ -0,0 +1,132 @@
...
+.. admonition:: Letting Google manage and protect your app signing key
+
+  The Google Play store's feature to manage app signing keys offers a way to
+  protect your app if your signing key is ever lost or compromised. You can read
+  more about it in `Google's documentation. <https://support.google.com/googleplay/android-developer/answer/7384423?hl=en>`_
+  For simplicity, this documentation relies on you to manage and protect your
+  app's signing key.
+
+  You can enable this feature later if you are willing to give Google the private
+  key corresponding to this app.
```
freakboy3742, line 103: From going through the release process, it's pretty clear that Google wants you to use google-managed keysigning. I'd argue it's better for these instructions to reflect that.

docs/how-to/publishing/android.rst
```diff
@@ -0,0 +1,132 @@
...
+allow you to choose the four APK files under the TODO directory.
+
+Scroll to **What's new in this release?**. If this is your first upload of the app,
+write the words, "Initial upload."
+
+Finally, click **Review.**
+
+This will prompt a variety of questions from the Google Play Store about the content
+of your app, including the absence/presence of advertising, its appropriateness for
+different age groups, and any embedded commercial aspects.
```
freakboy3742, line 126: I also had to save a draft, then visit the "Content Rating", "App Content" and "Pricing and Distribution"  tabs to fill out all sorts of details. There's a series of grayed out arrows; until they're all green, you can't publish. There wasn't a "just answer the questions" workflow.

It also raised a warning about being a non-optimized bundle. Short term, we need to flag this isn't a problem; long term, we should modify the publish command to produce multiple APKs.

docs/how-to/publishing/android.rst
```diff
@@ -0,0 +1,132 @@
...
+
+To distribute your app on the Google Play Store, you need to visit
+https://play.google.com/apps/publish/ and create an account.
+
+Once you've done that, click **Create Application**. Choose a language and
+write a brief app title, up to 50 characters. This should probably be the
+same as your app's Formal Name in the BeeWare ecosystem.
+
+This will take you to **Store Listing** section of your app. You will need to
+answer a variety of questions. Take your time to answer them well.
```
freakboy3742, line 79: "Your app metadata may be helpful here."

paulproteus:
Hi @freakboy3742 -- good feedback. I'll plan to take a fresh look and submit revisions tomorrow daytime my time, perhaps sooner.

docs/how-to/publishing/android.rst
```diff
@@ -0,0 +1,132 @@
...
+
+Creating a listing
+++++++++++++++++++
+
+To distribute your app on the Google Play Store, you need to visit
+https://play.google.com/apps/publish/ and create an account.
+
+Once you've done that, click **Create Application**. Choose a language and
+write a brief app title, up to 50 characters. This should probably be the
+same as your app's Formal Name in the BeeWare ecosystem.
```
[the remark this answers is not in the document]
paulproteus: I was really confused by this difference. The reason is that I already had one Google Play Store Android app in my Google Play Store console, whereas you didn't. Presumably our users will be more like you than like me in this regard. I'll be fixing that now :)

docs/how-to/publishing/android.rst
```diff
@@ -0,0 +1,132 @@
...
+  Before you sign the APK files, you need to :doc:`create a code signing
+  identity<../code-signing/android>`.
+
+Since you already have a code signing identity, we run the two Android SDK
+commands required to sign the APK files. First, we ensure the APK files are
+aligned to 32-bit boundaries, per the Android requirements.
+
+::
+
+  $ zipalign -p 4 -f filename.apk
```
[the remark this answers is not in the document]
paulproteus: FWIW, the reason I wrote it this way is that zipalign is a very safe operation IMHO, so adding another file creates more confusion than it's worth.

Having said that, the zipalign process is no longer needed with the Android App Bundle approach, so whatever. :)

docs/how-to/code-signing/android.rst
```diff
@@ -0,0 +1,38 @@
...
+where you retain full control over the signing key. There's another way, where you still
+sign the app this way, but the Google Play Store removes that signature and signs it with
+its own key. That can have its advantages, but is out of scope for this documentation.
+
+Generate a key
+--------------
+
+Run this command::
+
+  $ keytool -genkey -v -keystore my-release-key.jks -keyalg RSA -keysize 2048 -validity 10000 -alias my-alias
```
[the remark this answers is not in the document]
paulproteus: Fixed now, I believe.

docs/how-to/publishing/android.rst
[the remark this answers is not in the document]
paulproteus: Fixed now, I believe.

docs/how-to/publishing/android.rst
```diff
@@ -0,0 +1,132 @@
...
+    - Test it.
+    - Decide with Russell where we want to store these keys.
+
+Adding the app to the Google Play store
+---------------------------------------
+
+Adding the app requires a few steps. You need to have a Google Account and
+create a listing for your app. You will need to choose your app's signing
+configuration in the store. Finally, you will create a new release of the app,
+at which point you will upload the APK files that comprise your app.
```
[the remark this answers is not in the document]
paulproteus: Fixed now, I believe.

docs/how-to/publishing/android.rst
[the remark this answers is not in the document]
paulproteus: Fixed now, I believe.

docs/how-to/publishing/android.rst
paulproteus: Fixed now, I believe.

docs/how-to/publishing/android.rst
paulproteus: Fixed now, I believe.

docs/how-to/publishing/android.rst
paulproteus: +1; fixed now, I believe.

docs/how-to/publishing/android.rst
paulproteus: Fixed now. (Also, the Android App Bundle saves us from the non-optimized bundle problem! Yay.)

docs/how-to/publishing/android.rst
paulproteus: Fixed now

paulproteus:
Hi @freakboy3742 -- I've tested this pretty well and am happy for another review pass and/or merge if you like. Note that the switch to Android App Bundle format results in a change in the code **and** docs.

docs/how-to/publishing/android.rst
```diff
@@ -0,0 +1,192 @@
...
+in **upload-key-helloworld.jks** under ``.android`` within your home
+folder.
+
+.. tabs::
+
+  .. group-tab:: macOS
+
+    .. code-block:: bash
+
+      $ ~/.briefcase/tools/java/Contents/Home/bin/jarsigner -verbose -sigalg SHA1withRSA -digestalg SHA1 -keystore ~/.android/upload-key-helloworld.jks android/*/app/build/outputs/bundle/release/app-release.aab upload-key -storepass android
```
freakboy3742, line 63: To confirm - the '*' here is trying to capturie "all .aab files in the android folder" without explicitly naming the app, right? The possible problem here is that a briefcase project can contain multiple apps. 

Also - does this sign-in-place, or produce a new aab file somewhere?

docs/how-to/code-signing/android.rst
```diff
@@ -0,0 +1,77 @@
...
+
+
+This creates a 2048-bit key and stores it in a Java keystore located in the
+``.android`` folder within your home folder. Since the key's purpose is to be
+used as an upload key for the Google Play store, we call the key "upload-key".
+
+We use a password of **android**. This is the `default password for common
+Android keystores <https://developers.google.com/android/guides/client-auth>`__.
+You can change the password if you like. It is more important to limit who
+has access to the keystore file than to change the password.
```
freakboy3742, line 74: Any reference on “use the default password" as best practice? The general philosophy makes sense, but if we're going to recommend using a default password, I want to be sure we're giving good advice.

docs/how-to/publishing/android.rst
paulproteus: > To confirm - the '*' here is trying to capturie "all .aab files in the android folder" without explicitly naming the app, right? The possible problem here is that a briefcase project can contain multiple apps.

True -- I'm trying to capture "all .aab files in the android folder". And true -- this is a bad fit given the possibility of a briefcase project containing multiple apps. I hadn't considered that.

I'll fix that in a follow-up push in a sec. 

> Also - does this sign-in-place, or produce a new aab file somewhere?

In-place.

paulproteus: 96366039dd85175f7cb2ac6e352a8f4381049c4d

docs/how-to/code-signing/android.rst
paulproteus: Unfortunately, no -- this page here talks about creating and confirming a secure password for your keystore -- https://developer.android.com/studio/publish/app-signing -- followed by a different password for the key.

This is IMHO near-total insanity.

A Java keystore does nearly nothing, and the Java keystore password is approximately useless. See https://gist.github.com/zach-klippenstein/4631307 for demo code that shows how to change the password on a keystore without knowing the old password.

In our case, I'm using the PKCS #12 format for keystores, which may or may not have that specific problem, but it adds no value to the key password if the keystore only contains one key, which is how we suggest doing it.

As for the key's password... this is a classic ambiguity in current crypto recommendations. I'll spare you my standard rant for now, but I'm also happy to offer it to you if useful. My suggestion would be that the password adds no value if you're not sharing the key, and if you are sharing the key, do you really think the password you picked is solid enough to protect the key? If so, then you surely need to put it somewhere safe, because you surely can't remember it... in which case you have now opened a massive can of worms. Oops, I gave you 50% of the standard rant anyway.

PR #346, opened 2020-04-11
Android SDK root

freakboy3742:
* Refactors Android SDK support to move all the verification and download code into the integrations layer.
* Introduces an "AndroidSDK" object to provide an entry point for SDK path management.
* Allows an end-user to use their own SDK by specifying ANDROID_SDK_ROOT.


freakboy3742:
Closing this in favor of #347 - it's the same PR, but directly on beeware/beeware, which allows for PR chaining.

PR #347, opened 2020-04-12
Fixed #338 - Refactor AndroidSDK handling

### LINKED ISSUE #338: Autodetect pre-existing Android SDK installation
The Android backend currently automates the download of the Android SDK as part of the install process. 

However, it is possible that the user may already have Android Studio, or a working Android SDK install. 

If the user:
1) has defined `ANDROID_SDK_HOME` as an environment variable, or
2) has Android Studio installed in it's default location

then that install should be used, instead of downloading a new install.

The tests that `sdkmanager` exists, and that the right SDK tools have been installed, should be preserved.

freakboy3742:
* Refactors Android SDK support to move all the verification and download code into the integrations layer.
* Introduces an "AndroidSDK" object to provide an entry point for SDK path management.
* Allows an end-user to use their own SDK by specifying ANDROID_SDK_ROOT.


src/briefcase/integrations/android_sdk.py
```diff
@@ -0,0 +1,370 @@
...
+    sdk_root = command.os.environ.get('ANDROID_SDK_ROOT')
+    if sdk_root:
+        sdk = AndroidSDK(command=command, root_path=Path(sdk_root))
+
+        if sdk.exists():
+            # Ensure licenses have been accepted
+            sdk.verify_license()
+            return sdk
+        else:
+            print("""
```
paulproteus, line 35: I'm -0 on printing something out every time, but it's OK. I get why you're doing it. I don't know that I really have a better idea, but it seems a little sadly noisy.

src/briefcase/integrations/android_sdk.py
```diff
@@ -0,0 +1,370 @@
...
+*************************************************************************
+** WARNING: ANDROID_SDK_ROOT does not point to an Android SDK          **
+*************************************************************************
+
+    The location pointed to by the ANDROID_SDK_ROOT environment variable:
+
+     {sdk_root}
```
paulproteus, line 42: 🎉

tests/integrations/android_sdk/AndroidSDK/test_properties.py
```diff
@@ -0,0 +1,55 @@
+from pathlib import Path
```
paulproteus, line 1: I don't love the `test_properties` name. I don't have any better ideas. Feel free to leave it as-is :)

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+
+    # Mock-out the `sys` module so we can mock out the Python version in some tests.
+    command.sys = MagicMock()
+
+    # Use the `tmp_path` in `dot_briefcase_path` to ensure tests don't interfere
+    # with each other.
+    command.dot_briefcase_path = tmp_path / ".briefcase"
+
+    # Make the `os` module and `host_os` live.
+    command.os = os
```
paulproteus, line 25: I'd prefer not to provide the entire `os` module unless you need the whole module, but it's OK.

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+        license_dir.mkdir(parents=True)
+        (license_dir / "android-sdk-license").touch()
+
+    return _side_effect
+
+
+@pytest.mark.parametrize("host_os", ("ArbitraryNotWindows", "Windows"))
+def test_succeeds_immediately_in_happy_path(mock_command, host_os, tmp_path):
+    "If verify_android_sdk is invoked on a valid path, it does nothing."
+    # `sdkmanager` exists and has the right permissions, and
```
paulproteus, line 51: I think you're missing the word "If" on this line?

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+        license_dir = root_path / "licenses"
+        license_dir.mkdir(parents=True)
+        (license_dir / "android-sdk-license").touch()
+
+    return _side_effect
+
+
+@pytest.mark.parametrize("host_os", ("ArbitraryNotWindows", "Windows"))
+def test_succeeds_immediately_in_happy_path(mock_command, host_os, tmp_path):
+    "If verify_android_sdk is invoked on a valid path, it does nothing."
```
paulproteus, line 50: I'd prefer another phrase other than "valid path".

Perhaps:

```
"If verify_android_sdk() is invoked on a path with an Android SDK, it does nothing."
```

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+    sdk = verify_android_sdk(mock_command)
+
+    # No calls to download, run or unpack anything.
+    mock_command.download_url.assert_not_called()
+    mock_command.subprocess.run.assert_not_called()
+    mock_command.subprocess.check_output.assert_not_called()
+    mock_command.shutil.unpack_archive.assert_not_called()
+
+    # The returned SDK has the expected root path.
+    assert str(sdk.root_path) == str(root_path)
```
paulproteus, line 113: Line 113 asserts `str(x) == str(y)`, but line 135 asserts on the paths themselves. I propose you stick to asserting the paths are equal.

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+def test_user_provided_sdk(mock_command, tmp_path):
+    "If the user specifies a valid ANDROID_SDK_ROOT, it is used"
+    # Create `sdkmanager` and the license file.
+    root_path = tmp_path / "other_sdk"
+    tools_bin = root_path / "tools" / "bin"
+    tools_bin.mkdir(parents=True, mode=0o755)
+    (tools_bin / "sdkmanager").touch(mode=0o755)
+    licenses = root_path / "licenses"
+    licenses.mkdir(parents=True, mode=0o755)
+    (licenses / "android-sdk-license").touch()
```
paulproteus, line 96: I'd prefer you mock-out the license check; otherwise, this is a lot of boilerplate.

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+    mock_command.shutil.unpack_archive.assert_not_called()
+
+    # The returned SDK has the expected root path.
+    assert sdk.root_path == root_path
+
+
+def test_user_provided_sdk(mock_command, tmp_path):
+    "If the user specifies a valid ANDROID_SDK_ROOT, it is used"
+    # Create `sdkmanager` and the license file.
+    root_path = tmp_path / "other_sdk"
```
paulproteus, line 90: I'd prefer a variable name like `existing_sdk_path` to make it clearer.

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+    # The returned SDK has the expected root path.
+    assert str(sdk.root_path) == str(root_path)
+
+
+def test_invalid_user_provided_sdk(mock_command, tmp_path):
+    "If the user specifies an invalid ANDROID_SDK_ROOT, it is ignored"
+
+    # Create `sdkmanager` and the license file
+    # for the *briefcase* SDK.
+    root_path = tmp_path / ".briefcase" / "tools" / "android_sdk"
```
paulproteus, line 121: Similar comments as previous function.

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+def test_download_sdk(mock_command, tmp_path, host_os):
+    "If an SDK is not available, one will be downloaded"
+    root_path = tmp_path / ".briefcase" / "tools" / "android_sdk"
+
+    # Mock-out `host_os` so we only do our permission check on non-Windows.
+    mock_command.host_os = host_os
+
+    # The download will produce a cached file
+    cache_file = MagicMock()
+    cache_file.__str__.return_value = "/path/to/download.zip"
```
paulproteus, line 157: I forgot again why you have to mock-out `__str__`. Can you leave a comment about why? Can you use a `Path` object?

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+    )
+    mock_command.shutil.unpack_archive.assert_called_once_with(
+        "/path/to/download.zip",
+        extract_dir=str(root_path)
+    )
+
+    # The cached file will be deleeted
+    cache_file.unlink.assert_called_once_with()
+
+    # On non-Windows, ensure the mocked binary was made executabley
```
paulproteus, line 187: `executabley` is like `chocolatey`, which is very Windows-y ;) https://chocolatey.org/

Anyway, you probably want to delete the letter `y`. :)

**Also** I'd say: "ensure the **unpacked** binary" rather than ensure the "**mocked** binary"

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+    mock_command.shutil.unpack_archive.assert_not_called()
+
+    # The returned SDK has the expected root path.
+    assert sdk.root_path == root_path
+
+
+@pytest.mark.parametrize("host_os", ("ArbitraryNotWindows", "Windows"))
+def test_download_sdk(mock_command, tmp_path, host_os):
+    "If an SDK is not available, one will be downloaded"
+    root_path = tmp_path / ".briefcase" / "tools" / "android_sdk"
```
paulproteus, line 150: I'd prefer if this were called `expected_briefcase_android_sdk_path`.

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+
+    # Create non-executable `sdkmanager`.
+    root_path = tmp_path / ".briefcase" / "tools" / "android_sdk"
+    (root_path / "tools" / "bin").mkdir(parents=True)
+    (root_path / "tools" / "bin" / "sdkmanager").touch(mode=0o644)
+
+    # The download will produce a cached file
+    cache_file = MagicMock()
+    cache_file.__str__.return_value = "/path/to/download.zip"
```
paulproteus, line 213: Same remark as elsewhere; I continue to be confused about mocking-out `__str__`.

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+    mock_command.download_url.assert_called_once_with(
+        url="https://dl.google.com/android/repository/sdk-tools-unknown-4333796.zip",
+        download_path=mock_command.dot_briefcase_path / "tools",
+    )
+    mock_command.shutil.unpack_archive.assert_called_once_with(
+        "/path/to/download.zip",
+        extract_dir=str(root_path)
+    )
+
+    # The cached file will be deleeted
```
paulproteus, line 232: `deleeted` has one too many `e` IMHO

tests/integrations/android_sdk/test_verify_android_sdk.py
```diff
@@ -0,0 +1,281 @@
...
+    assert mock_command.shutil.unpack_archive.call_count == 0
+
+
+def test_detects_bad_zipfile(mock_command, tmp_path):
+    "If the ZIP file is corrupted, an error is raised."
+    root_path = tmp_path / ".briefcase" / "tools" / "android_sdk"
+
+    # The download will produce a cached file
+    cache_file = MagicMock()
+    cache_file.__str__.return_value = "/path/to/download.zip"
```
paulproteus, line 264: Same remark as elsewhere about weird mocking-out `__str__`.

Might make sense to move this to a pytest fixture, if we you have to keep doing this weird thing.

src/briefcase/integrations/android_sdk.py
freakboy3742: I agree that being noisy isn't generally a good thing; but in this case, it will only be noisy if (a) you have `ANDROID_SDK_ROOT` set and (b) it's not actually pointing at an Android SDK. If that's happening, I don't feel especially bad about yelling at the user to fix their setup.

tests/integrations/android_sdk/AndroidSDK/test_properties.py
freakboy3742: Yeah - I'm not 100% wild about it either; but it's a convention that I've used elsewhere in the codebase rather than have dozens of 1-test files.

tests/integrations/android_sdk/test_verify_android_sdk.py
freakboy3742: What's your reasoning here? This is essentially only restoring "default behavior"; i.e., that os exists and is usable. Generally, my preference is that is mocks *can* be avoided, they should be. Admittedly they're unavoidable in a lot of circumstances, but when they *can* be avoided, we might as well - that's one less discrepancy between production and test.

tests/integrations/android_sdk/test_verify_android_sdk.py
freakboy3742: Yeah... so... I completely agree. Python 3.5 doesn't :-) 
```
==============================================================================
FAIL: tests/integrations/android_sdk/test_verify_android_sdk.py::test_user_provided_sdk
  File "/Users/rkm/projects/beeware/briefcase/tests/integrations/android_sdk/test_verify_android_sdk.py", line 113, in test_user_provided_sdk
    assert sdk.root_path == root_path
AssertionError: assert PosixPath('/private/var/folders/9p/nzv40l_974ddhp1hfrccbpw80000gn/T/pytest-of-rkm/pytest-2/test_user_provided_sdk0/other_sdk') == PosixPath('/private/var/folders/9p/nzv40l_974ddhp1hfrccbpw80000gn/T/pytest-of-rkm/pytest-2/test_user_provided_sdk0/other_sdk')
 +  where PosixPath('/private/var/folders/9p/nzv40l_974ddhp1hfrccbpw80000gn/T/pytest-of-rkm/pytest-2/test_user_provided_sdk0/other_sdk') = <briefcase.integrations.android_sdk.AndroidSDK object at 0x103dce7b8>.root_path

```
That is:
```
/private/var/folders/9p/nzv40l_974ddhp1hfrccbpw80000gn/T/pytest-of-rkm/pytest-2/test_user_provided_sdk0/other_sdk
```
!=
```
/private/var/folders/9p/nzv40l_974ddhp1hfrccbpw80000gn/T/pytest-of-rkm/pytest-2/test_user_provided_sdk0/other_sdk
```
This exact test passes under Python 3.6+. I can only assume it's some obscure bug about comparison of paths that is causing the failure.

I'll add a FIXME that can be cleaned up once we drop Python 3.5 support (along with migrating to f-strings instead of `.format()`...)

tests/integrations/android_sdk/test_verify_android_sdk.py
freakboy3742: It's a little difficult to mock, as we're in the process of creating the object that will have the license check invoked on it; but I *can* use the pre-existing side effect method that sets up the signed license.

tests/integrations/android_sdk/test_verify_android_sdk.py
freakboy3742: We need to be able to call "unlink" on the file, so we either need to create an actual file so it can be unlinked, or a mock object that we can verify. However, because it's mocking a Path object, it's being used as an argument to some of the process calls, where we call str() on the path because Python 3.5 Pathlib support isn't consistent; and str() on a mock doesn't yield a useful result.

We *could* use a Path object - but we'd then also need to create the physical file that could be deleted.

tests/integrations/android_sdk/test_verify_android_sdk.py
freakboy3742: Same explanation as above :-)

tests/integrations/android_sdk/test_verify_android_sdk.py
freakboy3742: Ah, but I'm yeeting this file... :-)

tests/integrations/android_sdk/test_verify_android_sdk.py
freakboy3742: I'll keep that in mind.

PR #348, opened 2020-04-12
Fixes #340 - Add ability to select from available Android devices

### LINKED ISSUE #340: Improve handling of Android emulators and devices
The Android backend currently requires that a user:
* Manage the creation of an emulator
* Manage starting the emulator
* Specify a device ID for all run commands.

We should:
* Autodetect any emulators that already exist, or devices that are connected and enabled for debugging
* Provide an option to create a new emulator, including downloading chrome for the emulator "window"
* Manage the process of starting an emulator if it isn't already running.
* Enable specification of device by "name" (or some user-friendly handle), rather than device ID

freakboy3742:
This PR builds on the foundation laid down by #347.

This user provides the user with a list of available devices if they don't specify the `-d` option to `android run`.

The device list contains:
* All physical devices (including unauthorized devices)
* All emulators (running and idle)
* A "create new emulator" option.

The following behavioral changes have been made:
 * The semantics of `run android -d` have been modified to accept device IDs (in both the hexstring physical device names and `emulator-5554` format), as well as the `@robotFriend` format for specifying emulators.

* The value provided to `-d` is validated. If the provided value isn't a valid ID or AVD, an error is raised.

* If the provided device ID points at a device that hasn't been authorized for developer activity, an error is raised that provides a link to the official Android documentation on how to enable developer mode.

* If the emulator is not running, an error is raised providing details on how to start the device. Automating this is left as future work.

* If the user selects a completely new device, an error is raised with basic instructions on how to create a new emulator. Automating this is left as future work.


src/briefcase/integrations/android_sdk.py
```diff
@@ -5,11 +5,36 @@
...
 
 DEVICE_NOT_FOUND = re.compile(r"^error: device '[^']*' not found")
 
 
+class AndroidDeviceNotAuthorized(BriefcaseCommandError):
+    def __init__(self, device):
+        self.device = device
+        super().__init__(
+            """
+The device you have selected ({device}) has not had Developer Options
```
paulproteus, line 23: Pedantic: Google doesn't capitalize Developer Options like this.

Useful feedback: it's **USB debugging** that the user needs to enable. It's good IMHO to change the string to say "USB debugging".

src/briefcase/integrations/android_sdk.py
```diff
@@ -63,16 +91,12 @@ def verify_android_sdk(command):
...
+            url=sdk.sdk_url, download_path=command.dot_briefcase_path / "tools",
         )
     except requests_exceptions.ConnectionError:
         raise NetworkFailure("download Android SDK")
     try:
-        command.shutil.unpack_archive(
-            str(sdk_zip_path),
-            extract_dir=str(sdk.root_path)
-        )
+        command.shutil.unpack_archive(str(sdk_zip_path), extract_dir=str(sdk.root_path))
```
paulproteus, line 99: Pedantic: don't mix your code style pull requests with your pull requests that add features :P 

It's fine; I'm just making fun. :)

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+                universal_newlines=True,
+                stderr=subprocess.STDOUT,
+            ).strip()
+
+            # Process the output of `adb devices -l`.
+            # The first line is header information.
+            # Each subsequent line is a single device descriptor.
+            devices = {}
+            for line in output.split("\n")[1:]:
+                parts = re.sub(r"\s+", " ", line).split(" ")
```
paulproteus, line 296: I believe that `\t` is a better character to split on for name/rest, but it's OK.

I'm imagining, like:

```
name, property_string = line.split('\t', 1)
properties = property_string.split()
```

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+        """
+        Select a device to be the target for actions
 
-  * a ~12-16 character alphanumeric string (for a physical device); or
-  * a value like `emulator-5554` (for an emulator).
+        Interrogates the system to get the list of available devices
 
-If you do not see any devices, you can create and start an emulator by running:
+        If the user has specified a device at the command line, that device
+        will be validated, and then automatically selected
```
paulproteus, line 326: Pedantic: add a `.` at the end of this sentence.

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+        # Device choices is the full lookup list. Devices can be looked up
+        # by any valid key - ID *or* AVD.
+        device_choices = {}
+
+        # Iterate over all the running devices.
+        # If the device has an AVD, use ADB to get the emulator AVD name.
+        # If it is a physical device, use the device name.
+        # Keep a log of all running AVDs
+        running_avds = {}
+        for d, details in sorted(running_devices.items(), key=lambda d: d[1]["name"]):
```
paulproteus, line 352: Pedantic: Why sorted here?

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+                    avd=avd, name=name,
+                )
+                choices.append((d, full_name))
+
+                # Save the AVD as a device detail.
+                details["avd"] = avd
+
+                # Device can be looked up by device ID or AVD
+                device_choices[d] = full_name
+                device_choices["@" + avd] = full_name
```
paulproteus, line 368: Why prefix with `@`? I'd prefer to type `robotFriend` than `@robotFriend`, but I'd be OK with `@robotFriend` I guess.

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+                # Provided device_or_id isn't a valid device identifier.
+                if device_or_avd.startswith("@"):
+                    id_type = "emulator AVD"
+                else:
+                    id_type = "device ID"
+                raise InvalidDeviceError(id_type, device_or_avd)
+
+        # We weren't given a device/AVD; we have to select from the list.
+        # If we're selecting from a list, there's always one last choice
+        choices.append((None, "Create a new Android emulator"))
```
paulproteus, line 420: 🎉

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+                        # it must be a non-running emulator.
+                        return None, name, avd
+                else:
+                    # Specifier is a direct device ID
+                    avd = None
+                    device = device_or_avd
+
+                details = running_devices[device]
+                avd = details.get("avd")
+                if details["authorized"]:
```
paulproteus, line 403: Since `adb` only ever prints `unauthorized`, I'd prefer that `details` store `unauthorized=True` rather than `authorized=False`, to avoid making up our own nomenclature. That's just personal preference.

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+        # Choices is an ordered list of options that can be shown to the user.
+        # Each device should appear only once, and be keyed by AVD only if
+        # a device ID isn't available.
+        choices = []
+        # Device choices is the full lookup list. Devices can be looked up
+        # by any valid key - ID *or* AVD.
+        device_choices = {}
+
+        # Iterate over all the running devices.
+        # If the device has an AVD, use ADB to get the emulator AVD name.
```
paulproteus, line 348: You wrote "has an AVD", but I would say "is an AVD".

Perhaps instead:

```
If the device is an Android virtual device, use ADB to get the emulator AVD name.
```

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+                choices.append(("@" + avd, name))
+                device_choices["@" + avd] = name
+
+        # If a device or AVD has been provided, check it against the available
+        # device list.
+        if device_or_avd:
+            try:
+                name = device_choices[device_or_avd]
+
+                if device_or_avd.startswith("@"):
```
paulproteus, line 387: I still want to see `robotFriend` not `@robotFriend`, personally.

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+        choices.append((None, "Create a new Android emulator"))
+
+        # Show the choices to the user.
+        print()
+        print("Select device:")
+        print()
+        choice = select_option(choices, input=self.command.input)
+
+        # Proces the user's choice
+        if choice is None:
```
paulproteus, line 429: Dumb question: is this the only way to get `None` out of `select_option()`? Does it have a way to e.g. cancel, which would return `None`? I just fear using `None` as a sentinel value if it might be actually returned.

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+
+                # Return the device ID and name.
+                device = choice
+                name = device_choices[choice]
+                avd = details.get("avd")
+            except KeyError:
+                raise InvalidDeviceError("device ID", choice)
+
+        if avd:
+            print()
```
paulproteus, line 456: The `if` and `elif` are so similar that I wonder if you can unify them. Not a showstopper.

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+            print("    briefcase run android -d {device}".format(device=device))
+
+        return device, name, avd
+
+    def create_emulator(self):
+        """Create a new Android emulator.
+
+        """
+        print()
+        name = self.command.input("Emulator name: ")
```
paulproteus, line 473: I would personally suggest we always use the name "briefcasephone" or something, or at least that we provide that as a default. I believe many people will be surprised they have to choose a name here.

src/briefcase/integrations/android_sdk.py
```diff
@@ -212,61 +248,305 @@ def verify_emulator(self):
...
+        """
+        try:
+            output = self.run("emu", "avd", "name")
+            return output.split("\n")[0]
+        except subprocess.CalledProcessError as e:
+            # Status code 1 is a normal "it's not an emulator" error response
+            if e.returncode == 1:
+                return None
+            else:
+                raise BriefcaseCommandError(
```
paulproteus, line 544: Because we call this from other helper functions, in a way that the user isn't really aware of, I'd prefer that we not crash the `briefcase` process in this case. Not a showstopper.

src/briefcase/integrations/android_sdk.py
```diff
@@ -314,7 +586,14 @@ def install_apk(self, apk_path):
 
         Returns `None` on success; raises an exception on failure.
         """
-        return self.run("install", str(apk_path))
+        try:
+            self.run("install", str(apk_path))
+        except subprocess.CalledProcessError:
+            raise BriefcaseCommandError(
```
paulproteus, line 592: Have you seen this happen? If so, great. If not, then maybe we're not catching the error condition accurately.

src/briefcase/integrations/android_sdk.py
```diff
@@ -327,7 +606,14 @@ def force_stop_app(self, package):
         # In my testing, `force-stop` exits with status code 0 (success) so long
         # as you pass a package name, even if the package does not exist, or the
         # package is not running.
-        self.run("shell", "am", "force-stop", package)
+        try:
+            self.run("shell", "am", "force-stop", package)
+        except subprocess.CalledProcessError:
+            raise BriefcaseCommandError(
```
paulproteus, line 612: Given the comments above about `force-stop` exiting with 0 in all cases, I'm curious if you triggered this error. If not, then maybe we're not catching the error condition accurately.

The only condition in which I could imagine this happening is if the device has gone away. In that case, I'd prefer to show an error message about the device not being connected.

src/briefcase/integrations/android_sdk.py
```diff
@@ -342,29 +628,40 @@ def start_app(self, package, activity):
...
     Activity class not found while starting app.
 
     `adb` output:
 
-    {output}""".format(output=output))
+    {output}""".format(
+                        output=output
+                    )
+                )
+        except subprocess.CalledProcessError:
```
paulproteus, line 662: Have you seen this error occur? If not, I worry we're not catching any realistic error condition.

The way `BriefcaseCommandError` works is it squelches any other error reporting, so IMHO the `except` clause here makes it harder to debug any unknown-unknown-type issue.

src/briefcase/platforms/iOS/xcode.py
```diff
@@ -12,7 +12,7 @@
 )
 from briefcase.config import BaseConfig
 from briefcase.console import select_option
-from briefcase.exceptions import BriefcaseCommandError
+from briefcase.exceptions import BriefcaseCommandError, InvalidDeviceError
```
paulproteus, line 15: Unrelated to this PR :P but OK :)

tests/console/test_select_option.py
```diff
@@ -22,6 +22,26 @@ def test_select_option():
     assert result == 'second'
 
 
+def test_select_option_list():
```
paulproteus, line 25: Per previous comment, I'd prefer we change `option_list` to always be a list of tuples.

src/briefcase/console.py
```diff
@@ -16,20 +16,24 @@ def select_option(options, input=input, prompt='> ', error="Invalid selection"):
...
     :param prompt: The prompt to display to the user.
     :param error: The error message to display when the user provides invalid
         input.
     :returns: The key corresponding to the user's chosen option.
     """
-    ordered = list(
-        sorted(
-            options.items(),
-            key=operator.itemgetter(1)
+    if isinstance(options, dict):
```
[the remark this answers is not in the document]
freakboy3742: I agree with the general motivation; in this case "dict or list of items" is a common pattern in Python libraries (e.g., requests HTTP header handling). It's mostly an artefact of pre-Python3.5 dictionary key ordering.

src/briefcase/integrations/android_sdk.py
freakboy3742: There's 2 actions required (the 7 presses on build number dance, and then enabling USB debugging) - so it's worth mentioning both. The capitalization note is a good point, though.

src/briefcase/integrations/android_sdk.py
freakboy3742: ppfffffft :-P

src/briefcase/integrations/android_sdk.py
freakboy3742: ....except it's not a tab - it's a bunch of spaces.  At least it is in all my testing...

src/briefcase/integrations/android_sdk.py
freakboy3742: Python3.5 test ordering. Python 3.5 doesn't guarantee reliable dictionary order, so sorting is the easiest way to make it's always processed in the same order.

src/briefcase/integrations/android_sdk.py
freakboy3742: Seems a bit bikeshed to me. It works both ways, and "not unauthorized" would be a double negative...

src/briefcase/integrations/android_sdk.py
freakboy3742: However, the @ prefix is (a) used by `emulator` as an identifier, and (b) allows a reliable differentiation between an AVD identifier and a device identifier. This is significant when we're using `-d`, because we need to make sure there's no collision in the two name types.

src/briefcase/integrations/android_sdk.py
freakboy3742: A fair point; but select_choice always returns one of the choices, or raises an exception.

src/briefcase/integrations/android_sdk.py
freakboy3742: Yeah - the catch is all the extra processing that needs to be done on AVDs.

src/briefcase/integrations/android_sdk.py
freakboy3742: A reasonable concern; but in this case, if you ask for an AVD of a non-existent device ID, something has gone *badly* wrong.

src/briefcase/integrations/android_sdk.py
freakboy3742: This specific case, no - adding this was part of the refactoring that made `ADB.run` re-raise the `CalledProcessError` for anything other than an invalid device ID. I can't think of any obvious way that this would manifest on install - but as a safety mechanism, it's here.

src/briefcase/integrations/android_sdk.py
freakboy3742: Again - it's a safety mechanism, combined with the refactoring of the exception handling on `ADB.run()`.

src/briefcase/integrations/android_sdk.py
freakboy3742: Agreed the "unknown unknown" error handling isn't great; but we *will* get an explicit error that says what ADB command failed, so the debugging path (get them to manually re-run the command in question) should be clear.

src/briefcase/platforms/iOS/xcode.py
freakboy3742: Tangentially related... I introduced an InvalidDeviceError, so I'm making sure it's used consistently.

tests/console/test_select_option.py
freakboy3742: ack, but respectfully disagree in this case.

src/briefcase/integrations/android_sdk.py
freakboy3742: Ah - but the eventual plan is that there will be the ability to select different device profiles, sizes, etc (There's a TODO in the branch to create the actual devices).

PR #349, opened 2020-04-12
Fixes #341 - Add ability to create new Android devices

### LINKED ISSUE #341: Download and use an Android device skin when creating an Android emulator
At the moment, the documentation printed by `briefcase run android` on how to create a virtual Android device results in an emulator without any "skin".

The skin provides helpful visual context to the user to feel that they're testing their app on a mobile device, not just a small computer screen. See e.g. this pic ([source](https://www.tutorialspoint.com/android/images/android_p.jpg))

:tmp:android_p

See discussion of `skins` at https://github.com/paulproteus/Python-Android-support-old for info on where to get skins.

After updating briefcase, also update the screenshots in the BeeWare tutorial.

freakboy3742:
Builds on #348. 

Implements the ability to create a new Android device, if requested.

* Asks the user for a device name
* Creates a new `pixel` device
* Obtains the pixel_3a skin
* Modifies the AVD configuration to add the skin, hardware keyboard, extra disk space, etc.

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1,6 +1,7 @@
 import re
 import shutil
 import subprocess
+import time
 from pathlib import Path
 
 from requests import exceptions as requests_exceptions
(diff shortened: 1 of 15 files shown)


src/briefcase/integrations/android_sdk.py
```diff
@@ -145,8 +152,12 @@ def avdmanager_path(self):
 
     @property
     def emulator_path(self):
-        emulator = "emulator.bat" if self.command.host_os == "Windows" else "emulator"
-        return self.root_path / "tools" / "bin" / emulator
+        emulator = "emulator.exe" if self.command.host_os == "Windows" else "emulator"
```
paulproteus, line 155: Good catch. I confirm on my Windows VM that it's `emulator.exe`. I suspect you can just do `emulator` not `emulator.exe`, though.

freakboy3742: I think you're right; but explicit vs implicit, etc...

PR #351, opened 2020-04-13
Add "override" and "no-override" flags to create command

saroad2:
This is done in order to avoid user input while running the "create" command


saroad2:
No problem. This requires a little bit more work and removing everything I did here, therefore I'm closing this pull request and starting over with a global "--no-input" flag in a different pull request.

Stay tuned.

PR #364, opened 2020-04-18
Add commands dynamically

saroad2:
The reason for this change is that new commands won't need to be added hard-coded to *cmd.py*. Instead, we add each command to a commands list that the function `parse_cmdline` would read the commands from.

### DIFF
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -3,7 +3,7 @@ import sys
 from pathlib import Path
 
 from briefcase import __version__
-from briefcase.commands import DevCommand, NewCommand
+from briefcase.commands import BaseCommand, DevCommand, NewCommand
 from briefcase.platforms import get_output_formats, get_platforms
 
 from .exceptions import (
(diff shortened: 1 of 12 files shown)


saroad2:
Hey, if you want to keep the order of the commands -- it's no longer dynamic. Therefore I suggested removing this pull request except for the tests I added in *test_cmdline.py*. Those tests pass before the other changes and add more coverage to the cmd.

For your consideration.

freakboy3742:
if I *really* think outside the box, I can think of a world where we assign "weights" to each fo the commands ("new" has weight 0, "dev" has weight 10, and so on), and we sort the commands by weight; a new command could then inject itself into the presentation order. However, I'm not sure that overhead is necessarily worth it; Hence why I suggested alphabetical ordering as an alternative. 

However - is there a specific motivation here? I mean, it's a nice enough patch, albeit one with a slight change to the user experience (possibly not one that matters, but still - one where the original design was somewhat intentional) - but what motivated the change? The only obvious benefit I can see is that it would allow for dynamically registering new commands. Is this a simple case of "it looked like something I could clean up", or is there a something deeper going on?

saroad2:
I think the idea of the weights should work fine, but it feels like unnecessary complexity.

The motivation for the change was that I started implementing the `briefcase project` command and saw that I need to add it manually to *cmdline.py*. it felt a little bit wrong to me. I thought there should be some kind of registry for the commands.

It is not that important and we probably shouldn't be bothering ourselves too much about it. I really suggest you accept only the extra tests if the order of the commands is meaningful.

saroad2:
This PR is no longer needed.

I think that eventually, the right thing to do is to move *Briefcase* to use [click](https://github.com/pallets/click). I'll try and investigate a little more about it in the next few weeks.

PR #409, opened 2020-05-25
allow ad-hoc code signing

tlambert03:
Curious whether you would allow this?  I know that ad-hoc signing has significant limitations, but it can be slightly more useful than no code-signing during app development.


tlambert03:
sounds good!  I'll work on that

tlambert03:
tests and docs added, ready for re-review at your convenience!

Issue #392, opened 2020-05-02
Force update of new app template on `briefcase new`


freakboy3742:
The first time you run `briefcase new`, the briefcase template is cloned. 

However, on subsequent runs, the briefcase template isn't updated. This means updates to the base briefcase app template aren't picked up for new projects.

Interestingly, the *platform* templates *are* updated every time they are used. 

The logic from platform templates needs to be applied to the base app template.

freakboy3742:
#414 was a good attempt at this; however, I think that PR went a little too far. 

I think it would make sense to target a much smaller feature set - only make the options of specific questions (the multi-choice ones) dynamic. This would make it possible to add a new supported platform (e.g., pyside6) without needing to make a Briefcase release - which was the real driving motivation of this feature request; however, the list of questions that are actually asked would be static.

rmartin16:
This was partially addressed in https://github.com/beeware/briefcase/pull/1524 with the ability for GUI bootstrap plugins to inject arbitrary prompts or additional context for cookiecutter.

PR #414, opened 2020-06-01
New app wizard should use template to generate questions for user input

GenevieveBuckley:
Closes https://github.com/beeware/briefcase/issues/392

This PR attempts to derive the questions & answer options for the `briefcase new` app wizard from the cookiecutter template, instead of hard coding these.

It does this by reading the `cookiecutter.json` file that is cached to disk into a python dictionary, then iterating through all the keys in order to build the app context. Validation functions and other special handling functions are also stored in dictionaries, and used if there is a match between keys.

### DIFF
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -14,6 +14,7 @@ from urllib.parse import urlparse
 import requests
 from cookiecutter.main import cookiecutter
 from cookiecutter.repository import is_repo_url
+from cookiecutter.vcs import clone
 
 from briefcase import __version__, integrations
 from briefcase.config import AppConfig, GlobalConfig, parse_config
@@ -482,6 +483,10 @@ class BaseCommand(ABC):
             # fall back to using the specified template directly.
             try:
                 cached_template = cookiecutter_cache_path(template)
+                if not cached_template.exists():
+                    print('Fetching the template...')
+                    clone(template, checkout=branch,
+                          clone_to_dir=cached_template.parent)
                 repo = self.git.Repo(cached_template)
                 try:
                     # Attempt to update the repository
--- a/src/briefcase/commands/new.py
+++ b/src/briefcase/commands/new.py
@@ -11,7 +11,7 @@ import json
 from briefcase.config import PEP508_NAME_RE
 from briefcase.exceptions import NetworkFailure
 
-from .base import BaseCommand, BriefcaseCommandError
+from .base import BaseCommand, BriefcaseCommandError, cookiecutter_cache_path
 from .create import InvalidTemplateRepository
 
 VALID_BUNDLE_RE = re.compile(r'[a-zA-Z0-9-]+(\.[a-zA-Z0-9-]+)+$')
@@ -365,8 +365,9 @@ Select one of the following:
 
         :returns:
             Default dictionary of keys/values for cookiecutter,
-            Optional dictionary 
+            plus dictionary of extra explanation for cookiecutter parameters.
         """
+        cached_template = cookiecutter_cache_path(cached_template)
         # Make a dictionary of default cookicutter values
         with open(Path(cached_template) / "cookiecutter.json") as f:
             defaults = json.loads(f.read())
--- a/src/briefcase/commands/new.py
+++ b/src/briefcase/commands/new.py
@@ -1,6 +1,7 @@
 import re
 import subprocess
 from email.utils import parseaddr
+from pathlib import Path
 from typing import Optional
 from urllib.parse import urlparse
 
@@ -121,10 +122,21 @@ class NewCommand(BaseCommand):
         """
         Construct a valid module name from an app name.
 
-        :param formal_name: The app name
+        :param app_name: The app name
         :returns: The app's module name.
         """
-        module_name = app_name.replace('-', '_')
+        module_name = app_name.replace("-", "_")
+        return module_name
+
+    def make_module_from_formal_name(self, formal_name):
+        """
+        Construct a valid module name from app formal name.
+
+        :param formal_name: The formal application name
+        :returns: The app's module name.
+        """
+        app_name = self.make_app_name(formal_name)
+        module_name = self.make_module_name(app_name)
         return module_name
 
     def validate_bundle(self, candidate):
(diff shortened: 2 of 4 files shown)


GenevieveBuckley:
I'm not familiar with towncrier, so I'm not exactly sure why that CI check is failing.

GenevieveBuckley:
One issue I'm having trouble with now is `update_cookiecutter_cache` does not appear to be cloning the template git repository. Cloning works as expected on the briefcase master branch but not on my branch, which is a little confusing because I don't think I have touched any part of that code (except to move it so it executes *before* we try to build the app context).

The error is being caught and passed silently in `update_cookiecutter_cache`, but causes a problem later on because I assume that `cookiecutter.json` must be a file that lives in the cache directory. I didn't spot it earlier because I did have a cached template most of the time.

To reproduce:
1. Delete the cached `/Users/genevieve/.cookiecutters/briefcase-template` directory
2. Run `briefcase new`
3. Observe that the briefcase-template repository is not cloned to the cache directory, and an error is raised when the `cookiecutter.json` file can't be found in the cache.

freakboy3742:
TownCrier is the mechanism we use to make release notes remain up to date. Every PR needs to include a file in the `changes` directory that describes a release note entry for the change. So - in this case, you need to add a `414.bug.rst` file that describes the change. In this case, something like "The new app wizard now uses the template to generate the questions for user input". This will often be very close (or identical) to the ticket title; but it should be written for the external-user audience, rather than an internal-developer audience.

As for the templating issue: the code is "working as designed" - but I can completely understand why the "design" in this case could be counterintuitive (and counterproductive). 

If you walk the logic through `update_cookiecutter_cache()` (in base.py) you'll find that it only updates the cookiecutter if the repo has already been cloned. The return value is the path that is passed to Cookiecutter; that's either an up-to-date local template cache, or the original repository URL.

However, this means that on the first pass through, the cookiecutter checkout doesn't exist until the cookiecutter is actually executed. This doesn't affect the master branch because it doesn't expect the template to exist before the cookiecutter is invoked.

The `update_cookiecutter_cache()` method exists because by default, in no-input mode, cookiecutter uses the existing template, rather than updating/recloning. This is problematic for briefcase because we use template repositories with multiple branches, and we need to ensure templates are updated to the correct branch before they're used.

I'm guessing the fix here will be to force the repository to be cloned, and then *always* pass cookiecutter a local cache directory.

GenevieveBuckley:
Thank you for the explanations, on both counts!
* I've added a file for towncrier like you suggest, but it seems like the towncrier CI check is still unhappy?
* I think I can force cookiecutter to clone the repository, that should be ok. I'm planning to come back to do this when I have a bit of time, hopefully within the next week.

freakboy3742:
My apologies - I mistyped in my description about towncrier: the file should be called `414.bugfix.rst`, not `414.bug.rst`.

GenevieveBuckley:
Note: tests are failing because I still need to mock the repository clone command

GenevieveBuckley:
The holdup here is because I'm not sure how to test this properly.  Maybe there needs to be some extra fixture to handle this, or do you think is there another simpler way?

The suggestion to force cookiecutter to clone the repository each time works well for me, but testing is causing some trouble. Most of the existing tests use dummy paths (eg: "https://github.com/beeware/briefcase-tester-dummy-template.git") and I run into trouble because this url doesn't really exist, and neither does a local cache at `~/cookiecutters/briefcase-tester-dummy-template/`.

You can see an illustrative example of what I'm talking about by running the `test_default_template` function in `briefcase\tests\create\test_generate_app_template.py`, eg:
```
pytest tests/commands/create/test_generate_app_template.py::test_default_template
```
will complain about not being able to clone from the (non-existent) repository url.

If we try and skip cloning by adding `@mock.patch('briefcase.commands.base.clone')` to the `test_default_template` function and run pytest again, the create command is never called because there is no local repository `~/.cookiecutters/briefcase-tester-dummy-template/`

freakboy3742:
The first step is to transition from calling `clone` directly, and to proxying the call through the Command. You'll notice that utilities like `git` and `cookiecutter` aren't invoked as directly; they're accessed as `self.git` and `self.cookiecutter`. 

This provides a layer of abstraction so that allows you to do two things.

Firstly, it means you don't need to use `mock.patch`; you can overload `self.clone` on the test command with a Mock object. The net effect is similar, but it's slightly more reliable because the active codepath isn't being monkeypatched; it's using a mock object, but the way it *gets* that mock object is the same in test as it is in prod.

Secondly, you set up the `clone` command so that it [has a side effect when invoked](https://docs.python.org/3/library/unittest.mock-examples.html#side-effect-functions-and-iterables). That side effect can create files on disk in the same way that a "real" call to clone would do - or, at least, enough of a subset of files that is enough to test whatever is required. In this case, I suspect it's not much more than the prompt file from the template (and maybe the cookiecutter.json?).

Then, you focus the test on ensuring that it makes the required calls, but not that it actually *does* anything. Validate that it tries to *invoke* clone, and that the invocation uses the right arguments, but don't *actually* clone anything. Similarly, check that cookiecutter is invoked, but don't actually run the template. The only behaviour you *actually* run is behaviour that is under our control - in this case, parsing the prompt file and displaying it to the user - and you then validate that it parsed the right thing.

Does that make sense?

GenevieveBuckley:
Yes, I think that broadly makes sense, but I do not have time available to give to this for at least the next two weeks. 

(I don't expect you personally are considering this to be urgent, I just think it's usually better to have more information instead of less about expected timelines. You are welcome to ping me in a few weeks if I haven't checked back in to schedule a time to tackle it.)

freakboy3742:
@GenevieveBuckley Thanks for letting me know about timelines. I agree this isn't *urgent*, but it would definitely be a nice to have. At the very least, it would be nice to have the list of supported platforms in the template match the list that Briefcase displays, so I'll probably sneak in a small workaround fix for the short term until you can get this permanent fix in place.

freakboy3742:
Thanks for the contribution, but I'm going to close this PR, for two reasons. Firstly, it needs a big update due to age; but more importantly, I think the way it's splitting responsibility between the template and briefcase will end up being problematic. 

The validation/hinting logic is all on the briefcase side; which means it's not actually practical to add a completely new question to the template - you need active coordination between the template and briefcase itself.

On that basis, I think it would make sense to target a much smaller feature set here - only make the options of specific questions dynamic. This would make it possible to add a new supported platform (e.g., pyside6) without needing to make a Briefcase release - which was the real driving motivation of this feature request; however, the list of questions that are actually asked would be static.

PR #415, opened 2020-06-03
Create `package_name` template property

paulproteus:
This allows the Android template to rely on briefcase to convert
`-` in bundles into `_`, allowing them to be used as Android app
identifiers and Java package identifiers.

This addresses a problem where apps with `-` in their `bundle`
name, e.g., TravelTips, would not launch properly with the
briefcase Android tooling.

## Testing performed

On my own helloworld app, I changed the `bundle` in `pyproject.toml` to include a `-` character. Before this change, the app would not launch. After this change, it does.

Note that this also requires a corresponding change in the Android template.

### DIFF
--- a/changes/415.feature.rst
+++ b/changes/415.feature.rst
@@ -1,0 +1,1 @@
+Allow Android apps to use ``-`` in their bundle name; we now convert those to ``_`` in the resulting Android package identifier and Java package name.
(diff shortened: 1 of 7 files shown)


paulproteus:
Per chat, we'll go with `package_name`!

paulproteus:
I've re-done my testing, and it passes, so back to you @freakboy3742 :)

paulproteus:
Other remark -- getting this right is a prerequisite for splash screens :)

PR #425, opened 2020-06-16
Briefcase version metadata

### LINKED ISSUE #412: Add a mechanism for detecting if an code is running as a bundled app
Apologies if I missed this in the docs... curious if you have a recommended (cross-platform?) approach for detecting whether a python app is being run from inside of a briefcase app?  something like the equivalent of `getattr( sys, 'frozen', False )` in pyinstaller?

ForeverWintr:
Hi! I think this is a really cool project, and hope this small contribution will be of use. I intend it to resolve #412 by adding `Briefcase-Version` to app metadata. It can be retrieved as in your example in the linked issue. 

Feedback is welcome.


ForeverWintr:
Thanks @freakboy3742! I've taken a crack at the documentation, what do you think?

I'm not sure what best practice is for providing backwards compatibility in example code. I've indecisively included both your backwards compatible `metadata` import, and then printed with f-strings...

ForeverWintr:
Great, thanks for the review and edits!

PR #428, opened 2020-06-18
Permit Python 3.6 or 3.7 for Android tooling

paulproteus:


paulproteus:
> I'd also be interested in removing this code entirely, and instead relying on the availability of the necessary version branch in the template repo. That would be beneficial to all repos, not just Android, and means we wouldn't need to publish a new Briefcase version to support Python 3.9.

+1! Implemented. Please review and merge if you're willing. :D

PR #432, opened 2020-06-24
Refer to Android devices by `model`, not `device`

paulproteus:
Also refactor `select_target_device()` by extracting a function.

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -126,6 +126,64 @@ Delete {sdk_zip_path} and run briefcase again.""".format(
     return sdk
 
 
+def _get_choices_and_names(running_devices, all_emulators, avd_name_by_device_id):
+    '''Input:
+
+    - `running_devices`: Dict from ADB device ID to dict with 'model' property.
+    - `all_emulators`: List of emulator device names.
+    - `avd_name_by_device_id`: A dict from ADB device ID to AVD name.
+
+    Output:
+
+    Tuple of `choice_ids`, `full_names`.
+
+    `choice_ids` is a list of all running devices, followed by all non-running
+    AVD names prefixed with `@`.
+
+    `full_names` is dict from string (device ID *or* AVD) to a formatted description
+    of the device.
+    '''
+    choice_ids = []
+    full_names = {}
+
+    running_avds = set()  # Used to avoid adding an AVD twice to the list of choices.
+
+    # Handle devices that are running.
+    for device_id, device_info in sorted(running_devices.items(), key=lambda d: d[1]["name"]):
+        model = device_info["name"]
+        if avd_name_by_device_id.get(device_id):
+            avd = avd_name_by_device_id[device_id]
+            running_avds.add(avd)
+            # For emulators, the full_name shows the @{avd} name. This is because the
+            # ADB device ID is not stable over reboots for Android virtual devices.
+            printable_device_id = "@{avd}".format(avd=avd)
+            device_type = "emulator"
+        else:
+            printable_device_id = device_id
+            device_type = "device"
+
+        full_name = "{printable_device_id} ({model} {device_type})".format(
+            device_type=device_type, model=model, printable_device_id=printable_device_id,
+        )
+        full_names[device_id] = full_name
+        if device_type == "emulator":
+            full_names[printable_device_id] = full_name
+        choice_ids.append(device_id)
+
+    # Handle devices (emulator names) that aren't running.
+    for avd in all_emulators:
+        if avd in running_avds:
+            continue  # skip
+        choice_id = '@{avd}'.format(avd=avd)
+        choice_ids.append(choice_id)
+        full_name = "{choice_id} ({device_type})".format(
+            choice_id=choice_id, device_type=device_type,
+        )
+        full_names[choice_id] = full_name
+
+    return choice_ids, full_names
+
+
 class AndroidSDK:
     def __init__(self, command, root_path):
         self.command = command
@@ -319,18 +377,17 @@ class AndroidSDK:
                         key, value = part.split(":")
                         details[key] = value
 
+                    model = "Unknown"
                     if parts[1] == "device":
-                        name = details["device"]
+                        model = details["model"]
                         authorized = True
                     elif parts[1] == "offline":
-                        name = "Unknown device (offline)"
                         authorized = False
                     else:
-                        name = "Unknown device (not authorized for development)"
                         authorized = False
 
                     devices[parts[0]] = {
-                        "name": name,
+                        "name": model,
                         "authorized": authorized,
                     }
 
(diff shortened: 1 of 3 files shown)


src/briefcase/integrations/android_sdk.py
```diff
@@ -460,44 +472,32 @@ def select_target_device(self, device_or_avd):
         # Proces the user's choice
         if choice is None:
             # Create a new emulator. No device ID or AVD.
-            device = None
-            avd = None
-            name = None
+            device, avd, name = None, None, None
```
freakboy3742, line 475: Stylistically, I *really* dislike this pattern. One assigment, one line, unless you're actually unpacking a tuple.

src/briefcase/integrations/android_sdk.py
```diff
@@ -460,44 +472,32 @@ def select_target_device(self, device_or_avd):
...
         elif choice.startswith("@"):
             # A non-running emulator. We have an AVD, but no device ID.
-            device = None
-            name = device_choices[choice]
-            avd = choice[1:]
+            device, name, avd = None, full_names[choice], choice[1:]
```
freakboy3742, line 478: Same here - one assignment, one line.

src/briefcase/integrations/android_sdk.py
```diff
@@ -460,44 +472,32 @@ def select_target_device(self, device_or_avd):
...
-        if avd:
-            print("""
-In future, you can specify this device by running:
+            # Ensure the device is developer-enabled.
+            if not running_devices[choice]["authorized"]:
+                raise AndroidDeviceNotAuthorized(choice)
 
-    briefcase run android -d @{avd}
+            # Compute the 3-tuple that we return.
+            device, name, avd = choice, full_names[choice], avd_name_by_device_id.get(choice)
```
freakboy3742, line 490: Again here.

src/briefcase/integrations/android_sdk.py
```diff
@@ -126,6 +126,64 @@ def verify_android_sdk(command):
...
+
+    Output:
+
+    Tuple of `choice_ids`, `full_names`.
+
+    `choice_ids` is a list of all running devices, followed by all non-running
+    AVD names prefixed with `@`.
+
+    `full_names` is dict from string (device ID *or* AVD) to a formatted description
+    of the device.
```
freakboy3742, line 144: Why return both full_names *and* choice_ids? Won't choice_ids == full_names.keys()? (Especially since insertion order is guaranteed in Python 3.6 - and Android won't work on Python 3.5, so we get that one as a gimme.

freakboy3742: Answering my own question here - it looks like full_names can include more identifiers than is in choice_ids? If that's the case, that's worth clarifying in the explanation here.

src/briefcase/integrations/android_sdk.py
```diff
@@ -353,92 +410,47 @@ def select_target_device(self, device_or_avd):
...
-        for avd in self.emulators():
-            if avd not in running_avds:
-                name = "@{avd} (emulator)".format(avd=avd)
-                choices.append(("@" + avd, name))
-                device_choices["@" + avd] = name
-
-        # If a device or AVD has been provided, check it against the available
-        # device list.
+        # Compute mappings of running Android virtual devices to running device IDs.
+        avd_name_by_device_id = dict([
```
freakboy3742, line 420: Any reason to not use a dict comprehension here instead of a constructing a dict from a list comprehension of tuples?

src/briefcase/integrations/android_sdk.py
```diff
@@ -353,92 +410,47 @@ def select_target_device(self, device_or_avd):
...
+            (device_id, self.adb(device_id).avd_name())
+            for device_id in running_devices
+        ])
+        device_id_by_avd_name = dict([(avd_name, device_id) for device_id, avd_name in avd_name_by_device_id.items()])
```
freakboy3742, line 424: Same again here - why not a dict comprehension?

src/briefcase/integrations/android_sdk.py
```diff
@@ -126,6 +126,64 @@ def verify_android_sdk(command):
...
+    '''
+    choice_ids = []
+    full_names = {}
+
+    running_avds = set()  # Used to avoid adding an AVD twice to the list of choices.
+
+    # Handle devices that are running.
+    for device_id, device_info in sorted(running_devices.items(), key=lambda d: d[1]["name"]):
+        model = device_info["name"]
+        if avd_name_by_device_id.get(device_id):
```
freakboy3742, line 154: Not a fan of the "Null as false" use here. This is either an "if device_id in avd_name_by_device_id", or a try/catch.

src/briefcase/integrations/android_sdk.py
```diff
@@ -126,6 +126,64 @@ def verify_android_sdk(command):
     return sdk
 
 
+def _get_choices_and_names(running_devices, all_emulators, avd_name_by_device_id):
+    '''Input:
+
+    - `running_devices`: Dict from ADB device ID to dict with 'model' property.
+    - `all_emulators`: List of emulator device names.
+    - `avd_name_by_device_id`: A dict from ADB device ID to AVD name.
```
freakboy3742, line 134: Could `all_emulators` and `avd_name_by_device_id` be merged into a single dict with a None value if we don't know the device ID? 

If we can't - can we rename `avd_name_by_device_id` to a more functional name - like "active_emulators"?

src/briefcase/integrations/android_sdk.py
```diff
@@ -126,6 +126,64 @@ def verify_android_sdk(command):
...
+            # For emulators, the full_name shows the @{avd} name. This is because the
+            # ADB device ID is not stable over reboots for Android virtual devices.
+            printable_device_id = "@{avd}".format(avd=avd)
+            device_type = "emulator"
+        else:
+            printable_device_id = device_id
+            device_type = "device"
+
+        full_name = "{printable_device_id} ({model} {device_type})".format(
+            device_type=device_type, model=model, printable_device_id=printable_device_id,
```
freakboy3742, line 166: The "name" is used in the menu, but it's also used in the context of saying "starting device on {name}"; to that end, "Kogan Agora 9 device" is a more useful "name", with the device ID providing important context. 

To support that, I'd be inclined to list then name the other way around for physical devices - "Kogan_Agora_9 device (KABCDABCDA1513)" for a physical device (i.e., what it is to a human, with the ID only there for disambiguation).

The text for emulators can be simplified, too; "Android_SDK_build_for_x86" doesn't convey anything that "emulator" doesn't already convey. Not sure if I prefer "@beePhone (emulator)" or "Emulator (@beePhone)", though. Thoughts?

paulproteus:
I think this is really good feedback, and I want to make these changes, but I'm submitting #433 as a simpler implementation to get this out the door faster.

PR #433, opened 2020-06-25
Rely on Android device model, in device list

paulproteus:
Example output:

When emulator is running:

```
$ briefcase run android

Select device:

  1) @beePhone (Android SDK built for x86 emulator)
  2) Pixel 3a (94AAY0LNE1)
  3) Create a new Android emulator

> 
```

When emulator is not running:

```
Select device:

  1) Pixel 3a (94AAY0LNE1)
  2) @beePhone (emulator)
  3) Create a new Android emulator

> 
```

You had mentioned that "Android_SDK_built_for_x86" wasn't necessarily useful. IMHO it is -- it could be e.g. a x86_64 emulator instead of an x86 one.

Let's use this PR to get the output exactly right. Code refactoring can wait IMHO.

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -377,8 +377,8 @@ class AndroidSDK:
             if avd:
                 # It's a running emulator
                 running_avds[avd] = d
-                full_name = "@{avd} ({name} emulator)".format(
-                    avd=avd, name=name,
+                full_name = "@{avd} (running emulator)".format(
+                    avd=avd,
                 )
                 choices.append((d, full_name))
 
@@ -689,8 +689,8 @@ find this page helpful in diagnosing emulator problems.
                             # Found an active device that matches
                             # the AVD we are starting.
                             name = details["name"]
-                            full_name = "@{avd} ({name} emulator)".format(
-                                avd=avd, name=name,
+                            full_name = "@{avd} (running emulator)".format(
+                                avd=avd,
                             )
                             break
                         else:
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -688,7 +688,6 @@ find this page helpful in diagnosing emulator problems.
                         if device_avd == avd:
                             # Found an active device that matches
                             # the AVD we are starting.
-                            name = details["name"]
                             full_name = "@{avd} (running emulator)".format(
                                 avd=avd,
                             )
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -320,7 +320,7 @@ class AndroidSDK:
                         details[key] = value
 
                     if parts[1] == "device":
-                        name = details["model"]
+                        name = details["model"].replace("_", " ")
                         authorized = True
                     elif parts[1] == "offline":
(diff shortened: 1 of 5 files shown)


src/briefcase/integrations/android_sdk.py
```diff
@@ -320,7 +320,7 @@ def devices(self):
                         details[key] = value
 
                     if parts[1] == "device":
-                        name = details["device"]
+                        name = details["model"]
```
freakboy3742, line 323: Really minor possible cosmetic improvement - would replacing `_` with a literal space make sense here? We have a limited sample size, but displaying a user-visible name of `Pixel 3a`/`Kogan Agora 9` seems preferable to `Pixel_3a`/`Kogan_Agora_9`. We're not using this name as an identifier (AFAIK?), so it should be purely cosmetic change?

src/briefcase/integrations/android_sdk.py
```diff
@@ -377,7 +377,7 @@ def select_target_device(self, device_or_avd):
             if avd:
                 # It's a running emulator
                 running_avds[avd] = d
-                full_name = "@{avd} ({name} emulator)".format(
+                full_name = "emulator (@{avd} {name})".format(
```
freakboy3742, line 380: Cosmetically, I think the old construction for emulators read better. The content in brackets should be the qualifier, not the "meat" of the description.

src/briefcase/integrations/android_sdk.py
```diff
@@ -390,13 +390,14 @@ def select_target_device(self, device_or_avd):
...
-                device_choices[d] = name
+                full_name = "{name} ({d})".format(name=name, d=d)
+                choices.append((d, full_name))
+                device_choices[d] = full_name
 
         # Add any non-running emulator AVDs to the list of candidate devices
         for avd in self.emulators():
             if avd not in running_avds:
-                name = "@{avd} (emulator)".format(avd=avd)
+                name = "emulator (@{avd})".format(avd=avd)
```
freakboy3742, line 400: Same here - I think the old string works better.

src/briefcase/integrations/android_sdk.py
paulproteus: OK! Either way works for me, honestly.

src/briefcase/integrations/android_sdk.py
paulproteus: Sure :)

src/briefcase/integrations/android_sdk.py
```diff
@@ -377,8 +377,8 @@ def select_target_device(self, device_or_avd):
             if avd:
                 # It's a running emulator
                 running_avds[avd] = d
-                full_name = "@{avd} ({name} emulator)".format(
-                    avd=avd, name=name,
+                full_name = "@{avd} (running emulator)".format(
```
paulproteus, line 380: IMHO, this is a decrease in information, and I like `{name}` showing up b/c it lets me know if it's x86 vs. x86_64, etc. Up to you, though.

paulproteus:
Great to have you pushing commits! Should I leave it to you at this point to take lead on this and merge it when you like it?

freakboy3742:
Yeah - that should be it (once flake8 is happy). There will be some related updates to the docs patch to match the new text.

Issue #442, opened 2020-07-07
Error trying to connect to Docker daemon socket when building


DannyPinghero:
Trying out the basic commands with the empty / default app.  When running `briefcase build`, I get:

```
$ briefcase build

Ensure we have the linuxdeploy AppImage...
linuxdeploy-x86_64.AppImage already downloaded

[pub-ex] Building AppImage...

[pub-ex] Entering Docker context...
docker: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.40/containers/create: dial unix /var/run/docker.sock: connect: permission denied.
See 'docker run --help'.

Error while building app pub-ex.
```

This isn't _really_ an error / fault of briefcase - it's because when I installed docker initially I didn't bother to do the post install step of making a new group so I could call `docker` without `sudo` (documented here: https://docs.docker.com/engine/install/linux-postinstall/  or the helpful SO answer here: https://stackoverflow.com/questions/21871479/docker-cant-connect-to-docker-daemon ). 

_But_ it is relatively common (judging by personal experience + number of similar stack overflow posts), maybe the error message could include a mention / link to that?  Or the briefcase docs?  Or mention the `--no-docker` flag?  [The building docs here (https://docs.beeware.org/en/latest/tutorial/tutorial-3.html#building-your-application) mention that docker is used to build but it's relatively sparse - I only noticed the --no-docker flag grepping around in the source to see if I could hack in adding a `sudo` to the front of the docker commands just to get it to work]

DannyPinghero:
If anyone has thoughts on what they would like, I'm happy to try adding error messaging or updating the docs!  Or if it's felt this is not useful / not the responsibility of briefcase, totally understand!

freakboy3742:
Completely agreed - anything we can do to identify "obvious" sources of error and provide workarounds would be a welcome addition.

Docker is mentioned in [the AppImage reference docs](https://briefcase.readthedocs.io/en/latest/reference/platforms/linux/appimage.html) since that's the only backend that uses Docker. If we can elaborate on these docs, or link to them better, suggestions are welcome. 

However the ideal outcome is to not need the docs at all (if only because nobody ever reads the the docs unless they absolutely *have* to :-).

[...]

PR #443, opened 2020-07-07
Enhanced verification that Briefcase will be able to run docker commands

DannyPinghero:
Code now uses a preflight `docker info` command to check the availabilty of docker commands.  Tested only on linux
Addresses issue #442

Unfortunately I don't have access to a Mac environment or a Windows install that is capable of running docker (you need to have the professional edition or sign up for advanced telemetry to get WSL2 to run docker for windows :cry: ), so I am not sure if these changes are safe enough to be run cross-platform.  Appreciate any advice in this regard

Limitations: On linux (ubuntu 20.04 specifically) the docker commands will just fail with error code 1 in a variety of error cases.  The 2 I check for here I differentiate by examining the text of the error message - I don't know if docker considers these errors part of their API, so these checks may be brittle.  I think the worst that happens is it just hits the fallback to the generic error message, so this should be at worst extra code that can never get run, if docker changes the text of the errors.

### DIFF
--- a/src/briefcase/integrations/docker.py
+++ b/src/briefcase/integrations/docker.py
@@ -37,7 +37,7 @@ def verify_docker(command):
     :param command: The command that needs to perform the verification check.
     """
 
-    WRONG_DOCKER_VERSION_ERROR_MESSAGE = '''
+    WRONG_DOCKER_VERSION_ERROR = """
 Briefcase requires Docker 19 or higher, but you are currently running
 version {docker_version}. Visit:
 
@@ -45,48 +45,51 @@ version {docker_version}. Visit:
 
 to download and install an updated version of Docker.
 {extra_content}
-'''
-    UNKNOWN_DOCKER_VERSION_WARNING = '''
+"""
+    UNKNOWN_DOCKER_VERSION_WARNING = """
 *************************************************************************
 ** WARNING: Unable to determine the version of Docker                  **
 *************************************************************************
 
-   Briefcase will proceed, assuming everything is OK. If you experience
-   problems, this is almost certainly the cause of those problems.
+    Briefcase will proceed, assuming everything is OK. If you
+    experience problems, this is almost certainly the cause of those
+    problems.
 
-   Please report this as a bug at:
+    Please report this as a bug at:
 
-     https://github.com/beeware/briefcase/issues/new
+      https://github.com/beeware/briefcase/issues/new
 
-   In your report, please including the output from running:
+    In your report, please including the output from running:
 
-     docker --version
+      docker --version
 
-   from the command prompt.
+    from the command prompt.
 
 *************************************************************************
-'''
-    DOCKER_INSTALLATION_STATUS_UNKNOWN_WARNING = '''
+"""
+
+    DOCKER_INSTALLATION_STATUS_UNKNOWN_WARNING = """
 *************************************************************************
 ** WARNING: Unable to determine if Docker is installed                 **
 *************************************************************************
 
-   Briefcase will proceed, assuming everything is OK. If you experience
-   problems, this is almost certainly the cause of those problems.
+    Briefcase will proceed, assuming everything is OK. If you
+    experience problems, this is almost certainly the cause of those
+    problems.
 
-   Please report this as a bug at:
+    Please report this as a bug at:
 
-     https://github.com/beeware/briefcase/issues/new
+      https://github.com/beeware/briefcase/issues/new
 
-   In your report, please including the output from running:
+    In your report, please including the output from running:
 
-     docker --version
+      docker --version
 
-   from the command prompt.
+    from the command prompt.
 
 *************************************************************************
-'''
-    DOCKER_NOT_INSTALLED_ERROR_MESSAGE = '''
+"""
+    DOCKER_NOT_INSTALLED_ERROR = """
 Briefcase requires Docker, but it is not installed (or is not on your PATH).
 Visit:
 
@@ -96,7 +99,7 @@ to download and install git manually.
 {extra_content}
 If you have installed Docker recently and are still getting this error, you may
 need to restart your terminal session.
-'''
+"""
     try:
         # Try to get the version of docker that is installed.
         output = command.subprocess.check_output(
@@ -114,7 +117,7 @@ need to restart your terminal session.
             if int(version[0]) < 19:
                 # Docker version isn't compatible.
                 raise BriefcaseCommandError(
-                    WRONG_DOCKER_VERSION_ERROR_MESSAGE.format(
+                    WRONG_DOCKER_VERSION_ERROR.format(
                         docker_version=docker_version,
                         **docker_install_details(command.host_os),
                     )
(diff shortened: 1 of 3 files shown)


freakboy3742:
FWIW, I agree that checking error messages isn't ideal, but it works reliably, and if the messages ever change, we can update them over time. As long as it doesn't give an answer that is *wrong*, I think we can live with it; and in this case, docker info returning a status code of 0 is what we're really looking for. Everything else that is printed is helper text - which could be *more* accurate, but won't ever be *wrong*.

Using your implementation on macOS, `failure_output` contains the following if the server is down:
```
Client:
 Debug Mode: false

Server:
ERROR: Error response from daemon: dial unix docker.raw.sock: connect: connection refused
errors pretty printing info
```
with a return code of 1.

DannyPinghero:
> FWIW, I agree that checking error messages isn't ideal, but it works reliably, and if the messages ever change, we can update them over time. As long as it doesn't give an answer that is _wrong_, I think we can live with it; and in this case, docker info returning a status code of 0 is what we're really looking for. Everything else that is printed is helper text - which could be _more_ accurate, but won't ever be _wrong_.
> 
> Using your implementation on macOS, `failure_output` contains the following if the server is down:
> 
> ```
> Client:
>  Debug Mode: false
> 
> Server:
> ERROR: Error response from daemon: dial unix docker.raw.sock: connect: connection refused
> errors pretty printing info
> ```
> 
> with a return code of 1.


Ok, cool, works for me.
Will add that error string to the check as well.

Some of the tests expect just the one subprocess command to be run, so I'm going to refactor slightly, move these new checks into a separate function so they can be mocked and tested separately

DannyPinghero:
I moved some of the multiline strings just to the top of the functions - I had a little trouble adjusting for the indentation with the strings inline - happy to undo that if it goes against any internal style or anything

src/briefcase/integrations/docker.py
```diff
@@ -118,12 +96,108 @@ def verify_docker(command):
...
+            universal_newlines=True,
+            stderr=subprocess.STDOUT,
+        ).strip('\n')
+    except subprocess.CalledProcessError as e:
+        failure_output = e.output
+        if (
+            'Got permission denied while trying to connect to the Docker daemon socket' in failure_output
+        ):
+            print(LACKS_PERMISSION_ERROR_MESSAGE)
+            raise BriefcaseCommandError("docker lacks required permissions")
```
freakboy3742, line 189: The message of the `BriefcaseCommandError` will be displayed to the user, so you don't need to print the message and then raise a different one; just raise the error.

src/briefcase/integrations/docker.py
```diff
@@ -118,12 +96,108 @@ def verify_docker(command):
...
+        if (
+            'Is the docker daemon running?' in failure_output or  # error message on ubuntu
+            'connect: connection refused' in failure_output  # error message on macos
+        ):
+            print(DAEMON_NOT_RUNNING_ERROR_MESSAGE)
+            raise BriefcaseCommandError("docker daemon not running")
```
freakboy3742, line 195: Same again here; raise the error message directly.

src/briefcase/integrations/docker.py
```diff
@@ -118,12 +96,108 @@ def verify_docker(command):
...
+
+        print(GENERIC_ERROR_MESSAGE.format(error_message=failure_output))
+        raise BriefcaseCommandError(GENERIC_ERROR_MESSAGE.format(error_message=failure_output))
```
freakboy3742, line 198: Same again here; raise the error message directly.

src/briefcase/integrations/docker.py
```diff
@@ -118,12 +96,108 @@ def verify_docker(command):
...
+ensuring that it is running or debugging any issues, or try
+running Briefcase build with the --no-docker option
+(note this may require installing addtional dependencies)
+
+*************************************************************************'''
+    GENERIC_ERROR_MESSAGE = '''
+docker command failed with error: {error_message}'''
+
+    try:
+        _ = command.subprocess.check_output(
```
freakboy3742, line 178: Minor style convention - if the return value isn't being used, preference is to not assign the value at all, rather than assigning to a dummy variable.

tests/integrations/docker/test_verify_docker.py
```diff
@@ -113,3 +120,60 @@ def test_docker_unknown_version(test_command, capsys):
     output = capsys.readouterr()
     assert '** WARNING: Unable to determine the version of Docker' in output.out
     assert output.err == ''
+
+
+def test_docker_exists_but_process_lacks_permission_to_use_it(test_command, capsys):
```
freakboy3742, line 125: Given that there are multiple error messages that can trigger this, we should test all of them. This is a good candidate for a parameterized test.

tests/integrations/docker/test_verify_docker.py
```diff
@@ -54,6 +59,7 @@ def test_docker_doesnt_exist(test_command):
     )
 
 
+@mock.patch('briefcase.integrations.docker._verify_docker_can_run')
```
freakboy3742, line 62: Preference is to avoid using mock.patch, and validate *all* the calls made to services.

tests/integrations/docker/test_verify_docker.py
```diff
@@ -92,6 +98,7 @@ def test_docker_bad_version(test_command, capsys):
         verify_docker(command=test_command)
 
 
+@mock.patch('briefcase.integrations.docker._verify_docker_can_run')
```
freakboy3742, line 101: Again; preference is to avoid mock.patch.

DannyPinghero:
> This looks really solid. I've flagged a couple of fairly minor things with error handling and testing, but for the most part, this looks great.
> 
> The refactoring around error messages makes the patch itself a bit of a PITA to review (because the diff is a mess), but I agree it make the code a lot more readable.
> 
> As a heads up - I'll probably do an editing pass on the specific text of the error messages before I merge; it's just as easy for me to do that edit myself, rather than try to communicate edits and then check them. The error text you've got here is all really good, but there are some small details for consistency of voice and formatting that I'm trying to keep consistent.

Great!  Should be able to address your comments without issue; you just rewriting the error messaging yourself makes sense, np, was intended as a first draft of the messaging.

Re: messy diff - I could move the error message refactoring into a separate PR if you'd prefer?

tests/integrations/docker/test_verify_docker.py
DannyPinghero: I parametrized `test_docker_exists_but_is_not_running` based on this feedback - right now I only have one error string that triggers the failure state in this test.  I may be misunderstanding what you are referring to here otherwise.  
Would you instead prefer that 

`test_docker_exists_but_process_lacks_permission_to_use_it`
`test_docker_exists_but_is_not_running`
`test_docker_exists_but_unknown_error_when_running_command`

[the 3 failure modes of `_verify_docker_can_run`, of which the 2nd one has 2 strings it checks]

are all 1 parametrized test?

tests/integrations/docker/test_verify_docker.py
DannyPinghero: Here I'm not sure which would be preferred.  I could:

1) Remove the patch.  But since both the call to `docker --version` and the call to `docker info` will then get the return code 1 output, this test will no longer return the docker wrapper, `_verify_docker_can_run` will see a failure and raise an exception instead.  This may be desired, since I don't know if there is a reasonable set of circumstances in which `docker --version` fails and `docker info` succeeds

2) Make the side effect of this a more complicated function that switches the return code based on the arguments, so that `--version` can fail and `info` can succeed.  This would preserve the spirit of the original test.

Initially it sounds like you'd prefer 2; if so, can do.  Just wanted to get your thoughts on whether this is still a valid test circumstance

tests/integrations/docker/test_verify_docker.py
freakboy3742: No - I was referring to `docker_exists_but_is_not_running`, but I attached the comment to the wrong test title. There are two error messages that can cause that branch to execute, but you were only testing one of them. Sorry for the misdirection on that one.

PR #461, opened 2020-08-01
update gradle.py

praveen-manohar:
AAB (Android App Bundle) to upload on Google Play

## PR Checklist:


freakboy3742:
Thanks for the contribution. However, this isn't what a "publish" command should do. "Publish" is for uploading the app to an App Store (or other publication location). What you've defined here is "package" - and the Gradle backend already has a working "package" command (whose implementation is almost identical to what you've defined here).

Issue #470, opened 2020-08-13
can't build PySide2 app with briefcase 0.3.3


jgirardet:
**Describe the bug**
error message : ERROR: Could not find dependency: libshiboken2.abi3.so.5.13 
it's working en 0.3.2.
This missing dependendy can't be added via `system-require` since it's in app_package. Maybe related to #458 
Can't find a fix for now

**To Reproduce**
Create fresh Pyside2 app from briefcase new

**Screenshots**
```
(.venv) jimmy@cacahuete:~/rien/33/helloworld$ briefcase build -v
>>> docker --version
>>> docker info

[helloworld] Building AppImage...

[helloworld] Entering Docker context...
[...]

PR #484, opened 2020-09-06
Github actions to build test app bundles

GenevieveBuckley:
## The problem
Sometimes fixes for things in briefcase unintentionally cause other issues (eg: https://github.com/beeware/briefcase/issues/470).

## The solution
We want to add CI tests that bundle hello world apps for all supported configurations (toga/pyside2/pyqt/pursuedpybear). Then we should be able to catch any regressions accidentally introduced by new features or bugfixes.

## This PR
This PR includes a `make_bundle` github action that will try to build briefcase apps from a number of other test app repos. ~~It also uploads the app installers as github artifacts so they can be downloaded and opened for manual testing, if that's ever needed.~~

You can see some results [here](https://github.com/beeware/briefcase/actions/runs/248127343), note that the build failures on linux for pyside are probably all to do with the regression introduced in briefcase 0.3.3 (so that's expected, until the underlying problem gets fixed)

## Comments
I'm happy to hear comments from other people about whether there should be any other changes to the build matrix, eg: including builds with a newer version of Ubuntu, or an older version of Python.

## What needs to happen next
There's two things I can't do, because I'm not a maintainer for beeware. Here's what I think should happen next.

1. These four repositories should be forked to the beeware organisation. No other changes needed, just forking.
    * [hellotoga](https://github.com/GenevieveBuckley/hellotoga)
    * [hellopyside](https://github.com/GenevieveBuckley/hellopyside)
    * [hellopursuedpybear](https://github.com/GenevieveBuckley/hellopursuedpybear)
    * [traveltips](github.com/freakboy3742/traveltips/)

2. We then update [line 25](https://github.com/GenevieveBuckley/briefcase/blob/224e49e4c019bb0e24c622b0989cab94c868692d/.github/workflows/bundle_app.yml#L25) of `.github/workflows/bundle_app.yml` and replace "GenevieveBuckley" with "beeware".


GenevieveBuckley:
I'd say this is ready for review now, thanks!

GenevieveBuckley:
One final thought, perhaps it's better to only upload artifacts some of the time (maybe just on pull requests?). Thoughts welcome.

GenevieveBuckley:
> One final thought, perhaps it's better to only upload artifacts some of the time (maybe just on pull requests?). Thoughts welcome.

On reflection, it's really not sensible to generate, upload, and store so many artifacts (this action runs often). I've removed that part of the script.

I don't think it's necessary now that all of the test repositories that are built here have their own github actions that include uploading artifacts. In the unlikely event someone might want to manually test the packaged app from one of these, they can just go to that repo & trigger a new build.

freakboy3742:
Apologies for the long delay on dealing with this. This is a really good starting point; however, I think we can make better use of tox, as well as testing `briefcase new` (removing the need to have all the external projects). I've opened #700 as an alternate approach.

Issue #486, opened 2020-09-09
Add `--no-sign` as a no-op to Windows/Linux/Android packaging platforms


GenevieveBuckley:
**Is your feature request related to a problem? Please describe.**
It is annoying to have special handling for the syntax of `briefcase package` commands, depending on whether you are packaging for Mac or Windows/Linux.

As suggested by Russell in [this comment](https://github.com/freakboy3742/traveltips/pull/4#pullrequestreview-484707229)
> Adding `--no-sign` support for other platforms (even if it's no-op) would also simplify things nicely.

**Describe the solution you'd like**
We'd like to have `briefcase package` accept a `--no-sign` argument regardless of the OS platform it's running on, and just have it be a no-op (no operation, do nothing) for Windows/Linux/Android.

**Describe alternatives you've considered**
Right now I use if statements in my github actions to get around this.

**Additional context**
You can see the kind of special case handling we currently need to do to work around this in [these lines of the traveltips github action to package and upload installers](https://github.com/freakboy3742/traveltips/blob/e7c996760184ffb44cc6e95846115112dfb2da81/.github/workflows/bundle_app.yml#L29-L34)

ElijahAhianyo:
hi, first timer here, can i take this up?

freakboy3742:
@ElijahAhianyo Absolutely - go right ahead! If you need any pointers or suggestions, just ask - we're happy to provide any assistance we can.

ElijahAhianyo:
@freakboy3742  Great! I took a look at the src/briefcase/cmdline.py file and I figured adding an argument like so would do:

```
parser.add_argument(
        '--no-sign',
        metavar='',
        help="<fix help message here>"
    )
```

I'm not exactly sure if I'm on track or doing what's required. Do you have any suggestions on how this can be handled?

freakboy3742:
That's the right argument to add; but the wrong place. `cmdline.py` applies to *all* commands. As this is an argument that is specific to the Publish command, it only needs to be visible on the Publish command. 

If you take a look at the macOSAppPackageCommand in platforms/macOS/app.py, you'll see the mechanism by which `--no-sign` is currently added. If that mechanism is migrated up the inheritance tree to the base Publish command, it will become a shared command across all Publish commands. You'll also need to tweak the help text so it's not macOS specific.

[...]

PR #491, opened 2020-09-17
moved package command arguments specific to macOs to base package

ElijahAhianyo:
Items completed in reference to #486 :
- moved signing options as a macOS specific extension on the app and dmg commands to reference docs for the package command
- moved package command arguments(adhoc,identity and no-sign) specific to macOs to the base package command

### DIFF
--- a/changes/486.doc.rst
+++ b/changes/486.doc.rst
@@ -1,1 +1,0 @@
-moved signing options as a macOS specific extension on the app and dmg commands to reference docs for the package command
--- a/changes/486.doc.rst
+++ b/changes/486.doc.rst
@@ -1,0 +1,1 @@
+moved signing options as a macOS specific extension on the app and dmg commands to reference docs for the package command
--- a/changes/486.feature.rst
+++ b/changes/486.feature.rst
@@ -1,1 +1,2 @@
-moved package command arguments(adhoc,identity and no-sign) specific to macOs to the base package command
+Added signing options for all platforms. App signing is only implemented on
+macOS, but ``--no-sign`` can now be used regardless of your target platform.
--- a/changes/486.feature.rst
+++ b/changes/486.feature.rst
@@ -1,0 +1,1 @@
+moved package command arguments(adhoc,identity and no-sign) specific to macOs to the base package command
--- a/docs/reference/commands/package.rst
+++ b/docs/reference/commands/package.rst
@@ -43,3 +43,25 @@ running::
 
     $ briefcase update
     $ briefcase package
+
+
+publish
+-------
+
+``--no-sign``
+~~~~~~~~~~~~~
+
+Don't perform code signing on the app.
+
+``--adhoc-sign``
+~~~~~~~~~~~~~~~~
+
+Sign app with adhoc identity.
+
+``-i <identity>`` / ``--identity <identity>``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The code signing identity to use when signing the app.
+
+
+
--- a/docs/reference/platforms/macOS/app.rst
+++ b/docs/reference/platforms/macOS/app.rst
@@ -19,26 +19,3 @@ Splash Image format
 
 ``.app`` bundles do not support splash screens or installer images.
 
-Additional options
-==================
-
-The following options can be provided at the command line when producing
-``.app`` bundles.
-
-publish
--------
-
-``--no-sign``
-~~~~~~~~~~~~~
-
-Don't perform code signing on the ``.app`` bundles.
-
-``--adhoc-sign``
-~~~~~~~~~~~~~~~~
-
-Sign ``.app`` bundles with adhoc identity.
-
-``-i <identity>`` / ``--identity <identity>``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-The code signing identity to use when signing the ``.app`` bundles.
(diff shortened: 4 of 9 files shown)


freakboy3742 (changes requested):
Those code and docs changes look like exactly what we need - thanks!

One more tweak that is required (which is what the CI system is complaining about). Towncrier is the system we use to ensure that every change is included in the release notes; you'll need to drop a file named `changes/486.feature.rst` that gives a one line description of the new feature, as you'd expect to see it described in release notes (the number is the ticket you're solving; `.feature.` because it's new feature, rather than `bugfix`, `docs` or `misc`).

freakboy3742:
Ah - now you're getting caught by coverage :-)

We have a test suite to make sure all the features of Briefcase continue to work as advertised; when we run the test suite, we measure which lines of code are executed. If you add some new code, but don't add a test, the coverage test fails (because you've presumably added functionality, but no tests).

In this case, you've removed code from the macOS app backend (increasing coverage, because there's less code to run); but you've added code to the package command - decreasing coverage. 

So - you'll need to add a test that exercises that code. 

For some reason, there actually aren't any tests of the base `package` command at present; you'll need to create some - including tests to verify that the three options you've added work as expected. The tests for the `run` command should give you an idea of the sort of thing that is required.

ElijahAhianyo:
whew. I did add some tests for the package command. Wasn't so sure if I covered all the required tests. If there's anything you think should be added, kindly let me know. If not, that was quite an interesting challenge and I'm glad to have contributed :)

freakboy3742 (approved):
Fantastic - those tests are exactly what we need! 

I'll make a couple of minor tweaks prior to merging There's no need to have *both* a doc and a feature change marker; and the text of the feature needs to be ready for inclusion in the release notes, so that means capitalisation, and talking more about what changes *functionally*, rather than the code that was changed. I also spotted a tiny typo in the test setup.

Thanks for the contribution - it's most appreciated!

freakboy3742:
...and then I discover the hiccup. As you've submitted this PR from your master, rather than a branch, I can't push updates 😢 

It's generally good practice to work from a feature branch - it means you can have multiple PRs underway at the same time, and when the maintainer wants to push an update to your code, they can. 

In this case - the easiest fix will for me to give you the patch I wanted to apply:
```
diff --git a/changes/486.doc.rst b/changes/486.doc.rst
deleted file mode 100644
index d2d8569..0000000
--- a/changes/486.doc.rst
+++ /dev/null
@@ -1 +0,0 @@
-moved signing options as a macOS specific extension on the app and dmg commands to reference docs for the package command
\ No newline at end of file
diff --git a/changes/486.feature.rst b/changes/486.feature.rst
index 4df7c58..5b4d3ac 100644
--- a/changes/486.feature.rst
+++ b/changes/486.feature.rst
@@ -1 +1,2 @@
-moved package command arguments(adhoc,identity and no-sign) specific to macOs to the base package command
\ No newline at end of file
+Added signing options for all platforms. App signing is only implemented on
+macOS, but ``--no-sign`` can now be used regardless of your target platform.
diff --git a/tests/commands/package/conftest.py b/tests/commands/package/conftest.py
index 63e1231..0c008b6 100644
--- a/tests/commands/package/conftest.py
+++ b/tests/commands/package/conftest.py
@@ -7,7 +7,7 @@ from briefcase.config import AppConfig
 
 class DummyPackageCommand(PackageCommand):
     """
-    A dummy run command that doesn't actually do anything.
+    A dummy package command that doesn't actually do anything.
     It only serves to track which actions would be performend.
     """
     platform = 'tester'diff --git a/changes/486.doc.rst b/changes/486.doc.rst
deleted file mode 100644
index d2d8569..0000000
--- a/changes/486.doc.rst
+++ /dev/null
@@ -1 +0,0 @@
-moved signing options as a macOS specific extension on the app and dmg commands to reference docs for the package command
\ No newline at end of file
diff --git a/changes/486.feature.rst b/changes/486.feature.rst
index 4df7c58..5b4d3ac 100644
--- a/changes/486.feature.rst
+++ b/changes/486.feature.rst
@@ -1 +1,2 @@
-moved package command arguments(adhoc,identity and no-sign) specific to macOs to the base package command
\ No newline at end of file
+Added signing options for all platforms. App signing is only implemented on
+macOS, but ``--no-sign`` can now be used regardless of your target platform.
diff --git a/tests/commands/package/conftest.py b/tests/commands/package/conftest.py
index 63e1231..0c008b6 100644
--- a/tests/commands/package/conftest.py
+++ b/tests/commands/package/conftest.py
@@ -7,7 +7,7 @@ from briefcase.config import AppConfig
 
 class DummyPackageCommand(PackageCommand):
     """
-    A dummy run command that doesn't actually do anything.
+    A dummy package command that doesn't actually do anything.
     It only serves to track which actions would be performend.
     """
     platform = 'tester'
```
get you to make those changes, and update the PR.

For later PRs (there will be more PRs... right? 😄), making your edits in a branch will make this final tweak process much easier.

ElijahAhianyo:
Oh right. Thanks for the insight. If I'm getting you correctly, do I need to submit another PR(and closing this one) on a branch aside the master branch?

Also. Do I need to delete the doc change maker and modify the feature change marker as well?

Yes ! Of course there'll be more PRs. Looking forward to that

freakboy3742:
You should delete the doc marker, and modify the existing feature marker with the text from the patch I provided.

At this point, it's probably easier to make the change on this PR, and keep the advice about PRs from branches for your next contribution.

ElijahAhianyo:
okay. i just did that

freakboy3742 (approved):
Awesome - thanks for the patch!

Issue #493, opened 2020-09-22
Cannot building docker container on Fedora 32


RastaDill:
Hello. I stuck in Tutoria 3 (from BeeWare documentation). I tryit confguration files and other scaffolding, but have problem:
**Command:**
briefcase create

**Error:**
```
Installing collected packages: gbulb, pycairo, pygobject, travertino, toga-core, toga-gtk
Successfully installed gbulb-0.6.1 pycairo-1.19.1 pygobject-3.38.0 toga-core-0.3.0.dev23 toga-gtk-0.3.0.dev23 travertino-0.1.3
ERROR: Exception:
Traceback (most recent call last):
  File "/usr/lib/python3.8/shutil.py", line 788, in move
    os.rename(src, real_dst)
OSError: [Errno 18] Invalid cross-device link: '/tmp/pip-target-6f2ah_xw/lib/python/toga_core-0.3.0.dev23.dist-info' -> '/app/Hello World/Hello World.AppDir/usr/app_packages/toga_core-0.3.0.dev23.dist-info'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/pip/_internal/cli/base_command.py", line 228, in _main
    status = self.run(options, args)
  File "/usr/local/lib/python3.8/dist-packages/pip/_internal/cli/req_command.py", line 182, in wrapper
    return func(self, options, args)
  File "/usr/local/lib/python3.8/dist-packages/pip/_internal/commands/install.py", line 455, in run
    self._handle_target_dir(
  File "/usr/local/lib/python3.8/dist-packages/pip/_internal/commands/install.py", line 512, in _handle_target_dir
    shutil.move(
  File "/usr/lib/python3.8/shutil.py", line 798, in move
    copytree(src, real_dst, copy_function=copy_function,
  File "/usr/lib/python3.8/shutil.py", line 554, in copytree
    return _copytree(entries=entries, src=src, dst=dst, symlinks=symlinks,
  File "/usr/lib/python3.8/shutil.py", line 455, in _copytree
    os.makedirs(dst, exist_ok=dirs_exist_ok)
  File "/usr/lib/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/app/Hello World/Hello World.AppDir/usr/app_packages/toga_core-0.3.0.dev23.dist-info'
[...]

PR #500, opened 2020-10-07
Fix issue on fedora, caused by docker and SELinux

randomunrandom:
Problem
======

Briefcase doesn't work on Fedora 32 because of SELinux policy. It most likely doesn't work on any Linux distribution with enforced SELinux.

Solution
======

as it's described in [docker docs](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label), to modify the SELinux label of file or directory, which is being mounted, you need to add :z flag at the end. This will allow docker to properly mount volumes and briefcase to create and run docker image. This shouldn't interfere with other Linux distributions as it will be ignored.

Related
----------

Issue #493  

## PR Checklist:

### DIFF
--- a/changes/500.bugfix.rst
+++ b/changes/500.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed docker build error on Linux
--- a/src/briefcase/integrations/docker.py
+++ b/src/briefcase/integrations/docker.py
@@ -246,10 +246,10 @@ class Docker:
         docker_args = [
             "docker", "run",
             "--tty",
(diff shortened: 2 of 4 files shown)


freakboy3742 (approved):
Thanks for the PR. My testing seems to confirm that this is a no-op if you're not on SELinux, so I can't see any harm including it. 

I needed to make some small tweaks to the release note and inline documentation - but as your PR was submitted from your master branch, I couldn't update this PR. I've included your commits in #533.

PR #506, opened 2020-10-18
Add install command

saroad2:
Sometimes, you don't want to run the app in development mode, but just install the required dependencies for debugging reasons or if you're in the middle of a CI/CD process.

Now you can do that with the `briefcase install` command (which is basically `briefcse dev` without running the app)

## PR Checklist:

### DIFF
--- a/changes/506.feature.rst
+++ b/changes/506.feature.rst
@@ -1,0 +1,2 @@
+Added the `briefcase install` command, which installs the app dependencies without
+running it
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -3,7 +3,7 @@ import sys
 from pathlib import Path
 
 from briefcase import __version__
-from briefcase.commands import DevCommand, NewCommand, UpgradeCommand
+from briefcase.commands import DevCommand, InstallCommand, NewCommand, UpgradeCommand
 from briefcase.platforms import get_output_formats, get_platforms
 
 from .exceptions import (
@@ -43,7 +43,7 @@ def parse_cmdline(args):
     parser.add_argument(
         'command',
         choices=[
-            'new', 'dev', 'upgrade',
+            'new', 'dev', 'install', 'upgrade',
             'create', 'update', 'build', 'run', 'package', 'publish'
         ],
         metavar='command',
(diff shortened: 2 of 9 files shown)


freakboy3742 (changes requested):
I accept the use case; however, I'm not sure `briefcase install` is the right user interface for this. From a naming perspective, it sounds to me like "briefcase is going to install your app", which is definitely not what is going on.

I'd be inclined to merge this into `briefcase dev` as as `--no-run` option; possibly with `--no-run` implying `-d`. I'm open to other names for the option if `--no-run` doesn't make sense to you.

saroad2:
Inherently, I think `briefcase dev` and `briefcase install` are 2 different entities that doesn't necesserily need to be merged into one. `briefcase dev` means "run me in development mode" while `briefcase install` means "install all needed dependencies of the application.

Maybe we can change the name `briefcase install` to  `briefcase install-dependencies` or `briefcase install-deps` in short. I think it will be more clearer that way.

What do you think?

jgirardet:
I think there is a clear separation between "dev" (use the source) et "run" (use the bundle) and we should keep it.
I'm +0 on  `briefcase dev --no-run`  which clearly express what it does : dev without run (in dev mode). But  add confusion with `briefcase run` / `briefcase dev`. It's also not clear for new comers I think.
maybe `briefcase dev --setup-only` so there is no confusing `run` in the command

freakboy3742:
@saroad2 I don't think this is something that is fixed with a different name. 

For me, it comes down to the purpose of Briefcase as a whole. Briefcase is an app packaging tool. "Local" execution is not the main purpose of the tool; it's a "necessary evil", recognising that in order to have an effective development cycle, you're going to need to run the code locally without packaging it every time. 

I look at the "dev" command almost as a "mode" - "Run the app in dev mode". Ensuring the active "dev mode" accurately reflects the app's configuration is part of what "dev mode" needs to do prior to running; that setup process also happens to overlap with what is needed in the CI case.

Optimizing for the CI use case (which is, to my mind, what a standalone "install" command is doing in this case) seems like (a) duplication of functionality (since dev already needs to, and does, do this), and (b) making a first class feature of something that isn't "core functionality". 

Thinking around the problem, I think I'd be more accepting of adding "briefcase test", since testing an app *is* something that is part of the app development story (in my mind, anyway). My hesitation here is that there are many other tools that do a better job of running test environments (tox, for example), and I don't really want to get into the game of duplicating functionality that other tools already do better unless there is a clear benefit in doing so.

I'm definitely open to alternatives to `--no-run` (@jgirardet's suggestion of `--setup-only` would be fine, IMHO).

Another option would be a subcommand (`briefcase dev update`, with `briefcase dev run` being the interpretation default action when you run `briefcase dev`. That would also potentially allow for a `briefcase dev test` at some point in the future, if we wanted to go down that path.

saroad2:
I accept your suggestions. Since it means to override the entire PR, I'm closing it and open a new one.

PR #519, opened 2020-10-29
User can now define which template branch to use

saroad2:
**Why is it useful?**
Assume you would like to take the template from a development branch, which is not derived from a specific python version,
or if you would like to always use the same version, no matter which python version you are using.

Now you can do that by defining "tempalte_branch" in *pyproject.toml*.

## PR Checklist:


setup.cfg
```diff
@@ -55,7 +55,7 @@ install_requires =
     pip >= 20
     setuptools >= 45
     wheel >= 0.34
-    cookiecutter >= 1.0
+    cookiecutter >= 1.0,<1.7.0
```
freakboy3742, line 58: I'm sure there's a good reason for this - can you elaborate on what it is? Upper version pins always make me nervous, as they effectively mean the dependency is moving in a different direction to the project using it.

freakboy3742 (changes requested):
This looks really good (and a good feature addition).

One question/clarification requested inline; two other missing pieces are:
1. a test case exercising the new override behavior
2. documentation for the new option

setup.cfg
saroad2: This upper boundary was due to a bug in *cookiecutter*, that apparently has been fixed by now 😄 .

Removed it.

src/briefcase/commands/create.py
```diff
@@ -252,16 +252,19 @@ def generate_app_template(self, app: BaseConfig):
         # use a default template.
         if app.template is None:
             app.template = self.app_template_url
+        if app.template_branch is None:
+            app.template_branch = self.python_version_tag
```
freakboy3742, line 256: Something that occurred to me on a re-review - should this branch be executed if you define a custom template URL?  

I can see arguments both ways. On the one hand, requiring developers to use a "3.8" branch for their 3.8 custom template makes sense. This would technically be backwards incompatible, but I'm not *overly* concerned about that. 

On the other hand, the most likely use case I can imagine is "I made a custom template for my project, and I've only got one (master) branch." Requiring that users manually specify a branch of "master" will be a bit of a PITA, given the default behaviours of git.

Thoughts?

freakboy3742 (changes requested):
One small edge case that occurred to me in a re-review; plus there's still a need for docs. I'm happy to write the docs in this case if you'd prefer - but we need to resolve the design issue first.

src/briefcase/commands/create.py
saroad2: I prefer the user to create his/her custom template repository with the 3.6/3.7/3.8 branches by default and to specify explicitly otherwise.

ie. by default the branches should be as the python version, but the user can define if it's not the case.

saroad2:
Hey, I have no problem to document the new ability, but I'm not sure where is the place this documentation.
Please let me know where and I'll write this documentation.

freakboy3742:
The docs will be needed [somewhere near here](https://github.com/beeware/briefcase/blob/master/docs/reference/configuration.rst#template).

src/briefcase/commands/create.py
freakboy3742: Fair enough. I suspect custom template repositories are going to be an edge case (or, at least, I *hope* they will be) - so it won't hurt too much to force projects that need a custom template to version them, and explicitly opt *out* if they don't want to use versioning.

freakboy3742 (approved):
👍 I've got a couple of minor typos/cleanups in the docs, but otherwise this is good to go!

PR #520, opened 2020-10-31
Support defining an msi installer scope as "per machine"

### LINKED ISSUE #498: Winows: installation per-machine over per-user
At the moment, when packaging an MSI using Briefcase, the installation is **per-user**. I need to have the ability to make the installation **per-machine**.

Right now, the program is currently installed in the %appdata% folder - which is a per-user folder. the env variable translates into something that includes the username in it, and because we are deploying to "fresh" installs - no users have logged in yet. I want to be able to install the program regardless of whether the user that is going to use the program is logged in or not - so in the very least, the program needs to be installed in the "program files" folder.

saroad2:
Now users can define an MSI installer to be "per user" or "per machine".

Fixes #498

## PR Checklist:

### DIFF
--- a/changes/498.feature.rst
+++ b/changes/498.feature.rst
@@ -1,1 +1,1 @@
-User can now define an MSI file to be "per user" or "per machine"
+Windows MSI installers can be configured to be per-machine, system-wide installers.
--- a/changes/498.feature.rst
+++ b/changes/498.feature.rst
@@ -1,0 +1,1 @@
+User can now define an MSI file to be "per user" or "per machine"
--- a/docs/reference/platforms/windows/msi.rst
+++ b/docs/reference/platforms/windows/msi.rst
@@ -34,6 +34,17 @@ The following options can be added to the
 ``tool.briefcase.app.<appname>.windows`` section of your ``pyproject.toml``
 file.
 
+``system_installer``
+--------------------
+
+Controls whether the app will be installed as a per-user or per-machine app.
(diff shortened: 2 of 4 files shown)


freakboy3742 (changes requested):
Looks good; but needs a test to make sure the attribute is picked up correctly.

freakboy3742 (changes requested):
I was about to merge this, and it occurred to me that we're not doing any error handling on the user-provided value for `install_scope`. WiX will only accept `perUser` and `perMachine` as valid values. At the very least, we should be validating that the user-provided value is in the right format; however, I also wonder whether the setting that Briefcase exposes should be a boolean. Is there any other valid option? Should we simplify the interface to `system_installer=True` (or similar)?

saroad2:
You are absolutely correct, and I like the name `system_installer`. In the following days, I'll make the change in https://github.com/beeware/briefcase-windows-msi-template and change this PR accordingly.

freakboy3742:
I actually think the _template_ is fine - after all, it specifies the text that will be directly injected into the template. However, if you feel it's worth keeping the template in alignment with the pyproject config option, I won't object.

saroad2:
In the end, I came to the conclusion to keep the template as it is and only change the code here.

freakboy3742 (approved):
👍 I've pushed some minor tweaks (including docs), but otherwise this is good to go!

PR #522, opened 2020-11-01
Add no-run option to dev command

saroad2:
One can now install dev-mode dependencies without actually running the app.
This is good for testing and CI/CD purposes.

This is a retry of #506

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/dev.py
+++ b/src/briefcase/commands/dev.py
@@ -52,7 +52,9 @@ class DevCommand(BaseCommand):
         )
         parser.add_argument(
             '--no-run',
-            action="store_true",
+            dest="run_app",
+            action="store_false",
+            default=True,
             help='Do not run the app, just install dependencies.'
         )
 
@@ -109,7 +111,7 @@ class DevCommand(BaseCommand):
         self,
         appname: Optional[str] = None,
         update_dependencies: Optional[bool] = False,
-        no_run: Optional[bool] = False,
+        run_app: Optional[bool] = True,
         **options
     ):
         # Confirm all required tools are available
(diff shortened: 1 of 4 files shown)


src/briefcase/commands/dev.py
```diff
@@ -50,6 +50,11 @@ def add_options(self, parser):
             action="store_true",
             help='Update dependencies for app'
         )
+        parser.add_argument(
+            '--no-run',
+            action="store_true",
+            help='Do not run the app, just install dependencies.'
+        )
```
freakboy3742, line 57: Although `no-run` is the option we want, having the negative in the stored argument name is cumbersome (as it leads to logic like "if not no_run"). If you add `dest='run_app', action='store_false'` here, you can check for `args.run_app`, which will be true by default, and false if the `--no-run` flag is provided.

src/briefcase/commands/dev.py
```diff
@@ -133,18 +139,19 @@ def __call__(
         # installed. If a dependency update has been manually requested,
         # do it regardless.
         dist_info_path = self.app_module_path(app).parent / '{app.module_name}.dist-info'.format(app=app)
-        if update_dependencies or not dist_info_path.exists():
+        if update_dependencies or no_run or not dist_info_path.exists():
```
freakboy3742, line 142: It took me a moment to work out why `no_run` was being mentioned here. Rather than integrating the no_run check here, it might be better to do a once-off check modifying `update_dependencies` if no_run is specified. That simplifies the logic on this line, and if we ever need to add some other check against update_dependencies, we won't accidentally forget the impact of no-run.

freakboy3742 (changes requested):
Looks good; a couple of minor suggestions, mostly aiding readability.

freakboy3742 (approved):
👍 A solid addition - thanks!

PR #532, opened 2020-11-19
fix crash when running with verbose output

geogriff:
Some arguments are passed as pathlib.Path objects, so convert them to str first.


freakboy3742 (approved):
Thanks for the fix!

PR #534, opened 2020-11-25
fix .briefcase directory permissions in some cases for linux build

jessa0:
During `briefcase build linux`, if the `.briefcase` directory is not already present when docker is being used, docker will create the directory as root in order to mount it as a volume in the container, preventing the container from modifying its contents. This fixes that by ensuring the `.briefcase` directory has been created by the user briefcase is running as, before running docker.

In many cases, the `.briefcase` directory is created when e.g. installing the app support package, but when using a custom app support package path on the filesystem, the `.briefcase` directory doesn't get created before docker is run.

Tested this change with `briefcase build linux` on my own project.

### DIFF
--- a/changes/534.bugfix.rst
+++ b/changes/534.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed broken .briefcase directory permissions in some cases during linux builds.
--- a/src/briefcase/integrations/docker.py
+++ b/src/briefcase/integrations/docker.py
@@ -289,6 +289,10 @@ class Docker:
             else:
                 docker_args.append(arg)
 
+        # Ensure the .briefcase directory has been created before mounting it as
+        # a volume or else inside the container it will be owned by root.
+        self.command.dot_briefcase_path.mkdir(exist_ok=True)
+
         # Invoke the process.
         # Any exceptions from running the process are *not* caught.
         # This ensures that "docker.run()" behaves as closely to
--- a/tests/platforms/linux/appimage/test_build.py
+++ b/tests/platforms/linux/appimage/test_build.py
@@ -26,6 +26,9 @@ def first_app(first_app_config, tmp_path):
     (app_dir / 'usr' / 'app_packages' / 'secondlib' / 'second_a.so').touch()
     (app_dir / 'usr' / 'app_packages' / 'secondlib' / 'second_b.so').touch()
 
+    # Create home directory
+    (tmp_path / 'home').mkdir(parents=True, exist_ok=True)
+
     return first_app_config
 
 


freakboy3742 (changes requested):
I can see *what* you're doing here... however, I'm not completely clear *why*.

Assuming Docker is the first process to try and access `~/.briefcase`, permissions will definitely be a problem... but Docker shouldn't be trying to access that folder until *after* the template has been rolled out and tools have been downloaded, both of which should be touching (and thus creating) `~/.briefcase`. So - what sequence of events caused this to happen? 

To be clear - I'm not doubting that you have seen this problem. I just can't work out how - and that means it's a sufficiently complex path through the codebase that it calls for a specific test case that reproduces those conditions so that future refactoring won't break that use case.

geogriff:
I believe the only relevant options in my case were  `support_package`. IIUC, when using `support_package` with a fs path, the support package directory isn't copied to `.briefcase`, because there's no reason to do so. As for the `template` option, that isn't relevant because template unrolling happens in the `briefcase build` step which is after the first docker command has been run by the `briefcase create` command. Does that make sense?

freakboy3742:
Template unrolling definitely happens in the `create` step - the `Dockerfile` is part of the generated template content. However, your comment about manually specifying a support package might point to the cause - if you had a custom support package *and* a custom template, you wouldn't be caching anything, and as a result, the `~/.briefcase` folder wouldn't be created. It's definitely an edge case - I would expect most users wouldn't be using custom templates *and* custom support packages on their first outing, and as soon as you've done either of those things, you'll have a `.briefcase` folder - but it's definitely possible.

The PR is a good fix for this; however, it definitely needs a test to confirm that the directory is created.

geogriff:
> Template unrolling definitely happens in the `create` step - the `Dockerfile` is part of the generated template content.

Ha, good point. I must have misinterpreted the output.

> if you had a custom support package and a custom template, you wouldn't be caching anything, and as a result, the ~/.briefcase folder wouldn't be created

Hmm except I wasn't using a custom template. IIUC templates are cached in `~/.cookiecutters` rather than `~/.briefcase`.

Here's an elided version of the failing build output, for reference:
```
[boundery] Generating application template...
Using app template: https://github.com/beeware/briefcase-linux-appimage-template.git

[boundery] Installing support package...
Using custom support package python-linux-support/dist/Python-3.7-linux-x86_64-support.custom.tar.gz
Unpacking support package...

[boundery] Installing dependencies...
[boundery] Entering Docker context...

[boundery] Building Docker container image...
```
...
```
[boundery] Leaving Docker context.

[boundery] Installing application code...
Installing boundery...
Installing linux/generated_src...

[boundery] Installing application resources...

[boundery] Created linux/Boundery Client
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.7/dist-packages/briefcase/__main__.py", line 26, in <module>
    main()
  File "/usr/local/lib/python3.7/dist-packages/briefcase/__main__.py", line 11, in main
    command(**options)
  File "/usr/local/lib/python3.7/dist-packages/briefcase/commands/package.py", line 64, in __call__
    state = self._package_app(app, update=update, **full_options(state, options))
  File "/usr/local/lib/python3.7/dist-packages/briefcase/commands/package.py", line 32, in _package_app
    state = self.build_command(app, **full_options(state, options))
  File "/usr/local/lib/python3.7/dist-packages/briefcase/commands/build.py", line 60, in __call__
    self.verify_tools()
  File "/usr/local/lib/python3.7/dist-packages/briefcase/platforms/linux/appimage.py", line 157, in verify_tools
    self.linuxdeploy = LinuxDeploy.verify(self)
  File "/usr/local/lib/python3.7/dist-packages/briefcase/integrations/linuxdeploy.py", line 42, in verify
    linuxdeploy.install()
  File "/usr/local/lib/python3.7/dist-packages/briefcase/integrations/linuxdeploy.py", line 62, in install
    download_path=self.command.tools_path
  File "/usr/local/lib/python3.7/dist-packages/briefcase/commands/base.py", line 469, in download_url
    download_path.mkdir(parents=True, exist_ok=True)
  File "/usr/lib/python3.7/pathlib.py", line 1258, in mkdir
    self._accessor.mkdir(self, mode)
PermissionError: [Errno 13] Permission denied: '/home/build/.briefcase/tools'
```

So, the create step had finished, and fetching LinuxDeploy fails early in the build step.

I'm not super comfortable writing a test, as I don't know this project's unit testing preferences. (I have questions like: do we mock `dot_briefcase_path.mkdir` or turn this into `os.isdir` and `os.mkdir` and mock those?) Feel free to take that part over if you get to it.

PR #543, opened 2020-12-14
Error reporting support package

Agyey:
## PR Checklist:


Agyey:
Implement Architecture Check in custom support package URL and added test for the same.

src/briefcase/commands/create.py
```diff
@@ -315,6 +326,11 @@ def install_app_support_package(self, app: BaseConfig):
             # the support package URL.
             try:
                 support_package_url = app.support_package
+                # Check if the URL mentions an unsupported architecture
+                url_parts = list(urlsplit(support_package_url))
+                for key, value in parse_qsl(url_parts[3]):
+                    if key == 'arch' and value != 'x86_64':
```
freakboy3742, line 332: This is a very specific check that will (a) break if you provide a custom support package URL, (b) only be true for Linux, and (c) could even change on Linux if someone wanted to submit a support packages for a different platform. For example, support for Raspbian isn't out of the question, which would mean arm64 support would be valid. Apple platforms already need to support arm64.

To that end, we should be looking for failures cloning the support package, and providing guidance on ways to fix the problem if the template can't be found, rather than pre-emptively assuming that a specific key and value will be present in the URL.

freakboy3742 (changes requested):
Definitely a good attempt, but you've missed a lot of possible use cases where this approach won't work. Details on a better approach inline.

src/briefcase/commands/create.py
Agyey: Since a MissingNetworkResourceError would be raised when a support package is unavailable. Do you think the following change to the message would resolve the problem?

Current Message: 
   "Unable to download {url}; is the URL correct?"
New Message:
   "Unable to download {url}; is the URL correct? Suggestion : Compile the support package yourself, add a support_package reference to pyproject.toml file"

freakboy3742 (changes requested):
The problem with this approach is that MissingNeworkResourceError is raised whenever *any* file is missing - it's raised by `download_url`, which yes, is used when downloading the support file, but it's also used to download other resources (Java SDK, Android SDK, WiX, Linuxdeploy, and more). So the error message is overly specific in the way you've defined it. If the failure occurs downloading Java, for example, you shouldn't get an error suggesting that you should recompile your support package.

This is a case where you need to catch the MissingNetworkResource error and interpret what that means in the appropriate context.

Agyey:
I assume since `MissingNetworkResourceError` is passed only the url, we would need to pass the `resource_type` as an additional parameter, but this doesn't seem to be viable approach. Alternatively when a  `MissingNetworkResourceError` exception is raised in `install_app_support_package` an Exception could be raised to signify that the package url is invalid and a custom compile of the package is required.

### Original Code
```
                    support_filename = self.download_url(
                    url=support_package_url,
                    download_path=self.dot_briefcase_path / 'support'
                )
            else:
                support_filename = Path(support_package_url)
        except requests_exceptions.ConnectionError:
            raise NetworkFailure('downloading support package')
```

### Suggested Change
```suggestion
                    support_filename = self.download_url(
                    url=support_package_url,
                    download_path=self.dot_briefcase_path / 'support'
                )
            else:
                support_filename = Path(support_package_url)
        except MissingNetworkResourceError:
            raise InvalidSuportPackageURL(support_package_url)
        except requests_exceptions.ConnectionError:
            raise NetworkFailure('downloading support package')
```

changes/541.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+- Created a new exception class InvalidSupportPackageURL
+- MissingNetworkResourceError is caught in install_app_support_package and InvalidSupportPackageURL is raised
+- Test Case test_support_package_url_with_invalid_url added to check the new exception
```
freakboy3742, line 3: The purpose of this file is to provide the text for the release note associated with this feature. You shouldn't be describing the implementation approach, or the code changes - you should be describing the feature that has been added or bug that has been resolved, as experienced by the end user.

tests/commands/create/test_install_app_support_package.py
```diff
@@ -85,6 +85,17 @@ def test_install_custom_app_support_package_file(create_command, myapp, tmp_path
...
 
+def test_support_package_url_with_invalid_url(create_command, myapp):
+    "Invalid URL raise MissingNetworkResourceError"
+    # Provide an invalid package URL with unsupported arch
+    url = 'https://briefcase-support.org/python?platform=linux&version=3.6&arch=i686'
+    myapp.support_package = url
+
+    # Confirm that an Exception is Raised
+    with pytest.raises(InvalidSupportPackageURL):
+        create_command.install_app_support_package(myapp)
```
freakboy3742, line 96: From the look of it, this will do a live hit on the support site when the test is run. This is an externality on the test - if, for example, we added i686 support to briefcase-support.org, this test would start failing. The test will also fail if you're not currently connected to a network (e.g., if you're on a plane).

Any externalities should be isolated with mocks so that the test is reliable.

src/briefcase/commands/create.py
```diff
@@ -36,6 +36,17 @@ def __init__(self, filename):
...
 
+class InvalidSupportPackageURL(BriefcaseCommandError):
+    def __init__(self, url):
+        self.url = url
+        super().__init__(
+            "Invalid support package URL {url}; Compile the support package yourself, "
+            "add a support_package reference to pyproject.toml file".format(
+                url=url
+            )
+        )
```
freakboy3742, line 47: This is a user-facing error, so it's really important that it is both correct and actionable. In this case, the error could be caused: 

 * requesting a Python version that isn't supported
 * requesting an architecture that isn't supported
 * an invalid *custom* support package URL

"Invalid support package URL" isn't a helpful message for the first two cases; the user hasn't specified anything, so they haven't specified anything that is "invalid". 
 
"Compile the support package yourself" isn't a helpful message if you've already specified a custom support package.

We need to be able to differentiate between case 3 and case 1 & 2.

freakboy3742 (changes requested):
Getting closer! Comments inline.

changes/541.bugfix.rst
Agyey: Got it, I'll revise this

src/briefcase/commands/create.py
Agyey: For the cases where the Python Version and/or the Architecture isn't supported, do we maintain a list of the supported Versions and Architecture that can be referenced? 

Suggested Messages per Case :
- **For the Invalid Custom URL**
_Invalid support package URL_
-  **For Unsupported Architecture and Version**
_Unsupported Architecture/Version, please compile a support package yourself and add a support_package reference to pyproject.toml file_

tests/commands/create/test_install_app_support_package.py
Agyey: Should I provide a dummy url instead, like abc.xyz or should a invalid url structure be also checked?

freakboy3742: It's not the URL that is the problem - it's the fact that you're calling a network service. Here's the check: turn off wifi on your laptop, and run the test. The test should still pass. 

You need to add a mock for the network access so that *actual* network access isn't required, and we can control the response from the network response to produce a predictable response.

src/briefcase/commands/create.py
freakboy3742: We don't maintain in *internal* list of supported versions and architectures - a 404 from the support platform download is the canonical source of truth for the list of supported versions and architectures.

As for the messages: A good error message should have two parts - a description of how the problem manifested, and a suggested course of action. "Invalid" isn't a great term to use - *why* is invalid? Was it badly formatted? Did it contain characters that aren't permitted? Or (as in this case) did the requested resource not exist?

However, don't get too distracted by the specific text in this case - no matter what you write, I'll almost certainly tweak it, so it's not worth worrying about it too much. :-)

Agyey: Got it thanks

tests/commands/create/test_install_app_support_package.py
Agyey: Is there some material I can refer learn how to create a mock for the network? Or would I find a similar application in the other tests?

freakboy3742: We use unittest.Mock as a framework; there's plenty of documentation online for how to use that (including Python's own builtin documentation). The tests for tests/commands/base/test_download_url.py has an example of mocking the requests library.

freakboy3742:
@Agyey Just pinging this for a status update; we've had another volunteer express an interest in the issue this PR will be resolving. Are you intending to revisit this PR and finish it, or can we hand the work off to them?

Issue #424, opened 2020-06-13
Allow specification of output filename to package command


dybber:
Hi there

I would like to suggest that users can specify a destination filename when executing `briefcase package`. The reasoning being, that you often would want automated CI scripts to run `briefcase`, and such scripts needs and easy way to locate file. Instead of these scripts guessing the right format <formal_name>-<version>.dmg for example, it would be nice if they could just specify where they wanted the output to be located and with what name.

For instance, a user would include a line like this in a script:
```
briefcase package macOS dmg --no-sign --output=dist/mypackage.dmg
```
and then the script can just assume where it has to look. Currently, a script or a Makefile would have to look up the version number in pyproject.toml, to be able to guess the destination.

freakboy3742:
Thanks for the suggestion! I can definitely understand the problem you're describing; I guess the question is whether full customization like you describe is the best approach. Customization is always a double edged sword - yes, it gives the end user more control, but at the cost of more complicated internals (which can ultimately lead to more bugs). So, adding new options/features is something we need to consider carefully.

Looking for analogies with other packaging tools: Python's `wheel`, for example, doesn't give any flexibility over where it outputs wheels - it uses a fixed name that includes a version number, and places the output in the dist directory. However, *all* output artefacts will appear in the `/dist` folder, and `twine` can be told to upload all files in the `/dist` folder. Would having a single reliable output folder (like `dist`) improve the CI situation?

Another possible approach: #360 is already tracking a feature request to extract the app's version number from the command line using `briefcase project version` and/or `briefcase app version -a myapp`; would exposing other properties (like the output path) `briefcase app output_path -a myapp` (returning the path to the distributable artefact) satisfy your use case?

dybber:
I guess the latter approach would work, being able to read the version number is actually enough, as the user can then construct the output path from that. It's hard to parse a pyproject.toml file from commandline, to extract the version number.

However, for anyone else with this issue, here's how I ended up getting Travis to find the file:

```
  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then export DMGFILE=`bash -c "echo macOS/mypackage-*.dmg"`; fi
```
I use bash to be able to use glob-patterns, in that way I avoided knowing the exact version number, and then I could just do a `mv $DMGFILE dist/mypackage.dmg` afterwards.

freakboy3742:
[...]

Issue #360, opened 2020-04-17
Add "version" command to briefcase


saroad2:
Taking inspiration from [poetry](https://python-poetry.org/docs/pyproject/), we should implement a "version" command for "briefcase". For example, given the following *pyproject.toml* file:

```toml
[tool.briefcase]
project_name = "EddingtonGUI"
bundle = "com.eddington_gui"
version = "0.0.2.dev0"
url = "https://eddington_gui.com/eddington_gui"
license = "Apache Software License"
author = 'Sagi Shadur'
author_email = "saroad2@gmail.com"
```

The command `briefcase version` should output "0.0.2.dev0"

saroad2:
@freakboy3742 , can I implement this command?

freakboy3742:
+1 to the feature idea. Longer term, adding "bumping" options analogous to poetry's would also make sense.

I have a *mild* concern about differentiating between `briefcase version` and `briefcase --version`; from a UX perspective, it's not immediately obvious which is Briefcase's version, and which is the project's version.

Would gathering "project management" commands under `briefcase project` or `briefcase manage` namespace make sense to you as a user? That would make the version command `briefcase project version` or `briefcase manage version`. That provides a *little* more context to the "version" request; and also provides an entry point for other project management options in future.

saroad2:
I prefer `briefcase project version` much better than `briefcase manage version`, but I must say that I think users will understand the difference between `briefcase version` and `briefcase --version`. This is absolutely up to you.

Once you decide what syntax you want the command to be, I'll implement it.

freakboy3742:
In which case - lets go with `briefcase project version`.

saroad2:
I started to implement this feature and unfortunately hit a wall.
When I run `briefcase project version`, *briefcase* interpret *version* as the platform...
This is accruing because of the way *cmdline.py* parser is implemented.

As I see it, there are 2 optional solutions to the problem:

The first one to change the command to my initial suggestion of using the command `briefcase version`

The more complicated option is to split the `parse_cmdline` function to 3 stages:

1. Parse only the *command* first (new, dev, create, etc.)
2. If it is a platform-based command:
2.1 Parse platform and output format
3. Parse specific command flags

As always, this is up to your consideration.

freakboy3742:
I tend to agree that splitting up the command processing makes a sense. Parsing the platform is something that should only be occurring on the "platform-based" commands, so it makes sense to factor that out, especially if it permits other command types (like `project`)

[...]

PR #558, opened 2021-01-22
User define output dir

saroad2:
The user can now define in which directory to save *Briefcase*'s output.

**Why is it good for us?**
Assume you want to build an app using a CI/CD process. Moreover, assume you want to build an app for every platform possible (windows, mac, android, you name it).

Using the current method of Briefcase, each platform will save its output to a different directory and you would have to state for each job where the bundle is saved. This makes the CI/CD process much complicated than it should be.

**What the proposed solution?**
The user can now define a custom output directory and everything will be saved to this directory.
When not specified, the output directory would be the same as it is now, based on the platform.

Another cool thing that one might do is to define a custom output directory *for a specific platform*. This can also be done by the new feature.

## PR Checklist:


freakboy3742 (changes requested):
The initial problem description makes sense, and I agree that there is value in having the output destination be customizable (or, at least, something that can be determined programatically - which is the suggestion that was raised in the earlier discussion; see #360 and #424). 

Two smaller issues first: 
1. Naming conventions: I've been trying (to varing degrees of success) to follow a "Hungarian-light" approach with naming of paths - `*_dir` is a string that describes a file location; `*_path` is a Pathlib object. In this case, the user-specified data would be `output_dir` (or just `output`); but the internally maintained method would be `output_path`, with the conversion occurring as early in the process as possible.
2. This needs documentation, as it's a significant change to the external API.

Now for the larger concern/problem:

If I'm reading this right, this PR implements `output_dir` as an option specified on the app in `pyproject.toml`. That's a problematic thing to specify in project config, because the output directory is machine specific. If I specify an output path of `/Users/rkm/output`, that path won't be usable on anyone's computer except mine. This makes me think that the output path should be a command line option to a specific build command, rather than a hard-coded project option.

We also need to be clear whether this is a "final output" path, or a "working build directory" path. Using `python setup.py bdist_wheel` as an analogy, this is the difference between the `build` and `dist` directories. This PR mixes the two, which I'm not convinced is a good idea. Having projects separated by platform means we only need to guarantee that templates can co-habitate on a single platform (i.e., app X and app Y can both be rolled out into the same platform folder). If we make the build directory customizable, we also need to guarantee that app X on macOS won't collide with app X on Windows - and given the exponential complexity that introduces, I'm not sure it's a good idea. At the very least, we need to differentiate between build and dist directories, and ensure there is a platform-derived name appended to the `build` folder.

Of course, this then introduces the complexity that where you're building won't necessarily be where the output end up - which complicates the interaction between successive calls to `briefcase create`, `briefcase build`, `briefcase run` and `briefcase package`. 

For me, this also raises the question of whether we want to adopt a `dist`-style output folder as a default behavior. Although I originally implemented output landing in the platform directory, I do see value in keeping final distributable artefacts separate from the intermediate build folder (if only for the "upload everything in the dist folder").

Lastly, there's a consideration that *might* make this entire discussion moot - the use of the `publish` command. This discussion has been based on the assumption that briefcase needs to output into a consistent location so that output can be found reliably by a CI system for post-processing - but publication *could* (and, was, in my original design of the tool) be something managed by briefcase itself, uploading to a target folder. If uploading is managed by briefcase, there's no real need to customise anything, because the file name is internally visible. Alternatively, one mode of "publication" that might be useful for CI purposes could be "copy the output file to this destination folder", which meets the requirement of having a reliable output location, but handles it as a publication method rather than reworking the internals of Briefcase.

That's a lot of food for thought - I'd be interested in your thoughts.

freakboy3742:
I'm going to close this PR; firstly because it needs a lot of updating, but also because I'm not convinced the approach to the solution is correct. 

There's some discussion about this feature #424. Based on that discussion, I suspect the better approach will be:
* a "distutils"-like approach with a consistent known output folder; or
* a project-level command that outputs the CI output path for a platform.

My personal preference is for the former, but we can continue discussion on #424.

Issue #460, opened 2020-07-26
On Android, if your app is called `x.y.switch`, it fails to build


paulproteus:
**Describe the bug**

If you call your app `switch` (or another Java reserved word), you will be unable to build it on Android.

**To Reproduce**

You can reproduce by creating an app whose module_name is `switch`. For example, you can use a Toga demo app.

1. Download the toga source code, then `cd` into `examples/switch`
2. Run `briefcase create android && briefcase build android`
3. Wait for the build to complete (about 60 seconds if you've used Briefcase on this system to make Android apps)
4. See an error:

```
> Task :app:processDebugResources FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:processDebugResources'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.Workers$ActionFacade
   > Package 'org.beeware.switch' from AndroidManifest.xml is not a valid Java package name as 'switch' is a Java keyword.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 9s
15 actionable tasks: 15 executed

Error while building project.
```

**Expected behavior**

The app should build or briefcase should print an error at the `briefcase new` phase, or briefcase should print an error before trying to run `briefcase create android`.

**Environment:**
 - Operating System: macOS 10.15
 - Python version: 3.7
 - Software versions:
   - Briefcase: 0.3.3
   - Toga: current git (0.3.0pre21)

**Additional context**

freakboy3742:
Agreed that this is an edge case we need to catch - probably very early in the process.

App names will also be restricted by Python reserved words - you couldn't name your app "def" or "import" for example.

We also need to validate the parts of the bundle ID - `org.switch.goodname` will also be a problem on Android.

sarthaksharma058439:
Check this video .....he make calculator in beeware and deploy to android device

https://youtu.be/RllR6FRdG6w

junefish:
Hello!  I would like to try fixing this.

freakboy3742:
@junefish By all means - go ahead! If you need any pointers or suggestions, let us know!

KabirAcharya:
I believe my fork should fix this issue.

PR #565, opened 2021-02-14
Fix for Issue #460: On Android, if your app is called `x.y.switch`, it fails to build

KabirAcharya:
Stumbled upon this issue while looking for where I could attempt my first GitHub contribution.
After some testing involving Issue #460 I made this quick desk check.

Name: switch | BUILD FAILED
Bundle: com.switch | BUILD FAILED
Name: def | BUILD SUCCESS & DOES NOT RUN
Bundle: com.def | BUILD SUCCESS & RUNS

Using this information I have made changes to `briefcase new` which prevents users from setting a package name or bundle identifier which triggers this issue.

### DIFF
--- a/changes/562.misc.rst
+++ b/changes/562.misc.rst
@@ -1,1 +1,0 @@
-Drop python 3.5 support!
--- a/changes/562.misc.rst
+++ b/changes/562.misc.rst
@@ -1,0 +1,1 @@
+Drop python 3.5 support!
--- a/src/briefcase/commands/new.py
+++ b/src/briefcase/commands/new.py
@@ -16,7 +16,7 @@ from .create import InvalidTemplateRepository
 
 VALID_BUNDLE_RE = re.compile(r'[a-zA-Z0-9-]+(\.[a-zA-Z0-9-]+)+$')
 
-javareserved = [
+reservedwords = {
     "abstract", "assert", "boolean", "break", "byte", "case", "catch",
     "char", "class", "const", "continue", "default", "do", "double",
     "else", "enum", "extends", "final", "finally", "float", "for",
@@ -26,7 +26,7 @@ javareserved = [
     "super", "switch", "synchronized", "this", "throw", "throws",
     "transient", "try", "void", "volatile", "while", "true", "false",
     "null"
-]
+}
 
 
 def titlecase(s):
@@ -125,11 +125,15 @@ class NewCommand(BaseCommand):
             raise ValueError(
                 "A '{candidate}' directory already exists. Select a different "
                 "name, move to a different parent directory, or delete the "
-                "existing folder.".format(candidate=candidate)
+                "existing folder.".format(
+                    candidate=candidate,
+                )
             )
-        if iskeyword(candidate) or candidate in javareserved:
+        if iskeyword(candidate) or candidate in reservedwords:
             raise ValueError(
-                "App name may not contain Java or Python reserved words."
+                f"App name may not contain the reserved word '{candidate}'.".format(
+                    candidate=candidate,
+                )
             )
         return True
 
@@ -158,9 +162,11 @@ class NewCommand(BaseCommand):
                 "include letters, numbers, and hyphens."
             )
         for word in candidate.split("."):
-            if word in javareserved:
+            if word in reservedwords:
                 raise ValueError(
-                    "Bundle name components may not contain Java reserved words."
+                    "Bundle name component may not contain the reserved word '{candidate}'.".format(
+                        candidate=candidate,
+                    )
                 )
         return True
 
@@ -349,7 +355,7 @@ for your application. This name must be PEP508-compliant - that means the name
 may only contain letters, numbers, hyphens and underscores; it can't contain
 spaces or punctuation, and it can't start with a hyphen or underscore.
 
-Due to build requirements, the app name may also not be a Java or Python keyword.
+Due to build requirements, the app name should avoid reserved words in common programming languages.
 Based on your formal name, we suggest an app name of '{default_app_name}',
 but you can use another name if you want.""".format(
                 default_app_name=default_app_name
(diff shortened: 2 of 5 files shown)


KabirAcharya:
Not sure why I failed the towncrier-check but I have tested my build and it works as expected.

KabirAcharya:
*Read the docs*

KabirAcharya:
Sorry for the long list of redundant commits, very new to GitHub. Will keep format testing local from now on.

changes/562.misc.rst
```diff
@@ -1 +0,0 @@
-Drop python 3.5 support!
```
freakboy3742, line 1: This file exists for a reason - it shouldn't be deleted in your merge.

src/briefcase/commands/new.py
```diff
@@ -9,11 +9,25 @@
...
 from briefcase.exceptions import NetworkFailure
 
+from keyword import iskeyword
+
 from .base import BaseCommand, BriefcaseCommandError
 from .create import InvalidTemplateRepository
 
 VALID_BUNDLE_RE = re.compile(r'[a-zA-Z0-9-]+(\.[a-zA-Z0-9-]+)+$')
 
+javareserved = [
```
freakboy3742, line 19: A really minor thing - but this should probably be a set; it's a micro-optimisation in this particular case, but it's a good habit to use datastructures that match the way in which they will be used. In this case, we want membership, and don't care about order, so set is a much better fit.

freakboy3742, line 19: We also want to avoid calling out Java specifically. If you're not writing an Android app, it's going to be weird to be warning you about a language that you're not using. There are also other languages that we need to care about - at the very least Javascript and Objective C - so this should be a more generic "reserved" list, rather than *Java* reserved specifically.

src/briefcase/commands/new.py
```diff
@@ -113,7 +127,10 @@ def validate_app_name(self, candidate):
                 "name, move to a different parent directory, or delete the "
                 "existing folder.".format(candidate=candidate)
             )
-
+        if iskeyword(candidate) or candidate in javareserved:
```
freakboy3742, line 130: Nice usage of the Python standard library!

src/briefcase/commands/new.py
```diff
@@ -113,7 +127,10 @@ def validate_app_name(self, candidate):
...
+            raise ValueError(
+                "App name may not contain Java or Python reserved words."
```
freakboy3742, line 132: As above - we should avoid calling out specific languages, and just refer to reserved words. For bonus points, we should also clarify what the reserved word *is* - i.e, say `The app name "switch" cannot be used, as it is a reserved word`.

src/briefcase/commands/new.py
```diff
@@ -140,6 +157,11 @@ def validate_bundle(self, candidate):
                 "least 2 dot-separated sections, and each section may only "
                 "include letters, numbers, and hyphens."
             )
+        for word in candidate.split("."):
+            if word in javareserved:
+                raise ValueError(
+                    "Bundle name components may not contain Java reserved words."
```
freakboy3742, line 163: Same comment about the message here.

src/briefcase/commands/new.py
```diff
@@ -327,6 +349,7 @@ def build_app_context(self):
 may only contain letters, numbers, hyphens and underscores; it can't contain
 spaces or punctuation, and it can't start with a hyphen or underscore.
 
+Due to build requirements, the app name may also not be a Java or Python keyword.
```
freakboy3742, line 352: The guidance here should say to "app name should avoid reserved words in common programming languages".

freakboy3742 (changes requested):
Nice work! I've flagged a couple of minor tweaks (mostly around removing specific references to Java); but on the whole, this is a great first pass.

On top of those minor changes, there's one additional piece of validation required. You've added validation during the `new` command, but that command is a wizard for creating new projects. You can write a `pyproject.toml` from scratch, without using the wizard at all; and you can manually modify `pyproject.toml` after it has been generated by `briefcase new`. We should ensure that when you get to `briefcase create`, no matter how you created `pyproject.toml`, an invalid app or bundle name is rejected there as well. You'll notice, for example, that the GlobalConfig object validates version numbers for PEP440 compliance, and AppConfig validates PEP508 naming, as well as some other properties.

KabirAcharya:
@freakboy3742  It seems like `briefcase create` is completely missing validation (unless I'm missing something). Should validation be implemented as a module and applied to multiple commands?

freakboy3742:
@KabirAcharya No worries about the delay - you're a volunteer, so we understand that work/studies need to take priority. 

As for `briefcase create` - the command itself doesn't contain any validation, but the configuration itself has validation. `create` will load the configuration it can find; if that configuration isn't valid, it's the configuration that raises the error. This means validation is effectively shared by *all* commands (other than `new`) - but in practice, `create` is where you're going to actually see any validation.

[This test](https://github.com/beeware/briefcase/blob/master/tests/config/test_AppConfig.py#L99), for example, does a lot of checks of valid names; [this test](https://github.com/beeware/briefcase/blob/master/tests/config/test_AppConfig.py#L125) checks some known invalid names.

However, your inclination is correct - we should be using the same code for both the wizard and the config/command validator; factoring that out into a utility method that can be shared by both implementations (and tested independent of the new/create commands) would be desirable.

Viridescence:
If desirable, I could get working on this patch and merge it with testing for illegal windows filenames such as "CON" seeing as it has been 17 days. If @KabirAcharya wants to work on it, just say so! I understand you have a lot to do in university right now. As I am new to open source, would someone like to give me guidance for this patch?

KabirAcharya:
> If desirable, I could get working on this patch and merge it with testing for illegal windows filenames such as "CON" seeing as it has been 17 days. If @KabirAcharya wants to work on it, just say so! I understand you have a lot to do in university right now. As I am new to open source, would someone like to give me guidance for this patch?

I'll be on to it very soon. Sorry for the big delays. Merging is a very good idea. I looked at your pull and we both address a similar problem so merging is definitely the way to go.

freakboy3742:
Superseded by #617 and #674.

PR #573, opened 2021-02-26
Prevent invalid app names on Windows

Viridescence:
I am aware that this is too much of a special case to be committed, and I coded it for some practice. However, I thought I might as well submit a pull request.

 This code (line 123) prevents an app being called invalid filenames such as PRN or CON on windows, which would result in a long error not easily understandable:
image
image

The names CON, PRN, AUX, NUL, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, COM0, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, LPT9 and LPT0 are not allowed on windows.
This PR checks for these invalid names, and (checking the system by importing platform) throws a ValueError which tells the user exactly what the error is. It also adds a warning for mac/linux users who call their file these names, telling them that this makes them incompatible with Windows. These filenames make creating on mac/linux then downloading to windows impossible - the download is stuck in the middle and gives no error message with some downloading methods like Drive. For obvious reasons, this is confusing and problematic. This PR gives an early warning of it.

## PR Checklist:

Checked with flake8 and tox, as well as manually
Towncrier fragment added, but Docs have not been updated.

### DIFF
--- a/changes/573.misc.rst
+++ b/changes/573.misc.rst
@@ -1,0 +1,1 @@
+Added check for invalid filenames like CON on windows, and warning for non-windows users.
--- a/src/briefcase/commands/new.py
+++ b/src/briefcase/commands/new.py
@@ -1,5 +1,6 @@
 import re
 import subprocess
+import platform  # For Windows invalid filenames like CON
 from email.utils import parseaddr
 from typing import Optional
 from urllib.parse import urlparse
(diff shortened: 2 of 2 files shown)


Viridescence:
Hmm... I don't know why the smoketest is failing.

freakboy3742 (changes requested):
Thanks for the suggestion & patch - definitely a good idea to warn about this edge case.

The test case failures you're seeing appear to be due to beeware/briefcase#574, which is a very recent change in behavior with setuptools. I've just committed a patch for this; if you merge your feature branch with the upstream master, your tests should start passing.

On the subject of which - you're going to find that very hard to do, because you've submitted your pull request from your fork's master. A general pattern you'll find with git is that you should *never* commit to master - especially if you're on a fork of someone else's repository - because it makes long term maintenance a real headache. It's generally best to keep *your* master as a direct mirror of the upstream master, and then do all feature development in a branch, and submit pull requests from your feature branch to the upstream's master. This also means that the upstream maintainer can commit minor tweaks to your pull request (e.g., to fix a typo in a docstring, or some other minor change that is easier for the maintainer to fix directly than it is to go through a 'comment and wait for the submitter to edit' cycle. Using a branch also means you can have multiple pull requests under review at the same time.

Another minor procedural suggestion - when you're providing stack traces and console logs, it's *much* more helpful to provide them as text, rather than screenshots. Text can be searched; screenshots can't. A screenshot should only be used when the output is graphical (e.g., showing a font that isn't being displayed correctly).

Now for the patch itself: what you've done here is great, but there's more needed. There's a lot of overlap here between your patch, and PR #565 (which is a fix for #460). There is a class of names that shouldn't be legal, or should only be legal with a warning to the user. The feedback I provided on #565 also applies here: you've updated the *wizard*, which is great - but you don't have to use the wizard. You can manually write a briefcase.toml from scratch, without ever using the wizard. That means there's a second place that we need to check and provide a warning - during `briefcase create`. 

Related to this - we are clearly acquiring a non-trivial set of "these names aren't legal/advisable", which suggests to me that we should be factoring that check into a standalone method that encompasses *all* the possible checks. 

The one significant improvement that your patch has over #565 is the 'warning' behavior - which is a very nice touch. CON etc definitely needs to be prohibited on Windows, but if I know my app will never be used on Windows, and I have a legitimate reason to name my app CON, then there's no reason I shouldn't be allowed to do so, provided I'm adequately warned. I'd suggest a slight tweak to the behavior you've got here. In the Wizard, a warning to the console *could* be inadvertently ignored. I'd suggest adding a confirmation "Are you sure y/n" in the non-Windows case. This confirmation isn't required in the `briefcase create` version - once you've got an active project, a warning message is more than enough.

Lastly, the patch is also missing tests. Tests help us ensure that code quality is maintained over time, and we don't miss/forget any edge cases. As a result, we don't add any new features without a corresponding test.

Viridescence:
Thanks for the feedback! I will implement the changes  and use forks this time, and submit a new PR: If I submit the code, it can easily be copied, reviewed, and implemented into #565 , into briefcase create, and eventually into the standalone method.

PR #575, opened 2021-02-26
Fix listing output formats for macOS and iOS

### LINKED ISSUE #574: Output formats for a platform are not correctly detected on macOS and iOS
**Describe the bug**
The method `briefcase.platforms.get_output_formats` fails to list output formats for iOS and macOS after updating to setuptools v53.1.0 or higher.

**To Reproduce**
Run `briefcase create macOS app` in the terminal. This will fail with the error message:

```
Invalid format 'app'; (choose from: )
```

**Environment:**
 - Software versions:
   - briefcase 0.3.4
   - setuptools 53.1.0

**Additional context**
Since setuptools 53.1, entry detected by `pkg_resources.iter_entry_points()` are no longer lower-case but preserve the casing from setup.cfg.

Since briefcase always converts the argument to the this call to lower case, listing entry points of "macOS" and "iOS" (note the casing!) fails.

SamSchott:
Fixes #574.

There where two options to this:

1. Do not convert the platform argument to lower case when listing entry points. This fixes the listing when setuptools v53.1 and higher is installed but breaks it for earlier releases.
2. Switch to lower case entry points for all platforms. This allows compatibility with a larger number of setuptools releases and results in more uniform entrypoint names. The entrypoints for Windows and Linux are also lower case, after all.

I've opted for the second one but wouldn't mind either.

## PR Checklist:


freakboy3742 (approved):
Thanks for the diagnosis and fix!

PR #576, opened 2021-02-27
Catch invalid windows names and issue warning for other systems

Viridescence:
New PR, made with fork this time. Suggestions such as verification message implemented. Does not need to be immediately merged if not desirable, as a larger solution for invalid filenames may be found in conjunction with PR #565.

Prevent windows users from calling their files invalid names, and issue a warning for Mac/linux users.

(copied from previous PR) 
This code (line 123) prevents an app being called invalid filenames such as PRN or CON on windows, which would result in a long error not easily understandable:

The names CON, PRN, AUX, NUL, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, COM0, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, LPT9 and LPT0 are not allowed on windows.
This PR checks for these invalid names, and (checking the system by importing platform) throws a ValueError which tells the user exactly what the error is. It also adds a warning for mac/linux users who call their file these names, telling them that this makes them incompatible with Windows and adding a confirmation question. These filenames make creating on mac/linux then downloading to windows impossible - the download is stuck in the middle and gives no error message with some downloading methods like Drive. For obvious reasons, this is confusing and problematic. This PR gives an early warning of it.

## PR Checklist:

### DIFF
--- a/changes/576.misc.rst
+++ b/changes/576.misc.rst
@@ -1,0 +1,1 @@
+Added check for illegal filenames like CON or PRN in briefcase new. Prevents these names in Windows, and adds warnings for them in mac/linux.
--- a/src/briefcase/commands/new.py
+++ b/src/briefcase/commands/new.py
@@ -1,5 +1,6 @@
 import re
 import subprocess
+import platform  # For Windows invalid filenames like CON
 from email.utils import parseaddr
 from typing import Optional
 from urllib.parse import urlparse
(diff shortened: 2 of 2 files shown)


Viridescence:
@freakboy3742 Do you recommend this be put on hold until #565 is completed / a standalone function is created, and then added to their code? Or should I go ahead, fix, test and PR?

freakboy3742:
The author of #565 (@KabirAcharya) hasn't updated the patch in the 2 weeks since feedback was provided. It's not clear if they're just busy, or if they've lost interest; I've tagged them in this comment to see if we can elicit a status update. Since neither patch currently contains the abstracted "name is valid" mechanism, I'd say it's fair game for you to add that capability to this patch; if we don't hear from @KabirAcharya in a day or two, integrating the content of #565 into this patch would also be fair game.

Viridescence:
@freakboy3742 It seems @KabirAcharya is back and working on it, I'll implement it as an addition to their work once their PR is complete, unless they want some help.

freakboy3742:
Superseded by #685

PR #617, opened 2021-09-05
Added reserved python keywords

elliegraves:
Adding a reserved python keyword list. 
## PR Checklist:


src/briefcase/commands/new.py
```diff
@@ -114,7 +113,14 @@ def validate_app_name(self, candidate):
                 "name, move to a different parent directory, or delete the "
                 "existing folder.".format(candidate=candidate)
             )
-
+        if 'switch' == candidate:
```
freakboy3742, line 116: Is this block still needed? Won't the `pythonKeywordList` catch `switch`?

src/briefcase/commands/new.py
```diff
@@ -309,7 +315,7 @@ def build_app_context(self):
         """
         formal_name = self.input_text(
             intro="""
-First, we need a formal name for your application. This is the name that will
+First, we need a formal name for your app. This is the name that will
```
freakboy3742, line 318: Not sure if this was intentional; but since this is a long-form explanation, we can use the full name rather than an abbreviation.

tests/commands/new/test_validate_app_name.py
```diff
@@ -25,6 +25,9 @@ def test_valid_app_name(new_command, name):
         '_helloworld',  # leading underscore
         '-helloworld',  # leading hyphen
         'existing',  # pre-existing directory
+        'switch',
+        'false',  # added from the keywords_list
```
freakboy3742, line 29: You may want to check some different capitalisations here; `false` and `False` should both be rejected.

src/briefcase/commands/python_keywords_list.py
```diff
@@ -0,0 +1,35 @@
+# Create a list of reserved keywords
+pythonKeywordsList = [
```
freakboy3742, line 2: 1. This doesn't need to be a separate file; it's a list of constants, so it could be included in the file where it's used.
2. We need to check for non-Python keywords too - Java, C, Objective C and Javascript keywords should all be on this list, because of the potential for collisions in language backends. There will be a lot of overlap with the Python keywords (as these languages all share some heritage), but there are a couple of new ones - "throw" and "do" are the two obvious ones that stood out to me, but there might be some others.
3. Since it *is* a list of constants, convention is to flag this by making the name of the variable `ALL_CAPS`.
4. Python code style also tends towards "snake_case" rather than "camelCase" when there are multiple words in a variable or function name.
5. To help work out whether words are (or aren't) in this list, it would help if they were sorted alphabetically.
6. The code you have here is syntactically correct, but stylistically, Python code style usually has the closing brace/bracket/parenthesis at the same level as the indentation of the line that started it: i.e.,
```
long_list = [
    value,
    value,
]
```

changes/617.bugfix.rst
```diff
@@ -0,0 +1 @@
+Created a list of reserved python keywords and added a check for name violations when creating a new app.
```
freakboy3742, line 1: Release notes should be written in present tense, rather than past tense. They're going to form a document that describes all the shiny new features of the next release, so they should describe the benefit to the user: e.g., "New app names are now validated against reserved words.", rather than describing the code change.

freakboy3742 (changes requested):
Nice work - this is a solid first pass! I've left some comments inline. 

One larger scale problem to consider is that the `new` command is a "wizard" - it helps you get started from a clean slate. However, there's no requirement that you use it, and even if you do, you can modify the file that `new` creates after you've created it. If a user changes the name of their app to something illegal, that should be caught as well.

The validation logic will be identical - so that means we should break that code out into a utility function, test it independently, and invoke it from the two places it's needed.

src/briefcase/commands/new.py
```diff
@@ -96,18 +96,24 @@ def make_app_name(self, formal_name):
         return re.sub('[^0-9a-zA-Z_]+', '', formal_name).lstrip('_').lower()
 
     def validate_app_name(self, candidate):
+
```
freakboy3742, line 99: While this extra line is legal, it's not needed.

changes/617.bugfix.rst
freakboy3742: This comment still needs to be addressed.

src/briefcase/config.py
```diff
@@ -8,14 +8,125 @@
...
+    if app_name.lower() in RESERVED_KEYWORDS:
+        return True
+    return False
+
+
+def is_valid_app_name(app_name):
+    if not is_reserved_keyword(app_name) and is_valid_pep508_name(app_name):
+        return True
+    return False
+# This is the canonical definition from PEP440, modified to include named groups
```
freakboy3742, line 127: This comment should be associated with the PEP440 definition, rather than being stuck to the bottom of the `if_valid_app_name` definition,

src/briefcase/config.py
```diff
@@ -8,14 +8,125 @@
...
+    'while',
+    'with',
+    'yield',
+]
+
+
+def is_valid_pep508_name(app_name):
+    if PEP508_NAME_RE.match(app_name):
+        return True
+    return False
```
freakboy3742, line 114: The logic here is correct, but a little more complex than it needs to be. The 2 lines can be simplified to 
`return PEP505_NAME_RE.match(app_name)`

src/briefcase/config.py
```diff
@@ -8,14 +8,125 @@
...
+
+
+def is_reserved_keyword(app_name):
+    if app_name.lower() in RESERVED_KEYWORDS:
+        return True
+    return False
```
freakboy3742, line 120: Again, the logic here is correct, but a little more complex than it needs to be. The 2 lines can be simplified to 
`return app_name.lower() in RESERVED_KEYWORDS`

tests/commands/new/test_validate_app_name.py
```diff
@@ -9,7 +9,6 @@
         'hello42world',
         '42helloworld',  # ?? Are we sure this is correct?
         'hello_world',
-        'hello-world',
```
freakboy3742, line 12: Why has this test case been removed? It seems like an interesting alternative that should be included, both here, and in the test of `is_valid_app_name`.

tests/commands/new/test_validate_app_name.py
```diff
@@ -20,11 +19,16 @@ def test_valid_app_name(new_command, name):
...
+        'hello world',
+        'helloworld!',
+        '_helloworld',
+        '-helloworld',
+        'existing',
+        'switch',
+        'pass',
+        'false',
+        'False',
+        'YIELD',
```
freakboy3742, line 31: The trailing comments are useful here because they describe what specific property is being tested. We should restore the old comments, and add new comment for the new test cases.

tests/config/test_is_reserved_keyword.py
```diff
@@ -0,0 +1,37 @@
...
+    [
+        'BLACK',
+        'purple',
+        '9pink',
+        'Yellow',
+        'green',
+    ]
+)
+def test_is_not_reserved_keyword_violation(name):
+    "Test that names not in the keywords list return false."
```
freakboy3742, line 17: As above, having comments next to the test cases can be helpful to explain what exactly is being tested (and why). This is especially useful in this case, because a name like `hello world` is *not* a valid name - but it's acceptable here because the check is only validating reserved keywords.

freakboy3742 (changes requested):
We're definitely on the downward slope now! The only remaining issues I can see are related to having clear documentation around the test cases, plus a couple of minor code formatting issues.

freakboy3742 (changes requested):

freakboy3742 (approved):
I've made the last couple of tweaks myself since they were minor. Congratulations on becoming a code contributor to Briefcase!

Issue #612, opened 2021-09-01
Non-latin formal app names cause problems


freakboy3742:
If you use non-latin characters (e.g., `学口算`, or anything outside the ASCII set) in the formal app name, `briefcase new` fails. Similar problems occur if you manually define `pyproject.toml`, and then run `briefcase create`.

This is because briefcase tries to create a "Class Name", as well as other artefacts by processing the formal app name; these conversions fail with non-latin names.

**To Reproduce**
1. Run `briefcase new`
2. Specify an app name of `学口算`.

or
1. Run `briefcase new`
2. Modify `pyproject.toml` so that the formal app name is `学口算`
3. Run `briefcase create`

**Expected behavior**

The formal name is human-readable, so any unicode string should be valid.

NaftoliOst:
Hi, can I claim this issue

freakboy3742:
@NaftoliOst Absolutely! If you need any pointers or help getting started, let us know!

freakboy3742:
Via Discord, there is evidently a problem with non-latin descriptions as well. Any testing for this bug should include anything "human text-like" in the spread of tests.

guaiguaicai:
It’s so touching to see you guys improving Briefcase so seriously, thank you!

xyang-github:
@freakboy3742  It appears that the re.sub is removing all non-ascii characters. For example, entering Chinese characters will result in a empty string; entering a mixture of English letters and Chinese characters will result in a string that is just English letters. I've found a way to bypass this, but getting a cookiecutter.exceptions.FailedHookException now. Still looking at this, but if you have any pointers on how to attempt this kind of exception, I would appreciate it (sorry, this is my first attempt to contribute so still new to the process).

freakboy3742:
@xyang-github Thanks for that investigation - that definitely explains why we're seeing a problem. 

As for your question - it's hard to suggest a fix without knowing what it is you're trying. If you can submit what you have as a pull request, that gives us something to discuss - even if it fails tests, it's something concrete to use for a discussion.

However, looking at the general problem, we're going to need to work out how to convert a "free text" app name into a Python-compatible class name. When the text is purely latin, the re.sub approach works fine, but when the input text is all "invalid Python identifier" characters, it's going to be difficult to manufacture a class name.

[...]

PR #627, opened 2021-10-17
Develop

xyang-github:
Originally, the use of non-latin characters would cause the program to crash. I've tested this with Russian, Chinese and Japanese and they would all result in an 'index out of range' error. The reason for this error is due to the fact that re.sub function would create an empty string by removing all non-latin characters. 

However, allowing the use of non-latin characters will lead to another error ("failed hook exception"). Thus, it might be best to limit user input to latin characters unless there is a sufficient workaround.

Issue: #612

## PR Checklist:

- [X ] I have read the **CONTRIBUTING.md** file
- [ X] I will abide by the code of conduct

### DIFF
--- a/src/briefcase/commands/new.py
+++ b/src/briefcase/commands/new.py
@@ -81,7 +81,9 @@ class NewCommand(BaseCommand):
         :param formal_name: The formal name
         :returns: The app's class name
         """
-        class_name = re.sub('[^0-9a-zA-Z_]+', '', formal_name)
+        # class_name = re.sub('[^0-9a-zA-Z_]+', '', formal_name)  # creates an empty string for non-ascii
+        class_name = re.sub('\s', '', formal_name)  # changed reg expression to remove spaces
+
         if class_name[0].isdigit():
             class_name = '_' + class_name
         return class_name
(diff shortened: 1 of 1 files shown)


src/briefcase/commands/new.py
```diff
@@ -103,7 +106,7 @@ def validate_app_name(self, candidate):
         :returns: True. If there are any validation problems, raises ValueError
             with a diagnostic message.
         """
-        if not PEP508_NAME_RE.match(candidate):
+        if not PEP508_NAME_RE.match(candidate) and candidate.isascii():  # prevent error for non-ascii characters
```
freakboy3742, line 109: This indicates that we're not using the PEP508_NAME_RE correctly - the definition comes directly from PEP508, and we're not telling the regex to match ASCII specifically, so it will match unicode as well. Passing `re.ASCII` as a flag when compiling the regex should address this.

src/briefcase/commands/new.py
```diff
@@ -81,7 +81,9 @@ def make_class_name(self, formal_name):
         :param formal_name: The formal name
         :returns: The app's class name
         """
-        class_name = re.sub('[^0-9a-zA-Z_]+', '', formal_name)
+        # class_name = re.sub('[^0-9a-zA-Z_]+', '', formal_name)  # creates an empty string for non-ascii
+        class_name = re.sub('\s', '', formal_name)  # changed reg expression to remove spaces
```
freakboy3742, line 85: This won't be sufficient; "123 app" is a legal formal name, but `123app` won't be a legal class name.

PR #630, opened 2021-10-21
Added and updated validations

xyang-github:
Added more validation options that prevent non-latin characters from throwing an error. This includes:
 - Added validation to formal name to prevent non-latin characters, else this will throw an error
 - Added validation to app name to prevent it from starting with a number, else this will throw an error. If the default name starts with a number, the default app name will be a generic MainApp.
 - Changed class name to be derived from app name, as this will prevent non-latin characters and names starting with numbers

Related to issue #612

These changes will hopefully stop the program from crashing when using non-latin characters, or app and class names start with a number.

## PR Checklist:

- [ x] All new features have been documented
- [ x] I will abide by the code of conduct

### DIFF
--- a/src/briefcase/commands/new.py
+++ b/src/briefcase/commands/new.py
@@ -74,7 +74,7 @@ class NewCommand(BaseCommand):
             help='The cookiecutter template to use for the new project'
         )
 
-    def validate_formal_name(selfs, candidate):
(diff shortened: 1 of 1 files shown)


freakboy3742:
Closed in favour of #698.

PR #634, opened 2021-10-29
Develop

xyang-github:
Originally, the program would throw an error if a non-latin name was entered as the foreign name. This was due to a regex that resulted in a blank class name.

My fix circumvents this issue, by explicitly asking the user for a class name instead of deriving it completly from the formal name. Added a make_class_name method and a validate_class_method. Originally had planned to make the candidate class_name to be from the app_name, but later decided to keep it with the formal_name. If a non-latin formal name is entered, then the program will prompt the user to enter an appropriate class name via validation. The regex used should be consistent with camel case format. Two test files, one for each new class_name method, was created and passed when ran.

During this process, also realized that the regex for the app_name isn't sufficient. I changed the regex statement so that the app name cannot start with a digit or a capitalized letter. If the formal name is non-latin, then the make_app_name method will return a blank string, which will raise a validation error if user doesn't enter anything. Updated the test pages, and the changes made passed.

Reference to issue #612 

## PR Checklist:

### DIFF
--- a/src/briefcase/config.py
+++ b/src/briefcase/config.py
@@ -10,8 +10,7 @@ from .exceptions import BriefcaseConfigError
 
 # The restriction on application naming comes from PEP508
 PEP508_NAME_RE = re.compile(
-    r'^[a-z][a-zA-Z0-9._-]*[a-zA-Z0-9]$',
-    re.IGNORECASE
+    r'^[a-z][a-zA-Z0-9._-]*[a-zA-Z0-9]$'
 )
 
 # This is the canonical definition from PEP440, modified to include


src/briefcase/commands/new.py
```diff
@@ -103,10 +91,11 @@ def validate_app_name(self, candidate):
         :returns: True. If there are any validation problems, raises ValueError
             with a diagnostic message.
         """
-        if not PEP508_NAME_RE.match(candidate):
+        if not re.match('^[a-z][a-zA-Z0-9._-]*[a-zA-Z0-9]$', candidate):
```
freakboy3742, line 94: If this is the correct regex pattern for PEP508 compliance, we should update the PEP508 regex, not define a complex new regex on the fly.

src/briefcase/commands/new.py
```diff
@@ -117,6 +106,33 @@ def validate_app_name(self, candidate):
...
 
+    def make_class_name(self, formal_name):
+        """
+        Construct a candidate class name from a formal name.
+
+        :param formal_name: The formal name
+        :returns: The candidate app name; will appear blank if only non-latin characters
+        were entered in the formal_name
+        """
+        formal_name.title()
```
freakboy3742, line 117: This doesn't do anything. You're creating a title case version the string, and then discarding it.

tests/commands/new/test_make_class_name.py
```diff
@@ -9,7 +9,7 @@
         ('Hello! World', 'HelloWorld'),
         ('Hello_World', 'Hello_World'),
         ('Hello-World', 'HelloWorld'),
-        ('24 Jump Street', '_24JumpStreet'),
```
freakboy3742, line 12: Since this whole PR exists because of a specific bug report, we should include the case from that bug report in the testing.

src/briefcase/commands/new.py
```diff
@@ -117,6 +106,33 @@ def validate_app_name(self, candidate):
...
+        """
+        Determine if class name is valid
+
+        :param candidate: The candidate name
+        :returns: True. If there are any validation problems, raises ValueError
+            with a diagnostic message.
+        """
+        if not re.match('^[A-Z]+[a-zA-Z0-9_]+$', candidate):
+            raise ValueError(
+                "The class name must start with a capital letter in the CapWords format."
```
freakboy3742, line 130: The usual term for this is CamelCase.

freakboy3742 (changes requested):
This is on the right track; I've highlighted a couple of areas that need attention, but this is starting to look pretty good!

src/briefcase/commands/new.py
xyang-github: I went ahead and updated the pep508 regex.

src/briefcase/commands/new.py
xyang-github: I went ahead and include the title  method in the return statement. This is actually needed if the formal name is spelled with two lowercase words. I added a test to the test_make_class_name to show an example of this scenario where "hello world" formal name is changed to "HelloWorld" class name.

tests/commands/new/test_make_class_name.py
xyang-github: Good point. I added a couple of tests using non-latin characters. My changes have passed these tests, but let me know if you had anything more specific in mind and I will be glad to make the changes.

src/briefcase/commands/new.py
xyang-github: I changed the wording from CapWords to CamelCase.

xyang-github:
@freakboy3742  Apologies it has taken so long. I am reviewing the reasons why the checks were not successful.

PR #638, opened 2021-11-01
fix Non-latin formal app names cause problems #612

charlietrumm:
Allow the use of non latin characters in formal name of briefcase applications.

To create python class, it looks for only non-latin characters in formal name.


freakboy3742:
@charlietrumm Thanks for the PR; however, there's already another contributor halfway through a much more comprehensive fix for this problem (See #634); the solution you've proposed won't be sufficient in the general case. See the discussion on #612 for details.

If you'd like to get involved, I'd suggest picking a different ticket tagged `first-timers-only` and `up-for-grabs`.

PR #642, opened 2021-11-16
Determine Xcode location with `xcode-select -p`

### LINKED ISSUE #622: "briefcase build iOS" fails with "Xcode is not installed. You can install it......." on MacOS Big Sur with Xcode 12.5 installed and working.
Following the beeware "Hello World" online tutorial ends in failure at the final stage of trying to build for iOS, stating that Xcode is not installed, despite Xcode being installed and working ,as well as the Xcode command line tools. In my case I installed Xcode 12.5

It turn out the error is 100% due to the line where the Xcode location is hard coded :  **xcode_location='/Applications/Xcode.app'**. This is in _/beeware-tutorial/beeware-venv/lib/python3.8/site-packages/briefcase/integrations/xcode.py_.

I had downloaded Xcode using the Apple Installer and installed it on a different drive, i.e. not in /Applications, the default.

Simply changing the **xcode_location** to my real location fixed the problem, and it compiled 100% with 'briefcase build iOS' and ran with 'briefcase run iOS'.

samschott:
Fixes #622 by not hardcoding the Xcode path but determining it at runtime with `xcode-select -p`.

Note that `xcode-select -p` may return the location of the command line tools if those are installed but Xcode itself it not. This case will be properly handled when checking the output of `xcodebuild -version` later.

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/xcode.py
+++ b/src/briefcase/integrations/xcode.py
@@ -268,7 +268,7 @@ accepted before you can use those tools.
 You can accept these licenses by starting Xcode and clicking "Accept"; or, you
 can run:
 
-    sudo xcodebuild -license
+    $ sudo xcodebuild -license
 
 at the command line and accepting the license there.
 
--- a/src/briefcase/integrations/xcode.py
+++ b/src/briefcase/integrations/xcode.py
@@ -224,10 +224,14 @@ Re-run Briefcase once that installation is complete.
         if " is a command line tools instance" in e.output:
             raise BriefcaseCommandError("""
 Xcode may be installed, but the active developer directory is a
-command line tools instance. To make Xcode the active developer
-directory, run:
+command line tools instance. To make the default Xcode install the
+active developer directory, run:
 
-    $ sudo xcode-select --switch path/to/Xcode.app
+    $ sudo xcode-select --switch /Applications/Xcode.app
+
+Or, to use a version of Xcode installed in a non-default location:
+
+    $ sudo xcode-select --switch /path/to/Xcode.app
 
 and then re-run Briefcase.
 """)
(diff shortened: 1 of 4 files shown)


SamSchott:
This PR also fixes a couple of unrelated tests in `test_cmdline.py` which were failing due to changed CLI output.

src/briefcase/integrations/xcode.py
```diff
@@ -129,14 +129,34 @@ def ensure_xcode_is_installed(
...
     # when you install from the App store.
+
+    if xcode_location is None:
+
+        try:
+            output = command.subprocess.check_output(
+                ['xcode-select', '-p'],
+                stderr=subprocess.STDOUT,
+                universal_newlines=True
+            )
```
freakboy3742, line 147: Have you/did you confirmed the behavior of `xcode-select` if Xcode is not installed? I've got a VM installation I can use to confirm this, but if you've already done this check, then I won't bother. My concern is that there may be an Xcode-select stub that triggers the App Store if Xcode isn't there.

freakboy3742 (changes requested):
Seems relatively straightforward; one detail that needs to be confirmed inline.

src/briefcase/integrations/xcode.py
SamSchott: Yes, I've checked this in a VM myself. It errors out with the following message:

> xcode-select: error: unable to get active developer directory, use `sudo xcode-select --switch path/to/Xcode.app` to set one (or see `man xcode-select`)

freakboy3742 (approved):
An overdue review; but looks good. I've made a small tweak to the error message so that the "simple case" has unambiguous recovery instructions.

Issue #471, opened 2020-08-14
Briefcase can't install Python packages with binary modules on mobile platforms.


freakboy3742:
**Describe the bug**

Briefcase is able to package any an app that has a dependency that is a pure-Python package (i.e., a package that contains *only* Python code).

However, if you're targeting a mobile platform (iOS or Android), and the Python module has a binary component (i.e., it contains a C module), any attempt to import that module will fail, causing the app to crash.

Desktop platforms (macOS, Windows and Linux) are *not* affected by this problem.

**To Reproduce**
1. Set up a hello world application
2. Add a package with a binary dependency (e.g., numpy or pillow) to the `requires` list, and add an import for that code (e.g., `import numpy` or` import pil`)
3. Deploy the app to iOS or Android
4. Try to run the app.

**Expected behavior**

The import should succeed and the app should continue to run.

**Environment:**
 - Operating System: iOS or Android
 - Python version: Any
 - Software versions:
   - Briefcase: Any

**Additional context**

This isn't *purely* a Briefcase issue; pip and PyPI doesn't currently have any support for mobile platforms.

sarthaksharma058439:
Check this video .....he make calculator in beeware and deploy to android device

https://youtu.be/RllR6FRdG6w

liyanhrxy:
is there any other possible  idea to fix ?   Is there an alternative to this package(pycryptodome)？

freakboy3742:
I don't know enough about pycryptodome to comment; however, OpenSSL is available as a precompiled library, so any python binding to openSSL crypto primitives should work.

liyanhrxy:
`pycryptodomex` is a very important crypto library  in python ? Is it possible to add it to template like OpenSSL as a  precompiled library ??? it's troubles me for many times

freakboy3742:
It is *possible* to add any library you want to the support package by forking and modifying the build process; however, we're unlikely to include it in the *official* package. OpenSSL is included because it's needed to compile the Python standard library. 

The fix here isn't to add more and more binary libraries to the support package; it's to fix the underlying problem of binary wheels for mobile platforms.

sergiysavelyev:
[...]

PR #647, opened 2021-12-07
Build packages for Android

eliteraspberries:
Looking for feedback as this is very experimental. :)

Edit: This is in reference to #471.

## PR Checklist:


freakboy3742:
Wow - this is awesome... and a lot simpler than I thought it would be - thanks!

I'll need to take a closer look (especially into what  exactly androidenv is doing), but this looks like a really good start!

src/briefcase/commands/create.py
```diff
@@ -379,6 +379,40 @@ def install_app_support_package(self, app: BaseConfig):
...
+        and its path in LDFLAGS.
+
+        :param app: The config object for the app
+        """
+        env = dict(os.environ)
+        if self.platform == "android":
+            # sysconfig.get_config_var("INCLUDEPY")
+            includepy = os.path.join(
+                self.bundle_path(app), "app", "include",
+                "python{}m".format(self.python_version_tag),
```
rbignon, line 395: This path doesn't exist, is the trailing 'm' a mistake?

src/briefcase/commands/create.py
```diff
@@ -379,6 +379,40 @@ def install_app_support_package(self, app: BaseConfig):
...
+            )
+            assert os.path.exists(includepy), includepy
+            cppflags = env.get("CPPFLAGS", "").split(" ")
+            cppflags = [
+                "-I{}".format(includepy.replace(" ", "\\ ")),
+            ] + cppflags
+            env.update({"CPPFLAGS": " ".join(cppflags)})
+            # sysconfig.get_config_var("LIBDIR")
+            libdir = os.path.join(
+                self.bundle_path(app), "app", "libs", "arm64-v8a"
```
rbignon, line 405: Why is it hardcoded for arm64-v8a? It should be compatible with all available targets, as the emulator can run with a x86_64 arch for example.

eliteraspberries: Yes, good point. This will require changes to the beeware/Python-Android-support project to include four architecture-specific "app_packages" directories, as well as changes to the beeware/briefcase-android-gradle-template project to add the right directory to sys.path. I'll try to implement this.

src/briefcase/commands/create.py
eliteraspberries: The "m" is a suffix added by Python when it is configured with --with-pymalloc. It seems this is only valid for 3.6 and 3.7: https://github.com/beeware/Python-Android-support/blob/master/main.sh#L84 I'll fix this.

eliteraspberries:
I have been using the SDK from Android Studio, so I didn't notice that ANDROID_SDK_ROOT was not in pip's environment for everyone using the SDK downloaded by briefcase. This latest commit should fix that, hopefully.

eliteraspberries:
OK, with the latest commit (f94f62b) and beeware/briefcase-android-gradle-template/pull/44, briefcase builds for every ABI. I only tested this with NumPy, and only on arm64-v8a.

danyeaw:
@eliteraspberries I fixed the merge conflict and updated the PR. I would really like to see this merged because you went most of the way to get something working. I think this probably needs:

- [ ] Towncrier news snippet
- [ ] Doc updates to explain how third party libraries are supported with Android
- [ ] A test added to cover this if possible
- [ ] Merge https://github.com/beeware/briefcase-android-gradle-template/pull/44 which also needs a rebase

Is there anything else? How can I help?

freakboy3742:
@danyeaw There's one big additional requirement: it requires [a change to Python's distutils](https://github.com/beeware/briefcase/issues/471#issuecomment-982970698). 

That's the biggest impediment at present, because there isn't a "simple fix" we can apply - we need a change made upstream in CPython. Alternatively, we need to work out a way for an end user to patch distutils in a way that enables them to leverage those changes. That might mean we need to maintain a forked version of distutils in the short term while we attempt to get this change upstreamed - although I'm open to any other suggestions.

danyeaw:
@freakboy3742 Good point.

Since distutils as part of the standard library is now deprecated and is being updated / migrated as part of setuptools, maybe this hurdle might not be so high.

@eliteraspberries Can androidenv make use of the vendored version of distutils in setuptools if setuptools is installed? I think this is now the default behavior of setuptools as long as it isn't overriden with `SETUPTOOLS_USE_DISTUTILS=stdlib`.

[This change](https://github.com/pypa/distutils/pull/50) might make the androidenv patch of `sysconfig.py` not needed any longer, it looks like you can override ranlib if it is in `os.environ`.

freakboy3742:
I'm going to close this PR; thanks for the work you've put into this, but it looks like we have a path forward using the [Chaquopy](https://github.com/chaquopy) toolchain as a base for Android projects.

joooeey:
> I'm going to close this PR; thanks for the work you've put into this, but it looks like we have a path forward using the [Chaquopy](https://github.com/chaquopy) toolchain as a base for Android projects.

Is there a timeline, pull request, or issue for this change? We are looking at using BeeWare for a research project because of its small complexity and since it only requires one language. However, Numpy support is a must for us and I think we don't have the skills and resources to add support for binary packages ourselves. I'd love to know how this effort comes along.

mhsmith:
The central PR for this change is https://github.com/beeware/briefcase-android-gradle-template/pull/52, and we're aiming to release it by the end of August. 

Binary packages will initially only be available for Python 3.8, and will include [all the packages](https://chaquo.com/pypi-7.0/) currently supported by Chaquopy. Later we'll add packages for Python 3.9, 3.10 and possibly 3.7.

PR #653, opened 2022-01-08
Deprecate support for Python 3.6

freakboy3742:
Deprecate support for Python 3.6

## PR Checklist:


carlosperate:
With this deprecation in the v0.3.6 release, should the `python_requires` value in `setup.cfg` also be changed? Or has that been purfosely left at 3.6 for now to be changed in a future release?

freakboy3742:
@carlosperate Good catch! Looks like we overlooked that one. I've updated the reference in #680.

Issue #416, opened 2020-06-05
Linux AppImage build fails if run as root


freakboy3742:
Reported by @axel-kah

If you run a `briefcase create linux` build while logged in as root, it attemps to build the Docker container with HOST_UID and HOST_GID set to 0. This causes the `useradd` step to fail:
```
Step 12/16 : ARG HOST_UID
 ---> Running in 063a424d6a86
Removing intermediate container 063a424d6a86
 ---> 4657e96c180d
Step 13/16 : ARG HOST_GID
 ---> Running in a59336d3cdb7
Removing intermediate container a59336d3cdb7
 ---> 4fede95af7ee
Step 14/16 : RUN groupadd --gid $HOST_GID briefcase || true
 ---> Running in bb37131e7a73
groupadd: GID '0' already exists
Removing intermediate container bb37131e7a73
 ---> a94d6f409c76
Step 15/16 : RUN useradd --uid $HOST_UID --gid $HOST_GID brutus --home /home/brutus
 ---> Running in 55dc8b2957d8
useradd: UID 0 is not unique
The command '/bin/sh -c useradd --uid $HOST_UID --gid $HOST_GID brutus --home /home/brutus' returned a non-zero code: 4

Error building Docker container for helloworld.
```

This is because the default user in Docker is root, with UID 0.

We need to either:
a. Modify the docker container to be resilient to UIDs that already exist, or
b. Add error handling to prevent briefcase from being invoked with users that already exist.

axel-kah:
Exploring option a)

Seems like `useradd`and `usermod` both feature the `--non-unique` option to allow duplicate UIDs. When adding this option in the modified template, referenced in my `pyproject.toml`, then I could get away with this and `briefcase create` executes without errors. A subsequent `briefcase build` will bail out with the familiar problem that linuxdeploy is known to not being executable in containers (conditions apply?) linuxdeploy/linuxdeploy/issues/86. Applying the suggested fix (replacing the magic bytes) and chowning the appimage with my local useraccount will produce a working tutorial app.
If I leave the appimage with its root ownership, then the appimage does not work when run as root with "Run Once" showing an error message:
```
Failed to register AppImage in AppImageLauncherFS: could not open map file
```
Not sure if that is a problem with the created app image or with the AppImageLauncher run as root.

freakboy3742:
[...]

PR #667, opened 2022-02-08
Add ELF patcher for linuxtools

belthesar:
This PR addresses an awkward to trace issue on Linux systems running briefcase that also have AppImageLauncher installed, first reported via the old Gitter chat room, then again in Discord.  This issue was also observed in https://github.com/beeware/briefcase/issues/416#issuecomment-640744410

On Linux systems with AppImageLauncher, AppImages that are run are hooked to start AppImageLauncher via the `binfmt-misc` kernel module. When briefcase creates a Docker container to build AppImages, that kernel module's capabilities are passed through to the container. As the container image doesn't have AppImageLauncher (nor should it), we need to ensure that linuxtools launches as a regular ELF binary. 

By patching the ELF header to remove the flags from the header that would trip `binfmt_misc` to pass execution along to AppImageLauncher, the system launches `linuxtools` as a regular ELF binary. This works, and has been tested on a system that doesn't have AppImageLauncher installed. 

Signed-off-by: Cody Wilson <cody@codywilson.co>

### DIFF
--- a/src/briefcase/integrations/linuxdeploy.py
+++ b/src/briefcase/integrations/linuxdeploy.py
@@ -62,7 +62,7 @@ class LinuxDeploy:
                 download_path=self.command.tools_path
             )
             self.command.os.chmod(linuxdeploy_appimage_path, 0o755)
-            self.elf_header_patch()
+            self.patch_elf_header()
         except requests_exceptions.ConnectionError:
             raise NetworkFailure('downloading linuxdeploy AppImage')
 
@@ -79,7 +79,7 @@ class LinuxDeploy:
         else:
             raise MissingToolError('linuxdeploy')
 
-    def elf_header_patch(self):
+    def patch_elf_header(self):
         """
         Patch the ELF header of the AppImage to ensure it's always executable.
 
@@ -123,4 +123,4 @@ class LinuxDeploy:
                     # wrong and we should raise an exception.
                     raise MissingToolError("linuxdeploy")
         else:
-            print("linuxdeploy AppImage not found. Skipping ELF header patch.")
+            raise MissingToolError("linuxdeploy")
--- a/src/briefcase/integrations/linuxdeploy.py
+++ b/src/briefcase/integrations/linuxdeploy.py
@@ -101,7 +101,6 @@ class LinuxDeploy:
         - https://github.com/AppImage/AppImageKit/issues/828
         """
 
-
         if self.exists():
             with open(self.appimage_path, 'r+b') as appimage:
                 appimage.seek(ELF_PATCH_OFFSET)
--- a/src/briefcase/integrations/linuxdeploy.py
+++ b/src/briefcase/integrations/linuxdeploy.py
@@ -81,18 +81,18 @@ class LinuxDeploy:
 
     def elf_header_patch(self):
         """
-        Patch the ELF header of the AppImage to ensure it's always executable. 
+        Patch the ELF header of the AppImage to ensure it's always executable.
 
-        This patch is necessary on Linux hosts that use AppImageLauncher. 
+        This patch is necessary on Linux hosts that use AppImageLauncher.
         AppImages use a modified ELF binary header starting at offset 0x08
-        for additional identification. If a system has AppImageLauncher, 
+        for additional identification. If a system has AppImageLauncher,
         the Linux kernel module `binfmt-misc` will try to load the AppImage
-        with AppImageLauncher. As this binary does not exist in the Docker  
-        container context, we patch the ELF header of linuxdeploy to remove 
-        the AppImage bits, thus making the system treat it like a regular 
-        ELF binary. 
+        with AppImageLauncher. As this binary does not exist in the Docker
+        container context, we patch the ELF header of linuxdeploy to remove
+        the AppImage bits, thus making the system treat it like a regular
+        ELF binary.
 
-        Citations: 
+        Citations:
         - https://github.com/AppImage/AppImageKit/issues/1027#issuecomment-1028232809
         - https://github.com/AppImage/AppImageKit/issues/828
         """
--- a/src/briefcase/integrations/linuxdeploy.py
+++ b/src/briefcase/integrations/linuxdeploy.py
@@ -101,11 +101,6 @@ class LinuxDeploy:
         - https://github.com/AppImage/AppImageKit/issues/828
         """
 
-        patch = {
-            'offset': 0x08,
-            'original': bytes.fromhex('414902'),
-            'patch': bytes.fromhex('000000')
-        }
 
         if self.exists():
             with open(self.appimage_path, 'r+b') as appimage:
--- a/src/briefcase/integrations/linuxdeploy.py
+++ b/src/briefcase/integrations/linuxdeploy.py
@@ -1,6 +1,10 @@
 from requests import exceptions as requests_exceptions
 
-from briefcase.exceptions import MissingToolError, NetworkFailure
+from briefcase.exceptions import CorruptToolError, MissingToolError, NetworkFailure
+
+ELF_PATCH_OFFSET = 0x08
+ELF_PATCH_ORIGINAL_BYTES = bytes.fromhex('414902')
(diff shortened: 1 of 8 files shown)


belthesar:
It would appear PEP8 W291 doesn't like trailing whitepaces in docstrings. :sweat_smile: Updated PR to clear those out. Unsure if towncrier exited early because of flake8's exit code, this is my first time with that one.

freakboy3742:
I can guarantee towncrier will still break - it's expecting a standalone file in the `changes` directory. See [this note](https://briefcase.readthedocs.io/en/latest/how-to/contribute-code.html#add-change-information-for-release-notes) for details on what is needed

belthesar:
I could use a little feedback here. I made the patch function a bit safer by checking for the existence by wrapping the file open in an if statement checking if the file exists., and raising `MissingToolError`'s if they don't exist. However, the tests `test_upgrade_exists` and `test_verify_do_not_exist` fail.  I feel like raising an Exception if the patcher can't find the file or if it's a "bad file" makes sense, but based on looking over the tests, it doesn't make sense why things are failing.  Thoughts?

src/briefcase/integrations/linuxdeploy.py
```diff
@@ -77,3 +78,49 @@ def upgrade(self):
...
+        - https://github.com/AppImage/AppImageKit/issues/828
+        """
+
+        patch = {
+            'offset': 0x08,
+            'original': bytes.fromhex('414902'),
+            'patch': bytes.fromhex('000000')
+        }
+
+        if self.exists():
```
belthesar, line 106: In order to make test coverage pass, I had to wrap the file handle in a test to see if the file exists. Given that `elf_header_patch()` is always called via the `install()` method, and that `install()` has plenty of seatbelts of its own to prevent a file from not existing, I feel like this is _reasonably_ safe, but I'll leave that to y'all to decide if that's sufficient.

freakboy3742: I presume this path is no longer required?

belthesar: Oh, I left that in intentionally. Previously, I did not have an exception in there for the else clause. With that else clause, this tests if the file isn't there (to protect if something about how downloads are managed changes), which avoids a file handle (the tiniest of safety measures). I suppose it's not _completely_ necessary with the if/elif/else statement but then the file open handle would throw its own exception. 

If you think that it's better to get the file exception from the `open` than to receive a `MissingToolError` exception, I'll remove that.

freakboy3742: That explanation makes sense; my original thought was that the method should *never* be called if the binary doesn't exist - but I guess the extra layer of safety doesn't hurt, and it raises an exception regardless - but in your case, it's a little more direct about what has gone wrong.

src/briefcase/integrations/linuxdeploy.py
```diff
@@ -77,3 +78,49 @@ def upgrade(self):
...
+        Citations:
+        - https://github.com/AppImage/AppImageKit/issues/1027#issuecomment-1028232809
+        - https://github.com/AppImage/AppImageKit/issues/828
+        """
+
+        patch = {
+            'offset': 0x08,
+            'original': bytes.fromhex('414902'),
+            'patch': bytes.fromhex('000000')
+        }
```
freakboy3742, line 104: This is structured as a dictionary, but there isn't really any reason for it to be; they're effectively constants. We're not doing anything dynamic here, so we're not gaining anything from storing these as a dictionary; and a dictionary lookup will be *very* marginally slower, because there's an extra key lookup operation. 

I'd even argue they should be module-level constants - they're "fundamental" properties of the ELF binary, and they're duplicated in the test, rather than referring to these fundamental properties.

tests/integrations/linuxdeploy/test_LinuxDeploy__elf_patch.py
```diff
@@ -0,0 +1,79 @@
+from unittest.mock import MagicMock
```
freakboy3742, line 1: There's nowhere else for me to put this comment; but the test file name should match the method name; when you renamed the method name, the test file should come along for the ride.

src/briefcase/integrations/linuxdeploy.py
```diff
@@ -77,3 +78,49 @@ def upgrade(self):
...
+                    appimage.seek(0)
+                    print("Patched ELF header of linuxdeploy AppImage.")
+                # Else if the header is the patched value, do nothing.
+                elif appimage.read(len(patch['original'])) == patch['patch']:
+                    print("ELF header of linuxdeploy AppImage is already patched.")
+                else:
+                    # We should only get here if the file at the AppImage patch doesn't have
+                    # The original or patched value. If this is the case, the file is likely
+                    # wrong and we should raise an exception.
+                    raise MissingToolError("linuxdeploy")
```
freakboy3742, line 124: This isn't quite the right exception - MissingTool means an attempt was made to run a tool, but it couldn't be found. This is a different type of problem - it's effectively a "CorruptedToolError" - the tool "exists", but it's not "right" for some reason. This is something we should probably be checking more broadly, so adding a new CorruptedToolError in the exceptions file and raising it here would make sense to me. 

This also allows us to differentiate between *this* condition, and the condition on L126.

src/briefcase/integrations/linuxdeploy.py
```diff
@@ -77,3 +78,49 @@ def upgrade(self):
...
+        else:
+            raise MissingToolError("linuxdeploy")
```
freakboy3742, line 126: This one probably *is* a MissingToolError (albeit one that we should never see)

tests/integrations/linuxdeploy/utils.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Args:
+        appimage_path (Path): Path to the appimage to create.
+        mock_appimage_kind (str): The kind of mock appimage to create.
+            'original' creates an unpatched mock appimage.
+            'patched' creates a patched mock appimage.
+            'corrupt' creates a corrupted mock appimage.
+
+    Returns:
+        bytes_to_be_patched (bytes): The bytes to be patched of the created AppImage.
```
freakboy3742, line 23: These docs are great, but we use a slightly different format; see [this method](https://github.com/beeware/briefcase/blob/master/src/briefcase/commands/create.py#L75) for an example

freakboy3742 (changes requested):
This looks really good - nice work! I've flagged a couple of mostly minor and cosmetic things inline; but the core of this is really solid.

One minor admin note: ordinarily, when I've got a PR like this that is 95% good, and just needs a few tweaks, it's easier for me to make them myself (making a few typo changes is easier to Just Do It, rather than explain what needs to be done). However, as you've submitted this PR from your master branch, I can't do that. It's not worth changing for this PR; but in future, it makes the process a lot easier if you submit the PR from a branch on your Github fork. It will also make *your* life easier in the long run, because you won't need to do merge commits when your master and the project master diverge.

src/briefcase/integrations/linuxdeploy.py
belthesar: Makes sense to me! I've changed those to constants, and am using those in the tests as well.

tests/integrations/linuxdeploy/test_LinuxDeploy__elf_patch.py
belthesar: Also tracks! Resolved.

src/briefcase/integrations/linuxdeploy.py
belthesar: This makes sense to me as well. I didn't really feel like the exception made sense the best sense, but I didn't also feel comfort in expanding the list of exceptions without knowing if that's the right thing to do in the scope of the rest of the project. I have added a CorruptToolError as requested.

tests/integrations/linuxdeploy/utils.py
belthesar: I've updated the documentation to conform to the project's documentation pattern.

freakboy3742 (approved):
Awesome - thanks for the contribution!

Issue #560, opened 2021-01-26
Add PySide6 Support


Daveismus:
**Is your feature request related to a problem? Please describe.**
PySide6 was released

**Describe the solution you'd like**
Make it possible to create Pyside6 applications

**Describe alternatives you've considered**
-

**Additional context**

freakboy3742:
Absolutely! We'd love to support as many backends as possible. 

If anyone wants to work this issue, the change will be primarily in the [v0.3 branch of the briefcase-template project](https://github.com/beeware/briefcase-template/tree/v0.3) project, adding the stub code for a PySide6 project. Bringing #392 (initial PR in #414) to completion would also be helpful.

Noir-:
Is there any progress on this? Looks like all python distribution frameworks currently fail with pyside6 :/

freakboy3742:
@Noir- No, there hasn't been any progress. I'll also add that it's not an especially high priority for me personally; I don't use Qt or PySide. 

That said - Briefcase's "support" for GUI toolkits is almost 100% related to the wizard template. There's nothing in the packaging process itself that is tied to the GUI toolkit. As long as you can `pip install pyside6` (or whatever the invocation is), and the install is self contained, Briefcase should work - you just won't get the full templated start-of-project experience.

Noir-:
@freakboy3742 Ok I see. The GUI toolkit isn't necessary. Thanks for pointing that out. Is there anything of what I should take care? I suspect that the basic project structure from `birefcase new` is mandatory. Just trying to integrate my application into this but it fails, gonna open a new issue for that.

freakboy3742:
If you run `briefcase new` and specify "no GUI toolkit", you can use that as a base and add your own PySide6 code.

The stretch goal beyond that is to take whatever modifications you make to the "no GUI toolkit" template, and add those changes as a "pyside6" option in that template.

freakboy3742:
Closing as this was added in Briefcase 0.3.6.

PR #670, opened 2022-02-18
Add option for PySide6

cmaureir:
Include an option to select PySide6 from the list of UI  frameworks.

This patch was motivated by https://github.com/beeware/briefcase/issues/560 and it depends of this Pull Request https://github.com/beeware/briefcase-template/pull/27 which adds the template on the `briefcase-template`

## PR Checklist:

I didn't find a place where the non-default gui framework were documented, and because there is already a test for a PySide2 project, I didn't want to include another with only that little change.
In case it's necessary, please let me know I can do it without any problem.

### DIFF
--- a/changes/670.misc.rst
+++ b/changes/670.misc.rst
@@ -1,0 +1,1 @@
+Add PySide6 to the GUI Framework template options
--- a/src/briefcase/commands/new.py
+++ b/src/briefcase/commands/new.py
@@ -426,6 +426,7 @@ What GUI toolkit do you want to use for this project?""",
             options=[
                 'Toga',
                 'PySide2 (does not support iOS/Android deployment)',
+                'PySide6 (does not support iOS/Android deployment)',
                 'PursuedPyBear (does not support iOS/Android deployment)',
                 'None',
             ],


cmaureir:
Hey!

> Happy to add PySide6 support, and this patch (and the matching template patch) look good; once CI is passing, I'll be happy to > add that support

I hope my new file in `changes/` is correct and enough for the `towncrier` test to pass :dancers: 

Before jumping into the other questions, just some general information regarding versions and releases:
* PySide2 last OSS release was 5.15.2 (there was another for a special case 5.15.2.1, but not relevant) and it's still getting bug fixes, for commercial customers only due to the changes in the Qt Offering (5.15 is the last 5 release, and it's LTS). At the moment commercial license holders have 5.15.8 available. However, due to the [KDE Free Qt Foundation](https://kde.org/community/whatiskde/kdefreeqtfoundation/) there will be future releases (source code only), so people might still keep building wheels and using it. For example: the next 5.15.3 will be in the available in March.
* PySide6 reached roughly a module parity in 6.2 (LTS), at the moment we are focused on 6.3 (to be released in March), so personally I believe people should use 6.x without major issues.


> Can you provide any insight on whether continuing to support PySide2 is desirable? Has PySide2 been fully deprecated?

We have seen many users and projects still on PySide2 (Qt5), and overall the installations numbers we get from pypi are still considerable, for example yesterday PySide2 had [2.2k](https://pypistats.org/packages/pyside2) and PySide6 only [1k](https://pypistats.org/packages/pyside6) 

Being an LTS release, PySide2 5.15 is still relevant.

> Is there any good reason why a developer would want to start a new app with Pyside2 if they are starting today?

If a person is "starting a new project", I tend to only recommend PySide6 because it's the newest and better, but for some reason, maybe online questions, community, and OSs that still don't have Qt6/PySide6 people is choosing PySide2.

> If Pyside2 is  old/deprecated/unsupported, I'd rather replace PySide2 with Pyside6, rather than encourage users to continue to use the older code.

If you and the other  maintainers are OK with keeping both, I guess it still makes sense, but if you decide to support only PySide6 I will be happy as well :)

freakboy3742 (approved):
The existence of an LTS release and policy around Pyside2, plus the fact that download counts favour Pyside2 would seem to suggest that it's worth preserving Pyside2 for now. 

This patch looks great, and ready to merge as-is; however, there is a problem with the related briefcase-template patch. Comments on that PR.

PR #690, opened 2022-03-29
Updated code signing routines for macOS

### LINKED ISSUE #652: `briefcase package` is missing to sign long with libraries capabilities when required
**Describe the bug**

While building [Tumpa](https://tumpa.rocks) I found that created dmg is passing notarization but does start on any other Mac, saying "Developer cannot be verified". 
After fighting with Apple notarization system for many hours, I found that while running the `codesign` on the files, I will have to pass `library` as option.

I had to patch `briefcase/platforms/macOS/__init__.py` the `sign` function to have the following:

```diff
--- .venv/lib/python3.9/site-packages/briefcase/platforms/macOS/__init__.py	2022-01-07 08:48:12.000000000 +0100
+++ /tmp/__init__.py	2022-01-07 08:47:54.000000000 +0100
@@ -117,7 +117,7 @@
                     '--deep', str(path),
                     '--force',
                     '--timestamp',
-                    '--options', 'runtime',
+                    '--options', 'runtime,library',
                 ],
                 check=True,
             )
```

**Environment:**
 - Operating System: macOS Monterey, 12.1
 - Python version: Python 3.9.1
 - Software versions:
   - Briefcase: 0.3.5

dreavjr:
Changes:
- Expands the criteria of what counts as an executable for signing:  instead of just checking for a small list of file extensions, now checks for all files marked with executable flag, and all files in an enlarged subset of extensions. All files in the set that that have a MACH-O magic number are sent for signing
- Finds bundles inside the bundle for signing
- Changes how signing is executed: instead of applying `--deep` by default, signable paths are signed from the inside out, as recommended by Apple. Includes `--deep` as a failsafe if first attempt of signing fails due to dependencies
- Signs the full DMG after creating it

Minor changes:
- Prints the command used for signing instead of just the name of the file being signed
- `select_identity()` now suggests a signing identity if one is partially matched

The signing procedure now follows Apple's recommendations in https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution

Fixes #652

## PR Checklist:


dreavjr:
This file has step-by-step procedures for signing and notarizing a package, as well as some links to Apple documentation:
[apple-signing-example.txt](https://github.com/beeware/briefcase/files/8367218/apple-signing-example.txt)

freakboy3742 (commented):
Thanks for this PR. I've made some changes to match project preferences, and added some tests - but your starting point was invaluable for working out where improvement were needed.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -160,37 +181,64 @@ def select_identity(self, identity=None):
...
                 check=True,
             )
-        except subprocess.CalledProcessError:
-            print()
-            raise BriefcaseCommandError(
-                "Unable to code sign {path}.".format(path=path)
-            )
+        except subprocess.CalledProcessError as e:
+            errors = e.stderr.decode('utf-8', errors='replace')
+            if 'not signed at all' in errors:
```
freakboy3742, line 219: @dreavjr Are you able to provide an example of how to manufacture this message? I'm not sure if I can do anything to manufacture a test for this, but at the very least, I'd like to reproduce it manually.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -160,37 +181,64 @@ def select_identity(self, identity=None):
...
+                        stderr=subprocess.PIPE,
+                        check=True,
+                    )
+                except subprocess.CalledProcessError:
+                    raise BriefcaseCommandError(f"Unable to deep code sign {path}.")
+
+            elif any(
+                msg in errors
+                for msg in [
+                    'unsupported format for signature',
```
freakboy3742, line 233: @dreavjr Similarly - how do you reproduce this message?

src/briefcase/platforms/macOS/__init__.py
dreavjr: I'm away from my main machine right now, so I'm cannot give you an exact procedure. This error happened to me while processing Qt for Python 6 due to some violations the package has on the strictly hierarchical structure Apple expects. Thus, signing from the inside out was not enough to warrant that one of the binaries inside Qt for Python was correctly signed, and that one binary had to be signed with the "--deep" option. 

As I understood, this kind of situation is supposed to be rare, so we should not use "--deep" preemptively on everything, because that would perform a huge amount of redundant work.

src/briefcase/platforms/macOS/__init__.py
dreavjr: This message happens when we try to sign a file that is not executable code (meaning that we had a false positive on the filtering procedures that select which files are going to be signed). This may happen, for example, if some package uses the ".so" extension for some proprietary data format.

That happened to me on Qt for Python 6 because they had a couple of non-executable object files marked with the execution permission flag.

src/briefcase/platforms/macOS/__init__.py
freakboy3742: Ok - I've been able to generate this error packaging a PySide6 app.

src/briefcase/platforms/macOS/__init__.py
freakboy3742: Are you saying you get the "unsupported format" message on a vanilla Pyside6 project? I've just generated a "HelloWorld" project with the PySide6 template, and the only file in the bundle that needed special handling was `PySide6/Qt/lib/QtWebEngineCore.framework/Versions/A/QtWebEngineCore` - and that needed a `--deep` pass. 

I've tried:
* Creating a text file with a `.so` extension
* Creating a text file with a `.dylib` extension
* Creating a text file with no extension
* Creating all of the above, and marking them executable
* Creating a binary file with Mach-O magic, but garbage content after that, marked as executable

and passing them to `codesign`, and I don't get any errors.

dreavjr: When it happened to me it was due to a `.o` object file (wrongly?) marked as executable in PySide6. I can look for it as soon as I get access to my main computer. It is conceivable, though, that the behavior is dependent on codesign version. I'll try to reproduce the error on my side and let you know exactly which version I'm using.

dreavjr: @freakboy3742 I'm no longer being able to reproduce a situation that triggers this message. It's possible that it became redundant with the other checks I added upstream. Maybe we can safely remove this downstream check?

freakboy3742: Thanks - I guess it doesn't hurt to retain it just in case. I've flagged the test case to say "don't know how you generate it", so if anyone ever works it out, we can clarify.

Issue #640, opened 2021-11-07
Provide an increased verbosity debug option


freakboy3742:
When Briefcase works well, it's great. However, sometimes, things go wrong - and it's often difficult to diagnose why.

However, Briefcase is only a wrapper around a variety of command line calls. For diagnostic purposes, it would be helpful to be able to ask a user to turn up the verbosity setting, so they can see *exactly* what commands are being issued, in what environment, and what is being returned.

**Describe the solution you'd like**

A user can invoke:

    $ briefcase <command> --debug

To start Briefcase in debugging mode. This doesn't change anything about how Briefcase actually works; but every time it calls `subprocess.run()` or equivalent, it prints to the console:

* the full command line that was executed, in a way that can be copy-and-pasted to a terminal to reproduce the same effect
* the value of all environment variables at the time the command was executed
* the return value of the command
* if possible/appropriate, any console output that is hidden from the user in "normal" operation.

Debugging console output should be clearly differentiated from "normal" briefcase output; for example, by prefixing any debug command with `>>>`, and providing annotations to differentiate between environment variables, command line, output and return codes:
```
$ briefcase run android --debug
Starting emulator beePhone...
>>> Environment: 
>>>     ANDROID_SDK_ROOT=/Users/someone/.briefcase/...
>>>     JAVA_HOME=/Users/someone/.briefcase...
>>> Run:
>>>      /Users/someone/.briefcase/tools/android/emulator @emulator-5554 -dns-server 8.8.8.8
>>> Output:
>>>      ... some command output...
>>>      ... some command output...
>>> Return code: 0
```

**Additional context**

Linux commands can be issued inside Docker; we will need the output of the full Docker command line, not just the command that is executed inside docker.

tusharjayaram:
I'm a first timer and would like to contribute tot his code update for the project. I've read the contributing guidelines and plan to fork a repo locally in the next day or so. Let me know if there's anything else I need to do. Thanks!

freakboy3742:
@tusharjayaram Welcome! There's nothing else specific you need to do, other than "add the feature" :-) If you need any pointers or suggestions, yell out and we'll do what we can to help!

devdev29:
Is this issue still open to PRs? The feature seems to have been implemented already in the subprocess.py file.

freakboy3742:
@devdev29 Yes, it's still open; the runtime environment and return code isn't currently output as part of the current implementation.

freakboy3742:
[...]

Issue #676, opened 2022-03-02
iOS log streaming misses first seconds of log


freakboy3742:
The code to launch an iOS app in the simulator is synchronous, and goes through the following steps:
1. Start simulator app
2. Boot simulator image for requested device type
3. Uninstall old app version
4. Install new app version
5. Start app
6. Start log streaming

However, depending on how long it takes for step (5) to fully complete, and step (6) to fully start, there is a window of opportunity for logs to be lost. In my testing, `Running 'helloworld'` is often the first log message that is seen. This means any logs related to the setup of the Python environment (including PythonPath and NSLog configuration) can be lost. 

**To Reproduce**
1. Run a HelloWorld iOS example
2. Observe that the logs *may* not start with `PythonHome is: ...` 

**Expected behavior**
All system logs should be returned 

**Environment:**
 - Operating System: macOS, running iOS app in simulator
 - Python version: Any
 - Software versions:
   - Briefcase: v0.3.6

**Additional context**

The ideal option would be a "`log stream`, but starting N seconds ago" option, but I do not believe such an option exists. There *is* a "show the last N seconds of logs" option (`log show`), but it requires a separate call to the log stream - and even if `log show` is immediately followed by `log stream`, there is a window between the two calls where logs can be lost.

I suspect this means we'll need to use threading to start `log stream` *before* starting the app.

The same problem almost certainly exists on other platforms as well, so a generic "start app with log streaming" solution would be desirable.

PR #704, opened 2022-04-10
Enhance debug logging for commands executed in the shell

rmartin16:
## Changes
 - Fixes beeware/briefcase#640.
 - Implement logger to encapsulate all application logging including warnings and errors.
 - Update debug mode (i.e -v) to print the full command and any updated environment variables.
 - Introduce deep debug mode (i.e. -vv) to print full command, all environment variables, and command return code.
 - In the event a shell command returns a non-zero return code, the command output is now printed.

## PR Checklist:

-----------Original PR Content Below----------------------------------------------------------------------------------------------
Outstanding Questions
1) Is there any interest or questions in using `logging` like this? If not, no problem; it was mostly find/replace.
2) Ok with just using `-vv` for this debug mode?
3) Outputting _all_ environment variables is especially noisy....or did you intend just the injected env vars to be printed?

(Probably overly verbose thoughts/details below :)

__Logging__
These changes go beyond what #640 is requesting by implementing the `logging` package. By using `logging`, all output is funneled through the same code. This allows for much easier control of the output in the future; for example, `logging` allows us to easily add ">>>" to the front of all debug output by controlling the format of "debug" output in `logging` setup. Or wholesale changes to how warnings and error are displayed. This approach also (theoretically) allows for simpler output redirection than using `print`.

__Activating debug mode__
As opposed to creating a separate flag for debug mode, this simply adds to the output from the `-vv` switch. It seemed like it might be confusing to create another switch if multiple `v` flags are already supported.

__Managing command output__
As for command output, a lot of it is already sent to `stdout/stderr` since `subprocess.run` is being used mostly. I considered that `Popen` could be used instead with `stdout` and `stderr` piped back out and their outputs logged manually from `briefcase`....however, some of this output is pretty robust. For instance, building an APK has some fancy screen writing that's clearly using curses or similar. That would likely be lost or corrupted if all that is piped out of `subprocess`.
That said, `check_output` returns `stdout`....so that can be included pretty easily in debug mode. Although, I found at least one case (i.e trying to run android without vmx enabled) where getting `stderr` would probably be useful but is unavailable because `check_output` is used....any changes there seems pretty out of scope here.

__Return code__
Finally, the return code is nearly always zero as best I can tell.....since `check=True` and `check_output` causes `subprocess` to raise an exception for non-zero return codes. Nonetheless, these changes will include it after calls to `run`.


rmartin16:
@freakboy3742, take two on logging infrastructure. I created a small `Log` class that gets attached to the `command` along side `verbosity` (although, `verbosity` should probably be removed if this is acceptable). It simply encapsulates all the debug logging semantics and doesn't require any more exposure elsewhere to use it than the `command.visibility` property requires.

Additionally, since we're trying to use the `-v` flags for this, I thought it'd be reasonable to actually use them....as such:
 - One v flag:
   - executed command and its arguments
 - Two v flags:
   - executed command's output (unless already included without any flags)
   - executed command's return code
 - Three v flags:
   - environment variables for command

Also, inspired from `ssh`, changed up the debug log preface to `debug#> ` where `#` is the debug level.

Thoughts? If you feel this is still too over-engineered, I can go back to `print` statements :)

rmartin16:
Irrespective of the feedback related to the `Log` class mentioned above, I also made changes to ensure the command output and return code can always logged.
This is accomplished by:
 - intercepting any `CalledProcessError` exceptions from `run` and `check_output`
 - Debug-printing the output & return code stored in the exception
 - Re-raising `CalledProcessError` for the calling code to continue

As it relates to "capture all output, in the form that it is displayed to the user", I experimented a little bit with piping `stdout` and `stderr` back out for commands sent to `run` (by mimicking `run` with a call to `Popen`). This would allow us to capture and preserve the command output while simultaneously printing it to the screen. Although, like I mentioned before, we lose all the pretty-printing and what's printed to the screen looks the same as if the user was piping the command output to a file. I'll do a bit more research but I definitely didn't see a straightforward way to preserve the experience users have today while also capturing stdout/stderr for our own purposes.

freakboy3742 (changes requested):
tl;dr - Some tweaks and questions, but really happy with the direction this is heading.

I've dropped some comments inline; but some general comments:

* I broadly like the direction you've taken with verbosity level interpretations; however, I'd add some caveats. 

  - I'm not sure if the level 2 vs 3 distinction matters; if you're at the point where you're needing to debug to the level of evaluating return codes and command output, having the full environment list is probably also on the cards. I think collapsing both of these into level 2 is probably sufficient.

  - I've flagged inline the need for the *full* kwargs; I'd suggest that level 1 you could get away with just the "additions", but level 2 you need the full environment.
  
* As a purely cosmetic thing, I'm not sure I'm wild about the `DebugN>` prefix. I appreciate it's got precedent, but I'm not sure about the value the additional information it's trying to convey.

* Related `debug1` and `debug2` aren't especially enlightening names (and, to confuse matters more, the numerical values don't align with verbosity). Fitting in with the "only need 2 levels of verbosity" change, I suggest it might be better to name these `debug()` and `deep_debug()`.

* I think the compromise about not "capturing and logging" output for Popen et al is fine; if the output is already displayed as part of normal operation, there's not much utility in logging it as well.

* I'd be interested in your thoughts about whether we should be replacing `print()` calls entirely with `log.info()` (or similar). That's obviously a much bigger set of changes; if we decided to go down this path, I'd be ok with a partial conversion with a longer term goal of transitioning completely. 

* The one use case that I want to give particular attention is the one I mentioned earlier about iOS simulator errors - the class of error where it's difficult/impossible to reproduce because running a second time causes the error to go away. Essentially, we need to be able to ensure that in the event of a command failure, we *always* capture the output, and put it somewhere it can be accessed. That doesn't *necessarily* mean the console - logging to a dump file may be acceptable. Also, sometimes failure is expected - e.g., during initial command/environment detection, failure is an expected result. However, for most "normal operation" commands once this detection has taken place, we need to be able to log output if it will be hard/impossible to reproduce.

rmartin16:
@freakboy3742,
Good catch with the `subprocess` `kwargs`...I was apparently just overlooking that.

Two things before more code changes, though.

__1) One more pitch regarding env vars :)__
What if we just print the full list of environment variables _once_ at the beginning of the beeware command execution and then only print the env vars to be injected for each `subprocess` command? This would avoid printing all the env vars each time a `subprocess` command is run. Based on the my understanding, the env vars should not change once the beeware command execution starts.....although, if that doesn't hold true (or even just may not be true), then all env vars should be printed in entirety each time.

__2) Ephemeral/One-time Error Output__
I am not sure if I am missing something here regarding this statement:

> [W]e need to be able to ensure that in the event of a command failure, we always capture the output, and put it somewhere it can be accessed.

For commands running though `check_output()`, capturing the output is relatively straightforward since the return value is the command's output or the exception has it attached (although, `stderr` handling can complicate things here). However, for commands running through `run()`, `stdout` and `stderr` are being written to the screen in real time. As far as I know, `subprocess` does not support sending the command's `stdout` and `stderr` to the screen while also making that same output available later to the calling python code. That's why I was talking about piping all output for `run` commands back out in to `subprocess.py` so that we could manually write it to the screen and preserve that output for other purposes. (This was also why I started bringing up the compromise about degrading user experience given the fancy screen printing why Apple's/Google's/etc tools would be lost.)

Can you help me reconcile your thoughts with my understanding?


<BR><BR>
P.S. - I definitely support converting all `print()` calls to something like `Log.info()`....furthermore, I can create `Log.warning()` and `Log.error()` as well. This also brings up the `print()` calls that are part of requesting real-time user interaction....I think there is opportunity there as well to better consolidate and normalize putting that information on the screen (or not requesting user interaction as may be necessary). I'll at least include a commit to replace the print calls for current info, warning, and error logging.

freakboy3742:
> **1) One more pitch regarding env vars :)** What if we just print the full list of environment variables _once_ at the beginning of the beeware command execution and then only print the env vars to be injected for each `subprocess` command? This would avoid printing all the env vars each time a `subprocess` command is run. Based on the my understanding, the env vars should not change once the beeware command execution starts.....although, if that doesn't hold true (or even just may not be true), then all env vars should be printed in entirety each time.

Your instincts here are definitely coming from a good place, but they collide up against the harsh reality of how end-users actually behave :-) 

However, if you hunt through bug reports made against Briefcase, a disturbing number of them fall into the "I got an error what happened" bucket - that is, people who report that they experienced a problem, but don't tell you *what the problem actually was*. Or, if they do, they give you the barest minimum of the stack trace - select readings from the last line of console output.

It's also not uncommon to go through a long debugging process trying to work out why some obscure error has occurred, and after multiple back-and-forth interactions, they offer "oh, we also have a firewall that blocks all network access to even numbered IP addresses... could that be causing a problem?".

To that end, we need to make sure that when we provide diagnostic information, we make it as simple and foolproof as possible for them to give us *all* the diagnostic information. Printing the background environment at the start of execution (under a 'max logging' option) is definitely a good idea though - having an idea of the "clean baseline" environment is one of those things that could be helpful as an initial diagnosis. But splitting the diagnostic information into "background environment" and "explicit environment for this command" is asking for trouble - because we'll be in a perpetual cycle of saying "ok, so now can you give us the *rest* of the environment debugging information?". 

It occurs to me that one way to get the "best of both worlds" would be to differentiate between what is printed, and what is output to a log file. Since we now have a single entry point for logging, we can potentially output to multiple locations - console *and* a file. I know this is something that can be captured at the shell with pipes and tee, but you'd be surprised how many users *don't* know how to drive pipes - and in the case of Windows, the option for tee doesn't exist at all AFAIK. This is evidenced by the fact that most console logs from Windows users come in the form of *screenshots* of a console, not text.

Using this approach `briefcase create -vv` will output useful logging detail, but will also create a `briefcase.20220418T0958.create.log` file that contains *all* output from the run. The console log would just show the "environment additions"; but the log *file* would contain the *full* environment. Again - anything we can do to make this foolproof is desirable, and saying "run command with -vv and send us the log file" is on that list. 

> **2) Ephemeral/One-time Error Output** I am not sure if I am missing something here regarding this statement:
> 
> > [W]e need to be able to ensure that in the event of a command failure, we always capture the output, and put it somewhere it can be accessed.
> 
> For commands running though `check_output()`, capturing the output is relatively straightforward since the return value is the command's output or the exception has it attached (although, `stderr` handling can complicate things here). However, for commands running through `run()`, `stdout` and `stderr` are being written to the screen in real time. As far as I know, `subprocess` does not support sending the command's `stdout` and `stderr` to the screen while also making that same output available later to the calling python code. That's why I was talking about piping all output for `run` commands back out in to `subprocess.py` so that we could manually write it to the screen and preserve that output for other purposes. (This was also why I started bringing up the compromise about degrading user experience given the fancy screen printing from Apple's/Google's/etc tools would be lost.)
> 
> Can you help me reconcile your thoughts with my understanding?

It's almost certainly *possible* to capture stdout and stderr, because you can replace stdout and stderr with any file-like object (we do this on iOS and Android to redirect stdout/err to system logs). However, I don't think this is needed here; we usually only care about the literal output of the command when we're using `check_output()`. 

Thinking about the problem some more, it's not even than `check_output()` is returning errors - it's that the post-processing (e.g., parsing the output as JSON) that is the problem. So, there may not actually be any special handling required here - it's more a case of ensuring that everywhere that we're handing "successful" `check_output()` output, the parsing done on that output includes a "dump to log in case of error" handling path.
 
> P.S. - I definitely support converting all `print()` calls to something like `Log.info()`....furthermore, I can create `Log.warning()` and `Log.error()` as well. This also brings up the `print()` calls that are part of requesting real-time user interaction....I think there is opportunity there as well to better consolidate and normalize putting that information on the screen (or not requesting user interaction as may be necessary). I'll at least include a commit to replace the print calls for current info, warning, and error logging.

+1 to this - we have a `--no-input` option for CI operation; anything we can do to formalize/centralize this would be a good addition IMHO.

src/briefcase/commands/base.py
```diff
@@ -55,6 +55,42 @@ def __init__(self, platform):
...
+    """
+    Manage debug logging output driven by verbosity flags.
+    """
+    def __init__(self, verbosity=1):
+        # verbosity will be 1 more than the number of v flags from invocation
+        self.verbosity = verbosity
+        # value to be printed at the beginning of all debug output
+        self.debug_preface = "debug{level}> "
+
+    @staticmethod
```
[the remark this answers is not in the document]
rmartin16: oh...I have a (potentially bad) habit of making any method static if it doesn't reference `self`. Reverted to normal method.

rmartin16:
@freakboy3742,
I completed the effort to consolidate all screen printing and user interaction.

After implementing `Log.info()`, `Log.warning()`, and `Log.error()`, I replaced all calls to print with an appropriate `Log` call.

Normalizing and consolidating user interaction took more time, but I reorganized all the different entry points to request user input around `Console.text_input`. For the most part this was already the case, but there was clearly duplicated efforts going on. 
This was largely achieved by:
 - Retooling `Console.text_input()` to be the real workhorse and basis of user input
 - Retooling `Console.selection_input()` to handle all presentation of choices to users and use `text_input()` for actual user input
 - Reworked `Console.boolean_input()` to directly use `user_input()` and enforce `y/n` answers via a validation function
 - Additionally, any printing to the screen that wholly assumes a user is interacting should use `Console.print()`....which is a simple wrapper on `print()` but respects `Console.enabled`.
   - This is largely for showing real-time file download status or simulator start-up status.

I think most of that is uncontroversial. As best I can, I've tried to maintain the same user experience for soliciting user input.

rmartin16:
With all this infrastructure type stuff mostly outta the way, I can focus more on the business logic.

I think you're last comments helped me understand better where command output is important....yet unavailable. This would largely seemingly stem from `check_output()` when the output isn't what's expected....so the downstream code chokes on it.

I'm gonna spend some more time experimenting and considering options.....but my initial idea is turning `check_output()` into a context manager where calling code tries to act on the output within the `with` block....if errors occur, they can then bubble back up to `subprocess.py`.

I also want to start finalizing the scope of the this PR....since it seems to keep growing :)

Please let me know any significant issues you have with the recent commits (although, I wouldn't bother with too close a review of the code, since I still need to go through with a fine-tooth comb (and still address previous comments)....but the tests are passing :)

src/briefcase/commands/base.py
```diff
@@ -525,10 +571,10 @@ def download_url(self, url, download_path):
                         downloaded += len(data)
                         f.write(data)
                         done = int(50 * downloaded / total)
-                        print('\r{}{} {}%'.format('#' * done, '.' * (50-done), 2*done), end='', flush=True)
-            print()
+                        self.input.print('\r{}{} {}%'.format('#' * done, '.' * (50-done), 2*done), end='', flush=True)
```
freakboy3742, line 574: Not sure why this is going through `input`; it's ostensibly a progress bar, so it's 100% output functionality; and the lone "print" on the next line is what does the newline flush.

I'd be more inlined to wrap this as a literal "show progress" function - possibly a context manager where exiting does the implied newline at the end?

src/briefcase/console.py
```diff
@@ -21,6 +24,39 @@ def enabled(self):
     def enabled(self, enabled):
         self._enabled = enabled
 
+    @staticmethod
+    def titlecase(s):
```
freakboy3742, line 28: Generally, I'm not a fan of staticmethod. Using classes as a namespace is a hangover of languages like Java that don't have "top level" functions. In this case, there's "Console" specific about titlecase - it's a utility method that will work on any string. It makes more sense as a top level function so it can be invoked without having any illusion that it's part of Console.

src/briefcase/console.py
```diff
@@ -34,6 +70,17 @@ def boolean_input(self, question, default=False):
...
+
+        def normalize_input(user_input):
+            """Normalize user input to just the lowercase version of the first character."""
+            return user_input.lower()[:1]
+
+        def is_y_or_n(user_input):
+            """Validator for text_input() that user entered a literal 'y' or 'n'."""
+            if normalize_input(user_input) in {'y', 'n'}:
+                return True
+            raise ValueError("Please enter Y or N")
```
freakboy3742, line 82: This is... very complicated for a "is the lower case input y or n". 2 closures, one referencing the other... something that was, in the previous implementation, a 1 line lambda.

src/briefcase/console.py
```diff
@@ -45,122 +92,163 @@ def boolean_input(self, question, default=False):
...
+        If intro is specified, it will be presented once prior to the input prompt.
+
+        If prompt is not specified, a prompt of the form "{input_name} [{default}}: "
+            will be shown to the user.
+        If prompt is specified, it is shown to the user and input_name is ignored.
+
+        If validator is not specified:
+            the user's input is returned.
+            If the user enters nothing, the default value is returned if specified,
+                else an empty string is returned.
```
freakboy3742, line 190: When the documentation for your method contains multiple levels of if clauses, it's time to reconsider whether the design is correct :-)

src/briefcase/commands/base.py
```diff
@@ -55,6 +54,50 @@ def __init__(self, platform):
         )
 
 
+class Log:
```
freakboy3742, line 57: As a structural matter, I'd argue that this should be part of the new console module as well. Although it will only be used by Commands in practice, there's nothing Command-specific about what it's doing.

src/briefcase/integrations/android_sdk.py
```diff
@@ -496,20 +496,23 @@ def select_target_device(self, device_or_avd):
...
             # If input is disabled, and there's only one actual simulator,
             # select it. If there are no simulators, select "Create simulator"
             if len(choices) <= 2:
                 choice = choices[0][0]
             else:
-                raise BriefcaseCommandError(
-                    "Input has been disabled; can't select a device to target."
-                )
+                raise BriefcaseCommandError("""
+Input has been disabled and the device to target is ambiguous.
```
freakboy3742, line 512: It took me a couple of reads to determine what this error was trying to say - my initial read is that it was logically inconsistent - either input has been disabled, *or* the device is ambiguous. The old message was a lot more direct; although I'll admit the "use -d" change is a useful addition.

Adding the -d

src/briefcase/integrations/android_sdk.py
```diff
@@ -713,12 +716,12 @@ def start_emulator(self, avd):
 
             # Step 1: Wait for the device to appear so we can get an
             # ADB instance for the new device.
-            print()
-            print('Waiting for emulator to start...', flush=True, end='')
+            self.command.input.print()
+            self.command.input.print('Waiting for emulator to start...', flush=True, end='')
```
freakboy3742, line 720: This is another progress bar... (this time one with undetermined length); it's not an input condition.

freakboy3742 (changes requested):
There's a lot in this PR; a lot of it is mechanical updates of `print()`, and a lot of it... isn't. Agreed that we need to start thinking about landing this, though, because it could easily spiral into an explosion of scope.

I will give my +1 to the general idea of extracting input handling into into a general utility module... but the refactoring you've done along the way is somewhat excessive. The methods for input handling have gotten a lot more complicated, with all sorts of short-lived closures and partial function applications... but it's a little hard to see what extra functionality has been gained as a result. My inclination would be to revert those changes, refactor the methods *as is* to the new console module, and look at any redesign or refactoring as a later effort (or, to be even more extreme - revert this back to the original scope of "just deal with logging warning and errors", and deal with input and 'normal' output as a later cleanup).

If we *do* keep "all console output" as part of the scope of this PR, one last thing to consider is how we protect ourselves against future PRs inadvertently introducing `print()`. Ideally, any call to `print()` would fail during the test suite; replacing the `print()` builtin with `pytest.fail()` during test suite setup might be one approach. We have near complete code coverage, so any accidental usage of `print()` would then raise an exception in the codebase. 

Lastly, regarding `check_output()` checks - I think a context manager is probably overkill. The use case here will almost always be "do some parsing to the output of check_output, and the parsing fails". A `parse_output()` wrapper that is "check output but with a user-supplied parser function" would be more than sufficient. If the command fails, it raises an error; if the parsing fails, it raises a different error (and logs the actual output). I'd need to check, but we might even be able to go as far as wrapping some common parsing use cases - I know "regex for a version number" and "JSON" are two types of parsing that are used more than once. That said - I'd also be comfortable putting this idea on the list for a second pass at this problem.

rmartin16:
That's fair; I clearly underestimated how opinionated my refactor is. I was mostly targeting a single user input loop...but that goal may not be laudable. I'll revert the refactor and probably mostly just target replacing the "input" `print()` calls with `Console.print()`.

Thanks for the thoughts on `check_output()`; something like `parse_output()` does seem more apt for the job.

src/briefcase/commands/base.py
rmartin16: makes sense. moved to `console.py`.

src/briefcase/integrations/android_sdk.py
rmartin16: yeah....i mostly just wanted to add the `use -d` because it was confusing why it failed with `--no-input`.

src/briefcase/console.py
rmartin16: Reverted the input handling refactor.

src/briefcase/console.py
rmartin16: Reverted the input handling refactor.

src/briefcase/console.py
rmartin16: Reverted the input handling refactor. Simply moved this back to `new.py`.

src/briefcase/commands/base.py
rmartin16: I implemented a progress bar context manager in Console.

src/briefcase/integrations/android_sdk.py
rmartin16: Created a wait bar context manager in Console. Implemented each place this sort of thing is happening.

rmartin16:
Ok....reverted the input handling refactor; I left `input_text()` and `input_select()` in `NewCommand` since I felt their current naming could create confusion in `Console` as is. I just replaced their `print()` calls.

I introduced `ProgressBar` and `WaitBar` in `console.py`. These are pretty straightforward context managers....I'll add tests if we want to keep them. I also decided _not_ to have `ProgressBar` and `WaitBar` respect `Console.enabled`....it was debatable if `ProgressBar` should but definitely seemed inappropriate for `WaitBar`.

I'll turn now to evaluating `check_output` usage to see what output parsing can be generalized for them.


Additionally, I don't know how you feel about f-strings versus `"".format()`....but I'd be willing to convert them all to f-strings if you see any benefit....especially since I'm already touching all this code with strings.

src/briefcase/platforms/iOS/xcode.py
```diff
@@ -420,9 +416,11 @@ def run_app(self, app: BaseConfig, udid=None, **kwargs):
...
         # Start streaming logs for the app.
         try:
-            print()
-            print("[{app.app_name}] Following simulator log output (type CTRL-C to stop log)...".format(app=app))
-            print("=" * 75)
+            self.logger.info()
+            self.logger.info(
+                "[{app.app_name}] Following simulator log output (type CTRL-C to stop log)...".format(app=app)
+            )
+            self.logger.info("=" * 75)
```
freakboy3742, line 423: Absolutely not in scope for this PR, but worth being aware of if you're at all interested in continuing past this PR: #676 tracks a known issue with capturing log output that we need to address.

tests/integrations/docker/test_Docker__run.py
```diff
@@ -100,12 +100,12 @@ def test_simple_verbose_call(mock_docker, tmp_path, capsys):
             'world',
         ]
     )
-    assert capsys.readouterr().out == (
-        ">>> docker run --tty "
+    assert capsys.readouterr().out.splitlines()[-1] == (
+        ">>>     docker run --tty "
```
freakboy3742, line 104: What's the origin of this extra space?

tests/integrations/docker/test_Docker__run.py
```diff
@@ -100,12 +100,12 @@ def test_simple_verbose_call(mock_docker, tmp_path, capsys):
...
     )
-    assert capsys.readouterr().out == (
-        ">>> docker run --tty "
+    assert capsys.readouterr().out.splitlines()[-1] == (
+        "debug1>     docker run --tty "
         "--volume {platform_path}:/app:z "
         "--volume {dot_briefcase_path}:/home/brutus/.briefcase:z "
         "briefcase/com.example.myapp:py3.X "
-        "hello world\n"
+        "hello world"
```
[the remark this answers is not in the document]
freakboy3742: AFAICT this comment is still current.

tests/integrations/subprocess/test_Subprocess__Popen.py
```diff
@@ -36,9 +38,9 @@ def test_simple_call_with_path_arg(mock_sub, capsys, tmp_path):
...
 def test_simple_verbose_call(mock_sub, capsys):
     "If verbosity is turned out, there is output"
-    mock_sub.command.verbosity = 2
+    mock_sub.command.logger = Log(verbosity=2)
 
     mock_sub.Popen(['hello', 'world'])
 
     mock_sub._subprocess.Popen.assert_called_with(['hello', 'world'])
-    assert capsys.readouterr().out == ">>> hello world\n"
+    assert capsys.readouterr().out.splitlines()[-1] == ">>>     hello world"
```
freakboy3742, line 46: As with the other output test - this is a slightly different test, and the origin of the extra space isn't clear

tests/integrations/subprocess/test_Subprocess__check_output.py
```diff
@@ -36,9 +38,9 @@ def test_simple_call_with_path_arg(mock_sub, capsys, tmp_path):
...
 def test_simple_verbose_call(mock_sub, capsys):
     "If verbosity is turned out, there is output"
-    mock_sub.command.verbosity = 2
+    mock_sub.command.logger = Log(verbosity=2)
 
     mock_sub.check_output(['hello', 'world'])
 
     mock_sub._subprocess.check_output.assert_called_with(['hello', 'world'])
-    assert capsys.readouterr().out == ">>> hello world\n"
+    assert capsys.readouterr().out.splitlines()[-1] == ">>>     hello world"
```
freakboy3742, line 46: Same again on this one - test change, plus extra space

freakboy3742 (changes requested):
This is looking really good; I think we're getting close to the final strait (or at least close enough to something that is mergeable).

In addition to some of the things I've flagged inline, two larger items still remaining:
1. A changenote so that CI will run to completion
2. Test coverage. Most (all?) of this change will likely get test coverage indirectly, but I wouldn't be surprised if there are some missing test branches. We'll get a better feel for things when (1) is completed.

As for your other comments:
* I'm comfortable leaving input_text and input_select as is for now; but I also wouldn't complain if there was a later PR cleaning up that usage, as there's clearly some overlap in purpose.
* ProgressBar and WaitBar both look great, with the exception of the one cosmetic thing described in the comments
* At this point, this PR is a *huge* improvement (both structurally, and in terms of surfacing the environment) to the status quo; I'd be happy to consider any console_output improvement as a v2 goal.
* 100% of using f-strings. The `.format()` calls are a historical artefact, because most of Briefcase was originally written when we still supported Python 3.5. My general approach has been "fix them while you're in the area"; but if there was a wholesale "replace format with f-string" patch in my inbox, I wouldn't reject it. It's up to you how much f-string updating you want to include in this patch.

src/briefcase/commands/new.py
```diff
@@ -454,21 +452,21 @@ def build_app_context(self):
                 "GNU General Public License v3 (GPLv3)",
                 "GNU General Public License v3 or later (GPLv3+)",
                 "Proprietary",
-                "Other"
-            ],
```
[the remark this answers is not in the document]
rmartin16: I should have reverted this along with reverting the refactor...although, I thought it would be innocuous to use a tuple here since I wouldn't expect anything to need to change this list of values.

Not that I'm pushing for a tuple, so mostly out of curiosity, why so adamant this is a list? (always looking to learn :) thanks)

src/briefcase/commands/new.py
```diff
@@ -454,21 +452,21 @@ def build_app_context(self):
...
-                'PursuedPyBear (does not support iOS/Android deployment)',
-                'None',
-            ],
+            options=(
+                "Toga",
+                "PySide2 (does not support iOS/Android deployment)",
+                "PySide6 (does not support iOS/Android deployment)",
+                "PursuedPyBear (does not support iOS/Android deployment)",
+                "None",
+            ),
```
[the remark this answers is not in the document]
rmartin16: Reverted to a list.

src/briefcase/platforms/iOS/xcode.py
rmartin16: Thanks for letting me know. I'll take a look.

src/briefcase/integrations/subprocess.py
```diff
@@ -44,70 +44,112 @@ def final_kwargs(self, **kwargs):
...
         # Invoke subprocess.run().
         # Pass through all arguments as-is.
-        # All exceptions are propegated back to the caller.
-        if self.command.verbosity >= 2:
-            print(">>> {cmdline}".format(
-                cmdline=' '.join(shlex.quote(str(arg)) for arg in args)
-            ))
+        # All exceptions are propagated back to the caller.
+        self._log_command(args)
+        self._log_environment(kwargs)
```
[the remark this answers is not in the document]
rmartin16: Ensured that `kwargs` is also updated first.

tests/commands/base/test_parse_options.py
```diff
@@ -16,7 +16,7 @@ def test_parse_options(base_command):
         'required': "important",
     }
     assert base_command.input.enabled
-    assert base_command.verbosity
+    assert base_command.logger.verbosity
```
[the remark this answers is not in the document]
rmartin16: Updated to check for `1` now.

tests/integrations/subprocess/test_Subprocess__run.py
```diff
@@ -36,9 +38,9 @@ def test_simple_call_with_path_arg(mock_sub, capsys, tmp_path):
...
 def test_simple_verbose_call(mock_sub, capsys):
     "If verbosity is turned out, there is output"
-    mock_sub.command.verbosity = 2
+    mock_sub.command.logger = Log(verbosity=2)
 
     mock_sub.run(['hello', 'world'])
 
     mock_sub._subprocess.run.assert_called_with(['hello', 'world'])
-    assert capsys.readouterr().out == ">>> hello world\n"
+    assert capsys.readouterr().out.splitlines()[-1] == "debug1>     hello world"
```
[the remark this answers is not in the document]
rmartin16: Updated to verify the full output.

tests/integrations/subprocess/test_Subprocess__check_output.py
```diff
@@ -36,9 +38,9 @@ def test_simple_call_with_path_arg(mock_sub, capsys, tmp_path):
...
 def test_simple_verbose_call(mock_sub, capsys):
     "If verbosity is turned out, there is output"
-    mock_sub.command.verbosity = 2
+    mock_sub.command.logger = Log(verbosity=2)
 
     mock_sub.check_output(['hello', 'world'])
 
     mock_sub._subprocess.check_output.assert_called_with(['hello', 'world'])
-    assert capsys.readouterr().out == ">>> hello world\n"
+    assert capsys.readouterr().out.splitlines()[-1] == "debug1>     hello world"
```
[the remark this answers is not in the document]
rmartin16: Updated to verify the full output.

tests/integrations/subprocess/test_Subprocess__Popen.py
```diff
@@ -36,9 +38,9 @@ def test_simple_call_with_path_arg(mock_sub, capsys, tmp_path):
...
 def test_simple_verbose_call(mock_sub, capsys):
     "If verbosity is turned out, there is output"
-    mock_sub.command.verbosity = 2
+    mock_sub.command.logger = Log(verbosity=2)
 
     mock_sub.Popen(['hello', 'world'])
 
     mock_sub._subprocess.Popen.assert_called_with(['hello', 'world'])
-    assert capsys.readouterr().out == ">>> hello world\n"
+    assert capsys.readouterr().out.splitlines()[-1] == "debug1>     hello world"
```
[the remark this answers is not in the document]
rmartin16: Updated to verify the full output.

tests/integrations/docker/test_Docker__run.py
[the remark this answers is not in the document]
rmartin16: Updated to verify the full output.

src/briefcase/integrations/subprocess.py
```diff
@@ -44,70 +44,112 @@ def final_kwargs(self, **kwargs):
...
         environment will be copied, and the contents of env overwritten
         into that environment.
         """
-        # Invoke subprocess.check_output
-        if self.command.verbosity >= 2:
-            print(">>> {cmdline}".format(
-                cmdline=' '.join(shlex.quote(arg) for arg in args)
-            ))
+        self._log_command(args)
+        self._log_environment(kwargs)
```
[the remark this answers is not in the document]
rmartin16: Ensured that `kwargs` is updated first.

tests/integrations/docker/test_Docker__run.py
rmartin16: Updated this test to check entire output.

The extra spaces are there because command being executed is indented now.

This makes more sense in the context of all debug output; for instance:

```
❯ briefcase create -vv
>>> 
>>> Running Command:
>>>     docker --version
>>> Environment:
>>>     LESSCLOSE=/usr/bin/lesspipe %s %s
[OMITTED]
>>>     DISPLAY=:0
>>> Command Output:
>>>     Docker version 20.10.14, build a224086
>>> Return code: 0
[TRUNCATED]
```

tests/integrations/subprocess/test_Subprocess__Popen.py
rmartin16: Updated the test to compare entire output.

Extra whitespace is from command being indented. Further explanation in other comment.

tests/integrations/subprocess/test_Subprocess__check_output.py
rmartin16: Updated the test to compare entire output.

Extra whitespace is from command being indented. Further explanation in other comment.

tests/integrations/subprocess/test_Subprocess__run.py
```diff
@@ -36,9 +38,9 @@ def test_simple_call_with_path_arg(mock_sub, capsys, tmp_path):
...
 def test_simple_verbose_call(mock_sub, capsys):
     "If verbosity is turned out, there is output"
-    mock_sub.command.verbosity = 2
+    mock_sub.command.logger = Log(verbosity=2)
 
     mock_sub.run(['hello', 'world'])
 
     mock_sub._subprocess.run.assert_called_with(['hello', 'world'])
-    assert capsys.readouterr().out == ">>> hello world\n"
+    assert capsys.readouterr().out.splitlines()[-1] == ">>>     hello world"
```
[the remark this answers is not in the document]
rmartin16: Updated the test to compare entire output.

Extra whitespace is from command being indented. Further explanation in other comment.

src/briefcase/console.py
```diff
@@ -21,6 +65,11 @@ def enabled(self):
     def enabled(self, enabled):
         self._enabled = enabled
 
+    def print(self, *values, **kwargs):
```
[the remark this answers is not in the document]
rmartin16: I agree; as the use-case for printing that isn't "logging" is more clear, `prompt` makes more sense here. updated.

src/briefcase/console.py
```diff
@@ -117,8 +165,75 @@ def __call__(self, prompt):
...
         if not self.enabled:
             raise InputDisabled()
+        try:
+            return self._input(prompt)
+        except EOFError:
+            raise KeyboardInterrupt
 
-        return self._input(prompt)
+
+class ProgressBar:
```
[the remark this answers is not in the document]
rmartin16: I had considered this as well....and with additional testing, this is actually already a problem when you consider debug logging.

For instance:
```
❯ briefcase run android -d @beePhone -v
>>> 
>>> Running Command:
>>>     /home/user/.briefcase/tools/android_sdk/platform-tools/adb devices -l
>>> 
>>> Running Command:
>>>     /home/user/.briefcase/tools/android_sdk/emulator/emulator -list-avds
>>> 
>>> Running Command:
>>>     /home/user/.briefcase/tools/android_sdk/emulator/emulator -list-avds
Starting emulator beePhone...
>>> 
>>> Running Command:
>>>     /home/user/.briefcase/tools/android_sdk/emulator/emulator @beePhone -dns-server 8.8.8.8

Waiting for emulator to start....>>> 
>>> Running Command:
>>>     /home/user/.briefcase/tools/android_sdk/platform-tools/adb devices -l
.>>> 
>>> Running Command:
>>>     /home/user/.briefcase/tools/android_sdk/platform-tools/adb devices -l
.>>> 
>>> Running Command:
>>>     /home/user/.briefcase/tools/android_sdk/platform-tools/adb devices -l
.>>> 
>>> Running Command:
>>>     /home/user/.briefcase/tools/android_sdk/platform-tools/adb devices -l
>>> 
>>> Running Command:
>>>     /home/user/.briefcase/tools/android_sdk/platform-tools/adb -s emulator-5554 emu avd name

 booting...>>> 
>>> Running Command:
>>>     /home/user/.briefcase/tools/android_sdk/platform-tools/adb -s emulator-5554 shell getprop sys.boot_completed
[TRUNCATED]
```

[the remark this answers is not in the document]
rmartin16: So, any larger solution would also probably need some closer integration between `Log` and `Console`....at least so that `Log` calls could interact with any ongoing progress/wait bars. Perhaps making `Log` and `Console` singletons would make that more straightforward than trying to make them aware of each other's instantiations....absent a much different refactor.

At any rate, exposing the progress and wait bars from `Console` is probably a better starting place for this. I added two new methods in `Console` to simply return an instantiated progress/wait bar.

src/briefcase/commands/new.py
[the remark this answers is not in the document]
freakboy3742: So: 
* A list is a structure of variable length, where all the items are of the same type, and there is no semantic significance to position other than ordering. Items can be arbitrarily added to a list.
* A tuple is of fixed length, items can (but don't have to be) be of different type, and there is a specific semantic interpretation of each position. Items cannot be arbitrarily added to a tuple; if something changes from an 3-tuple to a 4-tuple, the addition of the fourth element may change the interpretation of the first three.

Following that:
* "Here are your options, please choose" is a list, because each option is type equivalent, and changing the order doesn't change the interpretation of each item, just the index of the item.
* Size/Position are tuples, because a size is always (width, height), and position is (x,y) or (x,y,z).
* RGBA Color is a tuple; RGB are (usually) integers, A is a float, but the order isn't exchangeable; adding A to an RGB triple changes the interpretation of R, G, and B.
* A server `(host, port)` declaration is a tuple; host is a string, port is an integer.

As for why it matters - this is definitely in the weeds of "strong opinions". 

Functionally, the two will behave the same in most circumstances. However, there's an (admittedly weak) argument to be made about what you're communicating to the API user about the significance of position and type in the data you store in a list, vs a tuple. 

There's also a pedagogical concern. Python is a wonderful language that I love dearly, but one of its weaknesses from a teaching perspective is that the core data types are so powerful that new programmers often aren't exposed to the concept of data structures in any meaningful way. For beginning programmers (and, frankly, a large number of experienced programmers), this doesn't even matter because the core Python data types are so damn good - but one of the "big concepts" that you need to grok as you advance in skill is how data structures work, and how the choice of data structure can impact performance. Tuple vs List is a teaching moments - it's a chance to seed the idea that there are usually multiple ways to solve any problem, and choosing the right data structure is one of the things you need to do. It's very low stakes from a functional perspective, but it's also relatively easy to explain, and it provides a foundation for later discussions about why you might choose a list over a dictionary, or use something else entirely (like a linked list, or something else that isn't in the standard library).

The pedagogical concern is the bigger concern in my specific case here. I've spent a lot of energy encouraging contributors of all levels of experience; because of that, I sweat the small stuff like formatting and code structure so that the resulting codebase follows best practices.

tests/integrations/docker/test_Docker__run.py
freakboy3742: Ah - that makes sense. Thanks.

src/briefcase/integrations/android_sdk.py
```diff
@@ -733,34 +735,32 @@ def start_emulator(self, avd):
...
+                                known_devices.add(device)
+
+                    # Try again in 2 seconds...
+                    self.sleep(2)
 
             # Phase 2: Wait for the boot process to complete
-            while not adb.has_booted():
-                if emulator_popen.poll() is not None:
-                    raise BriefcaseCommandError("""
+            with self.command.input.wait_bar(" booting...") as boot_wait_bar:
```
freakboy3742, line 760: Really cosmetic; but the space before ` booting` existed previously because it was all on one line; we can drop the space here because it's the start of a new line. It should also be capitalized.

src/briefcase/integrations/subprocess.py
```diff
@@ -53,61 +53,108 @@ def run(self, args, **kwargs):
...
-            **self.final_kwargs(**kwargs)
+            **subprocess_kwargs
         )
+
+    def _log_command(self, args):
+        """
+        Log the entire console command being executed.
+        """
+        cmdline = ' '.join(shlex.quote(str(arg)) for arg in args)
+        self.command.logger.debug("\nRunning Command:")
```
freakboy3742, line 129: Why the `\n` here? Is this to get around the "we might have been mid-progress bar" problem?

tests/integrations/subprocess/test_Subprocess__Popen.py
```diff
@@ -34,11 +36,30 @@ def test_simple_call_with_path_arg(mock_sub, capsys, tmp_path):
...
+    mock_sub.command.logger = Log(verbosity=3)
+
+    mock_sub.Popen(["hello", "world"])
+
+    mock_sub._subprocess.Popen.assert_called_with(["hello", "world"])
+
+    expected_output = ">>> \n>>> Running Command:\n>>>     hello world\n>>> Environment:"
+    # some env vars (eg PS1) can contain line breaks...so this tries to replicate
+    # Log._log()'s functionality to print multi-line content with the appropriate preface.
+    for env_var, value in os.environ.items():
```
freakboy3742, line 60: Testing against the live os environment will be an unreliable test, because every test environment will be different. We should stub the response from os.environ, include examples where the environment variable might have line breaks, and use a literal as the expected output.

tests/integrations/subprocess/test_Subprocess__check_output.py
```diff
@@ -34,11 +36,31 @@ def test_simple_call_with_path_arg(mock_sub, capsys, tmp_path):
...
+    mock_sub.command.logger = Log(verbosity=3)
+
+    mock_sub.check_output(["hello", "world"])
+
+    mock_sub._subprocess.check_output.assert_called_with(["hello", "world"])
+
+    expected_output = ">>> \n>>> Running Command:\n>>>     hello world\n>>> Environment:"
+    # some env vars (eg PS1) can contain line breaks...so this tries to replicate
+    # Log._log()'s functionality to print multi-line content with the appropriate preface.
+    for env_var, value in os.environ.items():
```
freakboy3742, line 60: As above, a mocked os.environ is needed here.

tests/integrations/subprocess/test_Subprocess__run.py
```diff
@@ -34,11 +36,30 @@ def test_simple_call_with_path_arg(mock_sub, capsys, tmp_path):
...
+    mock_sub.command.logger = Log(verbosity=3)
+
+    mock_sub.run(["hello", "world"])
+
+    mock_sub._subprocess.run.assert_called_with(["hello", "world"])
+
+    expected_output = ">>> \n>>> Running Command:\n>>>     hello world\n>>> Environment:"
+    # some env vars (eg PS1) can contain line breaks...so this tries to replicate
+    # Log._log()'s functionality to print multi-line content with the appropriate preface.
+    for env_var, value in os.environ.items():
```
freakboy3742, line 60: As above, a mocked os.environ is needed here.

tests/integrations/docker/test_Docker__run.py
```diff
@@ -81,32 +71,27 @@ def test_simple_call_with_path_arg(mock_docker, tmp_path, capsys):
...
+            "hello",
+            "world",
         ]
     )
     assert capsys.readouterr().out == (
-        ">>> docker run --tty "
-        "--volume {platform_path}:/app:z "
-        "--volume {dot_briefcase_path}:/home/brutus/.briefcase:z "
+        ">>> \n"
+        ">>> Running Command:\n"
```
freakboy3742, line 91: The extra newlines on these two intrigue me - where do they come from? `subprocess.run` will do a clearing newline, but that doesn't seem to match what is being output here

freakboy3742 (changes requested):
This is looking great; I think we're almost entirely into testing as the only thing preventing a merge. 

I've flagged a couple of minor test cleanups; also, as expected, the coverage has highlighted some extra lines that need tests. AFAICT one of them isn't actually your fault (the one in commands/dev.py); if you can't work out an easy way to test that one, let me know and I'll tweak it before merging. However, the others look like they are areas that need to be exercised in a test to ensure they're working as expected.

I'm not sure what to do about the "-v output during progress bar" problem; however, the behaviour as-of this patch is probably good enough for now - it's a cosmetic problem when you have the debug turned up to 11, so it's not unexpected that this will cause problems. I have some ideas involving tracking whether there is a "current" progress bar... I might have a tinker when I'm doing final review.

The only other question I've got is about something we discussed early on - the idea of using `-v` to show just the environment additions, and `-vv` to show the full environment. I suspect this is something I may need to play with just to see how it "tastes" in practice, but if you've got any thoughts, let me know.

src/briefcase/console.py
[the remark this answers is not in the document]
freakboy3742: A related problem - creating a progress bar while you're already inside a progress bar. Not likely to be something that is a real problem in practice, but the accounting we'd need to do to track "are we in a progress bar" would also help us identify if there is a need to flush output for debugging, etc.

src/briefcase/integrations/android_sdk.py
rmartin16: Updated.

src/briefcase/integrations/subprocess.py
rmartin16: This is a carryover from when I was trying to decide the best way to introduce empty lines in log output; whether to make empty calls to `Log` or put line breaks directly in to the message. Given all the existing empty `print` calls, I eventually just went with empty `Log` calls.

I updated this to remove the `\n` and call `logger.debug()` instead.

The addition of this empty log output was driven by demarcating the logging of multiple shell commands running back to to back.

For instance:
```
❯ briefcase create -vv
>>> 
>>> Running Command:
>>>     docker --version
>>> Environment:
>>>     CLUTTER_IM_MODULE=xim
[OMITTED]
>>>     _=/home/user/github/briefcase/venv-3.9/bin/briefcase
>>> Command Output:
>>>     Docker version 20.10.14, build a224086
>>> Return code: 0
>>> 
>>> Running Command:
>>>     docker info
>>> Environment:
>>>     CLUTTER_IM_MODULE=xim
[OMITTED]
>>>     _=/home/user/github/briefcase/venv-3.9/bin/briefcase
>>> Command Output:
>>>     Client:
>>>      Context:    default
>>>      Debug Mode: false
>>>      Plugins:
>>>       app: Docker App (Docker Inc., v0.9.1-beta3)
>>>       buildx: Docker Buildx (Docker Inc., v0.8.1-docker)
>>>       scan: Docker Scan (Docker Inc., v0.17.0)
>>>     
>>>     Server:
>>>      Containers: 106
>>>       Running: 0
>>>       Paused: 0
>>>       Stopped: 106
>>>      Images: 71
>>>      Server Version: 20.10.14
>>>      Storage Driver: overlay2
>>>       Backing Filesystem: extfs
>>>       Supports d_type: true
>>>       Native Overlay Diff: true
>>>       userxattr: false
>>>      Logging Driver: json-file
>>>      Cgroup Driver: cgroupfs
>>>      Cgroup Version: 1
>>>      Plugins:
>>>       Volume: local
>>>       Network: bridge host ipvlan macvlan null overlay
>>>       Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
>>>      Swarm: inactive
>>>      Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
>>>      Default Runtime: runc
>>>      Init Binary: docker-init
>>>      containerd version: 3df54a852345ae127d1fa3092b95168e4a88e2f8
>>>      runc version: v1.0.3-0-gf46b6ba
>>>      init version: de40ad0
>>>      Security Options:
>>>       apparmor
>>>       seccomp
>>>        Profile: default
>>>      Kernel Version: 5.4.0-109-generic
>>>      Operating System: Ubuntu 18.04.6 LTS
>>>      OSType: linux
>>>      Architecture: x86_64
>>>      CPUs: 4
>>>      Total Memory: 9.728GiB
>>>      Name: virt-ubuntu18
>>>      ID: 2O2U:B6J2:M55X:PQ6G:SGSF:3RDN:H2GE:XV37:3F6N:GRPV:JGY2:Y2D6
>>>      Docker Root Dir: /var/lib/docker
>>>      Debug Mode: false
>>>      Registry: https://index.docker.io/v1/
>>>      Labels:
>>>      Experimental: false
>>>      Insecure Registries:
>>>       127.0.0.0/8
>>>      Live Restore Enabled: false
>>>     
>>>     WARNING: No swap limit support
>>> Return code: 0

Application helloworld already exists; overwrite [y/N]? 
```

Current output for posterity:
```
❯ briefcase create -vv
>>> docker --version
>>> docker info

Application helloworld already exists; overwrite [y/N]?
```

tests/integrations/subprocess/test_Subprocess__Popen.py
rmartin16: ah, of course. i was so focused on if i could, i didn't think if i should....

Replaced `os.environ` with mock and a literal comparison.

tests/integrations/subprocess/test_Subprocess__check_output.py
rmartin16: Replaced `os.environ` with mock and a literal comparison.

tests/integrations/subprocess/test_Subprocess__run.py
rmartin16: Replaced `os.environ` with mock and a literal comparison.

tests/integrations/docker/test_Docker__run.py
rmartin16: This also goes back to the additional new line I introduced via `Subprocess._log_command()` and talked about above.

src/briefcase/integrations/subprocess.py
```diff
@@ -44,70 +44,112 @@ def final_kwargs(self, **kwargs):
...
-        )
+        try:
+            cmd_output = self._subprocess.check_output(
+                [
+                    str(arg) for arg in args
+                ],
+                **self.final_kwargs(**kwargs)
+            )
+        except subprocess.CalledProcessError as exc:
+            self._log_output(exc.output, exc.stderr)
```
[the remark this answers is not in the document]
rmartin16: I agree with potential for confusion from potentially disjointed output in the event of an error.

What may be possible here is to always force `stderr` to be redirected to `stdout` for `check_output` calls. Although, I am not sure if that's even safe or possible as is; it may be best suited for that to be evaluated more in depth within an effort to generalize parsing return values from `check_output`.

src/briefcase/commands/new.py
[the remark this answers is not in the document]
rmartin16: I see. That definitely makes sense. Thank you for your insight.

rmartin16:
All right; I think everything is in.

Code coverage
* I made sure anything new has coverage now.
* Additionally, I added new tests for `DevCommand.install_dev_dependencies`; I didn't really understand how the code had coverage currently....so, I more or less mimicked the testing for `CreateCommand.install_app_dependencies`.
* Finally, I added a mechanism in the top-level `conftest.py` to fail any test that invokes `print` outside of `briefcase.console`....however, I'm not entirely sure where this strategy might sit on the spectrum of "things you should ever consider doing"....I'm still trying to pick up best practices for testing.

Below is my list of potential follow on tasks/issues from this PR:
* Generalize `check_output` parsing and logging
* `check_output` failure can result in displaying `stdout` and `stderr` independently when they should be interleaved
* Add support to properly print to the screen during an active `ProgressBar` or `WaitBar`
* Convert format strings to f-strings
* Refactor input methods in `NewCommand` in to `Console`

Along with any additional comments you have for me, please let me know if I should squash all these commits....or any other housecleaning; thanks.

src/briefcase/integrations/subprocess.py
freakboy3742: Ok - that makes sense. I'll probably tweak this when I land so that a "clearing debug newline" (i.e., debug with *no* input) omits the `>>>` preface; this has the added benefit that if you call multiple commands in a row, there's even better demarcation between calls.

freakboy3742 (approved):
I've just pushed a couple of minor tweaks and cosmetic improvements that are easier for me to just make than to explain - but otherwise, this is fantastic! Thanks for all your hard work on this (including, but not limited to, putting up with my pedantry :-)

As for the follow-up items - the list you've provided looks good; although the stderr/stdout interleaving item is probably not a high priority (for the reason I flagged inline). 

Thanks again - may this be the first of many contributions!

Issue #386, opened 2020-04-29
Backfill icons from provided images


freakboy3742:
Briefcase currently requires that the user provide icons of in various formats. However, if you have a high-res PNG format, any other format can be generated by image conversion.

Briefcase should try to use a correctly named file in the right size and format if it's available; but if it isn't, it should try to backfill by converting the high-res PNG to the needed size and format. This should be clearly logged so the end user knows what has happened.

Pillow can be used for most of these operations, but Pillow doesn't currently support ICNS; however, macOS ships with a tool to generate ICNS files from PNGs.

## Creating an ICNS file:

Into a directory named `myicon.iconset`, put files with the following names and sizes:

* `icon_512x512@2x.png`
* `icon_512x512.png`
* `icon_256x256@2x.png`
* `icon_256x256.png`
* `icon_128x128@2x.png`
* `icon_128x128.png`
* `icon_32x32@2x.png`
* `icon_32x32.png`
* `icon_16x16@2x.png`
* `icon_16x16.png`

A 16x16@2x icon is 32x32 pixels; so yes - you need to have 2 copies of the same 32x32 icon (if you're paying close attention to the details you can specify a different icon; however, if you're autoscaling from other sizes, you're not paying that sort of attention).

Then, invoke: `iconutil -c icns myicon.iconset`. This will produce `myicon.icns`.

Andrei-Pozolotin:
`pillow` now knows about `iconutil`

https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#icns

https://github.com/python-pillow/Pillow/blob/master/src/PIL/IcnsImagePlugin.py#L306
```python
def _save(im, fp, filename):
    """
    Saves the image as a series of PNG files,
    that are then converted to a .icns file
    using the macOS command line utility 'iconutil'.
    macOS only.
    """
```

Andrei-Pozolotin:
and `linux` has `png2icns` /  `icns2png`

https://icns.sourceforge.io/

<details>

https://archlinux.org/packages/community/x86_64/libicns/

```
usr/
usr/bin/
usr/bin/icns2png
usr/bin/icontainer2icns
usr/bin/png2icns
usr/include/
usr/include/icns.h
usr/lib/
usr/lib/libicns.so
usr/lib/libicns.so.1
usr/lib/libicns.so.1.2.0
usr/lib/pkgconfig/
usr/lib/pkgconfig/libicns.pc
usr/share/
usr/share/man/
[...]

PR #716, opened 2022-04-22
Refs #386 - Backfill missing icons by resizing existing icons

ChicagoDev:
Created several methods / method signatures preparing to backfill icons. 
Open to any and all feedback. 
Thanks!

Backfilling icons

Refs #386 

## PR Checklist:

- [ X] I have read the **CONTRIBUTING.md** file
- [ X] I will abide by the code of conduct


src/briefcase/commands/create.py
```diff
@@ -575,6 +579,86 @@ def install_image(self, role, variant, size, source, target):
...
+    Feedback Requested:
+    
+    Should multiple image types as input be supported? jpgs or other?
+    
+    Do the function names, method signatures and return types seem on the right track?
+    
+    What variable or variables would give me information about the target platform?
+    
+    I'm open to feedback on anything
+
```
freakboy3742, line 600: General note - this sort of comment should be in the PR comment, not the code itself.

freakboy3742 (changes requested):
Thanks for this - it's a good start.

My immediate comment is about broad structure. You've put this into the `create` command, which makes sense in terms of where the code will be used, but doesn't in terms of coherence of purpose. 

The current code for CreateCommand.install_image() is right on the edge between "general utility vs create-specific" - I'd argue that what you should be aiming to do is to refactor CreateCommand.install_image() into a separate function that *isn't* part of the create command, and extends that command to fill in the "Unable to find variant" gaps.

To that end, I'd suggest this should actually be a in completely standalone module - `briefcase/images.py`. The API surface for install_image() doesn't have to change - but it will need to call into additional methods to determine if/how it can satisfy the missing variants. This will also simplify the testing process, as you don't need the overhead of instantiating a "create" command to verify you can downscale a 1024px jpg to a 256px png. 

As a top-down design, that then suggests first new API entry point is a `create_variant(target, size, sources)` method that you provide the available image data, and it does whatever it can to satisfy that request. v0 of this method would do nothing other than log the error message that is being logged today; v1 then starts adding up/downscale conversions using the available sources.

Regarding image formats - in an ideal world, you'd be able to use any input format, and convert to any output format. However, this is something we can iterate on; if v1 can only scale MacOS png from one size to another, that's fine; we can look at converting a Windows .ICO to macOS ICNS as a v2+ feature.

ChicagoDev:
I created images/images.py where I have the create_variant() function. One issue is that to import that file, it looks different than your imports. I would like feedback if I created images/images.py in the correct place. If not, where should they go?

For the API entry point, I pinpointed that location by deleting the .icns file in the app's resources folder. Then I determined line 577 would be the location where create_variant would need to be called. Is that the correct place, at least for the specific case of a missing .icns file on MacOS?

Open to any feedback. And I greatly appreciate your help. Thanks!

src/briefcase/commands/create.py
```diff
@@ -568,13 +573,22 @@ def install_image(self, role, variant, size, source, target):
                 # Copy the source image to the target location
                 self.shutil.copy(full_source, target)
             else:
-                print(
+
+                source_suffix = source_filename[-5:]
+                if source_suffix == '.icns' and full_role == 'application icon':
```
freakboy3742, line 578: I'd be inclined to put this logic (including the "can't find variant" logic) into `create_variant` - That way the API you've got is "create a .icns application icon variant of size X", which will fail if that variant can't be satisfied. What you've effectively specified here is "Create a variant, but only if it's is one of the specifically approved types". 

Why does this matter? Think longer term: v1 will add support for at least one variant creation type. In the version of the code you've got here, you'll need to make 2 change - add the variant creation code, and then modify the logic so that the request to make the variant will be made at all. If all the logic is in create variant, you're just expanding the capabilities of create_variant so it won't return an error/warning in as many situations.

src/briefcase/images/images.py
```diff
@@ -0,0 +1,5 @@
+
+def create_variant(target, size, sources):
```
freakboy3742, line 2: Thinking about how this will be used - this function has no context around what a "variant" is. Using a name like `create_image_variant()` would mean that no matter where it is imported, it will be clear what it is for. An "image-less" name would only make sense if you're anticipating having multiple entry points into the `images` module, so that the usage pattern would be `images.create_variant()`; that seems unlikely in this case.

freakboy3742 (changes requested):
More progress; a couple of comments inline. 

In terms of overall structure, creating a full images directory with images.py in it is probably overkill. Unless you're anticipating an `images/something_else.py` being needed at some time in the future, the extra folder is just complexity that doesn't add anything. Reducing this to a single `src/briefcase/images.py` is more than sufficient.

(As a side note: I know this is probably disheartening having this many comments about a 10 line patch - but this is also the end of the PR where we're setting up API ergonomics,. These are the little naming and structural things that we'll have to live with long term, so it's worth spending the time to get them right from the outset)

src/briefcase/commands/create.py
ChicagoDev: I have moved this into create_image_variant in my upcoming commit. Thanks!

ChicagoDev:
Not disheartening at all. I greatly appreciate all of your help!

ChicagoDev:
For commit b5a3fb0, some comments/questions:

Should file writing (mkdir and shutil.copy()) remain in create.py? And therefore, create_image_variane should return an image or icon set file?

Is size a tuple as expected?

Does using subprocess make sense for creating the icns file?

Any and all feedback appreciated. Thanks!

freakboy3742:
> Should file writing (mkdir and shutil.copy()) remain in create.py? And therefore, create_image_variane should return an image or icon set file?

I'd say the desired API contract for `create_image_variant()` is "will generate a file in a specified location". It's up to the invoker to nominate what that location is; `create_image_variant()` will create any directories or copy/create any files are needed to satisfy the API request.
 
> Is size a tuple as expected?

Yes - size will always be a `(width, height)` tuple (specified in pixels).

> Does using subprocess make sense for creating the icns file?

Preference would always be to use an API rather than invoking a subprocess - but if an API doesn't exist, a subprocess call to a utility that will reliably exist is an acceptable fallback. The `briefcase.integrations` module is a collection of tools being invoked in this way.

In this case, I don't think PIL allows creation of icns files (happy to be proven wrong on that front), so a call to subprocess seems the best approach.

ChicagoDev:
I committed mostly to backup my progress. I will continue to investigate the error mentioned in the commit. Unless something stands out to you? Always open to feedback. Thanks!

src/briefcase/images.py
```diff
@@ -0,0 +1,152 @@
...
+        except TypeError:
+            print("If formats is not None, a list or a tuple")
+
+
+    else:
+        print(f"No images in {source}")
+
+    return largest_image
+
+def create_icns(target: PosixPath, image: Image) -> None: #? : # I think it would be much more clear if we could return the .icns file here somehow, then we can only have the final file writing occur in the parent method.
```
freakboy3742, line 98: While a nice idea, it's likely impossible; you're calling out to iconutil, so you don't ever have the icns file in process memory.

freakboy3742:
Checking in on this - are you intending to continue progress on this PR, or can we close it?

freakboy3742:
Closing due to lack of activity.

PR #721, opened 2022-04-22
Remove old app code before installation

### LINKED ISSUE #720: Dist-info from previous versions is kept in the app path
**Describe the bug**
When running `briefcase update`, `CreateCommand.install_app_code()` will replace any old app code with the new version and write up-to-date dist-info to the app path. It will however not clear stale dist-info from previous versions.

A similar problem with dist-info for app packages was fixed in #645.

**To Reproduce**
Update the app code with `briefcase update` after incrementing the project version.

**Expected behavior**
Remove old dist-info.

**Environment:**
 - Software versions:
   - Briefcase: 0.3.6

**Additional context**

samschott:
Fixes #720.

## PR Checklist:

### DIFF
--- a/changes/720.bugfix.rst
+++ b/changes/720.bugfix.rst
@@ -1,1 +1,1 @@
-Clear old dist-info when updating app code.
+Application code and dist-info is now fully replaced during an update.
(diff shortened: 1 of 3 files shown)


freakboy3742 (approved):
👍  Looks good. I've made a couple of minor cosmetic tweaks, but this looks solid and the right approach in the circumstances.

PR #726, opened 2022-04-29
Convert all str.format() to f-strings

rmartin16:
As follow on from #704, this converts all `str.format()` calls to f-strings. Quite tedious....but very cathartic.

Given the codecov output, I tried to increase coverage where I thought it was reasonable.

### Note on Code Style
I noticed that my updates "collapsed" a lot of the `raise BriefcaseError` calls. If this was an important stylistic decision, let me know and I'll re-expand the errors.

For instance:
```python
raise BriefcaseCommandError(
    "This is an error message"
)
```
become
```python
raise BriefcaseCommandError("This is an error message")
```

Also, in some locations (notably `android_sdk.py`), I tried to somewhat normalize the multi-line strings within the file. Some of the changes introduce subtle changes to the strings with regards to line breaks....but they are all error messages that are ultimately passed back up to `__main__.py` and just printed to the screen prior to exit.

## PR Checklist:


src/briefcase/integrations/android_sdk.py
```diff
@@ -203,15 +193,11 @@ def install(self):
             # TODO: Py3.6 compatibility; os.fsdecode not required in Py3.7
             self.command.shutil.unpack_archive(os.fsdecode(sdk_zip_path), extract_dir=os.fsdecode(self.root_path))
         except (shutil.ReadError, EOFError):
-            raise BriefcaseCommandError(
-                """\
```
rmartin16, line 207: Implicit `\n` added. Reminder to review.

src/briefcase/integrations/android_sdk.py
```diff
@@ -263,37 +245,29 @@ def verify_license(self):
...
 
-    $ {sdkmanager} --licenses""".format(
-                    sdkmanager=self.root_path / "tools" / "bin" / "sdkmanager"
-                )
-            )
+    $ {self.root_path / 'tools' / 'bin' / 'sdkmanager'} --licenses""")
 
         if not license_path.exists():
-            raise BriefcaseCommandError(
-                """\
```
rmartin16, line 292: Implicit `\n` added. Reminder to review.

src/briefcase/integrations/android_sdk.py
```diff
@@ -263,37 +245,29 @@ def verify_license(self):
         if license_path.exists():
             return
 
-        self.command.logger.info(
-            "\n"
-            + """\
```
rmartin16, line 268: Implicit `\n` added. Reminder to review.

src/briefcase/integrations/android_sdk.py
```diff
@@ -263,37 +245,29 @@ def verify_license(self):
...
+    """)
         try:
             # Using subprocess.run() with no I/O redirection so the user sees
             # the full output and can send input.
             self.command.subprocess.run(
                 [os.fsdecode(self.sdkmanager_path), "--licenses"], env=self.env, check=True,
             )
         except subprocess.CalledProcessError:
-            raise BriefcaseCommandError(
-                """\
```
rmartin16, line 281: Implicit `\n` added. Reminder to review.

src/briefcase/integrations/android_sdk.py
```diff
@@ -936,22 +886,14 @@ def start_app(self, package, activity):
                     for line in output.split("\n")
                 )
             ):
-                raise BriefcaseCommandError(
-                    """\
```
rmartin16, line 940: Implicit `\n` added. Reminder to review.

rmartin16:
github hasn't pulled in rmartin16/briefcase@b44e256.....i'll check back later

src/briefcase/integrations/android_sdk.py
```diff
@@ -204,13 +198,11 @@ def install(self):
...
-Unable to unpack Android SDK ZIP file. The download may have been interrupted
-or corrupted.
+                f"""
+    Unable to unpack Android SDK ZIP file. The download may have been interrupted
+    or corrupted.
 
-Delete {sdk_zip_path} and run briefcase again.""".format(
-                    sdk_zip_path=sdk_zip_path
-                )
+    Delete {sdk_zip_path} and run briefcase again."""
```
rmartin16, line 205: FYI. Increasing indent to match practice in rest of the file.

src/briefcase/integrations/android_sdk.py
rmartin16: Converting this one to just `"""` is the same as what was before.

src/briefcase/integrations/java.py
```diff
@@ -142,63 +138,63 @@ def verify(cls, command, install=True):
...
+    {java_home}/bin/javac -version
 
-    from the command prompt.
+from the command prompt.
 
-    *************************************************************************
+*************************************************************************
 
-    """.format(java_home=java_home)
+"""
```
rmartin16, line 197: Match indent of other errors.

src/briefcase/integrations/wix.py
```diff
@@ -73,9 +73,11 @@ def verify(cls, command, install=True):
             not installed, raises MissingToolError.
         """
         if command.host_os != 'Windows':
-            raise BriefcaseCommandError("""
+            raise BriefcaseCommandError(
+                """
```
freakboy3742, line 77: I think some of the errors in this file may need the `\` trick to avoid extra newlines (plus also auditing the newlines at the end of each message)

src/briefcase/integrations/java.py
```diff
@@ -261,23 +257,18 @@ def install(self):
                 extract_dir=os.fsdecode(self.command.tools_path)
             )
         except (shutil.ReadError, EOFError):
-            raise BriefcaseCommandError(
-                """
+            raise BriefcaseCommandError(f"""
```
freakboy3742, line 260: I think some of the errors in this file may need the \ trick to avoid extra newlines (plus also auditing the newlines at the end of each message)

freakboy3742 (changes requested):
Whelp... that's a lot of patch :-)

As you've noted, this is almost entirely mechanical and uncontroversial. I've flagged a handful of places where this isn't true - they pretty much all stem from an attempt to introduce consistency of one form or another. I'm completely on board with the goal of consistency; we just need to make sure that we're converging on a consistent consistency :-)

On the specific query you raised in the PR - I have no particular preference for 
```
BriefcaseCommandError(
    "something"
)
```
over
```
BriefcaseCommandError("something")
```
as a code style; the former was usually only in place for readability because of long strings/format statements. However, there *is* a difference between:
```
BriefcaseCommandError(
    """something"""
)
```
and
```
BriefcaseCommandError("""
    something
""")
```
and
```
BriefcaseCommandError("""
something
""")
```
that we need to be careful about. As a result of your console PR, we now ensure that there is a clearing newline before the top-level command surfaces the BriefcaseCommandError message payload is output; so we need to be careful that the message in a BriefcaseCommand error *doesn't* start with a newline. The "stray \" that you've noted in reminder comments exists specifically for this purpose (although - in your defence - it hasn't been consistently used in the existing code, so it's entirely understandable that it may not have been obvious that this is what was going on). There's a similar issue at the *end* of those long errors; we need to make sure that we don't end up with lots of stray newlines at the end of output.

The good news is that AFAICT, the only places where these problems exist are in 
commands/base, integrations/android_sdk, integrations/java, integrations/wix, and integrations/Xcode - the common thread here is that they're integrations that have complex failure modes, so the error messages are long.

rmartin16:
Yeah...I was probably a little too flippant with introducing these new line-breaks. Although, having noticed quite a bit of subtle variation in the consistency of multi-line error strings throughout the package, I have a proposal.

Instead of trying to get the newlines and indents correct everywhere an error string is being created (although I certainly think best efforts should be made), what if we just format the error back at the top level.

Something like this:
```python
print(textwrap.dedent(str(e).strip()), file=sys.stdout if e.error_code == 0 else sys.stderr)
```

\<rant\> I am kinda feeling like Python is lacking a "pretty" way to create multi-line strings that 1) aligns with the code block, 2) doesn't require manual line-break insertion, and 3) doesn't require subtle line continuations to omit line-breaks. I usually go with the route in `docker.py` where these strings live at the top of the file.....but I really think this code benefits from having these strings in-line...if only because they offer great contextual info...and just the sheer amount of them.\<\/rant\>

If you're not too keen on this, though, (especially since this approach does nothing for non-error strings), I think I'm more inclined right now to just do the f-string conversion without trying to normalize any of these strings. Just leave all the line-breaks as they are right now and tackle that as one-offs or a different PR.

freakboy3742:
You won't get any argument from me that there's variation over the codebase; I've tried to avoid it, but it's a big codebase. I'm very open to suggestions for how to use APIs to make standardisation happen automatically.

Introducing `text.dedent()` (plus the stderr redirection) sounds like a great step in this direction - and something that your Log refactor makes very plausible. I've got a minor question/concern about the exact API you have in mind (whether e.error_code will be available in that context, or whether we should just make `log.error` write to stderr), but at a high level, I'm on board with the idea.

There's probably also room for abstracting the "banner" error messages (the big ***WARNING*** ones) so that they have consistent decoration as well.

rmartin16:
> You won't get any argument from me that there's variation over the codebase; I've tried to avoid it, but it's a big codebase.

I certainly understand; and apologies if I sounded judgemental....I intended only to make an observation.

I will update this to avoid meaningfully changing any of these strings and potentially pursue a different PR to introduce any level of standardization.

As for my example using `textwrap.dedent`, `e.error`, and writing to `stderr`.....I only intended that to represent a potential change to the way `BriefcaseError` exceptions are printed prior to exiting in a failure mode. However, a more broad application of this strategy within `Log` definitely makes sense.

To that end, I wanted to verify your commitment to this statement:
> Console output should be left-aligned

I've noticed several cases where a lone sentence will be indented; for instance, consider a lot of the printing in `AndroidSDK` concerning the SDK's licenses.

**Are you OK with enforcing left alignment for all output?** (this would respect indentation within something like the WARNING block as you call out)

This approach would make it possible to create much "prettier" multi-line strings in my opinion. For instance, 

```python
        ...
        command.logger.info("""
            This is line one of a message
            
            this is line three of a message
            """
```
Calling `dedent` on this output (via `Log`) would at least allow for code block alignment while avoiding explicit line-breaks via `\n`....however, leading and trailing line-breaks are still inferred here.

Anyway, this is a lot of hay ultimately just about some strings.... :)

src/briefcase/commands/base.py
```diff
@@ -35,22 +35,25 @@ class TemplateUnsupportedVersion(BriefcaseCommandError):
     def __init__(self, python_version_tag):
         self.python_version_tag = python_version_tag
         super().__init__(
-            f"Could not find template for Python {self.python_version_tag}.\n\n"
-            f"This is likely because Python {self.python_version_tag} "
-            "is not yet supported. You will need to:\n"
-            "  * Use an older version of Python; or\n"
-            "  * Define your own custom template.\n"
+            f"""
```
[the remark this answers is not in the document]
rmartin16: Removed the leading new line that was introduced with line continuation.

src/briefcase/integrations/android_sdk.py
rmartin16: Reverted this indent. Indent standardization will be considered later.

src/briefcase/integrations/android_sdk.py
rmartin16: Restored the line continuation.

src/briefcase/integrations/wix.py
rmartin16: This file is compact so I added line continuations to the `BriefcaseCommandError` exceptions to avoid the extra initial new line.

freakboy3742 (approved):
Looks good! I've made a couple of small consistency/code formatting tweaks addressing the issues like the indentation in android_sdk that you flagged; tl;dr is that yes, left aligned is definitely fine for all output, with the exception of the stuff that is in the WARNING "dialogs".

I'm happy enough with where this is to merge it; I'm happy to consider any subsequent refactoring to introduce dedent/err redirection as a follow up PR.

rmartin16:
> tl;dr is that yes, left aligned is definitely fine for all output, with the exception of the stuff that is in the WARNING "dialogs".

@freakboy3742, in way of some follow up, I pursued some initial work for this...but hit an interesting roadblock.

My approach was largely dependent on allowing "trailing" whitespace in multiline strings....such that these strings would be aligned with the code block....like in the example in my previous post. In that case, the lines without meaningful content need to have whitespace to match the indentation of the other lines. However, `flake8` doesn't like this at all (PyCQA/pycodestyle#45). Given their recommendation is to either drop all trailing whitespace errors or use `# noqa` for _all_ instances.....I'm not so sure this is a viable path....

PR #727, opened 2022-04-30
Ensure that Java detection errors contain JAVA_HOME

rmartin16:
Prior to building the Java detection error strings, the value for `JAVA_HOME` was being cleared; therefore, the error displayed to users was missing the value of `JAVA_HOME` that likely was causing the problem.

These changes ensure that the errors contain the value of `JAVA_HOME`.

## PR Checklist:


freakboy3742 (approved):
Good catch - thanks for the PR!

Issue #629, opened 2021-10-19
Add a stub app for Windows


freakboy3742:
We recently modified the macOS template to provide a stub app - a small binary executable that is used as an entry point for the app, which is responsible for starting the Python interpreter.

This has a number of benefits:

 * The app appears in the process monitor with the app name and icon, rather than "python" or a shell file
 * The app that is started is the same one that is granted permissions by the operating system, rather than needing to fork a process
 * It provides a hook to catch high-level app failures, such as a fail on start due to import errors. If the app fails, we can catch the failure and show a crash dialog.
 * It provides a hook for installing logging and other system-level handlers.

**Describe the solution you'd like**

The [briefcase-windows-msi-template](https://github.com/beeware/briefcase-windows-msi-template) contains a stub app that behaves in a similar way to the macOS app template.

This will involve writing a small C app using Winforms that starts the Python interpreter. The [implementation in the macOS template](https://github.com/beeware/briefcase-macOS-Xcode-template/blob/dev/%7B%7B%20cookiecutter.formal_name%20%7D%7D/%7B%7B%20cookiecutter.formal_name%20%7D%7D/main.m) will be a useful reference; although that code is Objective C, and uses the Cocoa APIs, the structure of code needed to start the Python interpreter will be largely the same.

The macOS template has 2 versions - a pre-built version where the binary is copied and renamed as needed; and an "Xcode project" version where the user can modify the app if they wish. This approach should be duplicated for Windows; the "MSI" target should use a pre-compiled binary, and a new "VisualStudio" target should provide a project that can be modified and compiled.

PR #731, opened 2022-05-04
Implement Popen Output Streamer

rmartin16:
As detailed in #676, the attempt to display the iOS simulator log can miss the first few seconds of log output. This is because the capturing of the log output is started after the simulator is started; therefore, there will always be a non-zero amount of time that log output is not being captured.

Log streaming for macOS and iOS simulator now start before the app is started; once the app is running, any queued log output is flushed to the screen and new log output is printed in real time.

## Implementation
 - A new function in `subprocess.py` loops on the `stdout` output of an established `Popen` process
   - The loop will properly exit when it detects the underlying process has exited
   - The loop will properly exit and terminate the underlying process if the user sends CTRL+C
 - Starting up the iOS simulator was updated to initiate `log stream` via `Popen`
 - Once the simulator is started, the new output streaming function is called to print all of the simulator log

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -245,18 +245,19 @@ class Subprocess:
         """
         try:
             while True:
-                # readline should always return at least a newline...if it returns an
-                # empty string, that should mean the underlying process is exiting/gone.
+                # readline should always return at least a newline (ie \n)
                 output_line = ensure_str(popen_process.stdout.readline())
-                # a return code will be available once the process returns one to the OS.
-                # by definition, that should mean the process has exited.
-                return_code = popen_process.poll()
-                # only return once all output is flushed and the process has exited.
-                if output_line == "" and return_code is not None:
-                    self._log_return_code(return_code)
-                    return
-                if output_line != "":
+                if output_line:
                     self.command.logger.info(output_line)
+                # UNLESS the underlying process is exiting/gone; then "" is returned
+                elif output_line == "":
+                    # a return code will be available once the process returns one to the OS.
+                    # by definition, that should mean the process has exited.
+                    return_code = popen_process.poll()
+                    # only return once all output has been read and the process has exited.
+                    if return_code is not None:
+                        self._log_return_code(return_code)
+                        return
 
         except KeyboardInterrupt:
             popen_process.terminate()
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -232,7 +232,7 @@ class Subprocess:
             **self.final_kwargs(**kwargs)
         )
 
-    def stream_output(self, popen_process):
+    def stream_output(self, label, popen_process):
         """
         Stream the output of a Popen process until the process exits.
         If the user sends CTRL+C, the process will be terminated.
@@ -241,15 +241,17 @@ class Subprocess:
         log file, then initiating a non-blocking process that populates that
         log, and finally streaming the original process's output here.
 
+        :param label: A description of the content being streamed; used for
+            to provide context in logging messages.
         :param popen_process: a running Popen process with output to print
         """
         try:
             while True:
                 # readline should always return at least a newline (ie \n)
+                # UNLESS the underlying process is exiting/gone; then "" is returned
                 output_line = ensure_str(popen_process.stdout.readline())
                 if output_line:
                     self.command.logger.info(output_line)
-                # UNLESS the underlying process is exiting/gone; then "" is returned
                 elif output_line == "":
                     # a return code will be available once the process returns one to the OS.
                     # by definition, that should mean the process has exited.
@@ -260,14 +262,22 @@ class Subprocess:
                         return
 
         except KeyboardInterrupt:
-            popen_process.terminate()
-            try:
-                popen_process.wait(timeout=3)
-            except subprocess.TimeoutExpired:
-                self.command.logger.warning(
-                    "Process will be forcibly killed since it did not exit gracefully when signaled."
-                )
-                popen_process.kill()
+            self.cleanup(label, popen_process)
+
+    def cleanup(self, label, popen_process):
+        """
+        Clean up after a Popen process, gracefully terminating if possible; forcibly if not.
+
+        :param label: A description of the content being streamed; used for
+            to provide context in logging messages.
+        :param popen_process: The Popen instance to clean up.
+        """
+        popen_process.terminate()
+        try:
+            popen_process.wait(timeout=3)
+        except subprocess.TimeoutExpired:
+            self.command.logger.warning(f"Forcibly killing {label}...")
+            popen_process.kill()
 
     def _log_command(self, args):
         """
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -9,10 +9,6 @@ class ParseError(Exception):
     """Raised by parser functions to signal parsing was unsuccessful"""
 
 
-class PopenStreamingError(Exception):
-    """Raised for errors from streaming Popen process output"""
-
-
 def ensure_str(text):
     """Returns input text as a string."""
     return text.decode() if isinstance(text, bytes) else str(text)
@@ -250,12 +246,15 @@ class Subprocess:
         try:
             while True:
                 # readline should always return at least a newline...if it returns an
-                # empty string, that should mean the underlying process is exiting/gone
+                # empty string, that should mean the underlying process is exiting/gone.
                 output_line = ensure_str(popen_process.stdout.readline())
+                # a return code will be available once the process returns one to the OS.
+                # by definition, that should mean the process has exited.
                 return_code = popen_process.poll()
+                # only return once all output is flushed and the process has exited.
                 if output_line == "" and return_code is not None:
                     self._log_return_code(return_code)
-                    break
+                    return
                 if output_line != "":
                     self.command.logger.info(output_line)
 
@@ -264,10 +263,10 @@ class Subprocess:
             try:
                 popen_process.wait(timeout=3)
             except subprocess.TimeoutExpired:
+                self.command.logger.warning(
+                    "Process will be forcibly killed since it did not exit gracefully when signaled."
+                )
                 popen_process.kill()
-
-        except Exception as e:
-            raise PopenStreamingError(f"{e!r}")
 
     def _log_command(self, args):
         """
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -9,6 +9,10 @@ class ParseError(Exception):
     """Raised by parser functions to signal parsing was unsuccessful"""
 
 
+class PopenStreamingError(Exception):
+    """Raised for errors from streaming Popen process output"""
+
+
 def ensure_str(text):
     """Returns input text as a string."""
     return text.decode() if isinstance(text, bytes) else str(text)
@@ -231,6 +235,39 @@ class Subprocess:
             ],
             **self.final_kwargs(**kwargs)
         )
+
+    def stream_output(self, popen_process):
+        """
+        Stream the output of a Popen process until the process exits.
+        If the user sends CTRL+C, the process will be terminated.
+
+        This is useful for starting a process via Popen such as tailing a
+        log file, then initiating a non-blocking process that populates that
+        log, and finally streaming the original process's output here.
+
+        :param popen_process: a running Popen process with output to print
+        """
+        try:
+            while True:
+                # readline should always return at least a newline...if it returns an
+                # empty string, that should mean the underlying process is exiting/gone
+                output_line = ensure_str(popen_process.stdout.readline())
+                return_code = popen_process.poll()
+                if output_line == "" and return_code is not None:
+                    self._log_return_code(return_code)
+                    break
+                if output_line != "":
+                    self.command.logger.info(output_line)
+
+        except KeyboardInterrupt:
+            popen_process.terminate()
+            try:
+                popen_process.wait(timeout=3)
+            except subprocess.TimeoutExpired:
+                popen_process.kill()
+
+        except Exception as e:
+            raise PopenStreamingError(f"{e!r}")
 
     def _log_command(self, args):
         """
--- a/src/briefcase/platforms/iOS/xcode.py
+++ b/src/briefcase/platforms/iOS/xcode.py
@@ -13,7 +13,6 @@ from briefcase.commands import (
 from briefcase.config import BaseConfig
 from briefcase.console import InputDisabled, select_option
 from briefcase.exceptions import BriefcaseCommandError, InvalidDeviceError
-from briefcase.integrations.subprocess import PopenStreamingError
 from briefcase.integrations.xcode import (
     DeviceState,
     get_device_state,
@@ -366,13 +365,10 @@ class iOSXcodeRunCommand(iOSXcodeMixin, RunCommand):
             raise BriefcaseCommandError(f"Unable to launch app {app.app_name}.")
 
         # Start streaming logs for the app.
-        try:
-            self.logger.info()
-            self.logger.info(f"[{app.app_name}] Following simulator log output (type CTRL-C to stop log)...")
-            self.logger.info("=" * 75)
-            self.subprocess.stream_output(simulator_log_popen)
-        except PopenStreamingError as e:
-            raise BriefcaseCommandError(f"Encountered error during log stream for app {app.app_name}: {e}")
+        self.logger.info()
+        self.logger.info(f"[{app.app_name}] Following simulator log output (type CTRL-C to stop log)...")
+        self.logger.info("=" * 75)
+        self.subprocess.stream_output(simulator_log_popen)
 
         # Preserve the device selection as state.
         return {
--- a/src/briefcase/platforms/iOS/xcode.py
+++ b/src/briefcase/platforms/iOS/xcode.py
@@ -352,7 +352,7 @@ class iOSXcodeRunCommand(iOSXcodeMixin, RunCommand):
             stderr=subprocess.STDOUT,
             bufsize=1,
         )
-        time.sleep(3)  # let log stream start up
+        self.sleep(0.25)  # let log stream start up
 
         self.logger.info()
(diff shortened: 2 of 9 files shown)


rmartin16:
@freakboy3742, before I move too much further forward with this proposal, can you please provide any thoughts you have?

I know you mentioned threading in #676 but I felt that `Popen` would give us better control over this (especially since we're already doing so much with `subprocess`). This approach makes it more straightforward to start showing output exactly when we want; and in the most general use case for this, I thought this was superior over kicking off a command in to another thread where it would definitely be trickier to control when the thread started printing....especially if we wanted to print more things in the main thread first.

src/briefcase/platforms/iOS/xcode.py
```diff
@@ -399,6 +402,22 @@ def run_app(self, app: BaseConfig, udid=None, **kwargs):
                 )
             )
 
+        # start streaming the simulator's log now to ensure all app output is captured
+        simulator_log_popen = self.subprocess.Popen(
+            [
+                "xcrun", "simctl", "spawn", udid,
+                "log", "stream",
+                "--style", "compact",
+                "--predicate", 'senderImagePath ENDSWITH "/{app.formal_name}"'.format(app=app)
```
freakboy3742, line 411: This can now be an f-string

src/briefcase/platforms/iOS/xcode.py
```diff
@@ -421,19 +440,9 @@ def run_app(self, app: BaseConfig, udid=None, **kwargs):
...
-                ],
-                check=True,
-            )
-        except subprocess.CalledProcessError:
-            raise BriefcaseCommandError(
-                "Unable to start log stream for app {app.app_name}.".format(app=app)
-            )
+            stream_popen_output(simulator_log_popen, self.logger.info)
+        except OutputStreamingError as e:
+            raise BriefcaseCommandError(f"Unable to start log stream for app {app.app_name}: {e}")
```
freakboy3742, line 445: To my reading, there's a couple of conditions other than "unable to start" that could trigger this case; this means we either need a more errors, or a more generic error.

freakboy3742 (changes requested):
Oooh... I like this a *lot*. Popen streaming is a *much* better approach than threading - in hindsight, it's so obvious that I'm kicking myself for not thinking of it.

Other than the things I've flagged inline the only thing that stands out is that `stream_popen_output` will likely need to be another augmentation on the subprocess module so that it can be easily intercepted for testing purposes. (i.e., `subprocess.stream_output(process)`).

src/briefcase/platforms/iOS/xcode.py
rmartin16: Converted to an f-string.

src/briefcase/platforms/iOS/xcode.py
```diff
@@ -338,6 +341,21 @@ def run_app(self, app: BaseConfig, udid=None, **kwargs):
...
+                "xcrun", "simctl", "spawn", udid,
+                "log", "stream",
+                "--style", "compact",
+                "--predicate", f'senderImagePath ENDSWITH "/{app.formal_name}"'
+            ],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            bufsize=1,
+        )
+        time.sleep(3)  # let log stream start up
```
rmartin16, line 356: I'm not sure how long to wait here....ideally, i suppose, it'd be nice to know when `log stream` is ready....but I don't think that's an option...

src/briefcase/platforms/iOS/xcode.py
rmartin16: That's true....although, I changed this up and expounded in the next post.

rmartin16:
I moved the stream function inside of `Subprocess` still taking a `Popen` process. However, I removed any sort of detection of "failure" from it. The more I thought about it, the more it felt out of place to try to interpret if the process was in a state of failure just because it had seemingly exited prior to the stream function being invoked. Instead, the stream function will look for output, print it, and exit as normal.

That said, if we wanted to detect failure, it seems more appropriate to do so back up where streaming is being invoked. However, even here, I was having trouble determining when `log stream` may have failed....or if it ran normally but looks like it failed. For instance, if I close the iOS simulator while `log stream` is still running, the `log stream` process is killed automatically and returns a non-zero code. However, this isn't really a failure mode but looks very similar to what might be considered one. (Replicating this on `main` results in a "unable to start log stream" error.) Finally, I'm not even sure if it's worth tracking if `log stream` itself failed....it seems other problems would likely show up elsewhere before `log stream` failed. That said, we can add checks on its return codes perhaps if you think it's worth it.

I implemented this log streaming for the iOS simulator and macOS apps. AFAICT, adb's `logcat` will print everything in its ring buffer log that matches the filter....so starting it early doesn't seem to offer a benefit. Linux and Windows apps seem to depend on normal printing `stdout` for logging. That said, let me know if I should add this anywhere else you know of.

src/briefcase/integrations/subprocess.py
```diff
@@ -206,6 +210,39 @@ def Popen(self, args, **kwargs):
...
+                    self.command.logger.info(output_line)
+
+        except KeyboardInterrupt:
+            popen_process.terminate()
+            try:
+                popen_process.wait(timeout=3)
+            except subprocess.TimeoutExpired:
+                popen_process.kill()
+
+        except Exception as e:
```
rmartin16, line 243: On one hand, this sort of catching feels somewhat overdone seems Im really not sure how `readline` or `poll` can error....on the other hand, re-wrapping whatever might come out in a more contextual error seems appropriate.

src/briefcase/integrations/subprocess.py
```diff
@@ -206,6 +210,39 @@ def Popen(self, args, **kwargs):
...
+            raise PopenStreamingError(f"{e!r}")
```
rmartin16, line 244: On one hand, this sort of catching feels somewhat overdone seems Im really not sure how readline or poll can error....on the other hand, re-wrapping whatever might come out in a more contextual error seems appropriate.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -32,6 +34,35 @@ def run_app(self, app: BaseConfig, **kwargs):
...
+                "--predicate",
+                f'senderImagePath=="{sender}"'
+                f' OR (processImagePath=="{sender}"'
+                ' AND senderImagePath=="/usr/lib/libffi.dylib")',
+            ],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            bufsize=1,
+        )
+        time.sleep(3)  # let log stream start up
```
rmartin16, line 64: same here

freakboy3742 (changes requested):
This is looking good. I've flagged a few things; mostly around ensuring error handling and flow paths are obvious. Some of them might be stemming from my own misunderstanding; if that's the case, it's probably an indication that more explanatory comments are needed.

I agree with your analysis that trying to detect failure is more trouble than it's worth; as indicated inline, we should verify that the process *starts*, but once it has started, I don't know that there's much we can meaningfully detect. 

I think you're right that this is only needed on iOS and macOS at present; Android adb logging certainly seems to be able to functionally capture "all" the logs. Linux logging is a direct stdout/stderr stream, so there's nothing extra needed; and until we fix #629, there isn't any log to stream on Windows.

src/briefcase/integrations/subprocess.py
```diff
@@ -206,6 +210,39 @@ def Popen(self, args, **kwargs):
...
+        """
+        try:
+            while True:
+                # readline should always return at least a newline...if it returns an
+                # empty string, that should mean the underlying process is exiting/gone
+                output_line = ensure_str(popen_process.stdout.readline())
+                return_code = popen_process.poll()
+                if output_line == "" and return_code is not None:
+                    self._log_return_code(return_code)
+                    break
```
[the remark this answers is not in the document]
rmartin16: breaking an `if` huh....ngl, though, `break`s can definitely be problematic...especially in nested code. thanks for sharing.

Updated to `return`.

src/briefcase/integrations/subprocess.py
```diff
@@ -206,6 +210,39 @@ def Popen(self, args, **kwargs):
...
+                if output_line != "":
+                    self.command.logger.info(output_line)
+
+        except KeyboardInterrupt:
+            popen_process.terminate()
+            try:
+                popen_process.wait(timeout=3)
+            except subprocess.TimeoutExpired:
+                popen_process.kill()
```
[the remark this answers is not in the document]
rmartin16: Good call. Added a warning log.

src/briefcase/integrations/subprocess.py
rmartin16: That's a good perspective. Removed the catch.

tests/integrations/subprocess/test_Subprocess__stream_output.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def test_output(mock_sub, popen_process, capsys):
+    "Readline output is printed"
+    mock_sub.stream_output(popen_process)
+    assert capsys.readouterr().out == "output line 1\n\noutput line 3\n"
+
+
+def test_output_debug(mock_sub, popen_process, capsys):
+    "Readline output is printed"
```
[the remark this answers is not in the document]
rmartin16: There isn't a difference. I added this test only to confirm that "debug" mode is the same as "non-debug" mode. Should only see a difference in "deep debug". Clarified that in the description....although, I suppose these tests could also be combined.

tests/integrations/subprocess/test_Subprocess__stream_output.py
```diff
@@ -0,0 +1,68 @@
...
+    popen_process.stdout.readline.side_effect = Exception("error reason")
+
+    with pytest.raises(PopenStreamingError, match=r"Exception\(\'error reason\'\)"):
+        mock_sub.stream_output(popen_process)
+    assert capsys.readouterr().out == ""
+
+
+def test_keyboard_interrupt(mock_sub, popen_process):
+    "Process is terminated if user sends CTRL+C"
+    popen_process.stdout.readline.side_effect = KeyboardInterrupt()
```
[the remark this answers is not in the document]
rmartin16: Good point. Updated.

src/briefcase/integrations/subprocess.py
```diff
@@ -206,6 +210,39 @@ def Popen(self, args, **kwargs):
...
+
+        :param popen_process: a running Popen process with output to print
+        """
+        try:
+            while True:
+                # readline should always return at least a newline...if it returns an
+                # empty string, that should mean the underlying process is exiting/gone
+                output_line = ensure_str(popen_process.stdout.readline())
+                return_code = popen_process.poll()
+                if output_line == "" and return_code is not None:
```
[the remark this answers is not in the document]
rmartin16: I was a little leery about this loop as well at first. However, I believe it hinges on assumptions that are already baked in to our other uses of `subprocess`.

- `check_output` is wrapper on `run` and `run` is a wrapper on `Popen`.
- `Popen` does some basic validation, sets up `stdin/out/err`, and then forks a process to run the OS command.
  - Strictly speaking, `Popen` can fail but it appears to be either invalid input combinations or something went really wrong at the OS level.
- From here, `run` blocks on `read\write` calls to `stdin/out/err` via `Popen.communicate()`.
  - These `read` call implementations are buried in layers of abstraction, but all the documentation of python `io` suggests that `read` returns when it hits an `EOF`.
  - For `readline`, the docs say that once an `EOF` is hit, `readline` will only ever return `""` or `b""` for subsequent calls.
  - For `stdout/err`, I am interpreting `EOF` to mean that the child process closed the pipes on its end to them.
- Once `Popen.communicate()` returns, `run` blocks on `wait()`.
  - True to its name, it waits until the child process finishes and provides a return code.
- If return code is non-zero, raise `CalledProcessError`
- Alternatively, return a `CompletedProcess` object with return code and `stdout/err`

So....in this way, this loop is a really dumbed down version of `run`, in that:
- At some point, `readline()` will need to start returning `""` because the process closed the pipe
- And the child process will need to return an code via `poll()`

If these assumptions aren't met from a child process, then we'll need the user to send CTRL+C.....however, that would be true even using `run` (at least without a `timeout` since we don't provide one).


As for the specific structure of the loop:
 - The quit condition is based on both `readline()` returning empty and `poll()` providing a return code.
 - This is mostly to prevent a race condition with flushing `readline` output to the screen and `poll()` getting a return code from the process.
 - Requiring both means we've flushed all the output and we have a return code.
 - That said, I think there are other configurations for the loop, but I don't think they fundamentally change the need for `readline` returns `""` and `poll()` is not `None`.
 - Finally, this `Popen` loop idiom was surprisingly popular when I was researching the best way to do this....although, it's always hard to tell if that's because everyone just copy/pastes the seemingly most popular thing...or if its actually the best way to do this.

src/briefcase/platforms/iOS/xcode.py
rmartin16: So, on my exceptionally slow virtualized instance of macOS, I did encounter an instance of missing the beginning of the log when this wait was 1 second.

While starting `log stream` before running the app at least makes it possible to display the entire log, I believe there is still a race condition. When `Popen` is invoked, it simply forks a process with instructions to run a command; so, I'm not sure there would necessarily be any guarantee of the state of that command once `Popen` returns control.

So, if `log stream` (for some bizarre reason) takes 10 seconds to start actually capturing the log entries, it's quite likely that the app will already be up and running. That's why, for instance, it would be nice if we could enter another loop sending `SIGUSR1` signals or something to `log stream` until it responded affirmatively that it's ready to actually stream logs. Then we'd know 100% it is safe to start the app and show all the log entries.

side note: i found this in `Popen` ;) thought it somewhat apt
```python
        # How long to resume waiting on a child after the first ^C.
        # There is no right value for this.  The purpose is to be polite
        # yet remain good for interactive users trying to exit a tool.
        self._sigint_wait_secs = 0.25  # 1/xkcd221.getRandomNumber()
```

src/briefcase/platforms/macOS/__init__.py
rmartin16: Left a comment about timeout above. I'll comment on `Popen` exceptions here.

I think that catching `Popen` exceptions is probably unnecessary. As I detailed the behavior of `run` in another comment, `run` is a wrapper on `Popen`. In that analysis, it became clear to me that if `Popen` itself raised an exception (sans `TimeoutError`), `run` would let that exception bubble up in to the caller's code. In all the calling code here, we only ever catch `CalledProcessError` for `run` which will only occur for non-zero return codes (and if `check=True`).

So, given this and the fact `Popen` exceptions are discovered at dev time (via invalid arguments) or reveal a deep OS level runtime issue, I think it makes sense not to try to catch any exceptions from `Popen`.

rmartin16:
Responded to comments.

Outstanding items:
 - Structure of `Popen` output streaming loop
 - Waiting after starting `log stream`
 - `Popen` exception catching

src/briefcase/platforms/iOS/xcode.py
freakboy3742: Ah - that makes sense. I guess we need to make the call between slow enough to guarantee that *really* slow machines will have started the log, and being slow enough to get in the way of app development. 

My gut tells me that 3 seconds is leaning a little to hard on the "accomodate slow machines" side; made worse by the fact that it's a 3 second wait before any other console output happens (so it's *right* on the cusp of a "did that work?" wait). If you're running macOS virtualised, you're definitely going to be seeing the worst case scenario, and most people doing macOS development won't be doing so in a virtualised container. I wonder if the 0.25 value you've found actually might be good enough in this case...

src/briefcase/platforms/macOS/__init__.py
freakboy3742: I agree it's unlikely to fail; but if it does, we have an opportunity to fail gracefully - in a way that lets the user know it's probably not their fault, and in a way that maximises the likelihood that their bug report will point at the actual cause of the problem. 

Consider - in this case, in the "caught" scenario, the bug report is "Briefcase is raising an error saying it can't start the log stream". In the "uncaught" scenario, the bug report is a `subprocess.CalledProcessError`, and if we're *really* lucky, they'll provide the full stack trace from which we might be able to work back and determine it's the log stream start that has failed. 

Catching the error and reporting the high-level problem is defensive programming on our part.

src/briefcase/integrations/subprocess.py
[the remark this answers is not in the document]
freakboy3742: Ok - this all makes sense, and the revised loop structure looks good to me. `poll()` is definitely less of a risk than `wait()` under the circumstances; and AFAICT the usage of `poll()` you've got here should be safe.

freakboy3742 (changes requested):
Getting very close; comments inline - at least one of which is a "🤷‍♂️... how much will be Enough?".

src/briefcase/platforms/iOS/xcode.py
rmartin16: I found that `0.25` seconds did work sometimes on my virtualized macOS...maybe 50/50. As you said though, this should definitely be the extreme example of delay on macOS....I can only even allocate a single CPU to the VM without a kernel panic :| so I'm fine with `0.25`....can always increase later.

rmartin16:
I reduced the wait time for `log stream` to `0.25` seconds.

I also responded inline about the `Popen` exceptions....just want to double check.

src/briefcase/platforms/macOS/__init__.py
freakboy3742: Huh... for some reason I thought Popen generated other errors (akin to `run()`) , but checking the docs, you're right - it's a fairly limited set of errors that would firmly fall in the "what the hell have you done to your computer?!?" edge case. On that basis, what you've got here is probably sufficient; if the user has borked their system enough that they're getting permission errors out of `log`, surfacing the raw error is probably OK.

freakboy3742 (approved):
Looks great! I picked up one last detail while doing a close check of the tests - in the event of the app failing to start, the log process would be left dangling. I'm fairly certain this would be cleaned up when the parent process dies; but it doesn't hurt to do the cleanup, and it was easy to factor out - plus it gave an opportunity to provide slightly more contextual cleanup messages.

Thanks for another great contribution!

PR #733, opened 2022-05-08
Support for Popen start_new_session Flag

rmartin16:
## Problem
Currently, all processes invoked by Briefcase are added to the process group of the python interpreter running Briefcase. This can be beneficial since it allows the OS to handle these processes as a single entity.

However, in certain cases, it is necessary for a spawned process to run independently of its parent process. For example, after the Android emulator is started and adb's `logcat` is running, the user is instructed to use CTRL+C to return to the command prompt. Sending `SIGINT` (via CTRL+C) will return the user to the prompt but also kills the process for the emulator (and any other spawned processes for that matter). This happens because the `SIGINT` is propagated to all processes in the group.

## Solution
Add `start_new_session=True` to the relevant `subprocess` calls that need to be treated independently of Briefcase itself.

This is now implemented for invoking the Android emulator to prevent CTRL+C for adb's `logcat` also killing the process for the emulator.

## Implementation
The `Popen` parameter `start_new_session` implements the POSIX syscall `setsid` which ensures a spawned process runs in a new session (and therefore, by definition, a new process group). (The `start_new_session` parameter was added to replace the idiom of passing `preexec_fn=os.setsid` to `Popen` since it has problematic edge cases....and it's being [deprecated](https://bugs.python.org/issue38435).) Since a `console` or controlling terminal is not shared across sessions, the new process will be unaffected by even terminating the originating terminal window.

Windows, though, is a different story. The approach of `subprocess` in the `stdlib` appears to be to expose several underlying process spawn APIs (largely dependent on the underlying platform) instead of an approach that might produce a single API where the underlying implementations aren't so exposed. At any rate, `start_new_session` is not respected on Windows.

To convert `start_new_session=True` for Windows, the two [process creation flags](https://docs.microsoft.com/en-us/windows/win32/procthread/process-creation-flags) below are used:
- `CREATE_NEW_PROCESS_GROUP`: Makes the new process the root process of a new process group. This also disables CTRL+C signal handlers for all processes of the new process group.
- `CREATE_NO_WINDOW`: Creates a new console for the process but does not open a visible window for that console. This flag is used instead of `DETACHED_PROCESS` since the new process can spawn a new console itself (in the absence of one being available) but that console creation will spawn a visible console window. (Related: [Issue 41619](https://bugs.python.org/issue41619))

This configuration conversion for Windows was added to `Subprocess.final_kwargs()`.

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -62,6 +62,7 @@ class Subprocess:
          * Converting any environment overrides into a full environment
          * Converting the `cwd` into a string
          * Default `text` to True so all outputs are strings
+         * Convert start_new_session=True to creationflags on Windows
         """
         # If `env` has been provided, inject a full copy of the local
         # environment, with the values in `env` overriding the local
@@ -85,6 +86,31 @@ class Subprocess:
         # returned as strings instead of bytes.
         if not ('text' in kwargs or 'universal_newlines' in kwargs):
             kwargs['text'] = True
+
+        # For Windows, convert start_new_session=True to creation flags
+        if self.command.host_os == 'Windows':
+            try:
+                if kwargs.pop('start_new_session') is True:
+                    if 'creationflags' in kwargs:
+                        raise AssertionError(
+                            "Subprocess called with creationflags set and start_new_session=True.\n"
+                            "This will result in CREATE_NEW_PROCESS_GROUP and CREATE_NO_WINDOW being "
+                            "merged in to the creationflags.\n\n"
+                            "Ensure this is desired configuration or don't set start_new_session=True."
+                        )
+                    # CREATE_NEW_PROCESS_GROUP: Makes the new process the root process
+                    #     of a new process group. This also disables CTRL+C signal handlers
+                    #     for all processes of the new process group.
+                    # CREATE_NO_WINDOW: Creates a new console for the process but does not
+                    #     open a visible window for that console. This flag is used instead
+                    #     of DETACHED_PROCESS since the new process can spawn a new console
+                    #     itself (in the absence of one being available) but that console
+                    #     creation will also spawn a visible console window.
+                    new_session_flags = self._subprocess.CREATE_NEW_PROCESS_GROUP | self._subprocess.CREATE_NO_WINDOW
+                    # merge these flags in to any existing flags already provided
+                    kwargs['creationflags'] = kwargs.get('creationflags', 0) | new_session_flags
+            except KeyError:
+                pass
 
         return kwargs
 
--- a/tests/integrations/subprocess/test_Subprocess__Popen.py
+++ b/tests/integrations/subprocess/test_Subprocess__Popen.py
@@ -3,11 +3,14 @@ import os
 import pytest
 
 from briefcase.console import Log
+from .conftest import CREATE_NEW_PROCESS_GROUP, CREATE_NO_WINDOW
 
 
-def test_call(mock_sub, capsys):
+@pytest.mark.parametrize("platform", ["Linux", "Darwin", "Windows"])
+def test_call(mock_sub, capsys, platform):
     "A simple call will be invoked"
 
+    mock_sub.command.host_os = platform
     mock_sub.Popen(['hello', 'world'])
 
     mock_sub._subprocess.Popen.assert_called_with(['hello', 'world'], text=True)
@@ -37,6 +40,68 @@ def test_call_with_path_arg(mock_sub, capsys, tmp_path):
         text=True,
     )
     assert capsys.readouterr().out == ""
+
+
+@pytest.mark.parametrize(
+    ("platform", "start_new_session", "popen_kwargs"),
+    [
+        ("Linux", None, {}),
+        ("Linux", True, {}),
+        ("Linux", False, {}),
+        ("Darwin", None, {}),
+        ("Darwin", True, {}),
+        ("Darwin", False, {}),
+        ("Windows", None, {}),
+        ("Windows", True, {'creationflags': CREATE_NEW_PROCESS_GROUP | CREATE_NO_WINDOW}),
+        ("Windows", False, {})
+    ]
+)
+def test_call_with_start_new_session(mock_sub, capsys, platform, start_new_session, popen_kwargs):
+    "start_new_session is passed thru on Linux and macOS but converted for Windows"
+
+    mock_sub.command.host_os = platform
+    mock_sub.Popen(['hello', 'world'], start_new_session=start_new_session)
+
+    if platform == "Windows":
+        mock_sub._subprocess.Popen.assert_called_with(
+            ['hello', 'world'],
+            text=True,
+            **popen_kwargs,
+        )
+        assert capsys.readouterr().out == ""
+    else:
+        mock_sub._subprocess.Popen.assert_called_with(
+            ['hello', 'world'],
+            start_new_session=start_new_session,
+            text=True,
+            **popen_kwargs,
+        )
+        assert capsys.readouterr().out == ""
+
+
+@pytest.mark.parametrize(
+    ("creationflags", "final_creationflags"),
+    [
+        (0x1, CREATE_NEW_PROCESS_GROUP | CREATE_NO_WINDOW | 1),
+        (CREATE_NEW_PROCESS_GROUP, CREATE_NEW_PROCESS_GROUP | CREATE_NO_WINDOW),
+        (0, CREATE_NEW_PROCESS_GROUP | CREATE_NO_WINDOW),
+    ]
+)
+def test_call_windows_with_start_new_session_and_creationflags(mock_sub, capsys, creationflags, final_creationflags):
+    "creationflags used to simulate start_new_session=True should be merged with any existing flags"
+
+    mock_sub.command.host_os = "Windows"
+
+    # use commented test below when merging creationflags is allowed
+    with pytest.raises(AssertionError, match="Subprocess called with creationflags set"):
+        mock_sub.Popen(['hello', 'world'], start_new_session=True, creationflags=creationflags)
+
+    # mock_sub._subprocess.Popen.assert_called_with(
+    #     ['hello', 'world'],
+    #     creationflags=final_creationflags,
+    #     text=True,
+    # )
+    # assert capsys.readouterr().out == ""
 
 
 def test_debug_call(mock_sub, capsys):
--- a/tests/integrations/subprocess/test_Subprocess__check_output.py
+++ b/tests/integrations/subprocess/test_Subprocess__check_output.py
@@ -4,11 +4,14 @@ from subprocess import CalledProcessError
 import pytest
 
 from briefcase.console import Log
+from .conftest import CREATE_NEW_PROCESS_GROUP, CREATE_NO_WINDOW
 
 
-def test_call(mock_sub, capsys):
+@pytest.mark.parametrize("platform", ["Linux", "Darwin", "Windows"])
+def test_call(mock_sub, capsys, platform):
     "A simple call will be invoked"
 
+    mock_sub.command.host_os = platform
     mock_sub.check_output(['hello', 'world'])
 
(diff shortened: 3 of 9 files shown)


tests/integrations/subprocess/test_Subprocess__Popen.py
```diff
@@ -5,9 +5,11 @@
 from briefcase.console import Log
 
 
-def test_call(mock_sub, capsys):
+@pytest.mark.parametrize("platform", ["linux", "darwin", "win32"])
+def test_call(mock_sub, capsys, platform):
```
rmartin16, line 9: I added the platform to all of the `test_call` cases just to ensure that `final_kwargs` doesn't do anything unexpected in the default case.

tests/integrations/subprocess/test_Subprocess__Popen.py
```diff
@@ -39,6 +41,52 @@ def test_call_with_path_arg(mock_sub, capsys, tmp_path):
...
+    "start_new_session is passed thru on Linux and macOS but converted for Windows"
+
+    mock_sub.command.sys.platform = platform
+    mock_sub.Popen(['hello', 'world'], start_new_session=start_new_session)
+
+    if platform == "win32":
+        mock_sub._subprocess.Popen.assert_called_with(
+            ['hello', 'world'],
+            text=True,
+            **({'creationflags': 0x8000200} if start_new_session is True else {}),
```
freakboy3742, line 56: It's preferable to avoid logic in tests as much as possible; this can be handled as an extra "popen_kwargs" parameter in `parametrize()` (with an empty dictionary for the non-win32 case)

src/briefcase/integrations/subprocess.py
```diff
@@ -86,6 +87,24 @@ def final_kwargs(self, **kwargs):
...
+                    #     of a new process group. This also disables CTRL+C signal handlers
+                    #     for all processes of the new process group.
+                    # CREATE_NO_WINDOW: Creates a new console for the process but does not
+                    #     open a visible window for that console. This flag is used instead
+                    #     of DETACHED_PROCESS since the new process can spawn a new console
+                    #     itself (in the absence of one being available) but that console
+                    #     creation will also spawn a visible console window.
+                    new_session_flags = self._subprocess.CREATE_NEW_PROCESS_GROUP | self._subprocess.CREATE_NO_WINDOW
+                    # merge these flags in to any existing flags already provided
+                    kwargs['creationflags'] = kwargs.get('creationflags', 0) | new_session_flags
```
freakboy3742, line 104: This could have some interesting behaviours if the user provides their own creationflags - e.g., if they provide DETACHED_PROCESS, they're going to get creation flags with both DETACHED_PROCESS *and* CREATE_NO_WINDOW. It might be worth raising the "you've also specified creationflags" as an error in this case - at least for now, it won't be a problem because we're not using any specialist creationflags; but if this ever changes in the future, the exception will make sure the problem is caught and doesn't cause any hard-to-diagnose errors.

tests/integrations/subprocess/test_Subprocess__Popen.py
```diff
@@ -39,6 +41,52 @@ def test_call_with_path_arg(mock_sub, capsys, tmp_path):
...
+        )
+        assert capsys.readouterr().out == ""
+
+
+@pytest.mark.parametrize(
+    ("creationflags", "mod_creationflags"),
+    [
+        (1, 0x8000201),
+        (134217728, 0x8000200),
+        (0, 0x8000200),
```
freakboy3742, line 73: We can use the original subprocess constants here so it's clear what is going on in the test.

tests/integrations/subprocess/test_Subprocess__check_output.py
```diff
@@ -40,6 +42,52 @@ def test_call_with_path_arg(mock_sub, capsys, tmp_path):
...
+    "start_new_session is passed thru on Linux and macOS but converted for Windows"
+
+    mock_sub.command.sys.platform = platform
+    mock_sub.check_output(['hello', 'world'], start_new_session=start_new_session)
+
+    if platform == "win32":
+        mock_sub._subprocess.check_output.assert_called_with(
+            ['hello', 'world'],
+            text=True,
+            **({'creationflags': 0x8000200} if start_new_session is True else {}),
```
freakboy3742, line 57: As above; the original parametrize can be simplified, and this argument can be made part of the parametrize.

tests/integrations/subprocess/test_Subprocess__check_output.py
```diff
@@ -40,6 +42,52 @@ def test_call_with_path_arg(mock_sub, capsys, tmp_path):
...
+        )
+        assert capsys.readouterr().out == ""
+
+
+@pytest.mark.parametrize(
+    ("creationflags", "final_creationflags"),
+    [
+        (0x1, 0x8000201),
+        (134217728, 0x8000200),
+        (0, 0x8000200),
```
freakboy3742, line 74: As above, we can use the raw constants here.

tests/integrations/subprocess/test_Subprocess__run.py
```diff
@@ -40,6 +42,52 @@ def test_call_with_path_arg(mock_sub, capsys, tmp_path):
...
+    "start_new_session is passed thru on Linux and macOS but converted for Windows"
+
+    mock_sub.command.sys.platform = platform
+    mock_sub.run(['hello', 'world'], start_new_session=start_new_session)
+
+    if platform == "win32":
+        mock_sub._subprocess.run.assert_called_with(
+            ['hello', 'world'],
+            text=True,
+            **({'creationflags': 0x8000200} if start_new_session is True else {}),
```
freakboy3742, line 57: As above, parametrize() cleanups

tests/integrations/subprocess/test_Subprocess__run.py
```diff
@@ -40,6 +42,52 @@ def test_call_with_path_arg(mock_sub, capsys, tmp_path):
...
+        )
+        assert capsys.readouterr().out == ""
+
+
+@pytest.mark.parametrize(
+    ("creationflags", "mod_creationflags"),
+    [
+        (0x1, 0x8000201),
+        (134217728, 0x8000200),
+        (0, 0x8000200),
```
freakboy3742, line 74: As above, raw constants can be used here.

freakboy3742 (changes requested):
Nice diagnosis of the underlying problem! I've flagged a couple of fairly minor things inline, mostly in the tests, but otherwise, another solid contribution.

src/briefcase/integrations/subprocess.py
```diff
@@ -86,6 +87,24 @@ def final_kwargs(self, **kwargs):
         if not ('text' in kwargs or 'universal_newlines' in kwargs):
             kwargs['text'] = True
 
+        # For Windows, convert start_new_session=True to creation flags
+        if self.command.sys.platform == "win32":
```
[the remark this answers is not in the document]
rmartin16: Updated.

src/briefcase/integrations/subprocess.py
```diff
@@ -86,6 +87,24 @@ def final_kwargs(self, **kwargs):
...
+                if kwargs.pop('start_new_session') is True:
+                    # CREATE_NEW_PROCESS_GROUP: Makes the new process the root process
+                    #     of a new process group. This also disables CTRL+C signal handlers
+                    #     for all processes of the new process group.
+                    # CREATE_NO_WINDOW: Creates a new console for the process but does not
+                    #     open a visible window for that console. This flag is used instead
+                    #     of DETACHED_PROCESS since the new process can spawn a new console
+                    #     itself (in the absence of one being available) but that console
+                    #     creation will also spawn a visible console window.
+                    new_session_flags = self._subprocess.CREATE_NEW_PROCESS_GROUP | self._subprocess.CREATE_NO_WINDOW
```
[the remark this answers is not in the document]
rmartin16: Updated.

src/briefcase/integrations/subprocess.py
rmartin16: Added a warning....let me know if it might go a little overboard :)

tests/integrations/subprocess/test_Subprocess__Popen.py
```diff
@@ -39,6 +41,52 @@ def test_call_with_path_arg(mock_sub, capsys, tmp_path):
     assert capsys.readouterr().out == ""
 
 
+@pytest.mark.parametrize("platform", ["linux", "darwin", "win32"])
+@pytest.mark.parametrize("start_new_session", [None, True, False])
+def test_call_with_start_new_session(mock_sub, capsys, platform, start_new_session):
```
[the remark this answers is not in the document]
rmartin16: Combined the parametrize calls.

tests/integrations/subprocess/test_Subprocess__Popen.py
rmartin16: Incorporated in to `parametrize`.

tests/integrations/subprocess/test_Subprocess__check_output.py
rmartin16: Added to `parametrize`.

rmartin16:
Sooo....concerning the constants `CREATE_NEW_PROCESS_GROUP` and `CREATE_NO_WINDOW` from `subprocess`....they are actually only available when python is literally on Windows. This is apparently how `subprocess` actually determines whether it's running on Windows or not. I was a little surprised at that... I did consolidate where those constants are specified in the tests and used them as though I would use constants from `subprocess`.

src/briefcase/integrations/subprocess.py
freakboy3742: I'd lean towards `AssertionError` here rather than a bare `Exception`. Definitely doesn't need a custom Exception class, though.

freakboy3742:
> Sooo....concerning the constants `CREATE_NEW_PROCESS_GROUP` and `CREATE_NO_WINDOW` from `subprocess`....they are actually only available when python is literally on Windows. 

Well that's... fun :-)

In which case, the approach you've taken here is good (constants in the test classes); the only thing I'd add would be a safety catch test that only runs on Windows, validating that the test constants and the "real" constants are the same.

freakboy3742 (changes requested):
This is looking great; a minor tweak to the exception class mentioned inline, plus the extra verification test for the constant, and this should be good to merge.

The only other thing I'd like to get is independent verification that it works on Windows (since I can't test that myself); maybe @t-arn can provide that?

src/briefcase/integrations/subprocess.py
rmartin16: Updated to `AssertionError`.

rmartin16:
Updated the Exception class and added the `subprocess` constants verification test.

My primary machine runs Windows (although it also dual boots) but Ive been doing my dev in a Ubuntu 18 VM. So, I was able to verify all this on Windows 10; either way, it would definitely be good to see it work on at least one other machine first.

Also, I would expect this problem to manifest on macOS as well. Although, I noticed the iOS and macOS `run_app` methods both use `open` to launch the simulator or app, respectively. I am admittingly ignorant if this is normal or even if it's required on mac, but the `open` command results in a process in at least a new process group and it is not attached to the originating terminal. Just FYI I guess in case the `open` command isn't always used.

freakboy3742 (approved):
Awesome - looks great! Pending independent confirmation that this fixes the problem on Windows, I'll merge.

And yes - the same problem *does* occur on macOS with the Android emulator; and this patch fixes it. It doesn't occur with the iOS simulator, because of the use of `open` (which, AFAICT, is the recommended way of starting the simulator app).

t-arn:
@freakboy3742 How do I test this PR on my machine? Do I need to clone the fork of rmartin16?

freakboy3742:
@t-arn Yes, you need to get a checkout of rmartin16's fork; then install the content of that checkout into your virtual environment.

t-arn:
OK, I can confirm that this PR works great on my Windows 11 machine. Ctrl-C stopps the logging and gives me back control of my command line and the emulator keeps running :-) Well done @rmartin16 !

PR #734, opened 2022-05-08
Let Subprocess Coalesce Environment Variables

rmartin16:
One of the augmentations to `subprocess` is treating any environment variables from the caller as overrides to the current environment....instead of using the passed environment variables as the totality of the environment for the call. Currently, some calls to `Subprocess` are coalescing the current environment with overrides instead of just passing the overrides.

These changes defer creating the entire environment to `Subprocess`.

One motivation for this change is less verbose logging in debug mode (i.e. `-v`) where only environment variable overrides should be printed; since these commands are sending the entire environment in, the entire environment is being logged.

## PR Checklist:

### DIFF
--- a/changes/734.misc.rst
+++ b/changes/734.misc.rst
@@ -1,0 +1,1 @@
+Uses of Subprocess now only pass environment variable overrides instead of the entire current environment with overrides already applied.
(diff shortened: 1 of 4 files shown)


freakboy3742 (commented):
Nice catch. I saw this during my review of your `-vv` patch, and put made a mental note that it should be cleaned up... but then forgot to actually write it down anywhere.

freakboy3742 (approved):

PR #735, opened 2022-05-10
Prototype: General Method to Run Command within a Wait Bar

rmartin16:
## Run Command with Wait Bar
I developed this idea while implementing the Popen output streamer. There are several places where commands are executed and awaited without necessarily any feedback to the users. So, why not use the `WaitBar` to at least show that Briefcase is actually waiting on something.

@freakboy3742, curious on your thoughts and if this idea is compelling at all?

## Implementation
- Added `run_with_waitbar` to `Subprocess`.
- This is basically a simplified copy of the real `subprocess.run` where we loop waiting for the command to finish.
- Note: It is only demonstrated right now with the `run iOS` command.

## Considerations
- Provides a very simple method to implement a `WaitBar` for a command that doesn't output anything but may take a minute.
- This would more or less expose Briefcase codebase to bugs that are/will be addressed in `subprocess`.
- Without a _much_ more sophisticated approach, a command may deadlock if the buffers for `stdout` or `stderr` fill.
  - This can happen because it is difficult to implement a robust mechanism to periodically read from the pipes without sustaining potentially blocking calls. (For instance, `subprocess` uses threading for this in some instances.)
  - However, in practice, this is largely avoidable by only wrapping commands with little or no output.
- If the idea of "running a command with a wait bar is in fact compelling but this implementation is too risky....it alternatively shouldn't be that hard to just throw a call to `subprocess.run` in to a thread and instead just loop while polling the thread.

## PR Checklist:


freakboy3742 (commented):
Initial impression: seems like an idea worth pursuing. It's a cosmetic change, but one that provides useful feedback during processes that take time to run (like simulator boot loops). 

The only real reservations that I have are around the handling of stderr/output. The use case I'm anticipating is something where the long-running process provides useful feedback during the launch process; this approach will obscure that output until either the process fully succeeds/fails, or the user kills the process. I'm not sure if that's actually going to be an issue in practice, though; that might be a case where we need to see it working in the real world and make a practical judgement on when it works and when it doesn't.

rmartin16:
Ahh...that's a really good point. I'll reopen this if I'm able to address that requirement in the future. Thanks.

PR #742, opened 2022-05-20
Stop Log Stream when App Exits 

### LINKED ISSUE #741: Stop log streaming when app exits
On macOS, you can run an app and see a log stream. However, if you quit the app (or the app crashes), the log stream remains, even though there's nothing to log any more.

On desktop platforms, when the application exits, the log stream should also exit. It should *not* be necessary to hit CTRL-C to terminate the log stream; although hitting CTRL-C to prematurely terminate the log stream should remain an option.

The best fix will require identifying the process ID of the app that is started, and incorporating a monitor of that process ID as part of the log stream. An alternative (but less desirable) option would be to define a "hang up" signal that an app can include in a log stream that will be interpreted by the log stream as a "stop logging" indicator.

Status quo. Require ctrl-c to terminate the log.

**Additional context**

At present, this only impacts macOS. However, if #628 or #629 is implemented, it is likely the same problem will occur on Linux and Windows, respectively. 

It won't ever impact mobile platforms, because apps don't ever "exit" in the desktop sense.

rmartin16:
Closes #741 by stopping `log stream` and allowing briefcase to exit when the app terminates.

A few important steps to achieve this. (I'll add tests once we agree on the implementation).

## App Process Identification
This is non-trivial...especially on macOS since `open` starts a wholly independent process. Therefore, one must rummage around in the process list to determine which process is the App. To stand a chance to make this portable, [`psutil`](https://psutil.readthedocs.io/en/latest/#) is required.

To identify a process of interest, we can use the process' binary/exe name/filepath, command line, or its name. The binary/exe or name seem more ambiguous than command line; for instance, on Windows, name is set to `pythonw.exe` and the binary is the full path to `pythonw.exe` (although, the path goes to the app's directory). That got me to the command line; since our use case requires executing an exact command line and then finding the process for the command line, this seemed most likely to be accurate. On macOS, the command passed to `open` isn't exactly what ends up as the command for the App process; macOS seems to run a binary inferred by the `open` argument. So, a partial match is performed against the command line to match a process.

With any of these identification methods, there is still the possibility of multiple instances of the App running... To get around that, the most recently created process is used.

## Stop Blocking in Popen Output Streamer
This actually gets back to #735.....since we're using the Popen output streamer for `log stream`, it blocks on `readline()`. And if we want to detect that the App closed, we can't have this blocking on the `stdout` pipe for `log stream`. So, I refactored the Popen output streamer in to a simple thread.

## Detecting the App Exited
I added a `stop_func` Callable arg to the Popen output streamer to periodically run to check if the Popen output streamer should stop, terminate the Popen process, and return.

As for the callable to use to check when a process has exited, I landed on simply using `psutil`'s check if a PID currently exists. While PID's can be reused, since we're constantly checking for the PID to stop existing, it is highly unlikely the OS will reassign the PID before we detect it doesn't exist. Alternatively, `psutil` has a more sophisticated [`is_running()`](https://psutil.readthedocs.io/en/latest/#psutil.Process.is_running) check for processes. However, that will require the `psutil` [`Process`](https://psutil.readthedocs.io/en/latest/#process-class) object to be exposed outside of `subprocess.py` and passed around within the larger project.

## PR Checklist:


rmartin16:
For posterity, process dumps from `psutil`:

### macOS
```json
{"cmdline": ["/Users/user/tmp/beeware-tutorial/mycoolapp/macOS/app/My Cool "
             "App/My Cool App.app/Contents/MacOS/My Cool App"],
 "connections": [],
 "cpu_percent": 0.0,
 "cpu_times": pcputimes(user=0.89249312, system=0.189621552, children_user=0.0, children_system=0.0),
 "create_time": 1653087867.017026,
 "cwd": "/",
 "environ": {<OMITTED>},
 "exe": "/Users/user/tmp/beeware-tutorial/mycoolapp/macOS/app/My Cool App/My "
        "Cool App.app/Contents/MacOS/My Cool App",
 "gids": puids(real=20, effective=20, saved=20),
 "memory_full_info": None,
 "memory_info": pmem(rss=60960768, vms=35163623424, pfaults=17212, pageins=832),
 "memory_percent": 0.7096767425537109,
 "name": "My Cool App",
 "nice": 0,
 "num_ctx_switches": pctxsw(voluntary=140333761435302, involuntary=0),
 "num_fds": 12,
 "num_threads": 7,
 "open_files": [<OMITTED>],
 "pid": 6150,
 "ppid": 1,
 "status": "running",
 "terminal": None,
 "threads": None,
 "uids": puids(real=501, effective=501, saved=501),
 "username": "user"}
```

### Windows
```json
{"cmdline": ["C:\\Users\\Russell\\tmp\\beeware-tutorial\\mycoolapp\\windows\\msi\\My "
             "Cool App\\src\\python\\pythonw.exe",
             "-m",
             "mycoolapp"],
 "connections": [<OMITTED>],
 "cpu_affinity": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15],
 "cpu_percent": 0.0,
 "cpu_times": pcputimes(user=0.390625, system=0.171875, children_user=0.0, children_system=0.0),
 "create_time": 1653086994.679865,
 "cwd": "C:\\Users\\Russell\\tmp\\beeware-tutorial\\mycoolapp",
 "environ": {<OMITTED>},
 "exe": "C:\\Users\\Russell\\tmp\\beeware-tutorial\\mycoolapp\\windows\\msi\\My "
        "Cool App\\src\\python\\pythonw.exe",
 "io_counters": pio(read_count=698, write_count=0, read_bytes=4016873, write_bytes=0, other_count=5499, other_bytes=252112),
 "ionice": <IOPriority.IOPRIO_NORMAL: 2>,
 "memory_full_info": pfullmem(rss=87347200, vms=66363392, num_page_faults=38645, peak_wset=90468352, wset=87347200, peak_paged_pool=697800, paged_pool=696848, peak_nonpaged_pool=42216, nonpaged_pool=42104, pagefile=66363392, peak_pagefile=69537792, private=66363392, uss=46157824),
 "memory_info": pmem(rss=87347200, vms=66363392, num_page_faults=38645, peak_wset=90468352, wset=87347200, peak_paged_pool=697800, paged_pool=696848, peak_nonpaged_pool=42216, nonpaged_pool=42104, pagefile=66363392, peak_pagefile=69537792, private=66363392),
 "memory_maps": [<OMITTED>],
 "memory_percent": 0.12727990692839594,
 "name": "pythonw.exe",
 "nice": <Priority.NORMAL_PRIORITY_CLASS: 32>,
 "num_ctx_switches": pctxsw(voluntary=16141, involuntary=0),
 "num_handles": 538,
 "num_threads": 19,
 "open_files": [<OMITTED>],
 "pid": 25904,
 "ppid": 12356,
 "status": "running",
 "threads": [<OMITTED>],
 "username": "JUPITER-WIN\\Russell"}
```

### Linux
```json
{"cmdline": ["/home/user/tmp/beeware-tutorial/helloworld/linux/Hello_World-0.0.1-x86_64.AppImage"],
 "connections": [],
 "cpu_affinity": [0, 1, 2, 3],
 "cpu_num": 2,
 "cpu_percent": 0.0,
 "cpu_times": pcputimes(user=0.16, system=0.07, children_user=0.0, children_system=0.0, iowait=0.0),
 "create_time": 1653087270.26,
 "cwd": "/",
 "environ": {<OMITTED>},
 "exe": "/home/user/tmp/beeware-tutorial/helloworld/linux/Hello_World-0.0.1-x86_64.AppImage",
 "gids": pgids(real=1000, effective=1000, saved=1000),
 "io_counters": pio(read_count=3455, write_count=4557, read_bytes=0, write_bytes=0, read_chars=19193752, write_chars=21072209),
 "ionice": pionice(ioclass=<IOPriority.IOPRIO_CLASS_NONE: 0>, value=4),
 "memory_full_info": pfullmem(rss=2592768, vms=23171072, shared=1544192, text=176128, lib=0, data=9826304, dirty=0, uss=1355776, pss=1378304, swap=0),
 "memory_info": pmem(rss=2592768, vms=23171072, shared=1544192, text=176128, lib=0, data=9826304, dirty=0),
 "memory_maps": [<OMITTED>],
 "memory_percent": 0.02482265331811818,
 "name": "Hello_World-0.0.1-x86_64.AppImage",
 "nice": 0,
 "num_ctx_switches": pctxsw(voluntary=2444, involuntary=13),
 "num_fds": 6,
 "num_threads": 2,
 "open_files": [<OMITTED>],
 "pid": 14601,
 "ppid": 2086,
 "status": "sleeping",
 "terminal": None,
 "threads": [<OMITTED>],
 "uids": puids(real=1000, effective=1000, saved=1000),
 "username": "user"}
```

src/briefcase/integrations/subprocess.py
```diff
@@ -244,26 +296,47 @@ def stream_output(self, label, popen_process):
...
+                while output_streamer.is_alive():
+                    if stop_func():
+                        self.cleanup(label, popen_process)
+                    time.sleep(0.5)
+            else:
+                output_streamer.join()
         except KeyboardInterrupt:
             self.cleanup(label, popen_process)
 
+    def _stream_output_thread(self, popen_process):
```
freakboy3742, line 319: If I'm reading this right, this is still a useful function to have around - looking forward to #629, we might even end up using it with the Windows launcher. I'd suggest keeping this "as is" as `stream_output_thread`, and modify the other method to be an "stream_output_thread_background" that has the termination callable.

src/briefcase/integrations/subprocess.py
```diff
@@ -26,6 +31,53 @@ def json_parser(json_output):
...
+        if command_list and proc_cmdline == command_list:
+            matching_processes.append(proc.info)
+        if command and proc_cmdline and proc_cmdline[0].startswith(command):
+            matching_processes.append(proc.info)
+
+    if len(matching_processes) == 1:
+        return matching_processes[0]["pid"]
+    elif len(matching_processes) > 1:
+        # return the ID of the most recently created matching process
+        return sorted(matching_processes, key=operator.itemgetter("create_time"))[-1]["pid"]
```
freakboy3742, line 76: Given that this is an explicitly ambiguous case, it's probably worth logging that we *might* have found the wrong process. I suspect this will be a very unlikely edge case - but then, one in a million chances happen nine times out of ten :-) On that basis, it's worth flagging to avoid any confusion when it (inevitably) happens.

src/briefcase/integrations/subprocess.py
```diff
@@ -26,6 +31,53 @@ def json_parser(json_output):
...
+
+    :param command_list: list of a command's fully qualified path and its arguments.
+    :param command: a partial or complete fully-qualified filepath to a command.
+        This is primarily intended for use on macOS where the `open` command
+        takes a filepath to a directory for an application; therefore, the actual
+        running process will be running a command within that directory.
+    :return: PID if found else None
+    """
+    matching_processes = []
+    for proc in psutil.process_iter(["cmdline", "create_time", "pid"]):
```
freakboy3742, line 65: Would be worth a small docstring here to clarify the purpose of this line - retrieving the cmdline, creation time and PID of all currently running processes.

src/briefcase/integrations/subprocess.py
```diff
@@ -26,6 +31,53 @@ def json_parser(json_output):
...
+    Checking if a PID exists is only a semi-safe proxy to determine
+    if a process is dead since PIDs can be re-used. Therefore, this
+    function should only be used via constant monitoring of a PID
+    to identify when the process goes from existing to not existing.
+
+    :param pid: integer value to be checked if assigned as a PID.
+    :return: True if PID doesn't exist; defaults to False if PID=None.
+    """
+    if pid is None:
+        return False
```
freakboy3742, line 47: Is this a good default? If the process dies really quickly, it won't get found on the PID search; so the PID will be None; this would seem to suggest that it will report as "process is not dead".

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -83,7 +84,8 @@ def run_app(self, app: BaseConfig, **kwargs):
         self.logger.info()
         self.logger.info(f"[{app.app_name}] Following system log output (type CTRL-C to stop log)...")
         self.logger.info("=" * 75)
-        self.subprocess.stream_output("log stream", log_popen)
+        app_pid = get_process_id_by_command(command=str(self.binary_path(app)))
```
freakboy3742, line 87: Looking for edge cases - if the process crashes quickly, `app_pid` will return None; but this will still try to display a log (and because of the behaviour of is_process_dead, will run forever). Should we perhaps check explicitly for app_pid == None (or raise an exception for "can't find PID"), and log that as a "can't stream because I can't find the app"?)

freakboy3742 (changes requested):
Apologies for the delay in the review. I've flagged a couple of minor details inline, but on the whole, this looks like it's headed in the right direction. 

I can't think of anything better than `psutil` to retrieve the PID spawned by `open`; and iOS/macOS both use the cmdline of the process as a basis for log filtering, so that aspect of this patch is no worse than what we've already got.

You've only looked at macOS here; the one thing I'll flag is that it's not worth spending too much time looking at Windows at this time, because #629 is on my near-term todo list, and that will dramatically change how the process manifests. As long as the broad approach looks like it will work (i.e., psutil can be invoked and interrogated *at all*), we can leave the specifics until we have a better app-run story on Windows.

src/briefcase/integrations/subprocess.py
rmartin16: Updated.

src/briefcase/integrations/subprocess.py
rmartin16: Makes sense. Although, access to a `logger` is not immediately available here. This hearkens back a bit to making `Console` and/or `Log` singletons so their universal access would be trivial.

In lieu of that, I simply added `logger` as an optional parameter to this function. I'm not particularly keen on this...but the other alternative I see is to pigeonhole this function inside `Subprocess` which I think I'm even less keen on.

Thoughts?

P.S. While this function is general in nature, the message I added to be logged does imply the process being sought is the user's Application. I think this compromise makes sense, though, given the nature of this project.

src/briefcase/integrations/subprocess.py
rmartin16: Yeah...my thinking was that if the caller doesn't know the process (as made evident by the absent PID), then it may be better to assume the process exists. This approach also leans towards maintaining previous behavior by log streaming regardless of the App's status as running or not.

That said, this default was a late addition to this PR; I think I'd rather push the interpretation of `pid=None` back up to the caller; they will have better context on the appropriate behavior.

src/briefcase/platforms/macOS/__init__.py
rmartin16: Good call. I added an error log to indicate this. I also refactored `macOSRunMixIn` just a bit to ensure the log streaming process is terminated on exit; it is safe to call `cleanup` multiple times or after the process exits itself.

rmartin16:
Made some updates or requested additional information/clarification to comments.

I tested the principle of this strategy of process identification based on command line in windows and linux using the commands currently used to invoke the user's app; they were successful in finding the process. However, I am certainly unsure how well that will carry over to future changes for those platforms. That said, we can always add more parameters to `get_process_id_by_command` to perform identification differently obviously.

I should be able to finalize this PR once I hear back.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -64,26 +65,33 @@ def run_app(self, app: BaseConfig, **kwargs):
...
+                self.subprocess.run(
+                    [
+                        'open',
+                        '-n',  # Force a new app to be launched
+                        os.fsdecode(self.binary_path(app)),
+                    ],
+                    check=True,
+                )
+            except subprocess.CalledProcessError:
+                raise BriefcaseCommandError(f"Unable to start app {app.app_name}.")
```
freakboy3742, line 81: Should probably do a `self.subprocess.cleanup()` here before returning. In practice it probably won't matter much as the process will be cleaned up by the parent being killed, but it doesn't hurt.

freakboy3742 (changes requested):
Functionally, this looks good; the only thing missing at this point are tests.

Also - as I indicated on your other PR, I've just merged #744 enabling black, so there will be a couple of merge conflicts; if you need any help merging that cleanly, let me know.

src/briefcase/integrations/subprocess.py
rmartin16: Gotcha. I'll leave this as internal to `Subprocess` then for now.

src/briefcase/platforms/macOS/__init__.py
rmartin16: Given I also need to call `cleanup()` in the event the app PID isn't found, I put this whole section in to a `try\finally` block to always clean up at the end.

src/briefcase/integrations/subprocess.py
```diff
@@ -235,26 +294,46 @@ def stream_output(self, label, popen_process):
...
+            output_streamer.start()
+            if stop_func:
+                while output_streamer.is_alive() or not stop_func():
+                    time.sleep(0.5)
+            else:
+                output_streamer.join()
         except KeyboardInterrupt:
+            pass  # allow CTRL+C to gracefully stop streaming
+        finally:
             self.cleanup(label, popen_process)
```
rmartin16, line 315: I moved `cleanup()` in to a `finally` block; my changes had originally introduced multiple exit points for a deceptively simple loop where `cleanup()` would need to be invoked for some exits but not others....although, it was definitely not clear when/why to call `cleanup()` unless you're pretty familiar with all the interactions.

The `cleanup()` function is safe to call multiple times and largely implements short circuits for already dead processes. I talk about this more with the changes to `cleanup` itself below.

src/briefcase/integrations/subprocess.py
```diff
@@ -263,12 +342,13 @@ def cleanup(self, label, popen_process):
             to provide context in logging messages.
         :param popen_process: The Popen instance to clean up.
         """
-        popen_process.terminate()
-        try:
-            popen_process.wait(timeout=3)
-        except subprocess.TimeoutExpired:
-            self.command.logger.warning(f"Forcibly killing {label}...")
-            popen_process.kill()
+        with popen_process:
```
rmartin16, line 345: Soo....`subprocess.Popen` interestingly does not expose a `close()` API...which would probably just be a wrapper on `__exit__()`. The consequence of not calling `__exit__()` is technically leads to leaking file/proxyfile handles and potentially creating zombie processes (although, as we know, these are cleaned up with Python exits). So, similar to ensuring we're calling `cleanup()` elsewhere, this ensures we also tell `Popen` to cleanup.

Since `Popen.__exit__()` is just calling `close()` on `stdout/err/in` and ensuring the process has exited, this should be safe to call multiple times therefore allowing callers to also open a `Popen` object via their own `with` without any issues. That is, as long as the expectation that calling `close()` multiple times, in general, in Python [remains safe](https://docs.python.org/3/library/io.html#io.IOBase.close).

src/briefcase/integrations/subprocess.py
```diff
@@ -263,12 +342,13 @@ def cleanup(self, label, popen_process):
...
+            popen_process.terminate()
+            try:
+                popen_process.wait(timeout=3)
+            except subprocess.TimeoutExpired:
+                self.command.logger.warning(f"Forcibly killing {label}...")
+                popen_process.kill()
```
rmartin16, line 351: Soo....`subprocess.Popen` interestingly does not expose a `close()` API...which would probably just be a wrapper on `__exit__()`. The consequence of not calling `__exit__()` is technically leads to leaking file/proxyfile handles and potentially creating zombie processes (although, as we know, these are cleaned up with Python exits). So, similar to ensuring we're calling `cleanup()` elsewhere, this ensures we also tell `Popen` to cleanup.

Since `Popen.__exit__()` is just calling `close()` on `stdout/err/in` and ensuring the process has exited, this should be safe to call multiple times therefore allowing callers to also open a `Popen` object via their own `with` without any issues. That is, as long as the expectation that calling `close()` multiple times, in general, in Python [remains safe](https://docs.python.org/3/library/io.html#io.IOBase.close).

src/briefcase/integrations/subprocess.py
rmartin16, line 315: I moved `cleanup()` in to a `finally` block; my changes had originally introduced multiple exit points for a deceptively simple loop where `cleanup()` would need to be invoked for some exits but not others....although, it was definitely not clear when/why to call `cleanup()` unless you're pretty familiar with all the interactions.

The `cleanup()` function is safe to call multiple times and largely implements short circuits for already dead processes. I talk about this more with the changes to `cleanup` itself below.

tests/integrations/subprocess/test_Subprocess__stream_output.py
```diff
@@ -30,42 +38,44 @@ def popen_process():
...
+    def send_ctrl_c():
+        raise KeyboardInterrupt()
 
-    # Response to the CTRL-C is a process cleanup.
-    mock_sub.cleanup.assert_called_once_with("testing", popen_process)
+    mock_sub.stream_output("testing", popen_process, stop_func=send_ctrl_c)
 
-    assert capsys.readouterr().out == ("output line 1\n" "\n")
+    assert capsys.readouterr().out == ("output line 1\n" "\n" "output line 3\n")
+    mock_sub.cleanup.assert_called_once_with("testing", popen_process)
```
rmartin16, line 78: Moving the output streaming to a thread means that CTRL+C will no longer actually interrupt streaming since Python only handles signals in the main thread. Therefore, after a CTRL+C now, the code will request the Popen process terminates, wait 3 seconds, and then kill the process. This means everything that can be printed in 3 seconds will and control will then return to the user.

tests/integrations/subprocess/test_Subprocess__stream_output.py
```diff
@@ -74,3 +84,14 @@ def test_process_exit_with_queued_output(mock_sub, popen_process, capsys):
...
+
+
+@pytest.mark.parametrize("stop_func_ret_val", (True, False))
+def test_stop_func(mock_sub, popen_process, stop_func_ret_val, capsys):
+    """All output is printed whether stop_func aborts streaming or not."""
+    mock_sub.stream_output(
+        "testing", popen_process, stop_func=lambda: stop_func_ret_val
+    )
+    assert capsys.readouterr().out == ("output line 1\n" "\n" "output line 3\n")
+    mock_sub.cleanup.assert_called_once_with("testing", popen_process)
```
rmartin16, line 97: Given we are only mimicking a real process for testing, I couldn't really think of a a way to more meaningfully test `stop_func`. I am definitely open to ideas.

tests/integrations/subprocess/test_Subprocess__stream_output.py
```diff
@@ -13,7 +21,7 @@ def popen_process():
     # continuously returning "" once it reaches EOF
     process.stdout.readline.side_effect = [
         "output line 1\n",
-        "\n",
+        sleep(0.1) or "\n",  # sneaky method to insert time passing in output
```
rmartin16, line 24: So, I'm not especially proud of this (except that its clever...but that usually just gets you in trouble...).

I kept hitting race conditions during testing. The output streamer would finish too quickly and `output_streamer.is_alive()` would usually return `False`. Therefore, the inner `while` loop wouldn't run. This sleep here avoids that.

src/briefcase/integrations/subprocess.py
```diff
@@ -235,25 +294,46 @@ def stream_output(self, label, popen_process):
...
+            if stop_func:
+                while output_streamer.is_alive() and not stop_func():
+                    time.sleep(0.5)
+            else:
+                output_streamer.join()
         except KeyboardInterrupt:
+            pass  # allow CTRL+C to gracefully stop streaming
+        finally:
             self.cleanup(label, popen_process)
+            output_streamer.join()
```
rmartin16, line 316: After initially deciding a final `join` on the thread wasn't necessary because `cleanup` literally ensures the process the thread is monitoring is killed, I changed my mind during testing.

In the case of the keyboard interrupt test, pytest wouldn't always be able to capture the output in `capsys` because it got flushed through `stdout` from the thread too late....meaning that `stream_output` was returning control before the thread was truly done....or in the least before its streams were flushed.

Additionally, in thinking about a real world case where this might be a problem.....imagine a process sends an immense amount of data to `stdout`. Then the output streamer thread is likely to end up with a backlog in the pipe to be read. If the process terminates at this point, that thread will continue reading that pipe and printing output to the screen. Therefore, it would probably be better to let the thread finish before we return control back upstream; otherwise, upstream code (being executed in the main thread) could then try to print messages that would be interleaved with the output streamer's printing.

OTOH, if that scenario were to happen, having all that output on the screen may or may not be desirable anyway. Therefore, we may want to consider also killing the thread after a bit the same way we do with the process. Python doesn't expose an API to kill threads but we could always pass a stop flag to the thread for it to monitor. Probably overkill....but putting it out there.

tests/platforms/macOS/app/test_run.py
```diff
@@ -8,13 +8,19 @@
...
     command.subprocess = mock.MagicMock()
     log_stream_process = mock.MagicMock()
     command.subprocess.Popen.return_value = log_stream_process
 
+    monkeypatch.setattr(
+        "briefcase.platforms.macOS.get_process_id_by_command", lambda *a, **kw: 100
+    )
+    monkeypatch.setattr(
+        "briefcase.platforms.macOS.is_process_dead", lambda *a, **kw: False
+    )
```
rmartin16, line 23: I struggled a bit to discover a good way to mock these two functions. I ultimately landed here but I'm definitely open to feedback on any best practice you're aware of.

rmartin16:
@freakboy3742, rebased on the black implementation (which is a really cool addition...love that all those checks run automatically cause I'd always forget to run flake before and inevitably github CI would find whitespace or something....also nice to have black's consistent style).

Everything should be in now. Testing turned up some interesting quirks....I put them all in individual comments we can use to discuss them.

setup.cfg
```diff
@@ -60,6 +60,7 @@ install_requires =
     requests >= 2.22.0
     GitPython >= 3.0.8
     dmgbuild >= 1.3.3; sys_platform == "darwin"
+    psutil ~= 5.9.0
```
rmartin16, line 63: PyCharm automatically added this.

As for the version, I think it is safe to require a recent version of `psutil` since I think the expectation is for users to use a dedicated virtual environment....although, I wouldn't be surprised if users installed briefcase in to their app's virtual environment.

As for the `~=`, I usually prefer it over `>=` but wasn't sure if you might have a reason to avoid it.

src/briefcase/integrations/subprocess.py
rmartin16: I actually ended up reverting this `with` block since closing the Popen process here while the thread may still be running can cause `readline` to throw an error (since `stdout` will have been closed). In general, not closing these resources certainly isn't the end of the world.....however, we may need to establish a more concrete lifecycle for them. Perhaps I'll look at this again in a future PR.

src/briefcase/integrations/subprocess.py
freakboy3742: I think your instinct is right that the `join` is worth it. It's an edge case, but if you've got a process dumping large volumes of logs, there's a good chance the reason is because you're trying to debug some odd crash behavior - so making sure that the stream is fully purged before control is returned will avoid confusion.

I see what you're saying about extra safeties around killing the thread, but I think we can put that in the "wait and see" category. In my testing, I haven't noticed a problem with thread termination, so I'm happy to land this as-is. If we start seeing reports of odd behavior, we can revisit that as a bugfix/improvement.

tests/integrations/subprocess/test_Subprocess__stream_output.py
```diff
@@ -13,7 +21,7 @@ def popen_process():
     # continuously returning "" once it reaches EOF
     process.stdout.readline.side_effect = [
         "output line 1\n",
-        "\n",
+        sleep(0.25) or "\n",  # sneaky method to slow the output streamer down
```
freakboy3742: Are you sure this works the way you think it does? The sleep will be evaluated when the side effect is defined, not when the side effect is realised; this would need to be in a lambda to be deferred. The only impact of this line will be to make the setup of the test case slower.

tests/integrations/subprocess/test_Subprocess__stream_output.py
freakboy3742: The reason for the change makes sense, and I can't think of an alternative approach here (at least, not one that I actually like. You could set up a thread for testing purposes that sleeps for a time then sends a SIGINT to the main thread... but... ewwww... 🤮 )

freakboy3742 (approved):
This looks great. I've made one tweak reducing the timeout to make the log monitor a little more responsive (0.5s was hitting the edge of my perception of "this isn't responding"). I've also removed the sleep in the test case - I'm not sure what effect you were seeing, but it must have been coincidence (or a side effect of a draft form of the test that you didn't submit). 

Overall - another great contribution - thanks very much!

tests/integrations/subprocess/test_Subprocess__stream_output.py
rmartin16: Ahhh....of course....i must have been working on this too long when I thought this would help. At any rate, I think there may still be race condition for whether codecov thinks the inner `sleep` in `stream_output` is covered or not....not a big deal either way.

PR #743, opened 2022-05-24
Refactor and simplify

danyeaw:
I did some refactoring chores to clean up:

- Explicitly raise from previous errors
- Return directly from inline variables
- Simplify if statements, use if expressions when available
- Converted a couple of strings to f-strings
- Use list comprehensions

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -281,14 +281,13 @@ class AndroidSDK:
                 cmdline_tools_zip_path,
                 extract_dir=self.cmdline_tools_path.parent
             )
-        except (shutil.ReadError, EOFError) as exc:
+        except (shutil.ReadError, EOFError) as e:
             raise BriefcaseCommandError(f"""\
 Unable to unpack Android SDK Command-Line Tools ZIP file. The download may have been interrupted
 or corrupted.
 
 Delete {cmdline_tools_zip_path} and run briefcase again.
-""") from exc
-
+""") from e
 
         # If there's an existing version of the cmdline tools (or the version marker), delete them.
         if self.cmdline_tools_path.exists():
@@ -382,7 +381,6 @@ its output for errors.
 
     $ {self.sdkmanager_path} --licenses
 """) from e
-
 
         if not license_path.exists():
             raise BriefcaseCommandError("""\
@@ -600,7 +598,6 @@ Input has been disabled; can't select a device to target.
 Use the -d/--device option to explicitly specify the device to use.
 """) from e
 
-
         # Proces the user's choice
         if choice is None:
             # Create a new emulator. No device ID or AVD.
@@ -625,8 +622,8 @@ Use the -d/--device option to explicitly specify the device to use.
                 device = choice
                 name = device_choices[choice]
                 avd = details.get("avd")
-            except KeyError as exc:
-                raise InvalidDeviceError("device ID", choice) from exc
+            except KeyError as e:
+                raise InvalidDeviceError("device ID", choice) from e
 
         if avd:
             self.command.logger.info(f"""
@@ -736,8 +733,8 @@ An emulator named '{avd}' already exists.
                     url=skin_url,
                     download_path=self.root_path,
                 )
-            except requests_exceptions.ConnectionError as exc:
-                raise NetworkFailure(f"download {skin} device skin") from exc
+            except requests_exceptions.ConnectionError as e:
+                raise NetworkFailure(f"download {skin} device skin") from e
 
             # Unpack skin archive
             try:
@@ -798,7 +795,7 @@ briefcase run android -d @{avd}
                     pass
 
         # Augment the config with the new key-values pairs
-        avd_config |= updates
+        avd_config.update(updates)
 
         # Write the update configuration.
         with avd_config_filename.open('w') as f:
@@ -815,8 +812,10 @@ briefcase run android -d @{avd}
         if avd not in set(self.emulators()):
             raise InvalidDeviceError("emulator AVD", avd)
         self.command.logger.info(f"Starting emulator {avd}...")
-        emulator_popen = self.command.subprocess.Popen([os.fsdecode(self.emulator_path), f'@{avd}', '-dns-server', '8.8.8.8'], env=self.env, stdout=subprocess.PIPE, stderr=subprocess.PIPE, start_new_session=True)
-
+        emulator_popen = self.command.subprocess.Popen(
+            [os.fsdecode(self.emulator_path), f'@{avd}', '-dns-server', '8.8.8.8'],
+            env=self.env, stdout=subprocess.PIPE, stderr=subprocess.PIPE, start_new_session=True
+        )
 
         # The boot process happens in 2 phases.
         # First, the emulator appears in the device list. However, it's
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -801,8 +801,8 @@ briefcase run android -d @{avd}
                 except ValueError:
                     pass
 
-        # Augment the config with the new new key-values pairs
-        avd_config.update(updates)
+        # Augment the config with the new key-values pairs
+        avd_config |= updates
 
         # Write the update configuration.
         with avd_config_filename.open('w') as f:
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -532,7 +532,7 @@ connection.
 
                 # Device can be looked up by device ID or AVD
                 device_choices[d] = full_name
-                device_choices["@" + avd] = full_name
+                device_choices[f"@{avd}"] = full_name
             else:
                 # It's a physical device (might be disabled)
                 full_name = f"{name} ({d})"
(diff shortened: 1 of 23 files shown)


src/briefcase/integrations/android_sdk.py
```diff
@@ -282,13 +281,14 @@ def install(self):
                 cmdline_tools_zip_path,
                 extract_dir=self.cmdline_tools_path.parent
             )
-        except (shutil.ReadError, EOFError):
+        except (shutil.ReadError, EOFError) as exc:
```
freakboy3742, line 284: Really minor consistency thing here; we've used `as e` everywhere except this once case.

src/briefcase/integrations/android_sdk.py
```diff
@@ -801,8 +797,8 @@ def update_emulator_config(self, avd, updates):
                 except ValueError:
                     pass
 
-        # Augment the config with the new new key-values pairs
-        avd_config.update(updates)
+        # Augment the config with the new key-values pairs
+        avd_config |= updates
```
freakboy3742, line 801: The |= operator wasn't added until Py3.9, so we can't use it yet.

src/briefcase/integrations/android_sdk.py
```diff
@@ -816,35 +812,27 @@ def start_emulator(self, avd):
...
-                adb = None
-                known_devices = set()
-                while adb is None:
-                    startup_wait_bar.update()
-                    if emulator_popen.poll() is not None:
-                        raise BriefcaseCommandError(f"""\
+        if avd not in set(self.emulators()):
+            raise InvalidDeviceError("emulator AVD", avd)
+        self.command.logger.info(f"Starting emulator {avd}...")
+        emulator_popen = self.command.subprocess.Popen([os.fsdecode(self.emulator_path), f'@{avd}', '-dns-server', '8.8.8.8'], env=self.env, stdout=subprocess.PIPE, stderr=subprocess.PIPE, start_new_session=True)
```
freakboy3742, line 818: There's definitely a loss of readability here by removing the newlines.

src/briefcase/integrations/docker.py
```diff
@@ -127,13 +127,9 @@ def verify_docker(command):
...
         command.logger.warning(DOCKER_INSTALLATION_STATUS_UNKNOWN_WARNING)
-    except FileNotFoundError:
+    except FileNotFoundError as e:
         # Docker executable doesn't exist.
-        raise BriefcaseCommandError(
-            DOCKER_NOT_INSTALLED_ERROR.format(
-                **docker_install_details(command.host_os)
-            )
-        )
+        raise BriefcaseCommandError(DOCKER_NOT_INSTALLED_ERROR.format(**docker_install_details(command.host_os))) from e
```
freakboy3742, line 132: Line length is unwieldy here.

src/briefcase/integrations/docker.py
```diff
@@ -186,16 +182,11 @@ def _verify_docker_can_run(command):
...
-            raise BriefcaseCommandError(LACKS_PERMISSION_ERROR)
-        elif (
-            # error message on Ubuntu
-            'Is the docker daemon running?' in failure_output
-            # error message on macOS
-            or 'connect: connection refused' in failure_output
-        ):
-            raise BriefcaseCommandError(DAEMON_NOT_RUNNING_ERROR)
+            raise BriefcaseCommandError(LACKS_PERMISSION_ERROR) from e
+        elif 'Is the docker daemon running?' in failure_output or 'connect: connection refused' in failure_output:
```
freakboy3742, line 186: Line length readability

src/briefcase/integrations/wix.py
```diff
@@ -29,10 +29,7 @@ def __init__(self, command, wix_home=None, bin_install=False):
             binaries-only install does not.
         """
         self.command = command
-        if wix_home:
-            self.wix_home = wix_home
-        else:
-            self.wix_home = command.tools_path / 'wix'
+        self.wix_home = wix_home or command.tools_path / 'wix'
```
freakboy3742, line 32: Logically correct, but I'd argue it's not at all obvious from reading what is going on here. The older if/else logic is a lot more explicit.

src/briefcase/integrations/xcode.py
```diff
@@ -469,12 +471,7 @@ def get_identities(command, policy):
             ['security', 'find-identity', '-v', '-p', policy],
         )
 
-        identities = dict(
-            IDENTITY_RE.match(line).groups()
-            for line in output.split('\n')
-            if IDENTITY_RE.match(line)
-        )
+        return dict(IDENTITY_RE.match(line).groups() for line in output.split('\n') if IDENTITY_RE.match(line))
```
freakboy3742, line 474: Readability (plus... is there any reason to not do this a dict comprehension with `{}`?)

src/briefcase/platforms/windows/msi.py
```diff
@@ -43,7 +43,7 @@ def support_package_url_query(self):
         return [
             ('platform', self.platform),
             ('version', self.python_version_tag),
-            ('arch', "amd64" if (struct.calcsize("P") * 8) == 64 else "win32"),
+            ('arch', "amd64" if struct.calcsize("P") == 8 else "win32"),
```
freakboy3742, line 46: This one is slightly deliberate - yes, the end result is "==8", but the real check is "is this a 64 bit pointer".

freakboy3742 (changes requested):
For the most part, very straightforward and welcome improvements (especially the passing of exception context).

There are a couple of themes in my review comments:

 * Consistency in the `as e` addition. It's inconsistent where you've used `as e`, `as error`, `as exception`; unless there's a syntactic ambiguity, I'd argue it's better to stick to `as e` across the board.
 * Line length. I've got flake8 set up to accept 119 character lines; but that shouldn't be taken as an invitation to make lines long just because you can :-). There's a lot of places (some of which I've flagged explicitly) where pulling out the newlines dramatically reduces readability. In particular, complex comprehensions benefit from being split over multiple lines because the operative clauses are need to be independently understood. Breaking a single 119 char line into 3 lines that are "the pattern for each item", "the list being iterated", "the condition on list items" makes it a lot easier to digest what is going on. The 119 line limit is really there because there are rare occasions where lines need to run long; unless you've got a good reason, black's 89 char line limit (80+10%) is a good rule of thumb to aim for.
 * Similarly, while 119 characters allows you to reduce short if/else statements to a single line, that's not always desirable; Yes, an if-then statement is 4 lines - but we're not paying Github by line count :-) If there's even slightly complex logic in an if/else statement, I'd argue using multiple lines is preferable, even if the single statement will fit in 119 characters. Broadly speaking, the only context in which I really like `X if Y else Z` as a syntax construction is in a comprehension; in almost every other situation, I find the condensed syntax harder to read.

danyeaw:
Hi @freakboy3742! Thanks so much for all the feedback, I made the updates. On the formatting side of things, have you considered using black? I know we discussed it in the past, but boy getting rid of the discussion on formatting is such a nice cognitive overhead reduction :+1::smile:.

src/briefcase/integrations/wix.py
```diff
@@ -154,14 +154,13 @@ def upgrade(self):
...
-        else:
+        if not self.managed_install:
             raise NonManagedToolError('WiX')
+        if self.exists():
+            self.command.logger.info("Removing old WiX install...")
+            self.command.shutil.rmtree(self.wix_home)
+
+            self.install()
+            self.command.logger.info("...done.")
+        else:
```
freakboy3742, line 165: Really minor - but given this PR has moved to an "if problem: return; do stuff" rather than "if ok: do stuff; else: return" pattern, this should probably be updated to do an "if not self.exists()" check.

src/briefcase/integrations/xcode.py
```diff
@@ -469,12 +470,13 @@ def get_identities(command, policy):
...
-            for line in output.split('\n')
-            if IDENTITY_RE.match(line)
-        )
+        return {
+            (
+                IDENTITY_RE.match(line).groups()
+                for line in output.split('\n')
+                if IDENTITY_RE.match(line)
+            )
+        }
```
freakboy3742, line 479: The brackets are a problem here; this will be returning a set of pairs, not a dict. (and I think this is the source of the smoke test failures)

freakboy3742 (changes requested):
One bug (picked up by the test suite), and one minor (and possibly optional, if you're not feeling it) cleanup flagged inline; otherwise this is looking good.

As for black - that's on my list to look at :-) I was primarily waiting until black declared themselves "non-beta"; that happened [earlier this year](https://github.com/psf/black/releases/tag/22.1.0), so it's now down to the mechanics.

I've just run black over the Briefcase codebase, and the changes are 80% `'`->'"' by volume; there are a handful of other changes that grate against me a little bit, but not as much as the overhead of having to review code format :-) (or, in some cases, can be made palatable by making a small tweak to syntax).

The biggest impediment to "just doing it" is working out what black looks like from a CI and repo config perspective. If we're going to adopt black, it needs to be applied automatically without any explicit interaction from users (because I don't want to adopt a tool and then spend every PR review saying "Could you run black over this code"). Investigating what 'best practice" looks like is the next step on this journey.

src/briefcase/integrations/wix.py
danyeaw: Updated!

src/briefcase/integrations/xcode.py
danyeaw: Nice catch!

danyeaw:
PR updated, thanks for the feedback!

> The biggest impediment to "just doing it" is working out what black looks like from a CI and repo config perspective. 
Mind if I submit a PR? I can set things up with pre-commit and black.

freakboy3742:
> Mind if I submit a PR? I can set things up with pre-commit and black.

Go right ahead! Make sure the PR has any docs (either in the PR, or ideally part of the contribution guide) to describe what contributors need to do to adapt.

freakboy3742 (approved):
Awesome - thanks for the cleanups!

PR #744, opened 2022-05-25
Enable Black and other pre-commit hooks

danyeaw:
Enables pre-commit hooks for:

- Black
- isort
- flake8
- docformatter
- pyupgrade

Adds a .git-blame-ignore-revs file for keeping git blame history. Removes flake8 from tox since it is part of the pre-commit hook. Updates the CI to include the pre-commit hooks as part of the beefore step.

## PR Checklist:


.git-blame-ignore-revs
```diff
@@ -0,0 +1,5 @@
+# Migrate code style to Black
```
freakboy3742, line 1: I *think* I understand what this file is doing... but just to be sure, can you clarify for me?

.github/workflows/ci.yml
```diff
@@ -30,6 +29,8 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: "3.9"
+    - name: Lint with Pre-commit
+      uses: pre-commit/action@v2.0.3
```
freakboy3742, line 33: Does this modify the code, or just run flake/black/etc checks and report an error if they haven't been run? If it *does* modify - who is the author of the commit?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,30 @@
...
+    rev: v4.2.0
+    hooks:
+      - id: check-toml
+      - id: check-yaml
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        additional_dependencies: [toml]
+  - repo: https://github.com/asottile/pyupgrade
```
freakboy3742, line 22: I've never used pyupgrade before - do we need to configure it so that it knows our minimum supported language? Or does it pick that up from setup.cfg?

CONTRIBUTING.md
```diff
@@ -6,5 +6,15 @@ Please be aware, BeeWare operates under a Code of Conduct.
...
 
-See also the Briefcase specific contributing guides for contributing [code](https://briefcase.readthedocs.io/en/latest/how-to/contribute-code.html) and [documentation](https://briefcase.readthedocs.io/en/latest/how-to/contribute-docs.html)
+See also the Briefcase specific contributing guides for contributing
+[code](https://briefcase.readthedocs.io/en/latest/how-to/contribute-code.html)
+and
+[documentation](https://briefcase.readthedocs.io/en/latest/how-to/contribute-docs.html)
 
+## Development Environment
+
+Run `pip install pre-commit` to install pre-commit hooks in to your git hooks.
```
freakboy3742, line 16: What happens if a user *doesn't* do this? (Because you can guarantee 90% of first time contributors won't, even if it *is* in the documentation...)

freakboy3742 (changes requested):
Looks fairly straightforward - although there's already a merge conflict because of the formatting changes from #743 :-)

A couple of questions/clarifications inline about the specifics on how will play out in practice.

.git-blame-ignore-revs
danyeaw: `git blame --ignore-revs-file .git-blame-ignore-revs` can be used to ignore large changes like our conversion to black and docformatter to prevent "ruining" the git history. It also can be set a global option in git itself with `git config --global blame.ignoreRevsFile .git-blame-ignore-revs`.

.github/workflows/ci.yml
danyeaw: Exits with an error if pre-commit would need to modify a file.

danyeaw:
> Looks fairly straightforward - although there's already a merge conflict because of the formatting changes from https://github.com/beeware/briefcase/pull/743 :-)

Merge conflict fixed, I rebased it!

CONTRIBUTING.md
```diff
@@ -6,5 +6,15 @@ Please be aware, BeeWare operates under a Code of Conduct.
 
 See [CONTRIBUTING to BeeWare](http://beeware.org/contributing) for details.
 
-See also the Briefcase specific contributing guides for contributing [code](https://briefcase.readthedocs.io/en/latest/how-to/contribute-code.html) and [documentation](https://briefcase.readthedocs.io/en/latest/how-to/contribute-docs.html)
+See also the Briefcase specific contributing guides for contributing
+[code](https://briefcase.readthedocs.io/en/latest/how-to/contribute-code.html)
+and
+[documentation](https://briefcase.readthedocs.io/en/latest/how-to/contribute-docs.html)
 
+## Development Environment
```
freakboy3742, line 14: This note should be added to the contribution guide that is linked on L10, rather than here.

If I'm reading the pre-commit docs right, we also need to instruct users to run `pre-commit install`; just installing pre-commit won't install the hooks in a way that they'll actually run.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - repo: https://github.com/PyCQA/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+        additional_dependencies: [flake8-eradicate==1.2.0]
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.2.0
+    hooks:
+      - id: check-toml
+      - id: check-yaml
```
freakboy3742, line 16: It would be worth auditing [the full list of available hooks](https://pre-commit.com/hooks.html) to see if there's anything else useful there. Other projects (like [pyscript](https://github.com/pyscript/pyscript/blob/main/.pre-commit-config.yaml) or [pip](https://github.com/pypa/pip/blob/main/.pre-commit-config.yaml)) may be a good reference here.

CONTRIBUTING.md
freakboy3742: We may also want to add some notes about what to expect if a local commit fails. AFAICT, the process is: 
* attempt a commit
* tests will run (show a console dump showing what to expect)
* Any files with formatting problems will have the black fixes applied, but unstaged
* Manually add those files, recommit.
* tests will run an pass (show a successful console dump)

freakboy3742 (changes requested):
I've had a chance to play around with pre-commit; I'm not in love with the onboarding experience (i.e., there's no way to signal to developers that they're missing something in their config) - but I'm not sure if there's anything we can do about that.

I've flagged two specific places that need some more attention in the specific configuration of pre-commit.

CONTRIBUTING.md
danyeaw: Done!

.pre-commit-config.yaml
danyeaw: Added a few more for trailing whitespace, newline for end-of-file, etc. :+1:

.github/workflows/ci.yml
danyeaw: Done, moved pre-commit to the first action run, then beefore follows. Another option is, we could make use of https://pre-commit.ci. An advantage is it runs faster and automatically creates PRs with the pre-commit hooks are out of date. This could be a future thing we look at, though.

.pre-commit-config.yaml
danyeaw: One additional one that would be good would be to add mypy as a pre-commit hook, even if we aren't planning on or planning on adding incremental type hints. It's analysis found errors, and we could resolve these and make it part of a future PR.

danyeaw:
Updates made! Hopefully I don't have to rebase this one again :pray: :smile:

freakboy3742 (approved):
I've made a couple of documentation tweaks and similar cosmetic cleanups, but otherwise, this looks good to go. There's a couple more pre-commit checks that might be worth integrating - but given the overhead associated with maintaining this patch (and the impact it has on other patches), I think it's worth landing as-is, and dealing with those additions as follow-on work.

Thanks for the prod (and the legwork) to make this happen!

danyeaw:
Thanks @freakboy3742 ! 🙌

Issue #528, opened 2020-11-15
ValueError when executing `briefcase run` with specific device (`-d`) for iOS


The-Judge:
**Describe the bug**
When launching `briefcase run iOS -d "iPhone 11"` (as suggested in https://docs.beeware.org/en/latest/tutorial/tutorial-5/iOS.html), the following happens:

```
$ briefcase run iOS -d "iPhone 11"

[helloworld] Generating application template...
Using app template: https://github.com/beeware/briefcase-iOS-Xcode-template.git
Using existing template (sha 185a566d3a5829784d5fd5b3e0273573b3d29de4, updated Sun Jun 21 12:41:13 2020)

[helloworld] Installing support package...
Using support package https://briefcase-support.org/python?platform=iOS&version=3.8
... using most recent revision
Python-3.8-iOS-support.b3.tar.gz already downloaded
Unpacking support package...

[helloworld] Installing dependencies...
Collecting toga-iOS>=0.3.0.dev20
  Using cached toga_iOS-0.3.0.dev25-py3-none-any.whl (34 kB)
Collecting toga-core==0.3.0.dev25
  Using cached toga_core-0.3.0.dev25-py3-none-any.whl (484 kB)
Collecting rubicon-objc>=0.3.0
  Using cached rubicon_objc-0.4.0-py3-none-any.whl (53 kB)
Collecting travertino>=0.1.3
  Using cached travertino-0.1.3-py3-none-any.whl (15 kB)
Installing collected packages: travertino, toga-core, rubicon-objc, toga-iOS
Successfully installed rubicon-objc-0.4.0 toga-core-0.3.0.dev25 toga-iOS-0.3.0.dev25 travertino-0.1.3

[helloworld] Installing application code...
Installing src/helloworld...

[helloworld] Installing application resources...
Unable to find src/helloworld/resources/helloworld-20.png for 20px application icon; using default
Unable to find src/helloworld/resources/helloworld-29.png for 29px application icon; using default
Unable to find src/helloworld/resources/helloworld-40.png for 40px application icon; using default
Unable to find src/helloworld/resources/helloworld-58.png for 58px application icon; using default
Unable to find src/helloworld/resources/helloworld-60.png for 60px application icon; using default
Unable to find src/helloworld/resources/helloworld-76.png for 76px application icon; using default
Unable to find src/helloworld/resources/helloworld-80.png for 80px application icon; using default
Unable to find src/helloworld/resources/helloworld-87.png for 87px application icon; using default
[...]

PR #745, opened 2022-05-25
fix: Trim simulator os when sorting for version

ChillarAnand:
`get_simulators` is returning `OS Version` and it is being used as a key for sorting which is causing an issue. Splitting it and using version number for sorting.

Updated `get_simulators` return value in tests.

Closes https://github.com/beeware/briefcase/issues/528

## PR Checklist:

### DIFF
--- a/changes/528.bugfix.rst
+++ b/changes/528.bugfix.rst
@@ -1,0 +1,1 @@
+Corrected a problem with the iOS simulator detecting the iOS version when only a device name is provided.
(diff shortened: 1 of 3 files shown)


freakboy3742 (approved):
Thanks for the fix! I've made a couple of minor modifications to ensure that case sensitivity is preserved, but this was a really helpful starting point!

As an aside - the CI check that was failing was "towncrier" - that's the tool that we use to auto-compile release notes. Every PR needs to include a file in the `changes` folder that describes what the PR does. I've added one for this PR; but if you submit other PRs in future (and I hope you will!) you'll need to do this to see the full results of the test suite.

PR #751, opened 2022-05-31
Improve `parse_config` error message

TG-Techie:
### Describe your changes in detail:
Improve an error message that seem to trip people up when starting (including me). This provides the developer with more information about why the command failed.  (see diff)

### Rational
I came back to beeware after several year and while following the instructions tripped up by running `briefcase create` in the wrong folder. The "From the helloworld directory, run:" line is quite easy to miss in the [Packaging for Distribution](https://docs.beeware.org/en/latest/tutorial/tutorial-3.html) guide.

examples of other people having the same issue:
- https://gitter.im/beeware/general?at=5e782503846f9b7ca61a3fc5
- https://github.com/beeware/briefcase/issues/553

## PR Checklist:
- [ ] All new features have been documented (N/A)

### DIFF
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -452,7 +452,10 @@ class BaseCommand(ABC):
                     )
 
         except FileNotFoundError as e:
-            raise BriefcaseConfigError("configuration file not found") from e
+            raise BriefcaseConfigError(
+                f"configuration {filename!r} file not found, "
+                "it may be in a different directory or not exist."
+            ) from e
 
     def download_url(self, url, download_path):
         """Download a given URL, caching it. If it has already been downloaded,


freakboy3742 (changes requested):
Thanks for the suggestion! Agreed that this is an error that comes up semi-regularly, and could be assisted with a better error message.

However, I'd suggest a better error message would be one that provides a "call to action", rather than restating what "not found" means. For example:

    "Configuration file not found. Did you run briefcase in the directory that contains your {filename!r}?"

This gives the root cause of the error, but also suggests the likely cause and remedy.

You'll also need to [add a change note](https://briefcase.readthedocs.io/en/latest/how-to/contribute-code.html#add-change-information-for-release-notes) to pass CI, as well as any tests that fail as a result of making this change (I would be surprised if there isn't at least one test failure).

freakboy3742:
Closing due to lack of activity.

Issue #395, opened 2020-05-17
Re-installing an Android app sometimes fails


paulproteus:
@freakboy3742 ran into an issue where, if the device is clean, it starts fine.

However, if the app already exists on the device, it fails:

```
Unable to install APK /Users/rkm/projects/beeware/demo/traveltips/android/Travel Tips/app/build/outputs/apk/debug/app-debug.apk on KAPHN4G9AA1513
```

I can't reproduce this yet, but I'd like to get to the bottom of it.

freakboy3742:
Some more details:
* I'm running TravelTips.
* `briefcase run android` first run builds and starts the app fine
* Second run (no rebuild - just run) fails with the error in the original report.
* Looking at the task list, it looks like the app has been partially stopped: it's still on the task list, but if you pick it from the task list it boots again.
* If you manually kill the app, and re-run; the same error occurs.
* If you manually *delete* the app, and re-run, it re-installs and runs fine.
* The whole process works fine on the emulator. It only fails on my physical device (a Kogan Agora 9)

sarthaksharma058439:
Check this video .....he make calculator in beeware and deploy to android device

https://youtu.be/RllR6FRdG6w

PR #752, opened 2022-06-01
Fixed re-installation of Android apps

mhsmith:
I reproduced #395, and saw this line in the device log:
```
2022-06-01 14:04:09.460 768-800/? W/PackageManager: Attempt to re-install com.example.helloworld without first uninstalling.
```

This is caused by missing the `-r` option of [`adb install`](https://developer.android.com/studio/command-line/adb#pm), which enables updating an existing app. From my tests, the emulator doesn't always require this option, but physical devices do.

If the app doesn't already exist on the device, the option is harmless, so I've made briefcase use it unconditionally.

## PR Checklist:

### DIFF
--- a/changes/395.bugfix.rst
+++ b/changes/395.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed re-installation of Android apps by adding ``-r`` option to ``adb install``.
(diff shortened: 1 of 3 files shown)


mhsmith:
The CI faliure on Python 3.11 was caused by an issue in 3.11.0b2, which was released yesterday (https://github.com/pytest-dev/pytest/issues/10008). It should be fixed in 3.11.0b3, which [will be released very soon](https://mail.python.org/archives/list/python-dev@python.org/thread/VZ222BFFAWW73DYFCR7NZR2C5M3TXD65/).

danyeaw (approved):
Nice one @mhsmith!

Issue #718, opened 2022-04-22
Library issues with Linux AppImage observed at runtime


freakboy3742:
A large number of reports have been received (#458, #470, #481, #504, #583, #618, #625, #631, #692) relating to AppImage support on Linux. This is a consolidation ticket to cover the general problem of Linux AppImage support.

**To Reproduce**

Reports are most common when using third party binaries (e.g., Pillow), or GUI libraries that have third party binaries (e.g, PySide2/6). However, problems have also been observed with Toga apps when deploying on recent versions of Linux (e.g., using Ubuntu 22.04)

The problem manifests in one of two ways.

A failure to find dependencies during build time:
```
-- Deploying dependencies only for ELF files -- 
Deploying files in directory /app/Hello World/Hello World.AppDir/usr/app_packages/PySide2 
WARNING: Not an ELF file: /app/Hello World/Hello World.AppDir/usr/app_packages/PySide2/_config.py 
WARNING: Not an ELF file: /app/Hello World/Hello World.AppDir/usr/app_packages/PySide2/QtScxml.pyi 
WARNING: Not an ELF file: /app/Hello World/Hello World.AppDir/usr/app_packages/PySide2/QtCore.pyi 
Deploying dependencies for ELF file in AppDir: /app/Hello World/Hello World.AppDir/usr/app_packages/PySide2/QtXmlPatterns.abi3.so 
Deploying dependencies for ELF file /app/Hello World/Hello World.AppDir/usr/app_packages/PySide2/QtXmlPatterns.abi3.so 
ERROR: Could not find dependency: libshiboken2.abi3.so.5.13 
```

Missing dependencies at runtime:
```
>>> /home/criaz/repos/briefcase-613/hello/linux/hello-0.0.1-x86_64.AppImage
Traceback (most recent call last):
  File "/tmp/.mount_hello-9Pm4Um/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/tmp/.mount_hello-9Pm4Um/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/tmp/.mount_hello-9Pm4Um/usr/app/hello/__main__.py", line 4, in <module>
    main().main_loop()
  File "/tmp/.mount_hello-9Pm4Um/usr/app/hello/app.py", line 27, in main
    return hello()
[...]

Issue #628, opened 2021-10-19
Add a stub app for Linux


freakboy3742:
We recently modified the macOS template to provide a stub app - a small binary executable that is used as an entry point for the app, which is responsible for starting the Python interpreter.

This has a number of benefits:

 * The app appears in the process monitor with the app name and icon, rather than "python" or a shell file
 * The app that is started is the same one that is granted permissions by the operating system, rather than needing to fork a process
 * It provides a hook to catch high-level app failures, such as a fail on start due to import errors. If the app fails, we can catch the failure and show a crash dialog.
 * It provides a hook for installing logging and other system-level handlers.

**Describe the solution you'd like**

The [briefcase-linux-appimage-template](https://github.com/beeware/briefcase-linux-appimage-template) should be modified to contain a stub app that behaves in a similar way to the macOS app template, showing a dialog if the Python code crashes.

This will involve writing a small C app in GTK that starts the Python interpreter. The [implementation in the macOS template](https://github.com/beeware/briefcase-macOS-Xcode-template/blob/dev/%7B%7B%20cookiecutter.formal_name%20%7D%7D/%7B%7B%20cookiecutter.formal_name%20%7D%7D/main.m) will be a useful reference; although that code is Objective C, and uses the Cocoa APIs, the structure of code needed to start the Python interpreter will be largely the same.

The macOS template has 2 versions - a pre-built version where the binary is copied and renamed as needed; and an "Xcode project" version where the user can modify the app if they wish. In the context of Linux, it should be acceptable to just use the "project" version, because it's easy to get the compiler toolchain in place. However, we need to document those extra system requirements, and add them to the Docker build environment.

freakboy3742:
For context - we've added a stub app for the flatpak backend; there doesn't appear to be any particular utility in adding the equivalent for the Appimage backend.

PR #754, opened 2022-06-03
Implement Linux/Flatpak output for Briefcase

nickzoic:
Following [this comment by Russell](https://github.com/beeware/briefcase/issues/718#issuecomment-1128444577) I thought I'd have a look into packaging as  Flatpak.  This is kind of a workaround for #718 maybe, but also having support for multiple Linux packaging solutions is a good thing in its own right.

Early days yet and this is very much a work in progress.
It builds the helloworld app and that's about it!

Much more work needs to be done to generate a good Flatpak manifest from the information in the pyproject.toml file but I thought it would be nice to have a PR for this work to live in.

## PR Checklist:


nickzoic:
Next steps:

* Generate more information for the FlatPak manifest directly from pyproject.toml
  * system_requires?
* generate a requirements.txt from the requires sections for use by flatpack-pip-generator
* add check / installer for tools (flatpak, flatpak-builder, flatpack-pip-generator)
* Eliminate flatpak-pip-generator so manifest is created directly from pyproject.toml
  * it'd be helpful if this was a python library not an executable
  * an executable tool would be helpful for non-briefcase flatpak users though.

danyeaw:
@nickzoic I am really excited about this feature!

> Eliminate flatpak-pip-generator so manifest is created directly from pyproject.toml

When we were developing Flatpak support in Gaphor, we didn't have much luck with getting reliable builds using flatpak-pip-generator. Instead we created a [simple bash script](https://github.com/flathub/org.gaphor.Gaphor/blob/master/depends.sh) that pip installed the needed packages and then put them in to a manifest in the right format. You might be able to take some inspiration from this and implement something similar in Python.

nickzoic:
> When we were developing Flatpak support in Gaphor, we didn't have much luck with getting reliable builds using flatpak-pip-generator.

Yeah, I don't know if it'll prove to be a good approach.  I just hand-created the `requirements.txt` for now, but had some issues with versioning I think, where the local build had 0.3.0dev34 and the flatpak build had 0.2.15 and that wasn't quite right, but hopefully that's resolvable.

I think if we were to use the flatpak-pip-generator -like approach we'd have to mostly rewrite it but it's a good start on manifest generation & I went with that for now in an attempt to get something at least partially working.

I'll have a look at your script too, thanks!  Looks a lot simpler than what it's currently doing ...

src/briefcase/platforms/linux/flatpak.py
```diff
@@ -1 +1,232 @@
...
+#
+# Issues:
+# * the flakpak-builder manifest file is either json or yaml, it'll parse JSON
+#   from a tempfile or /dev/stdin but it interprets include paths relative so
+#   use absolute paths if that's what you want to do.
+# * flatpak-builder is in C (!)
+# * flatpak-pip-generator is a python program which translates requirements.txt
+#   files into includable "python3-requirements.json" module *but* it isn't 
+#   distributed on pypi, isn't a library and doesn't load pyproject.toml so I
+#   suspect we're better off doing this stuff here instead.
```
freakboy3742, line 51: What Python version is being used here? One of the implied contracts of Briefcase packaging is that it uses the same Python version as the one doing the packaging. If I package this app on Ubuntu 22.04 (which uses Python 3.9 as a system Python IIRC), and then deploy on Fedora 28 which uses Python 3.6... what happens? What happens if my code uses the walrus operator? I'm not seeing anything in this flatpak definition that references a Python version.

nickzoic: Yep, the python version comes from the flatpak "runtime", which is kind of a base image for the flatpak, itself also a flatpak.

The runtime is specified in the flatpak manifest https://github.com/beeware/briefcase/pull/754/files#diff-c3953738384a9db4f1f0e7c5000448f194a3906c3260fd510190dfb02a05d887R131-R136 ...
I'm just using an arbitrary one, but I'd considered using a --option to let the user choose one.

We could also pick the appropriate runtime to match the python version being used for packaging, plus it doesn't look too difficult to create custom runtimes based on the existing ones so we could make & publish our own ones per python version (though I haven't tried this yet)

freakboy3742: What's the line between "tool provided by the Runtime" and "binary dependency required to make the app run"?

In particular - how would the introduction of something like #628 interact with the choice of runtime?

freakboy3742 (changes requested):

src/briefcase/platforms/linux/flatpak.py
```diff
@@ -1 +1,215 @@
...
+                        {"type": "dir", "path": str(Path("src").resolve())},
+                    ],
+                },
+                str(self.requirements_manifest_file(app).resolve()),
+            ],
+            "finish-args": [
+                "--socket=fallback-x11",
+                "--socket=wayland",
+            ],
+        }
```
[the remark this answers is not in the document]
nickzoic: Could be templated instead (like linux/appimage is) but this was the quickest & simplest way to get a first implementation up and running as a proof of concept.

src/briefcase/platforms/linux/flatpak.py
```diff
@@ -1 +1,215 @@
...
+
+    def build_app(self, app: BaseConfig, **kwargs):
+        """Build an application.
+
+        :param app: The application to build
+        """
+        self.logger.info()
+        self.logger.info(f"[{app.app_name}] Building Flatpak...")
+        try:
+            self.subprocess.run(
```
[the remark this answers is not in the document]
nickzoic: Yeah, maybe ... as you say, that way the dockerized build system is a known quantity.  And have the same --no-docker flag as linux/appimage.

Issue #662, opened 2022-01-28
AppImage loads Python code from project's sources, not from inside of AppImage


jhakonen:
**Describe the bug**
Running built AppImage loads Python sources from source tree instead of loading them from inside of the AppImage. This appears to happen when project's Python package is at project's root, and not in `src` folder. 

You may get this abnormal folder structure if you create your project as Poetry project (with `poetry new`) and later on migrate to Briefcase.

**To Reproduce**
1. Create new Toga project using default answers with:
```bash
briefcase new
cd helloworld
```
2. Move `helloworld` package to root level:
```bash
mv src/helloworld .
rmdir src
sed -i 's|src/helloworld|helloworld|' pyproject.toml
```
3. Then build AppImage:
```bash
briefcase build
```
4. Run the built AppImage:
```bash
linux/Hello_World-0.0.1-x86_64.AppImage
```
5. You should see this:
![helloworld](https://user-images.githubusercontent.com/1950698/151598132-9881fc3b-10e9-4ce0-82bd-6a8bf843c341.png)
6. Next in `helloworld/app.py` change window title from:
```python
self.main_window = toga.MainWindow(title=self.formal_name)
```
To:
```python
self.main_window = toga.MainWindow(title='You cannot see me!')
```
7. Finally run the AppImage again. Do not build it, just run the previously built AppImage:
```bash
linux/Hello_World-0.0.1-x86_64.AppImage
```
8. The title has changed, even though it shouldn't have:
![changed-title](https://user-images.githubusercontent.com/1950698/151598763-88432791-d126-40a6-b1ea-5f73df2c5b66.png)

**Expected behavior**
The title should have remained as "Hello World" as I didn't rebuild the AppImage.

**Environment:**
 - Operating System: Pop!_OS 20.04
 - Python version: 3.8.10
 - Software versions:
   - Briefcase: 0.3.5
   - Toga: 0.3.0.dev30

**Additional context**
This bug appears to depend on current working directory. If you cd to parent directory then the AppImage will load its sources correctly from inside of it:
```bash
cd ..
helloworld/linux/Hello_World-0.0.1-x86_64.AppImage
```

freakboy3742:
That's *very* weird.... thanks for the report.

The last bit of Additional Context is essentially the explanation of what is going on.

[...]

PR #756, opened 2022-06-05
Fix GTK Namespace Not Found Error in Linux

danyeaw:
## Summary
This PR fixes a portion of #718 and goes along with https://github.com/linuxdeploy/linuxdeploy-plugin-gtk/pull/33.

The issue is caused by missing typelib libraries for GTK which are not copied by linuxdeploy. The solution is to use the linuxdeploy-plugin-gtk to get the proper libraries installed by GTK. Although packaging GTK apps in older distributions worked, I think that was actually a case of AppImage bleed through to the host system libraries like in https://github.com/beeware/briefcase/issues/662, to be confirmed.

## PR Checklist:

## Triage of the Issue

When using the helloworld tutorial using a newer version of Linux, executing `briefcase run` results in the following error:

```bash
$ ~/P/b/helloworld briefcase run

[helloworld] Starting app...

  File "/tmp/.mount_Hello_um2Tym/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/tmp/.mount_Hello_um2Tym/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/tmp/.mount_Hello_um2Tym/usr/app/helloworld/__main__.py", line 4, in <module>
    main().main_loop()
  File "/tmp/.mount_Hello_um2Tym/usr/app/helloworld/app.py", line 27, in main
    return HelloWorld()
  File "/tmp/.mount_Hello_um2Tym/usr/app_packages/toga/app.py", line 272, in __init__
    self.factory = get_platform_factory(factory)
  File "/tmp/.mount_Hello_um2Tym/usr/app_packages/toga/platform.py", line 41, in get_platform_factory
    from toga_gtk import factory
  File "/tmp/.mount_Hello_um2Tym/usr/app_packages/toga_gtk/factory.py", line 1, in <module>
    from .app import App, DocumentApp, MainWindow
  File "/tmp/.mount_Hello_um2Tym/usr/app_packages/toga_gtk/app.py", line 13, in <module>
    from .keys import gtk_accel
  File "/tmp/.mount_Hello_um2Tym/usr/app_packages/toga_gtk/keys.py", line 3, in <module>
    from .libs import Gdk
  File "/tmp/.mount_Hello_um2Tym/usr/app_packages/toga_gtk/libs/__init__.py", line 1, in <module>
    from .gtk import *  # noqa: F401, F403
  File "/tmp/.mount_Hello_um2Tym/usr/app_packages/toga_gtk/libs/gtk.py", line 3, in <module>
    gi.require_version('Gtk', '3.0')
  File "/tmp/.mount_Hello_um2Tym/usr/app_packages/gi/__init__.py", line 126, in require_version
    raise ValueError('Namespace %s not available' % namespace)
ValueError: Namespace Gtk not available
```

## Research in to Linuxdeploy

As part of developing this PR, I read the docs for linuxdeploy and added an overview of what it is to the appimage docs. After I [pointed out](https://github.com/beeware/briefcase/issues/718#issuecomment-1126983063) the existence of the [linuxdeploy gtk plugin](https://github.com/linuxdeploy/linuxdeploy-plugin-gtk) a couple of weeks ago, I investigated how the main bash script works.

@nickzoic confirmed that he got the plugin to work. Although I was successful at getting the GTK plugin loaded, I was still getting the same Namespace Not Available message. I was somewhat expecting this result because the Namespace unavailable message indicates that Python is not able to load PyGObject introspection itself, not that libraries are missing.

## Troubleshooting PyGObject

```
$ cd linux/appimage/Hello\ World/Hello\ World.AppDir/bin
$ ./python3 -c "import gi"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'gi'
```
That's not good, gi can't be imported at all. I referenced the environmental variables here: https://docs.python.org/3/using/cmdline.html#environment-variables.
Let's try to add it manually by setting the PYTHONPATH:

```
$ PYTHONPATH=../app_packages ./python3 -c "import gi; from gi.repository import Gtk"
  File "<string>", line 1, in <module>
  File "/home/dan/Projects/beeware-practice/helloworld2/linux/appimage/Hello World/Hello World.AppDir/usr/app_packages/gi/importer.py", line 136, in load_module
    raise ImportError('cannot import name %s, '
ImportError: cannot import name Gtk, introspection  not found
```

This gives a hint that we are missing the typelib because the introspection is not found. Typelibs provide the namespace for GIRepository, so it makes sense that without them that we would be getting a namespace error. This error was being somewhat masked when we use gi.require_version('Gtk', '3.0') when the app normally loads.

The [briefcase-linux-appimage-template](https://github.com/beeware/briefcase-linux-appimage-template) and the Linux support package that goes along with provide Python and the AppDir structure only, and this is set up when you run `briefcase create`. They don't include any of the other system libraries that are added in when you run `briefcase build`. So missing typelibs is caused by linuxdeploy not bundling these extra dependencies. The only supported way to add in extra dependencies with linuxdeploy
is by creating a plugin.

In Linux, pkg-config is used to determine the library location without having to know where each distro installs each library. To find the location of the GTK typelibs for introspection we run:
```
$ pkg-config --variable typelibdir gobject-introspection-1.0
/usr/lib64/girepository-1.0
```
In Ubuntu, it would be in a different location: `/usr/lib/x86_64-linux-gnu/girepository-1.0/`
The linuxdeploy gtk plugin already finds other libraries, so to add the typelibs I added a new tree copy of the typelibs that installs them to the right location in the APPDIR.

## Other Solutions Considered

1. Implement the scripts to copy the typelibs directly in Briefcase. Although this would be possible, I think the implementation of how to import libraries to build an AppImage should stay with linuxdeploy.
2. Switch the default package format for Linux to Flatpak. Longer term this might be a good strategy, and @nickzoic already started working on it https://github.com/beeware/briefcase/pull/754, which is great!
3. Use PyInstaller to freeze the libraries and then package the app in to an AppImage. This might another possible solution for the future since it has a great set of import hooks which solve finding libraries for a wide set of dependencies. We would have to consider if we want to add another dependency to Briefcase, and it would be a large change to our AppImage implementation.

## Final Solution

You can now add a new `linuxdeploy_plugins` configuration in the linux portion of the config in the `pyproject.toml`, in order to have linuxdeploy load extra resources in to the AppImage bundle. This is needed for applications that depend on libraries that cannot be automatically discovered by linuxdeploy like GTK. For example,

    linuxdeploy_plugins = ['gtk']

Briefcase can take plugins in three different formats:

1. A plugin provided by Briefcase. Currently ``gtk`` is supported.
2. A URL to a plugin, for example  'https://github.com/linuxdeploy/linuxdeploy-gtk-plugin/master/linuxdeploy-gtk-plugin.sh'
3. A path to a plugin relative to the ``pyproject.toml`` file, like    'packaging/linuxdeploy-gtk-plugin.sh'

Some plugins need an environmental variable passed to configure it. For example, the ``gtk`` plugin needs the ``DEPLOY_GTK_VERSION`` environmental variable set to 3 or 4. If you use the ``gtk`` plugin provided by Briefcase,
it is set to 3 by default, but you can override it or set it for other plugins by setting the ``DEPLOY_GTK_VERSION`` environmental variable. For example, when using a relative path, you can set it like this:

    linuxdeploy_plugins = ['DEPLOY_GTK_VERSION=3 packaging/linuxdeploy-gtk-plugin.sh']


nickzoic:
I gave this a quick go with the default 'Hello World' app and once I added `'libgtk-3-dev', 'librsvg2-dev',` to the system_requires it worked first time!

danyeaw:
@nickzoic Thanks for giving it a test run and for all the groundwork you did on this issue!

Are you getting an error when you don't update the system_requires? It shouldn't need the extras now because I updated the gtk plugin to use default Ubuntu paths if those development libraries aren't available.

nickzoic:
No worries ... yeah, on e15d783f I was getting errors
 (missing `libgtk+-3.0.pc`, missing `librsvg-2.0.pc`,see below for details) at `briefcase build` time.  The plugin knows it wants them but they're not in the docker image.

The host is Ubuntu 20.04.4 but the libs were missing from the docker container so not sure if that matters or why it'd work differently since the container is always 18.04.

I had to add those -dev packages and redo `briefcase create` to get them into the docker container, after that the build worked.

```
$ briefcase build

[helloworld] Building AppImage...
Using linuxdeploy GTK plugin

[helloworld] Entering Docker context...
linuxdeploy version 1-alpha (git commit ID 097212a), GitHub actions build 80 built on 2022-05-03 17:10:24 UTC

-- Creating basic AppDir structure -- 
```
... Creating, Deploying, etc ...
```
-- Running input plugin: gtk -- 
[gtk/stdout] Installing AppRun hook
[gtk/stdout] Installing GLib schemas
[gtk/stdout] Installing GTK 3.0 modules
[gtk/stderr] Package gtk+-3.0 was not found in the pkg-config search path.
[gtk/stderr] Perhaps you should add the directory containing `gtk+-3.0.pc'
[gtk/stderr] to the PKG_CONFIG_PATH environment variable
[gtk/stderr] No package 'gtk+-3.0' found
[gtk/stderr] /home/brutus/.briefcase/tools/linuxdeploy-plugin-gtk.sh: there is no 'exec_prefix' variable for 'gtk+-3.0' library.
[gtk/stderr] Please check the 'gtk+-3.0.pc' file is present in $PKG_CONFIG_PATH (you may need to install the appropriate -dev/-devel package).
ERROR: Failed to run plugin: gtk (exit code: 1) 

Error while building app helloworld.
```

The `pyproject.toml` as created by `briefcase new` has:
```
system_requires = [
    'libgirepository1.0-dev',
    'libcairo2-dev',
    'libpango1.0-dev',
    'libwebkitgtk-3.0-0',
    'gir1.2-webkit-3.0',
]
```

danyeaw:
@nickzoic Did you try to delete `~/.briefcase/tools/linuxplugin-plugin-gtk.sh` to make sure a new one is pulled down?

nickzoic:
I hadn't thought of that!  I'd just been deleting the build directories, I'll give it a go ... oh no, it's not downloading correctly for me now, I end up with a linuxdeploy-plugin-gtk.sh full of HTML!  Looks like it failed in a way that wasn't an exception, I'll try and work out why.

I do think the idea of having state outside the build directory is fraught, though ... it's bound to come back and bite us at a later date.

danyeaw:
@nickzoic Whoops, thanks for helping to test and for all your contributions providing the basis of this PR in the first place!

I updated the download URL to fix the issue :+1:

docs/reference/platforms/linux/appimage.rst
```diff
@@ -41,6 +41,43 @@ AppImage:
    using Ubuntu 19.10, for example, you can expect that only people on the most
    recent versions of another distribution will be able to run your AppImage.
 
+Build Process
+=============
```
freakboy3742, line 45: It's definitely worth documenting this from a "how it works" perspective, but I wonder about the specific placement here. The information that will be more directly applicable are the details about Docker usage, Icon naming, and so on; this block is more in the camp of "something has gone wrong/this is what we're doing so that you can debug" territory. I think it might be better served by being at the *end* of this document, as part of an "Implementation details" section.

docs/reference/platforms/linux/appimage.rst
```diff
@@ -41,6 +41,43 @@ AppImage:
...
+In order to support GTK based apps like Toga in Linux, Briefcase also uses the
+`linuxdeploy-gtk-plugin
+<https://github.com/linuxdeploy/linuxdeploy-gtk-plugin>`__.
+This plugin goes through a series of steps to make sure that all the libraries
+are pulled in for the app to work properly once in the AppImage. The plugin
+goes through the following steps including:
+
+1.  Detects the GTK version to use, we manually set the version to 3
+2.  Installs itself as a hook in $APPDIR/apprun-hooks
+3.  Uses gsettings to set a light or dark adwaita theme
```
freakboy3742, line 69: I can see this is part of the plugin itself - but: what is the implication of this on the apps we distribution? Are all Briefcase GTK apps going to use the `adwaita` theme, no matter the user's local theme?

src/briefcase/platforms/linux/appimage.py
```diff
@@ -156,52 +157,69 @@ def build_app(self, app: BaseConfig, **kwargs):
...
+        :param app: The application to build
+        """
+
+        # For some reason, the version has to be passed in as an
+        # environment variable, *not* in the configuration...
+        env = {"VERSION": app.version}
+        plugins = []
+        if app.requires:
+            for dependency in app.requires:
+                if "toga-gtk" in dependency or "PyGObject" in dependency:
```
freakboy3742, line 178: Will this catch:
* `toga-gtk>=0.3.0.dev33`? (i.e., a version string on the dependency)?
* `pygobject` (i.e., case incompatibility?)
* An app that uses a different toolkit that wraps GTK, but doesn't directly import pygobject? (not sure I have any specific examples in mind, but I'm sure *someone* has built one)
?

On the last example - is there a reason you opted for dependency inspection rather than an explicit list of linuxdeploy plugins needed by the app?

src/briefcase/platforms/linux/appimage.py
```diff
@@ -156,52 +157,69 @@ def build_app(self, app: BaseConfig, **kwargs):
...
+                    self.logger.info("Using linuxdeploy GTK plugin")
+                    env["DEPLOY_GTK_VERSION"] = "3"
```
freakboy3742, line 180: Is this envvar something we can reliably set in code? What happens if someone uses GTK4?

src/briefcase/integrations/linuxdeploy.py
```diff
@@ -119,3 +119,71 @@ def patch_elf_header(self):
                 # The original or patched value. If this is the case, the file is likely
                 # wrong and we should raise an exception.
                 raise CorruptToolError("linuxdeploy")
+
+
+class LinuxDeployGtkPlugin:
```
freakboy3742, line 124: Is there any potential for a *generic* plugin interface here? I noticed there is a Qt plugin - it seems likely we may need to use that for PySide2/6 deployments

src/briefcase/integrations/linuxdeploy.py
```diff
@@ -119,3 +119,71 @@ def patch_elf_header(self):
...
+    name = "linuxdeploy_gtk_plugin"
+    full_name = "linuxdeploy_gtk_plugin"
+
+    def __init__(self, command):
+        self.command = command
+
+    @property
+    def plugin_name(self):
+        return "linuxdeploy-plugin-gtk.sh"
```
freakboy3742, line 133: What's the functional difference between `name`, `full_name`, and `plugin_name`?

freakboy3742 (changes requested):
Looks like a good start; I've got a couple of questions inline.

The other "bigger picture" questions I have are:

1. What the plugin is getting us that our existing library discovery isn't getting us? I *think* it's mostly the typelibdir stubs... is there anything else? What happens if someone uses a library that uses typelib stubs that *isn't* in the "blessed list" that the Gtk plugin captures?
2. What's the impact on the Qt story? Or for any other toolkit that we support?

danyeaw:
Thanks for the review and feedback, I really appreciate it!

Overall the scope of this contribution was to solve the immediate issue that many users in Linux are not able to package a Toga app using the tutorial. I agree that we should also discuss our longer term vision of our strategy for Linux packaging going forward. 

> 1. What the plugin is getting us that our existing library discovery isn't getting us? I _think_ it's mostly the typelibdir stubs... is there anything else?

If an AppImage should be completely self-contained with no guarantee that a user will have certain libraries (besides a working X Windows or Wayland system) installed, then I think our original implementation was missing other things beyond typelibs including:
1. GDK Pixbuf cache and module loaders. The cache has the module loaders paths hardcoded and is generated by the gdk-pixbuf-query-loaders binary. The module loaders are also dynamic libraries (*.so), but they weren't being pulled in by linuxdeploy without the plugin. I do not think loading images will work without this (without relying on GDK being installed on the host system).
2. Input Method support: immodules in GTK supports input characters that cannot be written directly by keyboard, for example to support languages like Chinese, Korean, and Japanese. This is similar to the one above, where there is a cache file that is generated using the gtk-query-immodules-3.0 executable.
3. Printbackends if an app wants to include print functionality. This may include things like file, cups, and lpr.

> What happens if someone uses a library that uses typelib stubs that _isn't_ in the "blessed list" that the Gtk plugin captures?

This solution only supports typelibs for GTK, if there are other libraries that need typelibs, we would have to add plugins for those or use a different option.

> 2. What's the impact on the Qt story? Or for any other toolkit that we support?

There is a [plugin](https://github.com/linuxdeploy/linuxdeploy-plugin-qt) we could add for Qt as well with a similar implementation. I think these two are by far the most complex to get working.

Why I discussed other solutions I considered is because I think we need to decide:
1. Do we want to support AppImage?
2. If we do, is our goal to reliably package most Toga and Qt apps? Or reliably package any app using any imported python module?

If our focus is on Toga and Qt, I think we can extend linuxdeploy. If our focus is on any app, then we shouldn't reinvent the wheel and we should make Briefcase a user friendly wrapper over PyInstaller. For Linux, Windows, and macOS, they have solved the analyzing python code and pulling in all required dependencies so that it can then be packaged problem.

src/briefcase/platforms/linux/appimage.py
danyeaw: Right now it is hard coded to support our Toga use case. The plugin does do some automatic detection, but our apps are not successfully being detected. This could be a future PR to extend support for GTK4.

docs/reference/platforms/linux/appimage.rst
danyeaw: Yes, that is a tradeoff of using AppImage. Themes have to be packaged with the app.

src/briefcase/platforms/linux/appimage.py
danyeaw: >  toga-gtk>=0.3.0.dev33? (i.e., a version string on the dependency)?

Yes, I think so, that is what I tested it with a toga-gtk is in that string.

> pygobject (i.e., case incompatibility?)

Updated the implementation to support this, good idea!

> An app that uses a different toolkit that wraps GTK, but doesn't directly import pygobject? (not sure I have any specific examples in mind, but I'm sure someone has built one)

No, but we could support this in the future if we had something in mind.

> On the last example - is there a reason you opted for dependency inspection rather than an explicit list of linuxdeploy plugins needed by the app?

I think the general use case this was implementing for was users either have a Toga or PyGObject app and those will be listed already in the requires. We could add another option for explicit plugins needed, but I am not sure that this is needed at this time unless we implement a much larger plugin system, and if we do that we could add it at that time.

src/briefcase/integrations/linuxdeploy.py
danyeaw: I think there definitely is a potential for something more general in the future if we add additional plugins. :+1:

src/briefcase/integrations/linuxdeploy.py
danyeaw: You are right, I don't think they are needed, I removed everything but the `plugin_name`. :+1:

docs/reference/platforms/linux/appimage.rst
danyeaw: Great suggestion, thanks! I moved the section to the end of the document.

freakboy3742:
> > 1. What the plugin is getting us that our existing library discovery isn't getting us? I _think_ it's mostly the typelibdir stubs... is there anything else?
> 
> If an AppImage should be completely self-contained with no guarantee that a user will have certain libraries (besides a working X Windows or Wayland system) installed, then I think our original implementation was missing other things beyond typelibs including:
> 
> 1. GDK Pixbuf cache and module loaders. The cache has the module loaders paths hardcoded and is generated by the gdk-pixbuf-query-loaders binary. The module loaders are also dynamic libraries (*.so), but they weren't being pulled in by linuxdeploy without the plugin. I do not think loading images will work without this (without relying on GDK being installed on the host system).
> 2. Input Method support: immodules in GTK supports input characters that cannot be written directly by keyboard, for example to support languages like Chinese, Korean, and Japanese. This is similar to the one above, where there is a cache file that is generated using the gtk-query-immodules-3.0 executable.
> 3. Printbackends if an app wants to include print functionality. This may include things like file, cups, and lpr.

That certainly explains some of the bug reports we've had - in particular the errors around loading images/icons.

I guess the follow up question is *why* are those pieces not being loaded? Is there any way that we could systematically ensure that pieces like this are found? What is the GTK plugin doing that is "special" in this regard? 

My understanding was that Linuxdeploy essentially followed the linking chain from every .so that it encountered. This fails as an automated technique as soon as dynamic loading occurs - but as long as you manually specify one of the entry points that is dynamically loaded, the link-following approach can continue.

It seems like you're suggesting that even that approach has it's limits, because there are additional files and resources that won't be picked up by this link-following process. This does make some sense - if there's a resource file that is required at runtime, it's going to be very difficult/almost impossible to automatically detect that a .so references that resource file.

The problem that we face is that we can include a plugin for GTK, and maybe even Qt, which will work for tutorial cases and simple apps - but if a user writes an app that uses some obscure library, they might end up in a situation where they can't package the app simply because there isn't a linuxdeploy plugin that works for their particular library.

Are we in a position (however theoretical) where we could reproduce what the GTK plugin is doing as a matter of configuration? Linuxdeploy doesn't have any real way to define a "configuration" - but we're not constrained in that way. We're in a position to provide almost arbitrarily complex lists of files/directories that need to be passed to the linuxdeploy invocation. If necessary, we could define our own "generic" plugin that uses configuration items in the pyproject.toml file - or we could provide a plugin definition as part of the project template, which would give end-users an easy way to customise what is/isn't included in their own apps. Based on what you know, does this approach sound even remotely viable?

> > What happens if someone uses a library that uses typelib stubs that _isn't_ in the "blessed list" that the Gtk plugin captures?
> 
> This solution only supports typelibs for GTK, if there are other libraries that need typelibs, we would have to add plugins for those or use a different option.
> 
> > 2. What's the impact on the Qt story? Or for any other toolkit that we support?
> 
> There is a [plugin](https://github.com/linuxdeploy/linuxdeploy-plugin-qt) we could add for Qt as well with a similar implementation. I think these two are by far the most complex to get working.

Agreed. A quick look at the qt plugin shows that it's a compiled binary, not just a shell script, which makes me question whether the 'custom plugin in the template' approach I described will work...

> Why I discussed other solutions I considered is because I think we need to decide:
> 
> 1. Do we want to support AppImage?
> 2. If we do, is our goal to reliably package most Toga and Qt apps? Or reliably package any app using any imported python module?

Addressing point (2) first - The end goal should be "any app, using any library"; however, I'm also comfortable with a 90% solution that only manages Toga, GTK and Qt apps with no "exotic" libraries as an interim step. Ideally, we'd be able to identify when an app hits this "exotic" boundary so we can report the problem to the user at time of build, rather than have the problem be discovered at runtime by end users.

As for point (1) - my enthusiasm for AppImage is driven primarily by the fact that it was the one I was able to get to work... for a definition of "work" that appears to be at least partially accidental. The current implementation has clearly missed some of edge cases, and as distributions diverge from what was present on Ubuntu 18.04, these edge cases are becoming more apparent. The biggest downside to AppImage appears to be that it is a passion project of a very small number of developers who... well, let's just say they don't consistently adhere to the standards of conduct that the BeeWare community expects of its members.

For me, the question is two-fold:
a. Can AppImage *ever* be made to work reliably - even in the 90% case? Will accomodating just the GTK and Qt plugins be enough to get us to the 90% case, or will we be perpetually chasing our tail? 
b. Are there viable alternatives? From an industry standpoint, Flatpak seems to have a lot more traction than AppImage and Snap; will it provide a more robust packaging solution? 

Supporting both AppImage *and* Flatpak is also an option, but we'd need to make a call as to which one is the default. 
 
> If our focus is on Toga and Qt, I think we can extend linuxdeploy. If our focus is on any app, then we shouldn't reinvent the wheel and we should make Briefcase a user friendly wrapper over PyInstaller. For Linux, Windows, and macOS, they have solved the analyzing python code and pulling in all required dependencies so that it can then be packaged problem.

I'm not especially enamoured by the idea of using pyinstaller - that would seem to be a bit of an existential question for Briefcase. Why does Briefcase exist if all it is doing is wrapping another python packaging tool? Working out how to package what is needed for a standalone app is kind of the *point* of Briefcase; if we can't do it, then we should pack up shop and move on :-) 

There is undeniably some overlap in what pyinstaller does and what Briefcase does/needs to do - to the extent that I'm sure we can learn from each other - but there's also enough different in the core approaches that I don't think we should give up.

danyeaw:
> It seems like you're suggesting that even that approach has it's limits, because there are additional files and resources that won't be picked up by this link-following process. This does make some sense - if there's a resource file that is required at runtime, it's going to be very difficult/almost impossible to automatically detect that a .so references that resource file.

Yes, that's my understanding as well.

> The problem that we face is that we can include a plugin for GTK, and maybe even Qt, which will work for tutorial cases and simple apps - but if a user writes an app that uses some obscure library, they might end up in a situation where they can't package the app simply because there isn't a linuxdeploy plugin that works for their particular library.

Yup, I agree that is the current limitation of our approach.

> Are we in a position (however theoretical) where we could reproduce what the GTK plugin is doing as a matter of configuration? Linuxdeploy doesn't have any real way to define a "configuration" - but we're not constrained in that way. We're in a position to provide almost arbitrarily complex lists of files/directories that need to be passed to the linuxdeploy invocation. If necessary, we could define our own "generic" plugin that uses configuration items in the pyproject.toml file - or we could provide a plugin definition as part of the project template, which would give end-users an easy way to customise what is/isn't included in their own apps. Based on what you know, does this approach sound even remotely viable?

I think it sounds viable, and I think a more generalized form of the first alternative I considered. However, we probably need to consider the user experience of this approach. It is very difficult figuring out all the things to include to get a certain dependency to work. PyInstaller did this by allowing for users to create "hooks". The hooks are then shared as part of PyInstaller itself and a contrib library, so once a hook for a certain troublesome library is created, then everyone who uses that library gets that solution from then on. 
 
> For me, the question is two-fold: a. Can AppImage _ever_ be made to work reliably - even in the 90% case? Will accomodating just the GTK and Qt plugins be enough to get us to the 90% case, or will we be perpetually chasing our tail?

Yes, I think we can get to a 90% solution, and I think this plus a 2nd PR to add Qt support would get us most of the way there.

> b. Are there viable alternatives? From an industry standpoint, Flatpak seems to have a lot more traction than AppImage and Snap; will it provide a more robust packaging solution?

> Supporting both AppImage _and_ Flatpak is also an option, but we'd need to make a call as to which one is the default.

I agree that Flatpak is the primary alternative, and I would vote that it is even the default.

> I'm not especially enamoured by the idea of using pyinstaller - that would seem to be a bit of an existential question for Briefcase. Why does Briefcase exist if all it is doing is wrapping another python packaging tool? Working out how to package what is needed for a standalone app is kind of the _point_ of Briefcase; if we can't do it, then we should pack up shop and move on :-)

One of the reasons that Python has had so much success in the data science field is that the community built on top of what already existed. Seaborn is built on top of matplotlib for example.

I think of packaging in a few phases:
1. Install dependencies for your app on the platform you are targeting
2. Collect all of those dependencies and other files needed to run the app
3. Package the app in a format that can be installed by someone else, including signing and notarizing

Briefcase might not have to solve all the problems for all 3 of those, but it could take the best of those things that exist and glue them together and make them easier for the user.

> There is undeniably some overlap in what pyinstaller does and what Briefcase does/needs to do - to the extent that I'm sure we can learn from each other - but there's also enough different in the core approaches that I don't think we should give up.

PyInstaller really only solves the 2nd one above, and only on the current platform, no cross-compilation to mobile for example.

nickzoic:
I don't know if there's any way to get from "requires toga-gtk" to "system_requires libgtk-3-dev and librsvg2-dev" automatically ... maybe python libraries *should* have such a thing, it'd be super handy instead of the traditional README.md declaration of "first, apt install all this crap, then ..."

(It seems like this is the kind of metadata that python packages should already have but ... no?  Seems weird there's no PEP for this at least ...)

We could maybe get from the 90% solution to the 95% solution for briefcase just by having a little table of "if you require *this* python library we know you'll need *those* system libraries".  The build process could then implicitly require that, or explicitly add it to the pyproject.toml file, or warn the user to add them, or *something*.

freakboy3742:
> I don't know if there's any way to get from "requires toga-gtk" to "system_requires libgtk-3-dev and librsvg2-dev" automatically ... maybe python libraries _should_ have such a thing, it'd be super handy instead of the traditional README.md declaration of "first, apt install all this crap, then ..."

We are already able to do this using the briefcase project template - if you specify that you want a Toga app, you [get a bunch of libraries included as system defaults](https://github.com/beeware/briefcase-template/blob/v0.3/%7B%7B%20cookiecutter.app_name%20%7D%7D/pyproject.toml#L40). If those two libraries are required, it's easy to add to the template; but from what I'm reading in the discussion, @danyeaw didn't need those libraries to get the app working. 

> (It seems like this is the kind of metadata that python packages should already have but ... no? Seems weird there's no PEP for this at least ...)
> 
> We could maybe get from the 90% solution to the 95% solution for briefcase just by having a little table of "if you require _this_ python library we know you'll need _those_ system libraries". The build process could then implicitly require that, or explicitly add it to the pyproject.toml file, or warn the user to add them, or _something_.

Binary requirements for "the whole of PyPI" is probably out of scope. There might be a possibility of defining `system-requires` requirements in packaging metadata as a Briefcase extension that packages could opt into - or, we could turn to Conda as a better source of defined binary dependencies.

nickzoic:
Oh, well, I do love a good Astronomically Out Of Scope :-)  If there's really no PEP maybe I should write one.

What I mean was a bit of a middle ground though: we currently know toga-gtk's requirements, no problem, but are there other common python packages we also should check for?

freakboy3742:
> Oh, well, I do love a good Astronomically Out Of Scope :-) If there's really no PEP maybe I should write one.
> 
> What I mean was a bit of a middle ground though: we currently know toga-gtk's requirements, no problem, but are there other common python packages we also should check for?

I guess it wouldn't be hard to generate a list of the "top N PyPI binary packages", build a map of those packages to their ubuntu 18.04 system package requirements, and then put those requirements in a data file somewhere (say, in the app template). 

However, I'm also acutely aware that that statement is also a good chunk of what Conda does out of the box, so... maybe it would better to spend that effort working out how to leverage what Conda can provide, rather than trying to reinvent the wheel (pun 100% intended).

nickzoic:
PS: It also seems to be what [PyInstaller hooks](https://github.com/pyinstaller/pyinstaller/tree/4b9fd01d7dc5a0aae3ba7c100437547915464f0e/PyInstaller/hooks) are for ...

src/briefcase/platforms/linux/appimage.py
```diff
@@ -156,52 +157,70 @@ def build_app(self, app: BaseConfig, **kwargs):
...
+        """
+
+        # For some reason, the version has to be passed in as an
+        # environment variable, *not* in the configuration...
+        env = {"VERSION": app.version}
+
+        plugins = []
+        if app.requires:
+            for dependency in app.requires:
+                if ("toga-gtk" or "pygobject") in dependency.lower():
```
freakboy3742, line 179: Um... are you sure this is doing what you think it is? AFAICT, this reduces to `"toga-gtk" in dependency.lower()`...

nickzoic: Ha, that's a late-night-too-much-coffee one if I've ever seen it :-)
(I had to check it wasn't me ...)

src/briefcase/platforms/linux/appimage.py
nickzoic: Yeah, it's meant to auto-detect but I found it wasn't ... the version would have to be the correct one for the toga-gtk version I suppose (see previous conversation about mapping libs to requirements)

src/briefcase/platforms/linux/appimage.py
danyeaw: Thanks for catching that, I successfully tested that "toga-gtk" is True :smile:

I updated this and added some additional parameterized tests to ensure that requires in different formats all load the plugin.

src/briefcase/platforms/linux/appimage.py
danyeaw: The autodetect doesn't work because it uses `ldd` to look for executables that are linked against the GTK libraries. Our only executable is Python itself which is not linked against GTK.

freakboy3742 (commented):
The testing that I've done around #757 has convinced me that adopting the GTK plugin is definitely worth adding. However, it's highlighted to me that plugins are a key part of the linuxdeploy ecosystem. I wonder if aiming generic might be the better approach here.

This PR makes a special case of the GTK plugin, and autodetects the need for that plugin. This has 2 downsides:
* The autodetection can fail.
* It only allows for the use of the "official" plugin.

The upside is that there's no need for any additional configuration items. 

As a counterproposal: We could add an `linuxdeploy_plugins` configuration item for the AppImage backend:

```
linuxdeploy_plugins = ['gtk', 'https://example.com/linuxdeploy-plugin-magic.sh']
```
This would allow end users to specify whatever combination of plugins that are necessary. We can include a "simple" list of shortcodes(like `gtk` and `qt`) that includes the officially supported plugins that expands to the official download URLs; but allow users to provide their own URLs (with a user-provided explicit plugin overriding the builtin). This would allow users to provide their own plugins; or their own bug fixed versions of plugins (e.g., your bugfix version of the GTK plugin).

This wouldn't allow us to handle the `DEPLOY_GTK_VERSION=3` environment variable; although we could possible handle this by:
1. Setting that environment variable for the user as part of the `gtk` 'shortcode" 
2. Ensure that variables in the environment used to invoke briefcase are passed to linuxdeploy (i.e., you'd need to invoke `DEPLOY_GTK_VERSION=3 briefcase build linux`)
3. Allow for `DEPLOY_GTK_VERSION=3 gtk` as a syntax for declaring plugins - i.e., use the shell convention of setting environment variables in the definition.  

(3) would be my preferred of those options, but is obviously more complex to implement.

The downside to this approach is that there is more manual configuration. We can slightly mitigate this by auto-adding the plugins in the base project template, but there's no escaping the fact that end-users will need to know a lot more about Linux packaging to be able to drive the tool effectively. Unfortunately, I'm not sure if this is completely avoidable. Linux packaging is... complex... and trying to completely avoid that is a 

Another option would be to split the difference - have autodetection of 'known, supported' plugins, but *also* allow for manual specification. This might also fit into the solution for #757 - if we know that certain packages must be deployed as source, we can provide that assistance as an automation; but we can also open the door to allow manual configuration when things go wrong. 

Any thoughts about this proposal?

danyeaw:
> As a counterproposal: We could add an `linuxdeploy_plugins` configuration item for the AppImage backend:
> 
> ```
> linuxdeploy_plugins = ['gtk', 'https://example.com/linuxdeploy-plugin-magic.sh']
> ```

I like your counterproposal! :+1:  For the configuration in the pyproject.toml, maybe it would look something like this?

```toml
[tool.briefcase.app.helloworld.linux]
requires = [
    'toga-gtk>=0.3.0.dev34',
]
system_requires = [
    'libgirepository1.0-dev',
    'libcairo2-dev',
    'libpango1.0-dev',
    'libwebkitgtk-3.0-0',
    'gir1.2-webkit-3.0',
]
linuxdeploy_plugins = [
    'gtk',
]
```

Where we could enter 3 things:

1. A known plugin like 'gtk' or 'qt'
2. A URL to a plugin like 'https://github.com/linuxdeploy/linuxdeploy-gtk-plugin/master/linuxdeploy-gtk-plugin.sh'
3. A path to a plugin that is part of the project, like 'packaging/linuxdeploy-gtk-plugin.sh'

I thought of doing both name and URL, but it seems like it would just make things more complex for the user.

> This wouldn't allow us to handle the `DEPLOY_GTK_VERSION=3` environment variable

I like your idea to support shell variables, but it may not be needed at this time. Ubuntu didn't have GTK4 in the repos until 21.04, and 22.04 is the first LTS version with it. I am assuming we would want to support 18.04 until 2023 and 20.04 until 2025 for AppImage compatibility. So we wouldn't support GTK4 until 2025, unless someone wants to use Docker to build it from source using JHBuild - [I have done this](https://github.com/gaphor/gaphor/blob/main/_packaging/appimage/Dockerfile), but it is not trivial. Not to pile on AppImage limitations, but GTK4 was released in December 2020, and it will then take 5 years to support the new toolkit. :frowning:

I would propose will hold off on this piece until we are a little more sure of our packaging strategy for Linux, and we see a need for passing the environmental variables. YAGNI :smile:

freakboy3742:
> I like your counterproposal! 👍 For the configuration in the pyproject.toml, maybe it would look something like this?
...
Yup - that's what I had in mind.

> Where we could enter 3 things:
> 
> 1. A known plugin like 'gtk' or 'qt'
> 2. A URL to a plugin like 'https://github.com/linuxdeploy/linuxdeploy-gtk-plugin/master/linuxdeploy-gtk-plugin.sh'
> 3. A path to a plugin that is part of the project, like 'packaging/linuxdeploy-gtk-plugin.sh'

I hadn't thought of "local path", but that's a good idea.
 
> I thought of doing both name and URL, but it seems like it would just make things more complex for the user.

Yeah - I had similar a thought. However, AFAICT, linuxdeploy plugin naming has a required scheme, so you can derive the name of the plugin from the name of the file to be executed; that makes `name = value` redundant (completely aside from what to use in the "name only " case).

The one additional detail that occurred to me - if the plugin is user-provided, we won't be able to cache it in `~/.briefcase/tools`. If I have one project using the "default" GTK plugin, and a second project using a custom GTK plugin, we would end up with an ambiguity over which version is being cached, based on which project was generated first. 

The docs for linuxdeploy say that one of the locations that is inspected for plugins is the location of linuxdeploy binary, or the location of the AppDir, which we can use to our advantage - depending on the order of resolution.

If linuxdeploy looks in the AppDir first, then a URL-specified plugin can be downloaded/copied and stored next to the AppDir as an artefact of the creation process. If the plugin is a system default, it can be downloaded and cached as you've implemented here.

If linuxdeploy looks at the same directory as linuxdeploy first, then we'll need to cache the system plugins in `~/.briefcase/tools/linuxdeploy-plugins`, and copy default plugins into the project. That way, the plugin will *always* be read from the AppDir location, we'll just have some redundant copies of the plugin on the system. We might be able to use a symlink to avoid the copy?
 
> > This wouldn't allow us to handle the `DEPLOY_GTK_VERSION=3` environment variable
> 
> I like your idea to support shell variables, but it may not be needed at this time. Ubuntu didn't have GTK4 in the repos until 21.04, and 22.04 is the first LTS version with it. I am assuming we would want to support 18.04 until 2023 and 20.04 until 2025 for AppImage compatibility. So we wouldn't support GTK4 until 2025, unless someone wants to use Docker to build it from source using JHBuild - [I have done this](https://github.com/gaphor/gaphor/blob/main/_packaging/appimage/Dockerfile), but it is not trivial. Not to pile on AppImage limitations, but GTK4 was released in December 2020, and it will then take 5 years to support the new toolkit. 😦
> 
> I would propose will hold off on this piece until we are a little more sure of our packaging strategy for Linux, and we see a need for passing the environmental variables. YAGNI 😄

Totally onboard with YAGNI as an analysis; but what will insert the GTK environment variable in the case of a user-customized GTK plugin? From my testing, the plugin crashes on Briefcase projects if the variable isn't set.

danyeaw:
@freakboy3742 The updated design is ready for a review when you get the chance. It supports the 3 methods of configuration that we discussed including allowing for environmental variables like `DEPLOY_GTK_VERSION=3 gtk` as a syntax for declaring plugins. It also supports plugins declared as URLs and local files for the project.

One portion I would appreciate feedback on is that I added tests, but didn't cover raising all errors. I am not sure if you are trying to achieve 100% code coverage for these cases.

To run the changes, you'll also need the modification of the PATH in https://github.com/beeware/briefcase-linux-appimage-template/pull/16.

Hopefully we can get this merged and then focus on less painful packaging formats like Flatpak! 😄

src/briefcase/integrations/linuxdeploy.py
```diff
@@ -127,3 +128,102 @@ def patch_elf_header(self):
...
+        with self.command.input.wait_bar("Removing the old linuxdeploy install..."):
+            self.file_path.unlink()
+
+
+class LinuxDeployPluginBase(LinuxDeployBase):
+    """Base class for linuxdeploy plugins."""
+
+    @property
+    def file_path(self):
+        return self.command.tools_path / "linuxdeploy_plugins" / self.filename
```
freakboy3742, line 163: Ok - this is going to be problematic. It works for "known" plugins like GTK and Qt, because there's a known canonical version; but it won't work for user plugins, because two different projects could have a plugin with the same name. 

I suspect that we will be able to locally cache for "known" plugins - but anything user provided will need to be downloaded/copied directly into the project folder.

freakboy3742 (changes requested):
Ok - the pieces are almost all here, but I'm not sure they're all in the right place.

Some initial comments
1. The handling of configuration files. As noted inline, the config file should be pretty dumb container; the config object shouldn't know anything about linuxdeploy specifics. 
2. The refactoring of verify_tools to be app sensitive. As noted inline, I don't think we can make that change in the way you've described here, because of interactions with other backends. 
3. There's the start of some good class abstraction in the linuxdeploy integration, but the naming of properties really threw me. There's a lot of legacy names that made perfect sense when the only plugin was linuxdeploy, but now that it could be the base linuxdeploy, a known plugin, a user URL plugin, or a local file, names like `linuxdeploy_download_url` don't make much sense - especially when it's referring to a local file. At the very least, there's docstrings required to clarify the purpose of each path; but I suspect the names themselves aren't great in a lot of cases.
4. It's also not clear to me how these plugins are being discovered at runtime. This might be a side effect of the naming - it's not clear to me which paths are "local plugin", which are "file_path", which are "download URL", and so on. 

In my testing, I also got "plugin not found" errors regardless of whether I used `gtk`, a url, or a local file. If this is working for you, I suspect it may be because you've got a cached gtk plugin in the tools directory - which means that although it's downloading the plugins, they're not actually being used as downloaded.

This then comes back to one of the earlier questions I had - we need to work out the priority of resolution of plugins. linuxdeploy defines a sequence of locations that it looks for plugins: 
* the directory containing the linuxdeploy binary
* the directory containing the AppImage
* the directories in the user's $PATH

but it's not clear what order those are checked. To get around versioning issues, we may need to rely on the directory containing the AppImage for *all* plugins, and symlink/copy them in as part of the "create" step, rather than as a tool validation in build - i.e., treat them as an input to the build, rather than a build tool.

docs/reference/platforms/linux/appimage.rst
```diff
@@ -92,6 +105,83 @@ a new environment that is completely isolated from your development
...
+Some plugins need an environmental variable passed to configure it. For
+example, the ``gtk`` plugin needs the ``DEPLOY_GTK_VERSION`` environmental
+variable set to 3 or 4. If you use the ``gtk`` plugin provided by Briefcase,
+it is set to 3 by default, but you can override it or set it for other plugins
+by setting the ``DEPLOY_GTK_VERSION`` environmental variable. For example,
+when using a relative path, you can set it like this:
+
+    linuxdeploy_plugins = ['DEPLOY_GTK_VERSION=3 packaging/linuxdeploy-gtk-plugin.sh']
+
+``linuxdeploy_gtk_version``
```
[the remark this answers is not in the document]
danyeaw: @freakboy3742 Can you please help me understand what you mean? Setting an environmental variable is not required?

danyeaw:
> The handling of configuration files. As noted inline, the config file should be pretty dumb container; the config object shouldn't know anything about linuxdeploy specifics.

Thanks for the feedback. My understanding that each command that briefcase runs is then sent downward to each platform/packaging type. I struggled with the fact that we are adding a bunch of logic to what was just a dumb configuration:

1. 3 different ways to define a plugin
2. Environmental variables nested in to the configuration

I started with just implementing this on the linuxdeploy and appimage side of things, but then for most commands you are reverifying and parsing the configuration, which didn't seem like a good solution. So I brought the smarts up to the config module to check and verify the configuration centrally. I understand not wanting to add linuxdeploy specifics to the config object as well. Do you have ideas on a better architecture? I guess we could create a new module or class that does linuxdeploy configuration verification, and then have all the linuxdeploy commands call that. Is that what you had in mind?

> The refactoring of verify_tools to be app sensitive. As noted inline, I don't think we can make that change in the way you've described here, because of interactions with other backends.

I think this is another architecture question which is hard for me to make a good decision on without guidance. To verify the plugins we need app information, since the plugins are tied to the app. You mentioned we could add another entry point, what would you like that to look like?

> In my testing, I also got "plugin not found" errors regardless of whether I used gtk, a url, or a local file. If this is working for you, I suspect it may be because you've got a cached gtk plugin in the tools directory - which means that although it's downloading the plugins, they're not actually being used as downloaded.

Did you make the changes from https://github.com/beeware/briefcase-linux-appimage-template/pull/16?

freakboy3742:
> > The handling of configuration files. As noted inline, the config file should be pretty dumb container; the config object shouldn't know anything about linuxdeploy specifics.
> 
> Thanks for the feedback. My understanding that each command that briefcase runs is then sent downward to each platform/packaging type. I struggled with the fact that we are adding a bunch of logic to what was just a dumb configuration:
> 
> 1. 3 different ways to define a plugin
> 2. Environmental variables nested in to the configuration

Sure - those properties are true of the interpretation of the config; but at the config level, it's a list of strings. When you're processing an AppImage, it does a merge to ensure that you have a single configuration that only contains the "common + linux + appimage" components, but it's still a dumb list of strings. `system_requires` is a list of strings too - and we don't do any validation of those (even to validate that it *is* a list of strings). 

Once you're *in* the AppImage backend, you can parse that list of strings in a meaningful way to extract whatever data you need. In the case of `system_requires`, the list is pretty much just joined and passed to Docker. In this case, we can convert the list of strings into a list of (environment variable dict, plugin identifier) pairs; and the plugin name can then be processed to determine how the plugin can be satisfied.
 
We could perhaps add validation - it might be a nice touch to improve error handling when the user provides input that can't be handled - but I'd view that more as an added (and orthogonal) bonus feature, rather than a core feature that is required for linuxdeploy plugin support.

> I started with just implementing this on the linuxdeploy and appimage side of things, but then for most commands you are reverifying and parsing the configuration, which didn't seem like a good solution. So I brought the smarts up to the config module to check and verify the configuration centrally. I understand not wanting to add linuxdeploy specifics to the config object as well. Do you have ideas on a better architecture? I guess we could create a new module or class that does linuxdeploy configuration verification, and then have all the linuxdeploy commands call that. Is that what you had in mind?

The architecture I'm thinking of is really not much more than a "parse_linux_deploy_plugin_config()" method, implemented once in the AppImage backend (either as a standalone method or as a mixin), and invoked by the various commands that need it.

> > The refactoring of verify_tools to be app sensitive. As noted inline, I don't think we can make that change in the way you've described here, because of interactions with other backends.
> 
> I think this is another architecture question which is hard for me to make a good decision on without guidance. To verify the plugins we need app information, since the plugins are tied to the app. You mentioned we could add another entry point, what would you like that to look like?

I was suggesting we need both a `verify_tools()` and a `verify_app_tools(app)` entry point. For all existing commands, `verify_app_tools()` is a no-op; for AppImage, we override to ensure that the required plugins are available. 

However, I'm also not convinced this is needed at all. `docker` and `linuxdeploy` itself are tools, and they have to exist to run the AppImage backend. While a linuxdeploy plugin *is* a tool (or is at least tool-adjacent), the fact that they're user customisable on a per-project makes them much closer in usage to the support packages. There are default support packages that it's worth caching; but end-users can define their own support packages, and they need to be installed into the project on a per-project basis. linuxdeploy plugins are much the same - there are default plugins we can cache; but ultimately, in order to fully support per-project customisability, we're going to need to install the plugins as part of the project creation process.

If we were *only* supporting official plugins, this wouldn't be the case - it's only the user-supplied plugins that make this an issue.

> > In my testing, I also got "plugin not found" errors regardless of whether I used gtk, a url, or a local file. If this is working for you, I suspect it may be because you've got a cached gtk plugin in the tools directory - which means that although it's downloading the plugins, they're not actually being used as downloaded.
> 
> Did you make the changes from [beeware/briefcase-linux-appimage-template#16](https://github.com/beeware/briefcase-linux-appimage-template/pull/16)?

Ah - no I didn't - missed that detail; thanks for the clarification.

However, I don't think that approach will work, for the reason I mentioned above. AFAICT, by making that change, there's effectively no difference between putting all the plugins in the tools directory, and putting linuxdeploy in the tools directory and the plugins in a plugins directory.

The underlying issue is that there could be multiple "GTK" plugins. Consider 2 projects on the same developer's system - one that uses the official GTK plugin, and one that uses a custom modified plugin. If the official plugin has been downloaded into the plugin folder, and it's always on the path, then AFAICT it's always going to be found, regardless of local conditions.

I think we're going to need to treat the plugins folder as a download optimisation, but at runtime, copy/link the plugin into the directory containing the AppDir if the system plugin is used; if the user specifies a local path, copy/link that file; and if the user specifies a URL, download that URL into the AppDir. That way, we're always relying on the "look in the AppDir" path, rather than anything tied to the PATH or the location of the linuxdeploy binary; and each project can make the decision over which plugins (and which versions of the plugins) are in use. 

We *might* be able to support the default plugins by downloading to the plugins folder, and then *optionally* including the plugin folder in the path... but frankly, that sounds really fragile for the sake of saving a tiny amount of disk space.

The other way we might be able to avoid this problem is if linuxdeploy's lookup order is "AppDir then PATH then linuxdeploy dir" - that way, any locally installed plugin would take priority over the "system" ones. However the linuxdeploy docs imply the order is the other way around. This would also mean that the plugins cache directory isn't needed, we could just drop plugins into the tools folder directly.

danyeaw:
> The underlying issue is that there could be multiple "GTK" plugins.

Is this use case supported by any of the other tools or support packages so I can reference an implementation?

I'm struggling to understand why I would need this as a user. If I want to make updates to a plugin, those updates wouldn't be app specific. I would want to use my updated GTK plugin for all my projects.

freakboy3742:
> > The underlying issue is that there could be multiple "GTK" plugins.
> 
> Is this use case supported by any of the other tools or support packages so I can reference an implementation?

It's not implemented by any of the other tools, because as I mentioned, this is the only case where a *tool* is optional. 

The reference implementation that *is* worth looking at is Support packages. Support packages are downloaded into the cache directory as an optimisation, but are installed into the package itself. The support package is copied *from* the cache folder, but it's never used *in* the cache folder. If the user specifies a local file, the support package is installed directly from that location. That's effectively what I'm describing here - we don't rely on the plugin being in the cache folder, except as a convenient store of pre-downloaded tools.

Thinking about it, I think there might be a bug in the case of a user-provided support package URL - in that case, the download will be cached into the cache folder. The risk of collision with the official support packages is *smaller*, because the support packages encourage custom version tags in the download name, but the possibility still exists. Collisions are inevitable with linuxdeploy plugins because every GTK plugin is named `linuxdeploy-plugin-gtk`, with no versioning or source-identifying marks.

> I'm struggling to understand why I would need this as a user. If I want to make updates to a plugin, those updates wouldn't be app specific. I would want to use my updated GTK plugin for all my projects.

While I'm sure that's likely in practice, it's not a given. If you've got a bugfix for a plugin that is needed in one project, but the same bugfix breaks a second project, you've got a problem. If we were *just* dealing with the official plugin, this problem wouldn't' happen; but as soon as we've got .

As one concrete example where this will become relevant - consider the case of the next bugfix that needs to be made to the GTK plugin. That will inevitably begin life as a local file or PR URL. The official GTK plugin also exists, and may have been used historically. If plugins were being always downloaded into the cache folder and used from that location, it would be very difficult to diagnose *which* plugin was being used at any given time, because you don't know which plugin is currently in the cache folder - the official version or the "fixed" version. Confirming that your fix is actually in place and working becomes more difficult. And, until that plugin has been verified as working, you don't want the plugin to be used inadvertently on a separate project that is being deployed - but if running a test causes the plugin cache to be overwritten/updated, you could easily deploy a production app using an in-development plugin.

However, if the plugin is installed into the project directory and used in that location, then you know exactly which plugin is currently in use (and, conveniently, would also allow you to edit the plugin in place to iterate on the fix), and there's no chance of a pre-production plugin being used on a production app.

danyeaw:
I submitted a working solution that integrated an official plugin 1 month ago and solved the issue I was having in Linux. I should have stopped with that. That was my PR, not a flexible plugin system with smart caching and custom env variable support.

Feel free to leave this PR open if you or others want to add to or change it, or close it if you want to. I'm not interested in spending more of my free time on this.

freakboy3742:
> I submitted a working solution that integrated an official plugin 1 month ago and solved the issue I was having in Linux. I should have stopped with that. That was my PR, not a flexible plugin system with smart caching and custom env variable support.
> 
> Feel free to leave this PR open if you or others want to add to or change it, or close it if you want to. I'm not interested in spending more of my free time on this.

Sure - totally understood. Thanks for your work on this; I'll wrap up the last bits once I've finished with the Windows packaging pieces I'm currently looking at.

freakboy3742:
For anyone looking at this PR - you can start a new project using the template from https://github.com/beeware/briefcase-template/pull/30 or add:
```
linuxdeploy_plugins = ['DEPLOY_GTK_VERSION=3 gtk']
```
to the linux configuration section of an existing project. No changes are needed to the `briefcase-linux-AppImage`.

rmartin16:
Worked out of the box on pop_os 22.04 🦾 

A few remarks. These may come down to expectations but but given the whole goal is to avoid target system dependencies, looking for clarification.

Below is the listed of loaded libraries for hello world with toga created within docker that _did not_ come from the AppImage:
```
python3 158594 russell  mem       REG                8,1  1306280   26742794 /usr/lib/x86_64-linux-gnu/libX11.so.6.4.0
python3 158594 russell  mem       REG                8,1   844624   26743407 /usr/lib/x86_64-linux-gnu/libharfbuzz.so.0.20704.0
python3 158594 russell  mem       REG                8,1   133224   26743300 /usr/lib/x86_64-linux-gnu/girepository-1.0/WebKit2-4.0.typelib
python3 158594 russell  mem       REG                8,1   149760   26743530 /usr/lib/x86_64-linux-gnu/libgpg-error.so.0.32.1
python3 158594 russell  mem       REG                8,1   153904   26743577 /usr/lib/x86_64-linux-gnu/libgraphite2.so.3.2.1
python3 158594 russell  mem       REG                8,1    34888   26743220 /usr/lib/x86_64-linux-gnu/libdatrie.so.1.4.0
python3 158594 russell  mem       REG                8,1    41152   26744454 /usr/lib/x86_64-linux-gnu/libthai.so.0.3.1
python3 158594 russell  mem       REG                8,1 15752576   26218205 /usr/lib/locale/locale-archive
python3 158594 russell  mem       REG                8,1   813128   26743966 /usr/lib/x86_64-linux-gnu/libfreetype.so.6.18.1
python3 158594 russell  mem       REG                8,1  2216304   26743105 /usr/lib/x86_64-linux-gnu/libc.so.6
python3 158594 russell  mem       REG                8,1    16336   26740571 /usr/lib/x86_64-linux-gnu/girepository-1.0/JavaScriptCore-4.0.typelib
python3 158594 russell  mem       REG                8,1    47472   26743898 /usr/lib/x86_64-linux-gnu/libmd.so.0.0.5
python3 158594 russell  mem       REG                8,1   137560   26743091 /usr/lib/x86_64-linux-gnu/libbrotlicommon.so.1.0.9
python3 158594 russell  mem       REG                8,1   194872   26743343 /usr/lib/x86_64-linux-gnu/libexpat.so.1.8.7
python3 158594 russell  mem       REG                8,1   166504   26744683 /usr/lib/x86_64-linux-gnu/libxcb.so.1.1.0
python3 158594 russell  mem       REG                8,1   298120   26743386 /usr/lib/x86_64-linux-gnu/libfontconfig.so.1.12.0
python3 158594 russell  mem       REG                8,1   940560   26743884 /usr/lib/x86_64-linux-gnu/libm.so.6
python3 158594 russell  mem       REG                8,1    89096   26743097 /usr/lib/x86_64-linux-gnu/libbsd.so.0.11.5
python3 158594 russell  mem       REG                8,1    26800   26742811 /usr/lib/x86_64-linux-gnu/libXdmcp.so.6.0.0
python3 158594 russell  mem       REG                8,1    18720   26742800 /usr/lib/x86_64-linux-gnu/libXau.so.6.0.0
python3 158594 russell  mem       REG                8,1   108936   26744727 /usr/lib/x86_64-linux-gnu/libz.so.1.2.11
python3 158594 russell  mem       REG                8,1    51512   26743093 /usr/lib/x86_64-linux-gnu/libbrotlidec.so.1.0.9
python3 158594 russell  mem       REG                8,1    68552   26744315 /usr/lib/x86_64-linux-gnu/libresolv.so.2
python3 158594 russell  mem       REG                8,1    14432   26744542 /usr/lib/x86_64-linux-gnu/libutil.so.1
python3 158594 russell  mem       REG                8,1    14432   26743261 /usr/lib/x86_64-linux-gnu/libdl.so.2
python3 158594 russell  mem       REG                8,1    21448   26744278 /usr/lib/x86_64-linux-gnu/libpthread.so.0
python3 158594 russell  mem       REG                8,1    30920   26744546 /usr/lib/x86_64-linux-gnu/libuuid.so.1.3.0
python3 158594 russell  mem       REG                8,1    14664   26744329 /usr/lib/x86_64-linux-gnu/librt.so.1
python3 158594 russell  mem       REG                8,1    27002   26742057 /usr/lib/x86_64-linux-gnu/gconv/gconv-modules.cache
python3 158594 russell  mem       REG                8,1   240936   26742580 /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
```

Most of these appear to be fine except for `girepository-1.0/WebKit2-4.0.typelib`, `girepository-1.0/JavaScriptCore-4.0.typelib`, and `libthai.so.0.3.1`.

`libthai` is required by `libpango` but AppImage's packaging explicitly [excludes](https://github.com/AppImage/pkg2appimage/blob/master/excludelist#L117) it....apparently for audicity. I don't know how to override that blocklist (or if we'd want to)....but I'm also not sure if most linux distros include this by default.

As for those typelibs coming from `girepository`, I cannot track down what's loading them. `linuxdeploy` doesn't detect them when packaging the AppImage and `LD_DEBUG=files` doesn't say they are being loaded when opening the AppImage... If i build the AppImage outside of docker, i get the same behavior with `linuxdeploy` and `LD_DEBUG` but now they are being loaded from the `girepository` from within the AppImage.....idk


As a sidenote, briefcase doesn't have much resiliency against the Docker image disappearing. This PR makes that situation worse, though, since it's passing a `CalledProcessError` exception back to the interpreter when the image disappears (instead of just erroring); I'll add a comment to at least catch that. I can also create a new issue for better handling this in general.

src/briefcase/platforms/linux/appimage.py
```diff
@@ -142,41 +142,85 @@ class LinuxAppImageBuildCommand(LinuxAppImageMixin, BuildCommand):
...
+                bundle_path=self.bundle_path(app),
+            )
 
+            self.logger.info("Configuring Linuxdeploy plugins...", prefix=app.app_name)
+            # We need to add the location of the linuxdeploy plugins to the PATH.
+            # However, if we arerunning inside Docker, we need to know the
+            # environment *inside* the Docker container.
+            if self.use_docker:
+                echo_cmd = ["/bin/bash", "-c", "echo $PATH"]
+                base_path = self.Docker(self, app).check_output(echo_cmd).strip()
```
rmartin16, line 173: This can bomb out if the Docker image disappears.

```
[helloworld] Configuring Linuxdeploy plugins...

>>> Running Command:
>>>     docker run --volume /home/russell/tmp/beeware/helloworld/linux:/app:z --volume /home/russell/.cache/briefcase:/home/brutus/.cache/briefcase:z --rm briefcase/com.example.helloworld:py3.10 /bin/bash
-c 'echo $PATH'
Unable to find image 'briefcase/com.example.helloworld:py3.10' locally
docker: Error response from daemon: pull access denied for briefcase/com.example.helloworld, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
>>> Return code: 125

Log saved to /home/russell/tmp/beeware/helloworld/briefcase.2022_07_21-14_19_42.build.log

  File "/home/russell/github/danyeaw/briefcase/venv-3.10/bin/briefcase", line 33, in <module>
    sys.exit(load_entry_point('briefcase', 'console_scripts', 'briefcase')())
  File "/home/russell/github/danyeaw/briefcase/src/briefcase/__main__.py", line 15, in main
    command(**options)
  File "/home/russell/github/danyeaw/briefcase/src/briefcase/commands/build.py", line 58, in __call__
    state = self._build_app(
  File "/home/russell/github/danyeaw/briefcase/src/briefcase/commands/build.py", line 39, in _build_app
    state = self.build_app(app, **full_options(state, options))
  File "/home/russell/github/danyeaw/briefcase/src/briefcase/platforms/linux/appimage.py", line 173, in build_app
    base_path = self.Docker(self, app).check_output(echo_cmd).strip()
  File "/home/russell/github/danyeaw/briefcase/src/briefcase/integrations/docker.py", line 324, in check_output
    return self._subprocess.check_output(
  File "/home/russell/github/danyeaw/briefcase/src/briefcase/integrations/subprocess.py", line 297, in check_output
    cmd_output = self._subprocess.check_output(
  File "/home/russell/.pyenv/versions/3.10.4/lib/python3.10/subprocess.py", line 420, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/home/russell/.pyenv/versions/3.10.4/lib/python3.10/subprocess.py", line 524, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['docker', 'run', '--volume', '/home/russell/tmp/beeware/helloworld/linux:/app:z', '--volume', '/home/russell/.cache/briefcase:/home/brutus/.cache/briefcase:z', '--rm', 'briefcase/com.example.helloworld:py3.10', '/bin/bash', '-c', 'echo $PATH']' returned non-zero exit status 125.
```

src/briefcase/integrations/linuxdeploy.py
```diff
@@ -124,6 +132,228 @@ def patch_elf_header(self):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.env = {}
+
+        if not self.file_name.startswith("linuxdeploy-plugin-"):
+            raise BriefcaseCommandError(f"{self.file_name} is not a linuxdeploy plugin")
+
+    @property
+    def plugin_id(self):
+        return self.file_name.split(".")[0].split("-")[2]
```
mhsmith, line 154: `plugin_id` is defined sometimes as a property and sometimes as a simple class variable. This makes `LinuxDeploy.plugins` a bit confusing, because it looks up `plugin_id` on the class, but when you go looking for its definition you might find the base class property, which can only be looked up on an instance, not a class.

I think it would be clearer to remove the `plugin_id` overrides in the GTK and Qt classes, and instead make `LinuxDeploy.plugins` set the known plugin names directly in a dict literal.

mhsmith (commented):
Also worked out of the box for me on Debian 10 (Buster), which was released in 2019. I looked through the linuxdeploy GTK plugin output and saw no concerning messages.

src/briefcase/platforms/linux/appimage.py
```diff
@@ -142,41 +142,85 @@ class LinuxAppImageBuildCommand(LinuxAppImageMixin, BuildCommand):
...
+                env.update(plugin.env)
+
+            # Construct a path that has been prepended with the path to the plugins
+            env["PATH"] = os.pathsep.join(
+                [os.fsdecode(plugin.file_path) for plugin in plugins.values()]
+                + [base_path]
+            )
+        except AttributeError:
+            self.logger.info("No linuxdeploy plugins configured.")
+            plugins = []
```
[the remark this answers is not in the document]
freakboy3742: Good catch!

src/briefcase/integrations/linuxdeploy.py
```diff
@@ -124,6 +132,228 @@ def patch_elf_header(self):
...
+        self._download_url = url
+
+        url_parts = urlparse(url)
+        self._file_name = url_parts.path.split("/")[-1]
+
+        # Build a hash of the download URL; this hash is used to
+        # idenfity plugins downloaded from different sources
+        self.hash = hashlib.sha256()
+        for part in url_parts:
+            self.hash.update(part.encode("utf-8"))
```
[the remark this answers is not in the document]
freakboy3742: Having 2 different URLs produce different hashes was deliberate. The following 3 URLs could all produce different versions of a plugin:

* `http://example.com/release/linuxdeploy-plugin-foobar.sh`
* `http://example.com/dev/linuxdeploy-plugin-foobar.sh`
* `http://example.com/archive/linuxdeploy-plugin-foobar.sh?version=1`

Hashing on just the URL would make the three appear to be the same plugin.

Admittedly, this is an edge case that is unlikely to be an issue in practice, but if it ever comes up, it would be very difficult to diagnose, so I figured we should be proactive. Documenting the why is definitely a good idea, though.

src/briefcase/integrations/linuxdeploy.py
freakboy3742: Yeah - I went back and forth on this. The inconsistency is potentially confusing; using a literal in `LinuxDeploy.plugins` means duplication of information. You're probably right that being confusing is a bigger risk than the duplication, though.

src/briefcase/platforms/linux/appimage.py
freakboy3742: I hit this error a couple of times, but I put it down to the fact that I was doing some "off script" rebuilds and cache refreshes. Definitely worth addressing, but I don't think it's a blocker for this PR specifically.

src/briefcase/integrations/linuxdeploy.py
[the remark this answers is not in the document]
freakboy3742: Oh wait - I see what you're saying now - just hash the URL literal, rather than using HMAC. Yeah - that's a good idea - we're not depending on any cryptographic properties of the HMAC. I was overthinking things because I had the parsed URL parts handy.

freakboy3742 (approved):
Following reviews from @mhsmith and @rmartin16, I'm going to call his done. Thanks to @danyeaw for the original work on this!

Issue #618, opened 2021-09-08
Pillow Related error when trying to execute appimage


daedalus-0:
Hi. Packaging my application works fine but when trying to run the produced appimage I get an error which I think is related to pillow because of the referenced library in the output. 

    Traceback (most recent call last):
      File "/tmp/.mount_qcomicOkkPRb/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
        return _run_code(code, main_globals, None,
      File "/tmp/.mount_qcomicOkkPRb/usr/lib/python3.9/runpy.py", line 87, in _run_code
        exec(code, run_globals)
      File "/tmp/.mount_qcomicOkkPRb/usr/app/qcomics/__main__.py", line 1, in <module>
        from qcomics.app import main
      File "/tmp/.mount_qcomicOkkPRb/usr/app/qcomics/app.py", line 9, in <module>
        from qcomics.reader import Reader
      File "/tmp/.mount_qcomicOkkPRb/usr/app/qcomics/reader.py", line 4, in <module>
        from PIL import Image
      File "/tmp/.mount_qcomicOkkPRb/usr/app_packages/PIL/Image.py", line 114, in <module>
        from . import _imaging as core
    ImportError: libtiff-102594ad.so.5.7.0: ELF load command address/offset not properly aligned

freakboy3742:
Agreed this definitely looks like a PIL problem. 

Have you also included libtiff in the `system_requires` definition? If you haven't, it's possible the problem is that PIL is trying to access a library outside the app image.

daedalus-0:
that was the issue, I missed one dependency, thanks a lot.

gambhiro:
@daedalus-0 what did your `system_requires` list look like? I ran into the same issue, but `system_requires = ["libtiff-dev",]` didn't solve it.

gambhiro:
For Linux I had to specify Pillow 7.2.0, the later versions produce the `libtiff ... offset not properly aligned` error when running the AppImage. I wonder what happened in Pillow that causes it.

Windows and MacOS works with Pillow 9.0.0.


``` toml
[tool.briefcase.app.helloapp.macOS]
requires = ["Pillow>=9.0.0"]

[tool.briefcase.app.helloapp.linux]
requires = ["Pillow==7.2.0"]
system_requires = ["libtiff-dev"]

[tool.briefcase.app.helloapp.windows]
requires = ["Pillow>=9.0.0"]

# And for Poetry:

[...]

Issue #458, opened 2020-07-22
can't use Pillow with briefcase 0.3.3 on linux


jgirardet:
**Describe the bug**
running an app with pillow crashes with the following error
```
[helloworld] Starting app...

Traceback (most recent call last):
  File "/tmp/.mount_Hello_dZoYVA/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/tmp/.mount_Hello_dZoYVA/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/tmp/.mount_Hello_dZoYVA/usr/app/helloworld/__main__.py", line 4, in <module>
    main()
  File "/tmp/.mount_Hello_dZoYVA/usr/app/helloworld/app.py", line 9, in main
    from PIL import Image
  File "/tmp/.mount_Hello_dZoYVA/usr/app_packages/PIL/Image.py", line 94, in <module>
    from . import _imaging as core
ImportError: libxcb-2dfad6c3.so.1.1.0: ELF load command address/offset not properly aligned
```

**To Reproduce**
1. clone https://github.com/jgirardet/briefpil
2. create env, activate, pip install briefcase
3. briefcase create && briefcase build && briefcase run

**Environment:**
 - Operating System: ubuntu 19.10
 - Python version: 3.8
 - Software versions:
   - Briefcase: 0.3.3
   - Pillow: 7.2

**Additional context**
adding  line 184 https://github.com/beeware/briefcase/blob/2259199dbe8b3049a92eccd27a7b94e6e6e558e0/src/briefcase/platforms/linux/appimage.py#L184:
```
so_folders.remove(self.appdir_path(app) / "usr/app_packages/PIL")
```
fixes the crash, I have no clear idea why.
full log: [scratch_1.txt](https://github.com/beeware/briefcase/files/4960467/scratch_1.txt)

freakboy3742:
Thanks for the report. The error suggests that the library adjustment process implemented by AppImage isn't working correctly. 

I'm a little surprised that removing the path from so_folders fixes the problem - that suggests that the contents of the app_packages folder may not need to be relocated. I'm not sure why that would be, though.

freakboy3742:
[...]

PR #759, opened 2022-06-10
Refs #718 - Modify AppImage builds to install dependencies from source

freakboy3742:
When building apps on Linux, users have reported errors of the form:

    libXXXX-<hash>.so.5.0.0: ELF load command address/offset not properly aligned

Apps that use Pillow are a common thread where this error is reported. in that instance, it's usually `libtiff` that raises the error (See #458, #618). 

These errors don't manifest when running on an Ubuntu 18.04 machine; they only appear when using a more recent machine (Ubuntu 22.04, Fedora 34).

I found [this comment from one of the linuxdeploy maintainers](https://github.com/linuxdeploy/linuxdeploy-plugin-conda/issues/39#issuecomment-792876427) which suggests there's something going on when the binary module is compiled. There's also [this comment on #458](https://github.com/beeware/briefcase/issues/458#issuecomment-1019271111) that suggests using an older version of Pillow will fix the error.

That led me to wonder if the issue is something about the interaction of manylinux binary wheels and Linuxdeploy. It's unlikely that an older version of Pillow would be doing anything radically different... but an older version of Pillow *would* be compiled from source, rather than being installed from a precompiled binary manylinux wheel.  

My testing has shown that installing all dependencies from source appears to fix the problem. This does impose the requirement that end-users need to specify more requirements in `system-requires`; and builds take longer (because all the binary modules need to be compiled); but... they *work*.

A little more investigation reveals that part of producing a ``manylinux`` wheel is running a tool called ``auditwheel`` - which does some library relocations that are similar to what linuxdeploy is doing (AFAICT).

So: this PR modifies the AppImage tooling to add `--no-binary :all:` to the arguments passed to pip install. This rejects binary wheels as an option, forcing all Python dependencies to be build from source.

It also modifies the `update` command on the AppImage backend so that it correctly inherits Docker wrapping behaviour (as well as the new pip argument handling). This would have led to incorrect behaviour if anyone tried to run `update` on a Linux app.

Tested was performed with the [Python support testbed](https://github.com/beeware/Python-support-testbed) project. This primarily exercises Pillow as a test case for this problem. The testbed will fail on the GTK family of tests without also using #756 - but the Pillow tests will pass. Without this patch, you get the "ELF load command address/offset not properly aligned" error on `libtiff` when PIL is imported.

I built the test AppImage on a macOS x86_64 machine in Docker. The AppImage worked on a Parallels install of:
* Ubuntu 18.04
* Ubuntu 22.04
* Fedora 34

Without this patch, the AppImage would work on Ubuntu 18.04, but fail on Ubuntu 22.04 and Fedora 34.
 
## PR Checklist:


freakboy3742:
Whelp... that's annoying. Pyside2 and Pyside6 are *only* available as binary wheels.

Back to the drawing board, I guess...

danyeaw:
> Whelp... that's annoying. Pyside2 and Pyside6 are _only_ available as binary wheels.

Drat! That was a good idea to explore.

freakboy3742:
I think the general approach is still sound - we will just need to poke a hole to allow for packages that only exist in binary form. My initial thought is either a new "requires_binary" setting, or some sort of annotation on the package (e.g. ":binary:pyside6"). I'd rather neither of these were needed... but then, I wish a lot of things about Linux packaging weren't true :-)

I may also need to spend some time with auditwheel to work out what it is doing to see if it's a universal problem, or something that only affects some packages (like Pillow).

The other approach to explore is conda. I don't know if conda does any processing similar to what auditwheel does in the default configuration; but in the extreme case, we could set up a custom channel that has packages built to Briefcase-compatible standards.

freakboy3742:
I'm going to close this; it's not quite as simple as I first thought (at least in part because of the issue picked up by CI). The good news is that on further investigation, I think there's a "documentation only" solution.

Issue #762, opened 2022-06-16
briefcase run android - Unable to create Android emulator


mozgokot:
Going trough beeware tutorial.

When trying to 
# briefcase run android

have an issue 

  Emulator name [beePhone]:

  Creating Android emulator beePhone...

  Unable to create Android emulator

The same in virt-env on AMD and Intel. Virtualization is on. Emulator in Android Studio works fine

PR #760, opened 2022-06-14
Support saving a detailed log to a file

rmartin16:
NOTE: #755 should be finalized and merged first so this PR can incorporate and accommodate those changes.

## Scope
Adds support to save all logging, including debug logging even if it isn't enabled, to a file. The functionality is automatically activated if briefcase exits due to an exception or if the new `--log` flag is provided.

## Design
The core of this approach uses Rich's "record" functionality to save all output in a buffer that is conditionally flushed to a file before briefcase exits. However, Rich will only "record" content that is actually printed; therefore, it was necessary to create a separate Rich `Console` to capture all logging.

Additionally, this new `Console` is configured to log to `os.devnull`; I did this to 1) avoid logging to an actual file before we knew if we wanted a file (although, really, it would be quite awkward to figure out where to put this file early enough) and 2) because this was the only thing I could figure out to get Rich to "record" the logging without any other side effects.

Open to any feedback while I polish up this PR.

## Outstanding Questions/Decisions
- Where should the logs be stored?
  - They are going to `~/.briefcase/logs/` right now.
- How long should logs live for?
  - Assuming logs should not live forever, what should be the maximum age? It's currently 14 days.
- How should they be named?
  - They are currently of the form `briefcase_2022_06_13-19_10_50.log`
- What should the logs contain?
  - I've added a header with these elements:
    - Date
    - Command line used to invoke `briefcase`
    - Briefcase version
    - Platform
    - Target platform for app
    - Target format for app
  - Then all of briefcase's logging is added via Rich's `log` functionality that adds the time and code location for each line.
  - Finally, the _full_ stacktrace for any exception.
  - I didn't add deep debug logs by default since env vars can contain sensitive information and it'd be annoying to try to sanitize the file before sharing. However, adding the `-vv` flag will add deep debug to the file.

## PR Checklist:

- [x] All new features have been tested
- [x] All new features have been documented
- [x] I have read the **CONTRIBUTING.md** file
- [x] I will abide by the code of conduct

PR #764, opened 2022-06-19
Log details of a fatal CalledProcessError even in non-debug mode

mhsmith:
This PR ensures that if Briefcase is terminated by a failed subprocess, then its command line, return code, stdout and stderr are logged to the console even in non-verbose mode.

This is implemented in `main` rather than in `integrations.subprocess`, because there are some subprocesses (e.g. `xcode-select --install`) where a non-zero return code is normal and is caught by the calling code. Such subprocesses are not affected by this PR.

Examples of issues where this would have helped:
* #762 -- We had to ask the user to re-run Briefcase with debug logging turned on.
* #395 -- The ADB output contained the error code "INSTALL_FAILED_ALREADY_EXISTS", and searching for that would have led directly to the solution.

## PR Checklist:

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,7 +2,6 @@
 source = src/briefcase
 omit =
     src/**/__init__.py
-    src/briefcase/__main__.py
 
 [report]
 show_missing = True
--- a/changes/764.feature.rst
+++ b/changes/764.feature.rst
@@ -1,0 +1,1 @@
+If Briefcase is terminated by a failed subprocess, then its command line, return code, stdout and stderr are now logged to the console even in non-verbose mode.
--- a/src/briefcase/__main__.py
+++ b/src/briefcase/__main__.py
@@ -1,14 +1,17 @@
+import subprocess
 import sys
 
 from .cmdline import parse_cmdline
 from .console import Log
 from .exceptions import BriefcaseError, HelpText
+from .integrations.subprocess import log_command, log_output, log_return_code
 
 
 def main():
     try:
         log = Log()
         command, options = parse_cmdline(sys.argv[1:])
+        log = command.logger
         command.parse_config("pyproject.toml")
         command(**options)
         result = 0
@@ -17,6 +20,7 @@ def main():
         log.info(str(e))
         result = e.error_code
     except BriefcaseError as e:
+        log_subprocess(log, e)
         log.error()
         log.error(str(e))
         result = e.error_code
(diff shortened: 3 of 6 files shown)


freakboy3742 (commented):
I definitely empathise with the use case, and I can see how this PR would address the problem you describe. My primary concern is that the while the output of the internal tools is very useful for *us* as developers of Briefcase, it isn't always especially meaningful for end-users - which is why we're wrapping those tools with a more approachable interface.

My immediate thought is that this may be made redundant by the work in #760. That PR will give us a complete log of everything that was done on a failed run without needing to ask the user to do an additional `--debug` run. It will also provide us a consistent user support process (can you send us the error log file) that will (hopefully) mean we get all those details in a form we can use (instead of a screenshot of the last 3 lines of the error message).

Any thoughts on whether there's a need for this in addition to what is provide by #760?

rmartin16:
I think the idea has merit....but I think its usefulness in practice may be limited.

For instance, if we consider that specific example where the Android emulator can't run because of some conflict with the host machine....then the emulator's output was already discarded back in `android_sdk.py` since it only ever existed within `emulator_popen.stdout/stderr` which is never read. That can be changed, of course....but that's just one example.

Moreover, I think a lot of the useful output is already shown in the terminal; so, I thin it is more about identifying when it is _not_ being show. The use of `subprocess.PIPE` outside of the Android emulator (and where the output isn't printed manually) is app signing on mac OS. The other situation where output is hidden is `check_output`; right now, the output is only ever printed as debug (error or otherwise). We could print the output more prominently if `check_output` raises an error.

So, I think there's potential here....but I also think the approach probably needs to be tweaked.

freakboy3742:
I think @rmartin16 has captured the core of my objection. While I agree that we should, wherever possible, empower our users to diagnose/debug problems on their own, and surfacing error output is a key part of that, I'm not convinced it's as simple as "surface all output in case of error". 

In the specific example you give (a failure starting the emulator), the useful error output is currently caught by a pipe and monitored by a series of `popen.poll()` calls. AFAICT, this PR wouldn't improve the situation for that case. To me, that suggests that a more targeted approach is required, looking at specific tasks that are error prone (like emulator startup), and ensuring that we aren't swallowing any details that might be useful.

mhsmith:
Thanks for the feedback. I'll take a more detailed look at the existing uses of the subprocess methods, and come back with an updated proposal.

mhsmith:
#760 is touching some of the same areas, so I'll wait for that PR to be merged before continuing.

freakboy3742:
@mhsmith Any thoughts on an updated version of this PR, or can we close this as wontfix/wontmerge?

mhsmith:
I think this issue should be revisited at some point, but I have higher priority things to work on so I'll close it for now.

PR #774, opened 2022-06-27
Docker usage tweaks

rmartin16:
- Revise wording in messages to users about docker usage.
- Containers created by `docker run` are now automatically deleted via `--rm`.
- A pseudo-tty is no longer created by `docker run`.

Using the `--tty` flag for `docker run` was then telling `pip` and `apt` to run as though they were printing to a terminal. Since the docker output is being streamed, this terminal output could become corrupted. Removing the `--tty` flag causes these programs to avoid outputing dynamic content or terminal control characters.

The inspiration for the `--rm` stems from discovering _a lot_ of orphaned containers in my dev environment. Given the containers created by `docker run` are not re-used (and actually re-using them is rather difficult), the `--rm` flag tells docker to delete the container with `docker run` exits.

## PR Checklist:

### DIFF
--- a/changes/774.feature.rst
+++ b/changes/774.feature.rst
@@ -1,0 +1,1 @@
+Temporary docker containers are now cleaned up after use. The wording of Docker progress messages has also been improved.
--- a/changes/774.misc.rst
+++ b/changes/774.misc.rst
@@ -1,1 +1,0 @@
-Corrects wording in messages about docker actions and temporary containers are now automatically deleted.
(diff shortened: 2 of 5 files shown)


freakboy3742 (approved):
👍 Good catch on both the ``--rm`` option, and the poor use of terminology in the Docker messages (and thanks to @mhsmith for the tty suggestion). 

My only comment is that this is more than a misc update - it's a substantive change worth a full inclusion in the release notes.

Issue #688, opened 2022-03-27
Briefcase installs 64-bit app to C:\Program Files(x86)\


t-arn:
**Describe the bug**
My Windows app with system_installer = true is installed in a subdirectory of C:\Program Files(x86)\

**To Reproduce**
1. set system_installer = true in Windows section of pyproject.toml
2. briefcase package winforms
3. install the msi package

**Expected behavior**
Since the Python support package is 64-bit, the app should be installed in a subdirectory of C:\Program Files\

**Environment:**
 - Operating System: Windows
 - Python version: 3.9
 - Software versions:
   - Briefcase: 0.3.6

freakboy3742:
Thanks for the report; from [a quick poke around](https://stackoverflow.com/questions/4304253/how-do-i-get-wix-to-generate-a-64-bit-msi), it appears that the underlying cause are some options in the WIX configuration file that are needed to flag the install product as being 64 bit; plus possibly an extra flag needed when invoking `candle`.

freakboy3742:
Fixed In #775.

PR #775, opened 2022-06-28
Fixes #382, #688 - Configure and correct the Windows install location

### LINKED ISSUE #382: User installable MSI packages?
The Mu Editor MSI file I managed to produce in the context of #381 requires administrative privileges for installation.

It would be very nice if `briefcase` had the option to produce user installable MSI files (assuming there is such a thing -- I suppose/hope so, given that lots of software is user installable, nowadays). If not, that could be considered a user-experience regression with respect to Mu users...

Which of user-install/system-install should be the default is a secondary thing for now.

Thoughts?

freakboy3742:
Modifies the MSI installer so that:

* It installs into ``Program Files``, rather than ``Program Files (x86)`` (since we only produce x64 binaries)
* If ``pyproject.toml`` does not express a preference for user or machine installs, the installer will ask the user which they want when the project is installed. If pyproject.toml contains an explicit `system_installer=True` or `system_installer=False`, the question will not be asked.

Requires the template changes in https://github.com/beeware/briefcase-windows-msi-template/pull/11. This can be tested by adding ``template_branch = "optional-user-install"`` in your pyproject.toml. **EDIT** These changes have now been merged into production.

To test the handling of installer scope, you will also need to delete any ``system_installer`` definition in pyproject.toml.

## PR Checklist:

### DIFF
--- a/changes/382.feature.rst
+++ b/changes/382.feature.rst
@@ -1,0 +1,1 @@
+Windows MSI installers can now be configured to ask the user whether they want a per-user or per-machine install.
--- a/changes/688.bugfix.rst
+++ b/changes/688.bugfix.rst
@@ -1,0 +1,1 @@
+Windows MSI installers now install in ``Program Files``, rather than ``Program Files (x86)``.
--- a/docs/reference/platforms/windows/msi.rst
+++ b/docs/reference/platforms/windows/msi.rst
@@ -43,7 +43,8 @@ installer; however, they are installed once and shared between all users on a
 computer.
 
 If ``true`` the installer will attempt to install the app as a per-machine app,
-available to all users. Defaults to a per-user install.
+available to all users. If ``false``, the installer will install as a per-user
+app. If undefined the installer will ask the user for their preference.
 
 ``version_triple``
 ------------------
--- a/src/briefcase/platforms/windows/msi.py
+++ b/src/briefcase/platforms/windows/msi.py
@@ -84,6 +84,10 @@ class WindowsMSICreateCommand(WindowsMSIMixin, CreateCommand):
             "version_triple": version_triple,
             "guid": str(guid),
             "install_scope": install_scope,
+            # Template forward compatibility flags
+            # 2022-06-29: #775 added the need to pass for -arch 64 to candle.exe;
+            # Briefcase v0.3.8 didn't use that flag.
+            "_use_arch64": True,
         }
 
     def install_app_support_package(self, app: BaseConfig):
--- a/src/briefcase/platforms/windows/msi.py
+++ b/src/briefcase/platforms/windows/msi.py
@@ -77,8 +77,8 @@ class WindowsMSICreateCommand(WindowsMSIMixin, CreateCommand):
         try:
             install_scope = "perMachine" if app.system_installer else "perUser"
         except AttributeError:
-            # system_installer not defined in config; default to perUser install.
-            install_scope = "perUser"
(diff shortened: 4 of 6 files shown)


freakboy3742:
The test failures are unfortunate, and point to an issue rolling this out. We can't land this PR until the template change has landed; but the template change requires the change in this PR. The two need to be rolled out simultaneously.

mhsmith:
The template projects currently have one branch per Python version. Would it make more sense for them to have a single branch, with a tag for each Briefcase version, and the Python version passed in using a template variable? That way:
* Briefcase and the template would always be consistent, and there would be no risk of a template update breaking all the old versions of Briefcase that people already have installed. 
* There would be no need to merge 4 branches for every template update.
* The user would still have the option of overriding the tag or branch in pyproject.toml.

t-arn:
Thank you Russel for fixing this issue!

I have one comment: in the PR description, you write: "(since we only produce x86 binaries)" 
I guess you meant to say: (since we only produce x64 binaries)...

freakboy3742:
> The template projects currently have one branch per Python version. Would it make more sense for them to have a single branch, with a tag for each Briefcase version, and the Python version passed in using a template variable? That way:
> 
> * Briefcase and the template would always be consistent, and there would be no risk of a template update breaking all the old versions of Briefcase that people already have installed.
> * There would be no need to merge 4 branches for every template update.
> * The user would still have the option of overriding the tag or branch in pyproject.toml.

A good idea on the surface - but with a couple of complications.

Firstly, the *current* template for *Windows* doesn't have any differences between versions - but there are significant differences in the iOS and macOS templates between versions. On iOS and macOS-Xcode, the differences are differences in the stub binary - between Python 3.5 and 3.10, there have been at least 2 changes in Python embedded startup that we've needed to accomodate. The macOS App template has an embedded binary which is, by definition, Python version dependent. At the very least, this "tag by Briefcase version" approach would need to be different depending on the platform. On top of that, I'm working on #629 right now which will (hopefully) add a stub binary for Windows (and probably a parallel "MSI-VSCode" template that will be subject to source code differences). 

Secondly, the existence of a 3.X branch is a way for us to gate whether we support 3.X in apps. This is especially important on Windows, because we use the Python Windows Embedded distribution directly, rather than a BeeWare-compiled support package. There was a long window where we weren't able to offer Python3.9 and 3.10 support, even though the "support library" was available. The template is the only piece we have consistent control over. This is less likely to be an issue as a result of Python.net 3 - they've made a change that makes the Python.net wheel an ANY wheel instead of a binary wheel - but I wouldn't rule out needing to gate Python versions in the future.

That said - you have hit on a fundamental problem: we don't currently have any way to version templates in line with Briefcase versions. In theory, the template change needed by this PR would represent a version bump because old versions of briefcase are not compatible with the "new" template required by this PR. Embedding the template hash/tag for a specific version in the Briefcase code would be one way to address this; but that would limit our ability to make smaller hot fixes to the template (or, at least, it would make hot fixes more difficult).

As a result of some shower thoughts last night, I think I have a way to work around this specific problem with some fancy footwork in the template; but I agree a more general approach to template versioning/upgrading may be required.

freakboy3742:
> I have one comment: in the PR description, you write: "(since we only produce x86 binaries)" I guess you meant to say: (since we only produce x64 binaries)...

🤦 Fixed. Thanks for the eagle eyes. Have you been able to verify this works for you?

PR #778, opened 2022-07-01
Restore Progress Bar to pip's Output

rmartin16:
This reverts the addition of the `--progress-bar=off` option for pip.

As I outline in more detail in my [comment](https://github.com/beeware/briefcase-linux-appimage-template/pull/15#discussion_r910277534) in beeware/briefcase-linux-appimage-template#15, when a tty is not available, pip (via rich) will not print a dynamic progress bar; instead, only the final state of the progress bar is printed. The final state of the progress bar is more familiar to people who use pip commonly and contains a few useful details about what happened.

## PR Checklist:


freakboy3742 (approved):
Agreed this is preferable behavior. It would be even better to have pip give an indication of how much is currently being downloaded in the background, but I can't see a way to accomodate that *and* the logging behavior.

PR #782, opened 2022-07-13
Include a stub app in Windows projects

### LINKED ISSUE #452: Can't use kivy with briefcase on windows
**Describe the bug**
`briefcase dev -d`  is working well
`briefcase run` fails

**To Reproduce**
1. clone https://github.com/jgirardet/kivybriefhello
2. briefcase dev -d => ok
3. briefcase run => fail no message
4. PS C:\Users\jim\rien\kivybrief\helloworld> & '.\windows\Hello World\src\python\python.exe' -m helloworld
```
[INFO   ] [Logger      ] Record log in C:\Users\jim\.kivy\logs\kivy_20-07-15_9.txt
[INFO   ] [Kivy        ] v1.11.0
[INFO   ] [Kivy        ] Installed at "C:\Users\jim\rien\kivybrief\helloworld\windows\Hello World\src\\app_packages\kivy\__init__.py"
[INFO   ] [Python      ] v3.7.6 (tags/v3.7.6:43364a7ae0, Dec 19 2019, 00:42:30) [MSC v.1916 64 bit (AMD64)]
[INFO   ] [Python      ] Interpreter at "C:\Users\jim\rien\kivybrief\helloworld\windows\Hello World\src\python\python.exe"
[INFO   ] [deps        ] Successfully imported "kivy_deps.angle" 0.1.9
[INFO   ] [deps        ] Successfully imported "kivy_deps.sdl2" 0.1.22
[INFO   ] [Factory     ] 184 symbols loaded
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_gif (img_sdl2, img_pil, img_ffpyplayer ignored)
[CRITICAL] [Window      ] Unable to find any valuable Window provider. Please enable debug logging (e.g. add -d if running from the command line, or change the log level in the config) and re-run your app to identify potential causes
sdl2 - ImportError: DLL load failed: Le module spécifié est introuvable.
  File "C:\Users\jim\rien\kivybrief\helloworld\windows\Hello World\src\\app_packages\kivy\core\__init__.py", line 63, in core_select_lib
    fromlist=[modulename], level=0)
  File "C:\Users\jim\rien\kivybrief\helloworld\windows\Hello World\src\\app_packages\kivy\core\window\window_sdl2.py", line 27, in <module>
    from kivy.core.window._window_sdl2 import _WindowSDL2Storage

[CRITICAL] [App         ] Unable to get a Window, abort.
Unable to start application 'helloworld'
```
This a usual error in kivy when missing dependencies. I don't think it's a kivy bug since `dev` is working well.
I looked at `app_packages` and every thing seems good.

**Environment:**
 - Operating System: windows 10
 - Python version: 3.7
 - Software versions:
   - Briefcase: 0.3.2
   - kivy: 1.11.0 (same thing with 2.0.0rc3)
  
I'm sorry ,I have no idea at all how to help on it.

### LINKED ISSUE #620: Add log streaming for Windows
#591 added app log tailing for every platform except Windows. This is due to the lack of an obvious way to access those logs because:

1. The way Briefcase packages apps for Windows doesn't provide an obvious way to capture logs for a running app
2. The way Briefcase starts the app on Windows isn't a good representation of how the app is actually started

As a result, it is difficult to diagnose why an app that works in development mode doesn't work in a packaged app.

At a minimum, `briefcase run` should surface stdout for the running app on Windows.

Ideally, the log of the app would also be available when the app has been started outside of Briefcase.

The suggested approach would be to introduce a "stub binary" similar to the one that macOS now uses. This stub binary could redirect stdout to somewhere readable. If that location is a logger, then it could also install a logging handler that writes logs to a known (and OS appropriate) location, and rotates those logs regularly.

### LINKED ISSUE #629: Add a stub app for Windows

### LINKED ISSUE #669: app_packages in the app bundle is not declared among the site packages
**Describe the bug**
The app_packages directory in the app bundle is not including among the site packages. I've found this prevents TensorFlow from loading its plugins, but I wonder if that may not induce bugs in other situations.

**To Reproduce**
1. Create an app with the code:
```
import logging
import site
import sys
logger = logging.getLogger('cookadream')
logger.setLevel(logging.INFO)
logger.addHandler(logging.FileHandler('/tmp/debug123.log'))
logger.info('sys.path = %s', sys.path)
logger.info('site.ENABLE_USER_SITE = %s', site.ENABLE_USER_SITE)
logger.info('site.USER_SITE = %s', site.USER_SITE)
logger.info('site.getsitepackages = %s', site.getsitepackages())
logger.info('site.getusersitepackages = %s', site.getusersitepackages())
```
2. Build and run the app with briefcase, check the contents of the file /tmp/debug123.log
3. app_packages appears in sys.path, but nowhere among the site packages

**Expected behavior**
Expected to see the bundle's app_packages both in the sys.path and somewhere among the site packages.

**Environment:**
 - Operating System: macOS 11.6.2
 - Python version: 3.8.9
 - Software versions:
   - Briefcase: 0.3.5

freakboy3742:
Introduces a stub binary as the entry point for Windows projects. This binary is an analog of what is used on macOS - a binary that starts, initialises the Python virtual machine, and starts an app module.

The name of the app module is read from the VersionInfo resource metadata that forms part of a Windows app. This metadata can be re-written on an existing app, so the app template can ship a template binary, and we can update that binary with the name of the app module (as well as other app metadata, such as version numbers and icons).

This ensures that when an app starts, it appears on the task manager as "app name", rather than "pythonw".

It ensures that the app appears on the taskbar etc with the app's icon, rather than the pythonw icon.

It also allows us to capture crashes, and expose the stack trace for the crash with a dialog box.

Lastly, it allows us to redirect stdout/err to a location that is visible. If there is an attached console, stdout/err are redirected to that console; if there is no console, a log file is written in the user's local AppData. These log files are rotated on app start.

To support this change, the `msi` backend has been renamed `app` - This better reflects the format of the output executable. MSI is a packaging format (the only one we support for Windows, at present), not an output executable format. A [briefcase-windows-app-template](https://github.com/beeware/briefcase-windows-app-template) repository has been created that contains stub binaries pre-compiled for each Python version. The older briefcase-windows-msi-template will be deprecated on the merge of this PR.

A Visualstudio backend (with a matching [briefcase-windows-VisualStudio-template](https://github.com/beeware/briefcase-windows-VisualStudio-template) repository) has also been created.

As part of these changes, the MSI template has also been modified, removing the x86/x64 compatibility shim (we only produce x64 binaries), and modifying the installation location to include the company (author) name in the install path. This is to better follow Windows conventions for app installation.

This stub binary will work with any Toga version; however, the stack trace dialog will work best with a version of toga that includes beeware/toga#1522.

* Fixes #629 
* Fixes #620

## PR Checklist:


src/briefcase/platforms/windows/__init__.py
```diff
@@ -1,5 +1,191 @@
...
+        :param app: The config object for the app
+        """
+        self.logger.info("Starting app...", prefix=app.app_name)
+        try:
+            # Start streaming logs for the app.
+            self.logger.info("=" * 75)
+            self.subprocess.run(
+                [os.fsdecode(self.binary_path(app))],
+                check=True,
+            )
```
rmartin16, line 90: I noticed that the new windows stub app output wasn't included in the briefcase log.....so, I added `stream_output=True` (and even `stderr=subprocess.STDOUT`) and I couldn't get this new output in to the briefcase log. The stub app is seemingly bypassing the pipe that python should be connecting to the app process' output....maybe windows is ignoring the pipe...?

Did you expect this? Or understand what might be happening?

rmartin16:
In general, everything basically worked for me. I went through both the `app` and `visualstudio` workflow from create->build->run->package->install->run from install and everything worked.

I downloaded a trial copy of Visual Studio Pro 2022 and installed with default settings; briefcase found the VS install path via `vswhere.exe`.

A few observations:
1) As noted in the comment, the output from the stub app as well as output from the app python's source can seemingly not be piped back through briefcase for logging.

2) Before the app is built, VS can't find some of the header files (e.g. `Python.h`). This is resolved, though, once the app is built once (though, it may require closing and re-opening the solution). I may just not understand VS well enough to see why this would be expected.

3) Debug builds do not work....but this feels more intentional since it seems like a debug version of python would be required in the support package.

4) The display in task manager felt a little awkward since it uses the project description as the grouper.
image

src/briefcase/platforms/windows/__init__.py
freakboy3742: ... I'm going to say some words that my mother will not be very happy to hear me say. 

Capturing stdout/err for a windows GUI app has taken me almost a week to get working *at all*. Windows GUI apps don't have a stdout/stderr by default, so we can't just redirect stdout/err. When you see the output in the terminal, you're not seeing stdout/stderr - you're seeing the output of a Windows specific device `CONOUT$`... and even then, you need to explicitly redirect stdout/err to `CONOUT$`. This is what the stub app does - if there's a console, it attaches and does the redirect; if there isn't it re-opens the stdout/err file descriptors as a file, and logs the output.

However, `subprocess.run()` doesn't have any handling for `CONOUT$`. It defines [`subprocess.STD_OUTPUT_HANDLE`](https://docs.python.org/3/library/subprocess.html#subprocess.STD_OUTPUT_HANDLE), but there doesn't appear to be any obvious way to capture this. From an initial search, [this stack overflow answer](https://stackoverflow.com/a/38749458) looks promising... if a little baroque.

PR #785, opened 2022-07-13
Include Prompts to the User in the Log File

rmartin16:
While we explicitly captured user's input to prompts, the prompts themselves were not captured in the log. Usually, the prompt can be inferred from surrounding logs but other times it is less clear. A simple example to demonstrate this is `briefcase create --log` when the app has already been created.

This PR simply includes the prompt in the log before the user's input value.

## PR Checklist:

### DIFF
--- a/changes/785.misc.rst
+++ b/changes/785.misc.rst
@@ -1,0 +1,1 @@
+The prompts to users are now included in the briefcase log file.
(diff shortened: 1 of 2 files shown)


freakboy3742 (approved):
👍

Issue #789, opened 2022-07-14
Android SDK Installation Path Cannot Contain Whitespace


rmartin16:
**Describe the bug**
Parts of the Android SDK do not support running from an installation path that contains whitespace, e.g. `/sdks/android sdk/`.

Briefcase's data directory was moved from `~/.briefcase` to OS-native locations in #777; for macOS, this location is `~/Library/Application Support/briefcase` and encountered this existing issue for Android SDK.

```
(venv-3.9) user@users-MacBook-Pro helloworld % briefcase run android -v

The Android tools provided by Google have license terms that you must accept
before you may use those tools.

>>> Running Command:
>>>     '/Users/user/Library/Application Support/briefcase/tools/android_sdk/cmdline-tools/latest/bin/sdkmanager' --licenses
>>> Environment Overrides:
>>>     ANDROID_SDK_ROOT=/Users/user/Library/Application Support/briefcase/tools/android_sdk
>>>     JAVA_HOME=/Users/user/Library/Application Support/briefcase/tools/java/Contents/Home
Error: Could not find or load main class Support.briefcase.tools.android_sdk.cmdline-tools.latest
>>> Return code: 1
```

In android/ndk#1400, it was noted this is unlikely to ever be fixed. And the internet is full of people encountering this [problem](https://stackoverflow.com/questions/37052934/android-sdk-location-should-not-contain-whitespace-as-this-cause-problems-with).

Of note, though, this is not limited to macOS.
On Linux, if you set `XDG_DATA_HOME=/mnt/my data/`, for example, you'll have the same problem.
On Windows, the user's home directory will contain a space if the user specifies a space in their "name" during account creation. Therefore, while the specific issue outlined above may not surface, other issues with Android SDK and whitespace could occur on Windows.

**To Reproduce**
Steps to reproduce the behavior (on macOS):
1. Install affected briefcase version: 
- `pip install git+https://github.com/beeware/briefcase.git@baf2942d08e09f8bb61fcecafb148904bedec6c2`
2. Run `briefcase new --no-input && cd helloworld && briefcase run android`

**Expected behavior**
The Android SDK should properly run from briefcase's default installation location regardless of platform or user/system configuration.

PR #790, opened 2022-07-14
Revert OS-native Data Directory

rmartin16:
In light of #789, this reverts the implementation of OS-native data directories in #777.

## PR Checklist:


rmartin16:
Given this problem stems from one tool being problematic, it feels like the nuclear option. After all, briefcase would ideally use these directories....and using `~/.briefcase` does not entirely ameliorate this problem because platforms do not entirely enforce the user's home path to not contain spaces; this is especially true on Windows.

The first alternative I considered was switching to `platformdirs.site_data_path()`.....however, on macOS that's just `/Library/Application Support/briefcase`....so we're in the same boat.

Some other possible alternatives:
1) Only on macOS, continue using `~/.briefcase`
2) Find a new location to install Android SDK for all platforms
3) Require users to manually install Android SDK (booo....)

freakboy3742:
Raising another option: we could *partially* revert this. Linux and Windows paths aren't affected by this problem unless the user has a space in their username, and if they do, the current `.briefcase` location will also be a problem. It's only macOS that is a problem across the board, so keeping the `.briefcase` location on macOS would be an improvement over v0.3.8 behavior on 2 of 3 platforms.

Another option worth putting into the mix (possibly as a long-term fix) is to allow for the user to define `.BRIEFCASE_HOME`. That location would take precedence over anything else, and would allow users who can't avoid having spaces in their usernames to have a workaround, as long as they can write to some other location.

PR #791, opened 2022-07-15
Use cache dirs rather than data dirs, and allow path customisation.

### LINKED ISSUE #374: Use appdirs to determine download cache directory instead of hardcoding ~/.briefcase

### LINKED ISSUE #789: Android SDK Installation Path Cannot Contain Whitespace

freakboy3742:
#777 modified the location of the Briefcase data store from `~/.briefcase` to a platform-specific data folder. 

However, it was then discovered that Android tooling *cannot* be stored in a location that has spaces in the path (#789). This is a problem because the default data location on macOS is `~/Library/Application Support`.

To address this, this PR modifies the default storage location from the user data folder to the user cache folder. This is `~/Library/Caches` on macOS, so it is "Android safe". The cache folder [was originally considered](https://github.com/beeware/briefcase/pull/777#pullrequestreview-1026800105) as a location, but rejected based on a comment about how the caches folder may be arbitrarily deleted. However, further investigation shows no evidence that this happens arbitrarily as a result of OS operation; and by way of social proof, pip also uses the cache location. 

Another benefit of the caches locations is that `~/Library/Application Support` is a location that is backed up by default by macOS, whereas `~/Library/Caches` is not; so this gives a small benefit to those using macOS backup services. 

Another related change - macOS convention is to use a bundle ID as part of the cache path; this isn't explicitly supported by platformdirs, but it's easy to add.

However, this change by itself won't fix the problem if the user has a space in their username. While this is uncommon, it is possible; to avoid the problem, this PR also introduces a ``BRIEFCASE_HOME`` environment variable that lets the user specify an arbitrary path for their Briefcase data. While this is a direct workaround for the Android "space-in-path" problem, it has the added benefit that it allows users that have disk space concerns to use an alternate storage location for Briefcase content.

Checks have been added to ensure that the user doesn't have a space in their data path. 

* Fixes #789
* Fixes #374 

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -126,7 +126,7 @@ class BaseCommand(ABC):
         self.host_arch = platform.machine()
         self.host_os = platform.system()
 
-        self.base_path = base_path
+        self.base_path = Path(base_path)
 
         # If a home path is provided during construction, use it. This usually
         # indicates we're under test conditions.
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -140,7 +140,11 @@ class BaseCommand(ABC):
         # isn't defined, use a platform-specific default data path.
         if data_path is None:
             try:
-                data_path = os.environ["BRIEFCASE_HOME"]
+                data_path = Path(os.environ["BRIEFCASE_HOME"])
+                if not data_path.exists():
+                    raise BriefcaseCommandError(
+                        "The path specified by BRIEFCASE_HOME does not exist."
+                    )
             except KeyError:
                 if self.host_os == "Darwin":
                     # macOS uses a bundle name, rather than just the app name
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -153,6 +153,22 @@ class BaseCommand(ABC):
                     appauthor="BeeWare",
                 ).user_cache_path
 
+        if " " in os.fsdecode(data_path):
+            raise BriefcaseCommandError(
+                f"""
+The location Briefcase will use to store tools and support files:
+
+    {data_path}
+
+contains spaces. This will cause problems with some tools, preventing
+you from building and packaging applications.
+
+You can set the environment variable BRIEFCASE_HOME to specify
+a custom location for Briefcase's tools.
+
+"""
+            )
+
         self.data_path = Path(data_path)
 
         self.tools_path = self.data_path / "tools"
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -140,8 +140,10 @@ class BaseCommand(ABC):
         # isn't defined, use a platform-specific default data path.
         if data_path is None:
             try:
-                data_path = Path(os.environ["BRIEFCASE_HOME"])
-                if not data_path.exists():
+                briefcase_home = os.environ["BRIEFCASE_HOME"]
+                data_path = Path(briefcase_home).resolve()
+                # Path("") converts to ".", so check for that edge case.
+                if briefcase_home == "" or not data_path.exists():
                     raise BriefcaseCommandError(
                         "The path specified by BRIEFCASE_HOME does not exist."
                     )
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -117,23 +117,49 @@ class BaseCommand(ABC):
     def __init__(
         self,
         base_path,
-        home_path=Path.home(),
-        data_path=PlatformDirs(appname="briefcase", appauthor="BeeWare").user_data_path,
+        home_path=None,
+        data_path=None,
         apps=None,
         input_enabled=True,
     ):
+        # Some details about the host machine
+        self.host_arch = platform.machine()
+        self.host_os = platform.system()
+
         self.base_path = base_path
-        self.home_path = home_path
-        self.data_path = data_path
+
+        # If a home path is provided during construction, use it. This usually
+        # indicates we're under test conditions.
+        if home_path is None:
+            home_path = Path.home()
+        self.home_path = Path(home_path)
+
+        # If a data path is provided during construction, use it. This usually
+        # indicates we're under test conditions. If there's no data path
+        # provided, look for a BRIEFCASE_HOME environment variable. If that
+        # isn't defined, use a platform-specific default data path.
+        if data_path is None:
+            try:
+                data_path = os.environ["BRIEFCASE_HOME"]
+            except KeyError:
+                if self.host_os == "Darwin":
+                    # macOS uses a bundle name, rather than just the app name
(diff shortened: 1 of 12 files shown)


src/briefcase/commands/base.py
```diff
@@ -117,24 +117,66 @@ class BaseCommand(ABC):
...
+            home_path = Path.home()
+        self.home_path = Path(home_path)
+
+        # If a data path is provided during construction, use it. This usually
+        # indicates we're under test conditions. If there's no data path
+        # provided, look for a BRIEFCASE_HOME environment variable. If that
+        # isn't defined, use a platform-specific default data path.
+        if data_path is None:
+            try:
+                data_path = os.environ["BRIEFCASE_HOME"]
```
rmartin16, line 143: I think it might be good to ensure that whatever is in `BRIEFCASE_HOME` can be created as a directory or already exists as a directory via `mkdir(parents=True, exists_ok=True)`....otherwise, the bomb out later could be other than intuitive. (may even be good to see if `touch()` is successful but that may be overkill.)

freakboy3742: This was *partially* intentional; if the user specifies something incorrect, I don't want to start arbitrarily creating directories. However, verifying that the directory actually exists is a good idea.

src/briefcase/commands/base.py
```diff
@@ -117,24 +117,66 @@ class BaseCommand(ABC):
...
+                    app_name = "org.beeware.briefcase"
+                else:
+                    app_name = "briefcase"
+
+                data_path = PlatformDirs(
+                    appname=app_name,
+                    appauthor="BeeWare",
+                ).user_cache_path
+
+        if " " in os.fsdecode(data_path):
```
rmartin16, line 156: If we really want to be absolute, checking all characters against `string.whitespace` may be best.....but I think I say this mostly tongue in cheek.

freakboy3742: Yeah... but you just know *someone* is going to put a tab in their username for giggles...

rmartin16: or a non-breaking space ;)

src/briefcase/commands/base.py
```diff
@@ -117,24 +117,70 @@ class BaseCommand(ABC):
...
+        self.home_path = Path(home_path)
+
+        # If a data path is provided during construction, use it. This usually
+        # indicates we're under test conditions. If there's no data path
+        # provided, look for a BRIEFCASE_HOME environment variable. If that
+        # isn't defined, use a platform-specific default data path.
+        if data_path is None:
+            try:
+                data_path = Path(os.environ["BRIEFCASE_HOME"])
+                if not data_path.exists():
```
rmartin16, line 144: Unfortunately, `Path("")` returns `PosixPath('.')`....so this lets an empty `BRIEFCASE_HOME` sneak by. I wouldn't consider this especially important if I hadn't tried `export <ENV VAR>=""` too many times in the past instead of `unset <ENV VAR>`.

freakboy3742: Good catch - I've added an explicit catch of this case.

PR #795, opened 2022-07-21
Allow print() calls from non-Briefcase code during tests

### LINKED ISSUE #794: PDB console output is swallowed under test conditions
PDB can be a useful tool for debugging when behavior isn't as expected in a test suite

As a result of recent changes to console handling, we monkeypatch `builtins.print` to ensure that there are no stray `print()` calls in code under test. 

However, this has the side effect of breaking pdb, as pdb uses `print` to output to the console - so any debug output is swallowed by the monkeypatched print.

**To Reproduce**
1. Set a breakpoint in a test case
2. Run the test case
3. Try to step through 

There's no error - but the expected output of pdb isn't displayed. The pdb *prompt* is displayed, and there is an echo of user commands, but the responses to those commands aren't displayed:

```
(venv3.10) rkm@stirtoni briefcase % pytest tests/integrations/linuxdeploy/test_LinuxDeploy__verify_plugins.py
.
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
(Pdb) tool_path
(Pdb) n
(Pdb) 
(Pdb) n
(Pdb) 
(Pdb) q

Ran 1 tests in 10.51s

OK
```
**Expected behavior**

It should be possible to use pdb to debug a test case.

**Environment:**
 - Operating System: All
 - Python version: All
 - Software versions:
   - Briefcase: 0.3.8

**Additional context**

I suspect it may be possible to make the monkeypatch redirect back to print if pdb is currently active (by detecting pdb in the stack trace somewhere).

rmartin16:
d0b74c6 removed the fallback in the tests' monkeypatched `builtin.print` to actually print if the calling code wasn't in Briefcase. My assumption at the time must have been "well, `console.py` no longer uses `print()` directly, so no need to support `print()` at all during tests".

This PR updates the monkeypatch to continue failing tests if `print()` is called from any Briefcase code but uses of `print()` outside of Briefcase code are allowed. This specifically addresses #794 to allow `pdb` to use `print()`.

- Fixes #794.

## PR Checklist:

### DIFF
--- a/changes/794.misc.rst
+++ b/changes/794.misc.rst
@@ -1,0 +1,1 @@
+Allow non-Briefcase code to use print() during tests.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,19 +21,22 @@ def mock_git():
     return git
 
 
-def monkeypatched_print(*arg, **kwargs):
-    """Raise an error for calls to print."""
+# alias print() to allow non-briefcase code to use it
+_print = print
+
+
+def monkeypatched_print(*args, **kwargs):
+    """Raise an error for calls to print() from briefcase."""
     frame = inspect.currentframe().f_back
     module = inspect.getmodule(frame.f_code)
 
     # Disallow any use of a bare print() in the briefcase codebase
-    if (
-        module.__name__.startswith("briefcase.")
-        and module.__name__ != "briefcase.console"
-    ):
+    if module.__name__.startswith("briefcase."):
         pytest.fail(
             "print() should not be invoked directly. Use Log or Console for printing."
         )
+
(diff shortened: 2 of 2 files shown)


freakboy3742 (approved):
Awesome - thanks!

PR #797, opened 2022-07-22
Ensure custom support URLs don't collide in the download cache

freakboy3742:
This came up in the context of the work of #756.

At present, support packages are all cached in `<data path>/support`, regardless of origin.

This means that:
* An official macOS 3.10 b3 support package
* https://example.com/development/Python-3.10-macOS-support.b3.tar.gz
* https://example.com/release/Python-3.10-macOS-support.b3.tar.gz
* https://example.com/archive/?platform=macOS&release=b3 (returning Python-3.10-macOS-support.b3.tar.gz)

will all save the same file: `<data path>/support/Python-3.10-macOS-support.b3.tar.gz`. Since this location is used as a download cache, the first version of the package that is downloaded will become the version that is used, regardless of local project configuration. If you have 2 apps, one using the default location, and one using a custom location, the support package that will be used on a specific user's machine will depend on which app they built first. This would also cause a problem if a single project changes the URL for the custom support package but doesn't change the package name; or adds a custom support URL to a project and the URL using the same filename as a default support package.

#756 addresses this with linuxdeploy plugins by differentiating between "official" URLs and custom URLs; and adding a hash of the URL to the cache path for custom URLs. 

This PR applies the same approach to support URLs. Using this approach, the four examples provided above would cache as:
* `<data path>/support/Python-3.10-macOS-support.b3.tar.gz` (same as before)
* `<data path>/support/4dd6df92767f88ba4309d3690bb16f7c7e915c6f1ac46cc79495f145c1d07b26/Python-3.10-macOS-support.b3.tar.gz`
* `<data path>/support/5f7b7dbfdd014859b4032fc3d99e298fa783a88c37e05c45ece1f5f40c337b00/Python-3.10-macOS-support.b3.tar.gz`
* `<data path>/support/0016f4c533b48e84f03663aa7af8249a99f85c21255812d819ed1f46325acbdb/Python-3.10-macOS-support.b3.tar.gz`

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/create.py
+++ b/src/briefcase/commands/create.py
@@ -1,3 +1,4 @@
+import hashlib
 import os
 import platform
 import shutil
@@ -315,6 +316,21 @@ class CreateCommand(BaseCommand):
                 self.logger.info(f"Using support package {support_package_url}")
 
             if support_package_url.startswith(("https://", "http://")):
+                if custom_support_package:
+                    # If the support package is custom, cache it using a hash of
+                    # the download URL. This is needed to differentiate to support
+                    # packages with the same filename, served at different URLs.
+                    # (or a custom package that collides with an official package name)
+                    download_path = (
+                        self.data_path
+                        / "support"
+                        / hashlib.sha256(
+                            support_package_url.encode("utf-8")
+                        ).hexdigest()
+                    )
+                else:
+                    download_path = self.data_path / "support"
+
                 try:
                     self.logger.info(f"... pinned to revision {app.support_revision}")
                     # If a revision has been specified, add the revision
@@ -336,7 +352,7 @@ class CreateCommand(BaseCommand):
                 # in the user's briefcase support cache directory.
                 support_filename = self.download_url(
                     url=support_package_url,
-                    download_path=self.data_path / "support",
+                    download_path=download_path,
                 )
             else:
                 support_filename = Path(support_package_url)
@@ -356,9 +372,9 @@ class CreateCommand(BaseCommand):
             with self.input.wait_bar("Unpacking support package..."):
                 support_path = self.support_path(app)
                 support_path.mkdir(parents=True, exist_ok=True)
-                # TODO: Py3.6 compatibility; os.fsdecode not required in Py3.7
                 self.shutil.unpack_archive(
-                    os.fsdecode(support_filename), extract_dir=os.fsdecode(support_path)
+                    support_filename,
+                    extract_dir=support_path,
                 )
         except (shutil.ReadError, EOFError) as e:
             raise InvalidSupportPackage(support_package_url) from e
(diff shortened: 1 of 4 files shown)


src/briefcase/commands/create.py
```diff
@@ -315,6 +316,21 @@ def install_app_support_package(self, app: BaseConfig):
                 self.logger.info(f"Using support package {support_package_url}")
 
             if support_package_url.startswith(("https://", "http://")):
+                if custom_support_package:
+                    # If the support package is custom, cache it using a hash of
+                    # the download URL. This is needed to differentiate to support
+                    # packages with the same filename, served at different URLs.
```
mhsmith, line 322: If a `revision` argument is added to the URL below, should it be included in the hash? The standard BeeWare URL always returns different filenames for different revisions., but a custom URL might not.

tests/commands/create/test_install_app_support_package.py
```diff
@@ -214,22 +298,42 @@ def test_install_pinned_custom_app_support_package_url_with_args(
...
-        download_path=create_command.data_path / "support",
+        download_path=create_command.data_path
+        / "support"
+        / "eb2519992f3776ae59dd0eb1ed6c79fa446a438e2bf4a6d5d61f1adad922f2dd",
         url="https://example.com/custom/support.zip?cool=Yes&revision=42",
     )
 
+    # Confirm the right file was unpacked
+    create_command.shutil.unpack_archive.assert_called_with(
+        support_file,
```
mhsmith, line 333: Since `support_file` has a different hash to `download_path`, at first sight this doesn't look as if the right file was unpacked, but that's explained by the way that `download_url` is mocked.

 It would be clearer if the mock `download_url` function was made a little more realistic by basing its return value on its `download_path` parameter, rather than a pre-determined value. 

In fact, we could go further by making the mock `download_url` actually create the file as a side-effect. That way, all copies of the ZipFile creation code (lines 308-310) could be factored out into the mock function.

src/briefcase/commands/create.py
freakboy3742: Good catch. Fixed.

tests/commands/create/test_install_app_support_package.py
freakboy3742: Yeah - that's a good idea. I'll make this a proper download side effect function. This is also as good a time as any to add some generic utility functions for download side effects and setting up dummy files.

freakboy3742: Ok - I've added some utilities to mock file creation and downloads. These functions could be used to clean up a lot of the test codebase, but we can introduce the use of these methods piecemeal (or set them as tasks for contributors who want to get their feet wet with a simple task)

freakboy3742:
@mhsmith Ready for a re-review whenever you're ready.

mhsmith (approved):

PR #803, opened 2022-07-28
Fixes #662 - Set a current working directory when running apps

### LINKED ISSUE #662: AppImage loads Python code from project's sources, not from inside of AppImage

freakboy3742:
By default, the Python interpreter will include the current working directory in the PYTHONPATH. As a result, if you create an app named `helloworld`, and create a `helloworld.py` in the root directory of the project, `briefcase dev` would pick up the `helloworld.py` in the project root, not the actual app code. 

This would also affect Linux AppImage apps. As Linux AppImages are an embedded Python interpreter, the folder where you invoke the AppImage would implicitly be on the PYTHONPATH. If you gave a `helloworld` AppImage to an end-user who has `helloworld.py` in their local folder, the *end-user's* `helloworld.py` would be executed.

macOS and Windows aren't affected by this problem, as they use an embedded interpreter that doesn't add CWD to the PythonPath. However it doesn't hurt to set the CWD for those run calls.

iOS and Android aren't affected because they're running on device in a sandbox, so they can't be affected by local code.

Linux Flatpaks are affected by this in a slightly different way, as they run with a CWD of `~`; so a `helloworld.py` in the user's home folder would get picked up. Adopting #628 (implemented in beeware/briefcase-linux-flatpak-template#3) would address this, as it would make Flatpaks isolated in the same way as Windows/macOS apps.

This PR addresses the problem by setting the CWD to the user's home folder when running the app, or when in dev mode. This doesn't provide perfect isolation in dev mode, but it's significantly better than the status quo, and leads to easier-to-diagnose end-user questions of "Why is Briefcase finding code in my home folder?", rather than "Why *isn't* briefcase finding the code in my project?" or "Why is briefcase picking up a stale version of my app?". Running apps aren't affected by this decision (other than the Flatpak leakage mentioned above, which is fixed in the linked template PR).

By way of debugging assistance, this PR also adds the CWD into the logged subprocess output.

beeware/briefcase-linux-appimage-template#17 includes a fix on the AppImage binary to force better isolation of the `AppImage itself.

## PR Checklist:

### DIFF
--- a/changes/662.bugfix.rst
+++ b/changes/662.bugfix.rst
@@ -1,0 +1,1 @@
+Apps now have better isolation against the current working directory. This ensures that code in the current working directory isn't inadvertently included when an app runs.
--- a/src/briefcase/commands/dev.py
+++ b/src/briefcase/commands/dev.py
@@ -87,7 +87,15 @@ class DevCommand(BaseCommand):
         try:
             # Invoke the app.
             self.subprocess.run(
-                [sys.executable, "-m", app.module_name],
+                [
+                    sys.executable,
+                    "-c",
+                    (
+                        "import runpy, sys;"
+                        "sys.path.pop(0);"
+                        f'runpy.run_module("{app.module_name}", run_name="__main__", alter_sys=True)'
+                    ),
+                ],
                 env=env,
                 check=True,
                 cwd=self.home_path,
--- a/src/briefcase/commands/dev.py
+++ b/src/briefcase/commands/dev.py
@@ -90,7 +90,7 @@ class DevCommand(BaseCommand):
                 [sys.executable, "-m", app.module_name],
                 env=env,
                 check=True,
-                cwd=self.platform_path,
+                cwd=self.home_path,
             )
         except subprocess.CalledProcessError as e:
             raise BriefcaseCommandError(
--- a/src/briefcase/commands/dev.py
+++ b/src/briefcase/commands/dev.py
@@ -1,6 +1,7 @@
 import os
 import subprocess
 import sys
+from pathlib import Path
 from typing import Optional
 
 from briefcase.config import BaseConfig
@@ -89,6 +90,7 @@ class DevCommand(BaseCommand):
                 [sys.executable, "-m", app.module_name],
                 env=env,
                 check=True,
+                cwd=self.platform_path,
             )
         except subprocess.CalledProcessError as e:
             raise BriefcaseCommandError(
@@ -98,7 +100,11 @@ class DevCommand(BaseCommand):
     def get_environment(self, app):
         # Create a shell environment where PYTHONPATH points to the source
         # directories described by the app config.
-        return {"PYTHONPATH": os.pathsep.join(app.PYTHONPATH)}
+        return {
+            "PYTHONPATH": os.pathsep.join(
+                os.fsdecode(Path.cwd() / path) for path in app.PYTHONPATH
+            )
+        }
 
     def __call__(
         self,
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -6,6 +6,7 @@ import subprocess
 import sys
 import threading
 import time
+from pathlib import Path
 
 import psutil
 
@@ -217,6 +218,7 @@ class Subprocess:
 
         # Otherwise, invoke run() normally.
         self._log_command(args)
+        self._log_cwd(kwargs.get("cwd"))
         self._log_environment(kwargs.get("env"))
 
         try:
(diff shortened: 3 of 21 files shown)


mhsmith:
> This PR addresses the problem by setting the CWD to the user's home folder when running the app, or when in dev mode. This doesn't provide perfect isolation in dev mode, but it's significantly better than the status quo, and leads to easier-to-diagnose end-user questions of "Why is Briefcase finding code in my home folder?", rather than "Why isn't briefcase finding the code in my project?" or "Why is briefcase picking up a stale version of my app?".

I think unexpectedly running something in the home folder could still really confuse a user. And it's not unusual for someone to drop a file in their home directory and then forget about it for months.

Ideally we would simply tell Python not to add the current directory to sys.path. Unfortunately, it looks like there isn't an option for this until the upcoming Python 3.11, which adds a `-P` option that does exactly this. In older Python versions the only way I can find is to use `-I`, which has some other undesired effects.

However, we could still achieve the desired effect by changing the Python command line from this:
```
python -m helloworld
```
To this:
```
python -c 'import runpy, sys; sys.path.pop(0); runpy.run_module("helloworld", run_name="__main__")'
```

If that works for `briefcase dev`, then I don't see any reason to change the current directory for `briefcase run` either, because just remaining in the directory where the command was executed means one less thing for the user to learn. 

Of course, for `briefcase` commands the current directory is always the project root directory, but an AppImage could be launched from anywhere. And adding a `cd` command to the launch script, as https://github.com/beeware/briefcase-linux-appimage-template/pull/17 does, would make it impossible for the app to contain code that depends on the current directory. For example, relative paths on the app's command line wouldn't work correctly. So if the `python -c` technique works for `briefcase dev`, I think it should be used for AppImage as well.

freakboy3742:
I completely agree that `cwd=~` isn't an ideal solution - it's more a case of the "best of a bunch of bad options" (at least within the scope of using cwd as the solution). 

However, I think you're right that the `runpy` approach is a better one - as long as we also add `alter_sys=True`. Without that extra argument, `sys.modules['__main__'].__package__` no longer identifies the app module used to start the app. Toga uses this as one (of multiple) approaches it uses to identify the running app. With `alter_sys=True`, I can't see any difference between `runpy` and `-m module`.

I think it's still worth retaining the `cwd` change though, if only to disconnect any perception that the app is running in the project directory. We get plenty of support queries that amount to "the app runs fine in dev mode, but breaks in app mode", which come down to absolute vs relative path handling; using a path that is very clearly *not* in the project can only help to reinforce this.

mhsmith:
> I think it's still worth retaining the `cwd` change though, if only to disconnect any perception that the app is running in the project directory. We get plenty of support queries that amount to "the app runs fine in dev mode, but breaks in app mode", which come down to absolute vs relative path handling; using a path that is very clearly not in the project can only help to reinforce this.

Good point: I've seen from the Chaquopy issue tracker that a lot of beginner programmers don't understand the concept of a working directory at all. This should help catch that kind of bug earlier.

mhsmith (approved):

PR #805, opened 2022-07-29
If support_path is missing from briefcase.toml, don't download a support package

mhsmith:
Chaquopy installs all of its runtime libraries with a Gradle plugin, so it won't need a support package. This PR provides a way for a template to disable support packages entirely.

## PR Checklist:


freakboy3742 (approved):
I've made a small tweak to the arguments passed to `_unpack_support_file()` which avoids the need for the dummy retrieval of `support_path()`; but otherwise, this looks great.

PR #810, opened 2022-08-04
Allow nested wait bars by updating onscreen message

rmartin16:
Allows a Wait Bar to be initiated while a Wait Bar is already active. Previously this would result in an Exception from Rich related to multiple Live displays. So support this, the `message` for the currently active Wait Bar is updated and a new manager is entered; in this way, as each Wait Bar exits, the previous Wait Bar is restored.

Nested Wait Bar Demo

```python
from time import sleep
from briefcase.console import Console
#
console = Console()
#
with console.wait_bar("Wait Bar 1..."):
    console.prompt("Inside the first wait bar")
    sleep(1)
    #
    with console.wait_bar("Wait Bar 2 (transient)...", transient=True):
        console.prompt("Inside the second wait bar")
        sleep(1)
        #
        with console.wait_bar("Wait Bar 3..."):
            console.prompt("Inside the third wait bar")
            for idx in range(3):
                console.prompt(idx)
                sleep(.5)
            console.prompt("Leaving third wait bar")
        #
        sleep(1)
        # raise Exception("error")
        console.prompt("Leaving second wait bar")
    #
    sleep(1)
    console.prompt("Leaving first wait bar")
```

## PR Checklist:

### DIFF
--- a/tests/commands/create/test_install_image.py
+++ b/tests/commands/create/test_install_image.py
@@ -51,7 +51,7 @@ def test_no_requested_size(create_command, tmp_path, capsys):
     )
 
     # The right message was written to output
-    expected = "\nInstalling input/original.png as sample image... done\n"
+    expected = "Installing input/original.png as sample image... done\n\n"
     assert capsys.readouterr().out == expected
 
     # The file was copied into position
@@ -62,7 +62,7 @@ def test_no_requested_size(create_command, tmp_path, capsys):
 
 
 def test_no_requested_size_invalid_path(create_command, tmp_path, capsys):
-    """If the app specifies an no-size image that doesn't exist, an error is
+    """If the app specifies a no-size image that doesn't exist, an error is
     raised."""
     create_command.shutil = mock.MagicMock()
     create_command.shutil.copy.side_effect = FileNotFoundError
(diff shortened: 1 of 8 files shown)


tests/commands/create/test_install_image.py
```diff
@@ -51,7 +51,7 @@ def test_no_requested_size(create_command, tmp_path, capsys):
     )
 
     # The right message was written to output
-    expected = "\nInstalling input/original.png as sample image... done\n"
+    expected = "Installing input/original.png as sample image... done\n\n"
     assert capsys.readouterr().out == expected
```
rmartin16, line 55: Sooo....I don't really understand this change. When I use the Wait Bar, I don't see line breaks like this or any changes from how it worked in the past. My best guess is something subtle has changed about the way the dynamism of the Wait Bar is interacting with `capsys`...

freakboy3742: I can't spot any differences in "real world" output either; capsys is doing enough weirdness that I'm happy to live with this as an explanation for now.

freakboy3742 (approved):
I've added one additional test around the defaulting of the `bufsize` argument, but otherwise, this looks solid.

PR #817, opened 2022-08-09
Avoid deadlocks while joining the output streamer

### LINKED ISSUE #809: On Windows, pressing Ctrl-C during an Android `build` or `run` causes briefcase to hang
Since the recent Briefcase logging changes, I can no longer interrupt an Android `build` or `run` on Windows. @rmartin16: You might be interested in this.

The `build` command launches two java.exe processes: one Gradle client and one Gradle daemon. Pressing Ctrl-C causes the daemon process to exit, but there is no feedback in the console, and the "Building" waitbar continues to animate. Pressing Ctrl-C a second time displays the message `Terminate batch job (Y/N)?`, but entering `y` has no effect. Killing the other java.exe process (the Gradle client) has no effect either. The only way I can get back to the command line prompt is to kill both briefcase.exe and python.exe. (This happens in both the cmd and MSYS2 terminals.)

There's a similar problem when `briefcase run` is displaying the Logcat. Ctrl-C kills the adb.exe process, but there is no feedback in the console, and Briefcase continues to run until I kill it with something outside of the terminal. (This only happens for me in a cmd terminal: MSYS2 works fine.)

**Expected behavior**
In all situations, Ctrl-C should promptly shut everything down and return to the command line prompt.

**Environment:**
 - Python version: 3.8
 - Software versions:
   - Briefcase: 8df2d112e

rmartin16:
Interrupting the joining of a thread with CTRL+C can lead to deadlocking certain versions of Python (python/cpython#89437).

Changes:
 - Replaces calls to `join()` the streamer thread with waiting on `is_alive()`.
 - Instantiate the streamer thread as a daemon to prevent Python itself trying to join the thread.
 - Remove all logic from streamer thread that is not necessary to stream output.
   - This limits any race conditions between the streamer finishing and Python exiting.
   - This was noticeably a problem when the streamer thread was trying to print the return code.
 - As seen in Python itself, give some [cushion of time](https://github.com/python/cpython/blob/main/Lib/subprocess.py#L911) for the CTRL+C to propagate to child processes.
   - python/cpython#70130
   - This is necessary since subprocess commands run by `briefcase` can exit differently based on whether they are aborted by the user or encounter a different error.
   - For instance, when `adb` is streaming logs, it will emit a return code of `0xC000013A` when exiting from CTRL+C.

 - Fixes #809

CC: @mhsmith 

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -5,8 +5,8 @@ import shlex
 import subprocess
 import sys
 import threading
-import time
 from pathlib import Path
+from time import sleep, time
 
 import psutil
 
@@ -277,6 +277,7 @@ class Subprocess:
             if process.stderr and kwargs["stderr"] != subprocess.STDOUT:
                 stderr = process.stderr.read()
             return_code = process.poll()
+        self._log_return_code(return_code)
 
         if check and return_code:
             raise subprocess.CalledProcessError(return_code, args, stderr=stderr)
@@ -367,7 +368,7 @@ class Subprocess:
             [str(arg) for arg in args], **self.final_kwargs(**kwargs)
         )
 
-    def stream_output(self, label, popen_process, stop_func=None):
+    def stream_output(self, label, popen_process, stop_func=lambda: False):
         """Stream the output of a Popen process until the process exits. If the
         user sends CTRL+C, the process will be terminated.
 
@@ -385,19 +386,22 @@ class Subprocess:
             name=f"{label} output streamer",
             target=self._stream_output_thread,
             args=(popen_process,),
+            daemon=True,
         )
         try:
             output_streamer.start()
-            if stop_func:
-                while output_streamer.is_alive() and not stop_func():
-                    time.sleep(0.1)
-            else:
-                output_streamer.join()
+            # joining the thread is avoided due to demonstrated
+            # instability of thread interruption via CTRL+C (#809)
+            while not stop_func() and output_streamer.is_alive():
+                sleep(0.1)
         except KeyboardInterrupt:
-            pass  # allow CTRL+C to gracefully stop streaming
+            # allow time for CTRL+C to propagate to the child process
+            sleep(0.25)
         finally:
             self.cleanup(label, popen_process)
-            output_streamer.join()
+            streamer_deadline = time() + 3
+            while output_streamer.is_alive() and time() < streamer_deadline:
+                sleep(0.1)
 
     def _stream_output_thread(self, popen_process):
         """Stream output for a Popen process in a Thread.
@@ -411,13 +415,7 @@ class Subprocess:
             if output_line:
                 self.command.logger.info(output_line)
             elif output_line == "":
-                # a return code will be available once the process returns one to the OS.
-                # by definition, that should mean the process has exited.
-                return_code = popen_process.poll()
-                # only return once all output has been read and the process has exited.
-                if return_code is not None:
-                    self._log_return_code(return_code)
-                    return
+                return
 
     def cleanup(self, label, popen_process):
         """Clean up after a Popen process, gracefully terminating if possible;
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -5,8 +5,8 @@ import shlex
 import subprocess
 import sys
 import threading
+import time
 from pathlib import Path
-from time import sleep, time
 
 import psutil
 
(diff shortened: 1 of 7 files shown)


src/briefcase/integrations/subprocess.py
```diff
@@ -385,19 +385,23 @@ def stream_output(self, label, popen_process, stop_func=None):
...
-            pass  # allow CTRL+C to gracefully stop streaming
+            # allow time for CTRL+C to propagate to the child process
+            sleep(0.25)
         finally:
             self.cleanup(label, popen_process)
-            output_streamer.join()
+            streamer_deadline = time() + 3
+            while output_streamer.is_alive() and time() < streamer_deadline:
+                sleep(0.1)
+            self._log_return_code(popen_process.poll())
```
rmartin16, line 405: This waiting around for the streamer is rather arbitrary.

In any expected case, `is_alive()` will return `False` and we continue on.

It's at least theoretically possible `is_alive()` returns `True`. This should mean the streamer is flooding the console with output or `readline()` is stuck waiting for `EOF`......or something worse.

That said, waiting around forever (like we were) would probably be worse....and moving forward prematurely presents the possibly to mix subprocess output with `briefcase` output....although, at that point, the output wouldn't likely be very valuable anyway.

tests/integrations/subprocess/test_Subprocess__stream_output.py
```diff
@@ -97,5 +69,34 @@ def test_stop_func(mock_sub, popen_process, stop_func_ret_val, capsys):
     mock_sub.stream_output(
         "testing", popen_process, stop_func=lambda: stop_func_ret_val
     )
-    assert capsys.readouterr().out == ("output line 1\n" "\n" "output line 3\n")
+    assert capsys.readouterr().out == "output line 1\n" "\n" "output line 3\n"
     mock_sub.cleanup.assert_called_once_with("testing", popen_process)
+
+
+def test_stuck_streamer(mock_sub, popen_process, monkeypatch, capsys):
```
rmartin16, line 76: definitely open to any feedback for potential improvement

freakboy3742: Yeah - this is hard to test; this is about as good as it's going to get AFAICT. 

I've pushed a change that mocks `time.time()`; this allows us to shortcut the loop so it doesn't take a full 3 seconds to finish. Ideally, we'd mock `time.sleep()` as well - but if you do that, the parent thread never releases control to the child thread (at least, it doesn't in my testing), so *some* sleeping is required to test the behaviour of the subthread.

freakboy3742 (approved):
This looks good to me, and works well in my testing. I'll wait for @mhsmith to confirm it works for him (since he was the original reporter), then merge once he's given a thumbs up.

tests/integrations/subprocess/test_Subprocess__stream_output.py
```diff
@@ -63,22 +50,17 @@ def test_keyboard_interrupt(mock_sub, popen_process, capsys):
...
     send_ctrl_c.side_effect = [False, KeyboardInterrupt]
 
     mock_sub.stream_output("testing", popen_process, stop_func=send_ctrl_c)
 
-    assert capsys.readouterr().out == ("output line 1\n" "\n" "output line 3\n")
+    assert (
+        capsys.readouterr().out == "output line 1\n"
+        "\n"
+        "output line 3\n"
+        "Stopping...\n"
```
freakboy3742, line 62: Ok - this is really weird; I'm not seeing the CI failure on any of my test configurations. Investigating...

freakboy3742: Found the source of the problem. Since the streamer thread isn't doing any substantial work, it was completely draining and exiting before we did our second stop check. The emergence of this behavior will be completely non-deterministic, as it depends on OS-level thread scheduling; however, we can strongly encourage the right behavior by adding a sleep to gently suggest to the OS that the streamer thread should suspend for a bit.

mhsmith (commented):
Works fine for me in both MSYS2 and cmd, when interrupting both Gradle and Logcat. Just one maintainability comment:

src/briefcase/integrations/subprocess.py
```diff
@@ -385,19 +386,27 @@ def stream_output(self, label, popen_process, stop_func=None):
...
+                time.sleep(0.1)
         except KeyboardInterrupt:
-            pass  # allow CTRL+C to gracefully stop streaming
+            self.command.logger.info("Stopping...")
+            # allow time for CTRL+C to propagate to the child process
+            time.sleep(0.25)
         finally:
             self.cleanup(label, popen_process)
-            output_streamer.join()
+            streamer_deadline = time.time() + 3
```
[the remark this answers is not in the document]
rmartin16: Can you please help me understand why these waits should be the same?

As best I can tell, they will happen sequentially and quite independently of each other. Further, I was initially considering making this final wait shorter....lest the user gets frustrated that their first CTRL+C wasn't working out and they just started sending ever more...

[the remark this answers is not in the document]
mhsmith: Sorry, you're quite right, I got confused by the threading. The first timeout is to shut down the subprocess, and the second timeout is to shut down the thread reading the process output, which may be delayed slightly in buffers. So this is fine the way it is.

rmartin16:
I didn't want to add any more changes to this at this point but the originating issue of interrupting an Android build or the emulator got me trying to interrupt arbitrary tasks.

I realized the output streamer is masking CTRL+C; therefore, a lot of upstream code is perceiving a failure leading to alerting the user and creating a briefcase log. Instead, the streamer should probably be re-raising the `KeyboardInterrupt` to generate the CTRL+C response that happens outside of a Wait Bar (and the streamer more generally). I can pursue this in a separate PR...although IMO that probably shouldn't hold up a release.

Additionally, fyi, I found a real-world use-case where `cleanup()` actually calls `kill()`. Turns out Docker does not just exit in response to the `terminate()`. When I sent a CTRL+C while building an AppImage in Docker, I initially thought it didn't take because (as it turns out) 3 seconds is _really_ long time when you're trying to decide if CTRL+C worked....so, I started sending more which just caused briefcase to completely abort. That may be at least some evidence that these waits should be decreased since users already asked briefcase to exit and it doesn't feel like that is happening in the moment. The log for `Stopping...` helps but in the case of an AppImage build, it gets lost in the flood of ongoing console logs. Alternatively, after the first time, a user may realize it did actually work and they just need to be more patient next time.

freakboy3742:
@rmartin16 Agreed that handling the KeyboardInterrupt in an output stream so that it doesn't generate a log would be preferable; but also agreed that (as well as any other improvements around streamer interaction) can be handled as future work.

PR #821, opened 2022-08-10
Android: filter logcat by PID

### LINKED ISSUE #814: Improve Android logging

When running an Android app, the logcat filtering currently displays a few specific tags from all apps. The problem is:
* If you have multiple apps on the device, you may see logs from all of them mixed together.
* There may be important information in other tags, such as Java or JNI crashes, which Briefcase doesn't display at all.

Briefcase should log all tags, but only for the app the user is working on. This would match the default behavior of Android Studio.

Depending on what's best supported by the tools, this may involve filtering by process ID, user ID, application ID, or something else.

mhsmith:
Fixes #814.

## PR Checklist:


src/briefcase/integrations/android_sdk.py
```diff
@@ -1265,11 +1254,11 @@ def logcat(self):
                     "-s",
                     self.device,
                     "logcat",
-                    "-s",
-                    "MainActivity:*",
-                    "stdio:*",
-                    "Python:*",
-                ],
+                    "--pid",  # This option is available since API level 24.
+                    pid,
```
mhsmith, line 1258: I've updated https://github.com/beeware/briefcase-android-gradle-template/pull/52 to set the minimum API level to 24.

src/briefcase/console.py
```diff
@@ -55,7 +55,9 @@ class Printer:
     """Interface for printing and managing output to the console and/or log."""
 
     # Console to manage console output.
-    console = RichConsole(highlighter=RichConsoleHighlighter(), emoji=False)
+    console = RichConsole(
+        highlighter=RichConsoleHighlighter(), emoji=False, soft_wrap=True
+    )
```
mhsmith, line 60: I had to add `soft_wrap` to prevent Rich from wrapping everything at 80 columns in an MSYS2 terminal regardless of the window size. I'm not sure when this started happening: going back to commit 8df2d11 didn't make any difference. It might even have been caused by yesterday's Windows update. 

@rmartin16: Have you ever seen a problem like this? And can you confirm that this change should be harmless?

src/briefcase/integrations/android_sdk.py
```diff
@@ -1265,11 +1254,11 @@ def logcat(self):
...
+                ]
+                # Filter out some noisy and useless tags.
+                + [f"{tag}:S" for tag in ["EGL_emulation"]],
```
mhsmith, line 1261: `EGL_emulation` is a tag which generates a message [like this](https://stackoverflow.com/questions/70376084) about once per second on the API level 31 emulator.

src/briefcase/platforms/android/gradle.py
```diff
@@ -212,20 +213,21 @@ def run_app(self, app: BaseConfig, device_or_avd=None, **kwargs):
         with self.input.wait_bar("Installing new app version..."):
             adb.install_apk(self.binary_path(app))
 
-        self.logger.info("Starting app...", prefix=app.app_name)
-        with self.input.wait_bar("Clearing device log..."):
-            adb.clear_log()
```
mhsmith, line 217: Now that we're filtering by PID, there's no need to clear the log to exclude earlier instances of the same app. And leaving the log intact allows the user to go back and examine it later with `adb` or Android Studio.

The PID filtering is done within the device and causes no noticable delay, even when the log buffer contains thousands of messages.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1281,3 +1270,15 @@ def logcat(self):
             # Exit normally since the user was instructed to use CTRL+C.
             if e.returncode not in {-2, 0xC000013A}:
                 raise BriefcaseCommandError("Error starting ADB logcat.") from e
+
+    def pidof(self, package):
+        """Return the PID of the given app as a string, or None if it isn't
+        running."""
```
mhsmith, line 1276: The PID is returned as a string because it'll only ever be used as a string.

mhsmith:
When testing this, please use a physical device, because they're likely to have more variation in their command-line tools.

src/briefcase/console.py
rmartin16: I haven't seen this....but it (obviously) implies that Rich cannot determine the size of the terminal....although, if it cannot determine that, it probably knows next to nothing about the terminal...

I can't get `briefcase` installed using MSYS2; `psutils` says `cygwin` isn't supported.

Out of curiosity, what does `python -m rich.diagnose` print?

As far as `soft_wrap=True`, it won't break anything...but it's unfortunate since its prior setting would prevent words bring broken over line breaks and increased readability IMO.

If this change stays, you can also remove `Printer.console.soft_wrap = True` from `tests/conftest.py`.

rmartin16: ahh...ok...learning new things today. MSYS2 can run windows binaries...seems to be a main selling point...anyway

I can replicate it using Python installed via Windows. This is probably coming from Rich setting `legacy_windows=True`.

<details>
<summary>python.exe -m rich.diagnose</summary>

```
$ ./venv-win/Scripts/python.exe -m rich.diagnose
+---------------------- <class 'rich.console.Console'> -----------------------+
| A high level console interface.                                             |
|                                                                             |
| +-------------------------------------------------------------------------+ |
| | <console width=79 None>                                                 | |
| +-------------------------------------------------------------------------+ |
|                                                                             |
|     color_system = None                                                     |
|         encoding = 'cp1252'                                                 |
|             file = <_io.TextIOWrapper name='<stdout>' mode='w'              |
|                    encoding='cp1252'>                                       |
|           height = 25                                                       |
|    is_alt_screen = False                                                    |
| is_dumb_terminal = False                                                    |
|   is_interactive = False                                                    |
|       is_jupyter = False                                                    |
|      is_terminal = False                                                    |
|   legacy_windows = True                                                     |
|         no_color = False                                                    |
|          options = ConsoleOptions(                                          |
|                        size=ConsoleDimensions(width=79, height=25),         |
|                        legacy_windows=True,                                 |
|                        min_width=1,                                         |
|                        max_width=79,                                        |
|                        is_terminal=False,                                   |
|                        encoding='cp1252',                                   |
|                        max_height=25,                                       |
|                        justify=None,                                        |
|                        overflow=None,                                       |
|                        no_wrap=False,                                       |
|                        highlight=None,                                      |
|                        markup=None,                                         |
|                        height=None                                          |
|                    )                                                        |
|            quiet = False                                                    |
|           record = False                                                    |
|         safe_box = True                                                     |
|             size = ConsoleDimensions(width=79, height=25)                   |
|        soft_wrap = False                                                    |
|           stderr = False                                                    |
|            style = None                                                     |
|         tab_size = 8                                                        |
|            width = 79                                                       |
+-----------------------------------------------------------------------------+
+--- <class 'rich._windows.WindowsConsoleFeatures'> ----+
| Windows features available.                           |
|                                                       |
| +---------------------------------------------------+ |
| | WindowsConsoleFeatures(vt=False, truecolor=False) | |
| +---------------------------------------------------+ |
|                                                       |
| truecolor = False                                     |
|        vt = False                                     |
+-------------------------------------------------------+
+------ Environment Variables -------+
| {                                  |
|     'TERM': 'xterm',               |
|     'COLORTERM': None,             |
|     'CLICOLOR': None,              |
|     'NO_COLOR': None,              |
|     'TERM_PROGRAM': 'mintty',      |
|     'COLUMNS': None,               |
|     'LINES': None,                 |
|     'JUPYTER_COLUMNS': None,       |
|     'JUPYTER_LINES': None,         |
|     'JPY_PARENT_PID': None,        |
|     'VSCODE_VERBOSE_LOGGING': None |
| }                                  |
+------------------------------------+
platform="Windows"
```
</details>

rmartin16: with that in mind, maybe `soft_wrap` should be set with something like ` shutil.get_terminal_size((1, 1)) == (1, 1)`

mhsmith: > MSYS2 can run windows binaries...seems to be a main selling point.

Yes, I'm using the standard Windows Python from python.org, I just use MSYS2 to get a Unix-like shell.

`python -m rich.diagnose` gives similar output to you (width = 79, height = 25), and shows the environment variables `COLUMNS` and `LINES` both as None. Both of those variables are actually set to the correct values in Bash, but not exported, so they're not available to Python. If I do export them, then `rich.diagnose` displays the correct width and height.

Even if they're not exported, `os.get_terminal_size()` returns the correct result anyway:
```
os.terminal_size(columns=105, lines=61)
```
So I don't know where that's coming from or why Rich was failing to use it.

> it won't break anything...but it's unfortunate since its prior setting would prevent words bring broken over line breaks and increased readability IMO

Word wrapping might be a bit more readable, but it has the disadvantage that when someone copies the text from the terminal into an issue report, it'll include line breaks that weren't there in the original. For example, if a phrase gets split up by a line break, it might prevent it from being found in a search for that phrase, either using GitHub's search  function or a web browser's "search within page" command.

src/briefcase/integrations/android_sdk.py
freakboy3742: Interesting... I'm not seeing that on my Android 31 emulator or my Android 8.1 test device. The internet seems to suggest it's a common piece of line noise, though, so we might as well filter it.

src/briefcase/console.py
freakboy3742: I agree with @mhsmith; while word wrapping is nice if you're dealing with text, we're dealing with content that is frequently mixed with computer-generated content or URLs, and the exact spacing and newlines matter in that context. 

Beyond that - my reading and testing seems to suggest `soft_wrap` is effectively a no-op in this case, so I'm OK with adding it if it cleans up a problem case on MSYS2.

src/briefcase/integrations/android_sdk.py
freakboy3742: But what if I want to perform complex arithmetic analysis on my PID??!?

freakboy3742 (approved):
Fairly unsurprising changes (other than the `soft_wrap` thing).

This is definitely a big improvement on the current log output; interesting that it reveals a bunch of `Briefcase/*` log output that I'd completely forgotten about. 

It might be worth exposing command line options to customise the log filter; but I'm happy to consider that as a future enhancement.

src/briefcase/integrations/android_sdk.py
mhsmith: I guess it depends on what graphics acceleration mode the emulator is using.

PR #832, opened 2022-08-18
Only run sdkmanager --list_installed when creating a log file

mhsmith:
Currently we run `sdkmanager --list_installed` within `verify_tools` in every Briefcase Android command. But this takes at least 1.5 seconds to run on my machine, even in the relatively small SDKs generated by Briefcase itself. When using a user-provided SDK which is affected by [this issue](https://issuetracker.google.com/issues/242831048), it can take more than 10 seconds. 

Since the purpose of this command is to provide information to help diagnose a failure, I've changed it to only run when creating a log file.

## PR Checklist:


tests/platforms/android/gradle/test_verify_tools.py
```diff
@@ -1,38 +0,0 @@
-import os
-from unittest import mock
-
-import pytest
-
-from briefcase.platforms.android.gradle import GradleBuildCommand
-
-
-@pytest.fixture
-def build_command(tmp_path, first_app_config):
```
mhsmith, line 10: This file lost all of its tests some time ago, so this fixture is no longer being used.

src/briefcase/__main__.py
```diff
@@ -10,6 +10,7 @@ def main():
     command = None
     try:
         command, options = parse_cmdline(sys.argv[1:])
+        log = command.logger
```
freakboy3742, line 15: It took me a moment to work out what was going on here. I think this points at an architectural flaw; we should be passing `log` to `parse_cmdline`, and using that log instance, rather than capturing a replacement log instance. However, that's a much more involved piece of dental work that doesn't need to block this PR.

freakboy3742 (approved):
Looks good - I've made a couple of minor comments inline (and implemented those fixes because it was easy enough to do so). If you're OK with these changes, I'll merge.

I think there's a couple of opportunities for larger scale refactors. The first one I mentioned inline - using a single log instance, and passing that through the calls, rather than capturing the log after it has been re-created.

The second is to develop a consistent API for a "tool integration". The `upgrade` command  does a partial ad-hoc version of this; this PR adds a second interface that requires an individual command to register that it has additional details to provide. If we had a single registry of tools, we could track which tools have been verified, iterate over all tools to dump versions/status or perform upgrades.

I don't think either of these changes need to hold up this PR thought; I'm mostly flagging them as the start of some tech debt that we might eventually need to pay down.

mhsmith:
Thanks, your changes all look good.

> using a single log instance, and passing that through the calls, rather than capturing the log after it has been re-created.

Agreed, and the same is true of the AndroidSDK object, which is currently instantiated multiple times if Briefcase does multiple `build` / `update` / `run` sub-commands. The AndroidSDK currently uses `self.command` to access various lower-level services like the logger. Refactoring it to bypass the command object and access these things directly would be quite disruptive, so it might be easier to just make sure that there's a single AndroidSDK instance with a reference to the top-level command object.

PR #836, opened 2022-08-21
Support multiple Rich renderables in console

rmartin16:
## Summary
While our discussions during the initial implementation of Rich largely decided that displaying multiple Rich renderables in the console at once was not necessary, I have found exceptions.

The primary use-case for this functionality is downloading a file while the Wait Bar is running. Right now, all the tools' installations are carefully crafted to start the Wait Bar after the tool has been downloaded.....but we're only doing this because the download progress bar could not be displayed while the Wait Bar was running.

This PR allows arbitrary use of `Console.progress_bar` and `Console.wait_bar` (as well as arbitrary Rich renderables if they are implemented).

<details>
<summary> Example Demo (paste in to REPL)</summary>

```python
from time import sleep
from briefcase.console import Console, Log

logger = Log()
console = Console()
sleep_time = 1

def pause(loop_count=5):
    for idx in range(loop_count):
        logger.info(f"{idx}")
        sleep(sleep_time)

def main():
    with console.wait_bar("Building..."):
        pause(3)
        with console.wait_bar("Installing..."):
            logger.info("Changed wait bar message")
            pause(3)
            logger.info("Starting progress bar with multiple tasks")
            with console.progress_bar() as bar:
                task_1_id = bar.add_task("", total=100)
                task_2_id = bar.add_task("", total=100)
                for idx in range(10):
                    if idx == 4:
                        logger.warning("warning for initial progress bar")
                    if idx == 6:
                        with console.progress_bar() as bar_new:
                            logger.info("Starting a new progress bar inside existing one")
                            new_task_id = bar_new.add_task("", total=100)
                            for _ in range(4):
                                bar_new.update(new_task_id, advance=25)
                                sleep(sleep_time)
                        logger.info("Finished nested progress bar")
                    bar.update(task_1_id, advance=20)
                    bar.update(task_2_id, advance=10)
                    sleep(sleep_time / 2)
            logger.info("Finished initial progress bar with multiple tasks")
            logger.info("removing dynamic elements", prefix='asdf')
            pause(2)
            console.remove_dynamic_elements()
            logger.info("Dynamic elements removed")
            pause(3)
            console.restore_dynamic_elements()
            logger.info("Restored dynamic elements", prefix='asdf')
            pause(3)
            logger.info("Wrapping up install")
        logger.info("Back to 'Building...' wait bar message")
        pause()
    logger.info("All dynamic elements wrapped up; exiting.", prefix="asdf")

main()

```

</details>

## Design
This started out as a curiosity and I ultimately nerd sniped myself...but it was nice getting to know Rich better because it really is so well done.

When a Rich renderable is started in isolation, it will use a Live Display that is reserved just for it. This prevents starting another Rich renderable the same way because Rich limits the [number of active Live Displays to one](https://rich.readthedocs.io/en/latest/live.html#nesting-lives). Therefore, the basis of this approach here is externalizing the Live Display that any renderable will use and managing it alongside the renderables.

Next, the content for the Live Display needs to be managed independently of the Live Display itself. I initially tried to introspect the Live Display to determine what's currently being displayed when I wanted to add or remove renderables. However, Will has said (in an issue/discussion I can no longer find) that the Display and its renderables should be treated more like HTML. Once it is displayed, that's it; if you want to change it, you don't introspect the HTML or modify it directly (unless youre sneaky js i guess). You build a new HTML document using externally managed data. So, I created (more or less) a stack of renderables that can be displayed. You can only add renderables on top of what's already being displayed; so, this isn't entirely generalized to allow doing whatever you want...but this limitation seems reasonable given our use-case.

Finally, I exposed methods to stop and start the Live Display from anywhere in Briefcase. These will be necessary to accommodate cmd.exe's `Terminate batch job` message. I can imagine exposing a context manager that allows you to wrap certain calls while ensuring dynamic elements are turned off....but given I'm imagining the cmd issue to be resolved directly in `subprocess.py`, I wouldn't use such a manager there. But it would be simple enough to create if needed later. (Although, thinking more about this...such a context manager would allow easily prompting the user inside a Wait Bar....as a thought.)

On a sidenote, these changes got rid of the mysterious extra new line that was showing up in wait bar tests...

## PR Checklist:

### DIFF
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -9,7 +9,9 @@ from datetime import datetime
 from pathlib import Path
 
 from rich.console import Console as RichConsole
+from rich.console import Group
 from rich.highlighter import RegexHighlighter
+from rich.live import Live
 from rich.markup import escape
 from rich.progress import (
     BarColumn,
@@ -302,24 +304,75 @@ class Console:
         self.print = printer
         self.input = printer.console.input
         self.enabled = enabled
-        self._wait_bar: Progress = None
+
         # Signal that Rich is dynamically controlling the console output.
-        # Therefore, all output must be printed to the screen by Rich to
-        # prevent corruption of dynamic elements like the Wait Bar.
+        # Therefore, all output must be printed to the console by Rich to
+        # prevent corruption by dynamic elements like the Wait Bar.
         self.is_output_controlled = False
 
-    def prompt(self, *values, markup=False, **kwargs):
-        """Print to the screen for soliciting user interaction.
+        self._live_display: Live = None
+        self._live_display_stack = []
+        self._wait_bar: Progress = None
 
-        :param values: strings to print as the user prompt
-        :param markup: True if prompt contains Rich markup
+    def _live_display_add(self, renderable):
+        """Add a renderable to the top of the Live Display."""
+        if renderable not in self._live_display_stack:
+            self._live_display_stack.append(renderable)
+            self._live_display_update()
+
+    def _live_display_remove(self, renderable):
+        """Remove a renderable from the Live Display."""
+        with contextlib.suppress(ValueError):
+            self._live_display_stack.remove(renderable)
+        self._live_display_update()
+        # Print non-transient renderables so their final refresh remains in
+        # the console. When a renderable is removed from the Live Display,
+        # it is also completely removed from the console.
+        if not renderable.live.transient:
+            self.print(renderable)
+
+    def _live_display_update(self):
+        """Update the Live Display with the current stack of renderables."""
+        if not self._live_display:
+            self._live_display = Live(
+                refresh_per_second=15,
+                transient=True,
+                console=self.print.console,
+            )
+        if self._live_display_stack:
+            self.is_output_controlled = True
+            self._live_display.update(
+                Group(*self._live_display_stack[::-1]),
+                refresh=True,
+            )
+            self._live_display.start()
+        else:
+            self._live_display.stop()
+            self.is_output_controlled = False
+
+    def remove_dynamic_elements(self):
+        """Remove dynamic console elements such as the Wait bar.
+
+        Only remove with the intent to restore. Otherwise, when the
+        context managers for the currently active elements exit, screen
+        corruption may occur.
         """
-        if self.enabled:
-            self.print(*values, markup=markup, stack_offset=4, **kwargs)
+        if self._live_display:
+            self._live_display.stop()
+            # Force creation of new Live Display when it is restored.
+            # This ensures the Live Display doesn't corrupt console
+            # content created after the existing one was stopped.
+            self._live_display = None
+            self.is_output_controlled = False
 
+    def restore_dynamic_elements(self):
+        """Restore previously removed dynamic console elements."""
+        self._live_display_update()
+
+    @contextlib.contextmanager
     def progress_bar(self):
         """Returns a progress bar as a context manager."""
-        return Progress(
+        progress_bar = Progress(
             TextColumn("  "),
             SpinnerColumn("line", speed=1.5, style="default"),
             BarColumn(bar_width=50),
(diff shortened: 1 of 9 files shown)


src/briefcase/console.py
```diff
@@ -302,24 +304,72 @@ def __init__(self, printer=Printer(), enabled=True):
...
+            )
+        if self._live_display_stack:
+            self.is_output_controlled = True
+            self._live_display.update(Group(*self._live_display_stack[::-1]))
+            self._live_display.start()
+        else:
+            self._live_display.stop()
+            self.is_output_controlled = False
+
+    def remove_dynamic_elements(self):
```
rmartin16, line 350: I didn't implement this idea....but this could also set a "do not allow starting the live display" flag here that would be unset in `restore_dynamic_elements()`. This would allow adding renderables but they wouldn't be displayed until the restore call.

Given the use cases, it seems reasonable that a new renderable will be added after calling remove.

freakboy3742 (commented):
This seems to make makes sense for what it is; my concern is that I'm not 100% sure I follow the *why* of this change. 

The use case you've provided (multiple Rich renderables)... doesn't sound like something that is particularly desirable. You mention "downloading a file while the waitbar is showing" - maybe my memory is fading with age, but I can't think of an example where that has come up in recent history, or a case where we'd want to do that. We use a progress bar for downloads, and if you're downloading a file, that's the thing you're currently doing. We use waitbars for tasks of indeterminate length, like calls to build a project; and if you're building the project, that's the thing you're currently doing. I can't think of any places where you're in the middle of a build *and* you need to start a download (or vice versa).

More broadly, I'm not sure multiple animated elements at once is something we'd want to do. The only use case I can think of where it might be desirable is if were we to support downloading multiple files at the same time - but that's not something we support (nor are we likely to be able to support it easily - it's more of a "GUI implemented in text" problem than a traditional command line problem.

That said - based on your description of the design, I'm not sure if I should be reading this PR as a case of aligning on how we should be managing live elements, regardless of whether we use multiple live elements at the same time.

So - is this a case of 
1. Refactoring to support a use case that we're not likely to actually need in practice
2. Refactoring to "do things the right way (according to Rich)", even if we're not likely to actually use the functionality that might afford
3. Refactoring to support a use case that I've completely forgotten about

rmartin16:
That's fair; I should have added to the PR description that there are definitely legitimate arguments to reject this.

**Immediate use-case**
The impetus for these changes was largely the `Installing...` Wait Bar. Oftentimes, the download step also does the majority of the actual "installing" while the Wait Bar is only active for small steps after. So, there is at least some argument the `Installing...` Wait Bar should be active during the download....but that argument could be reasonably considered weak.

**Argument against**
However, the Progress Bar and Wait Bar can be seen more as mutually exclusive....where only one should be in the console at a time. This guiding principle definitely has merit. I think whether the Wait Bar could represent a larger indeterminate length process while a simultaneous Progress Bar represents a sub-step of determinate length comes down to opinion and expectations of the UI. I admit multiple independent animations may not always be a good choice.

**Rich implementation**
I wouldn't say this better aligns the codebase with a "right way" to implement Rich; however, this is how multiple renderables stuck to the bottom of the console would be managed. These changes definitely generalize how Rich is implemented and better positions the codebase to respond to future use-cases that may necessitate multiple renderables.

**Final thoughts**
I was ultimately surprised the implementation ended up being relatively simple while generalizing and opening possibilities for Rich use-cases. That said, such use-cases _do_ seem unlikely....especially if my "immediate use-case" example isn't a change you'd want to see. So, some extra complexity without any real immediate gains....but I probably would've implemented something closer to this originally if I'd been more familiar with Rich at the time. Even if rejected, this implementation will, of course, always be here if needed in the future :)

freakboy3742:
Ok - on the basis that we can't make a "better architecture" argument for this, I'm gone to reject this one. In terms of the specific example, I'd argue the download/install case should 2 steps - there's a (slow, but known length) download, followed by a (generally quick, but strictly unknowable) "install" process.

Thanks for the work (and the explanations) - as you say, the code will still be here if we need it in future.

PR #842, opened 2022-08-24
Add iOS log filter for messages coming from extension modules.

freakboy3742:
When the ctypes module is dynamically loaded, the source of log messages written to NSLog by Python is reported as the extension module, rather than the name of the app binary. 

This mirrors the filter that is required for macOS apps, where libFFI is dynamically linked; on macOS, libffi is the source of NSLog messages.

## PR Checklist:

### DIFF
--- a/src/briefcase/platforms/iOS/xcode.py
+++ b/src/briefcase/platforms/iOS/xcode.py
@@ -351,6 +351,15 @@ class iOSXcodeRunCommand(iOSXcodeMixin, RunCommand):
             ) from e
 
         # Start log stream for the app.
+        # The following sets up a log stream filter that looks for:
+        #  1. a log sender that matches that app binary; or,
+        #  2. a log sender of that is a Python exception module,
+        #     and a process that matches the app binary.
+        # Case (1) works when the standard libary is statically linked,
+        #   and for native NSLog() calls in the bootstrap binary
+        # Case (2) works when the standard library is dynamically linked,
+        #   and ctypes (which handles the NSLog integration) is an
+        #   extension module.
         simulator_log_popen = self.subprocess.Popen(
             [
                 "xcrun",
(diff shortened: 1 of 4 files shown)


src/briefcase/platforms/iOS/xcode.py
```diff
@@ -351,6 +351,15 @@ def run_app(self, app: BaseConfig, udid=None, **kwargs):
             ) from e
 
         # Start log stream for the app.
+        # The following sets up a log stream filter that looks for:
+        #  1. a log sender that matches that app binary; or,
+        #  2. a log sender of that is a Python exception module,
```
mhsmith, line 356: Should this be "extension module"?

src/briefcase/platforms/iOS/xcode.py
```diff
@@ -362,7 +371,9 @@ def run_app(self, app: BaseConfig, udid=None, **kwargs):
                 "--style",
                 "compact",
                 "--predicate",
-                f'senderImagePath ENDSWITH "/{app.formal_name}"',
+                f'senderImagePath ENDSWITH "/{app.formal_name}"'
+                f' OR (processImagePath ENDSWITH "/{app.formal_name}"'
+                ' AND senderImagePath ENDSWITH "-iphonesimulator.so")',
```
mhsmith, line 376: Why not just include all messages where `processImagePath ENDSWITH "/{app.formal_name}"`? Is it because it would be too noisy? If so, it would be worth explaining that in a comment.

mhsmith (commented):
I'm not set up to test iOS yet, so just a couple of comments from reviewing the code:

src/briefcase/platforms/iOS/xcode.py
freakboy3742: ... No, I've invented a whole new type of module. It gets around all that tedious code writing and just raises the inevitable exception for you :-)

(oops - fixed :-)

src/briefcase/platforms/iOS/xcode.py
freakboy3742: The *process* will generate lots of messages that aren't user-originated (all the usual noise generated by the device notifying system-level events). We only care about the messages that come from the user's code (the sender). 

We could *almost* get away with *just* filtering on the `senderImagePath`. This would work on the simulator, and on any phone that only had 1 Python app running; but if you had a second app generating log messages, those would end up being interspersed with with the app you were trying to log.

Regardless - I'll add a note about this.

Issue #699, opened 2022-04-05
Add runtime tests for final Briefcase outputs


freakboy3742:
Briefcase has an extensive unit test suite; however, there's no formal test that the final product actually "works". Errors in templates, or changes in platform tooling occasionally cause errors

**Describe the solution you'd like**

Additional steps in the build matrix that verify that Briefcase will actually generate working output for each of the supported platforms.

Ideally, we would also have automated testing that the apps start and shut down; but as an initial step, just verifying that we can generate an app would be an improvement.

**Describe alternatives you've considered**

Nothing; but anything will be better than 100% manual testing (the status quo).

**Additional context**

#484 was an initial attempt at this.

The same tests should also be performed on the initial app template (beeware/briefcase-template), plus the platform templates (beeware/briefcase-macOS-app-template, etc)

mhsmith:
A couple of recent breakages which would have been caught by a simple build test:
* [briefcase-template#32](https://github.com/beeware/briefcase-template/pull/32): AppImage configuration was mislabeled in pyproject.toml.
* [#816](https://github.com/beeware/briefcase/pull/816): Flatpak CPython installation was broken by changes in the support package mechanism.

mhsmith:
Actually, I see there are build tests in the CI now, but these problems weren't caught for other reasons:
* [briefcase-template#32](https://github.com/beeware/briefcase-template/pull/32): Template updates should trigger a Briefcase build test. But even if they don't, this should still have shown up in the checks of the next Briefcase PR (probably [#803](https://github.com/beeware/briefcase/pull/803)). I guess the reason it didn't is that the CI uses `--no-docker` for AppImage builds.
* [#816](https://github.com/beeware/briefcase/pull/816) was missed simply because there's no CI test for Flatpak.

freakboy3742:
+1 to both of these. 

In addition to the Flatpak test, we should add a test for iOS and Android - neither of those are being tested either. 

We should also modify the tests to actually *run* the binary that is produced - we've had some regressions where Briefcase will successfully generate an app, but the app won't start. The complication here is that we need a "unit test mode" for the generated stub apps - at present, iOS and Android apps in particular don't have a concept of apps "stopping" (other than a crash), and the log streamer doesn't catch those exit conditions. Xcode has the concept of a unit test target - an app that doesn't follow the usual app lifecycle; we might be able to exploit this. I presume Gradle will have something similar.

mhsmith:
[...]

PR #844, opened 2022-08-24
Improve CI testing configurations of build products

### LINKED ISSUE #839: Modify build image of AppImage template to use Ubuntu 20.04
According to Github, by Dec 1 2022 they will [formally deprecate the Ubuntu-18.04 Actions runner](https://github.com/actions/runner-images/issues/6002). 

This puts Briefcase in an interesting situation. On the one hand Briefcase as used by end users is (mostly) unaffected - we encourage the use of Docker for building AppImages, and the Ubuntu 18.04 image will continue to exist. 

However, as a project, *we* have a problem. Due to historical issues using Docker in CI, we run our CI in `--no-docker` mode - which means that in December, we'll no longer be able to run CI for Linux. 

Update the template and documentation to use Ubuntu 20.04 as the base image. 

This is something we've done before ([transitioning from 16.04 to 18.04 as the base image](https://github.com/beeware/briefcase-linux-appimage-template/commit/a172088d46f86b0a95799e8f03bc04149741c765). Interestingly, this was almost 2 years ago to the day, which matches the 2 year LTS release cadence of Ubuntu.

In addition to continued CI support, this has the added benefit that we'll have access to more recent packages in the base image.

1. Remove AppImages from our testing setup, but otherwise do nothing. Ubuntu 18.04 AppImage images will continue to work; they'll just be untested.
2. Modify CI to use Docker, but continue to use 18.04 as the base image. 
3. Deprecate the AppImage backend in favour of Flatpak.

**Additional context**

This may have some overlap with #824 - a switch of this magnitude requires some sort of control from a user perspective.

freakboy3742:
Github Actions is scheduled to deprecate the ubuntu-18.04 test runner; this means we'll no longer be able to use `--no-docker` in CI. The use of `--no-docker` in CI has also led to some edge cases slipping through our testing, so this is probably desirable anyway.

Since we need to revisit our CI config anyway, this is also an opportune time to improve CI coverage of other testing platforms. This config change adds iOS, Android, Linux Flatpak, macOS Xcode and Windows VisualStudio tests.

It *doesn't* include a full execution of the final product, so it's not a complete solution to #699.

Fixes #839 
Refs #699

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,7 +137,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        platform: ['macos-latest', 'windows-latest', 'ubuntu-18.04']
+        platform: ['macos-latest', 'windows-latest', 'ubuntu-latest']
         framework: ['toga', 'pyside2', 'pyside6', 'ppb']
         include:
         - platform: macos-latest
@@ -148,10 +148,9 @@ jobs:
           briefcase-data-dir: ~\AppData\Local\BeeWare\briefcase\Cache
           pip-cache-dir: ~\AppData\Local\pip\Cache
           docker-cache-dir: C:\ProgramData\DockerDesktop
-        - platform: ubuntu-18.04
+        - platform: ubuntu-latest
           briefcase-data-dir: ~/.cache/briefcase
           pip-cache-dir: ~/.cache/pip
-          briefcase-args: --no-docker
           # cache action cannot cache docker images (actions/cache#31)
           # docker-cache-dir: /var/lib/docker
     runs-on: ${{ matrix.platform }}
@@ -171,7 +170,7 @@ jobs:
       with:
         python-version: ${{ env.python_version }}
     - name: Install dependencies
-      if: matrix.platform == 'ubuntu-18.04'
+      if: matrix.platform == 'ubuntu-latest'
       run: |
         sudo apt-get update -y
         sudo apt-get install -y python3-gi python3-gi-cairo gir1.2-gtk-3.0 python3-dev libgirepository1.0-dev libcairo2-dev pkg-config
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,24 +131,28 @@ jobs:
       run: |
         tox -e py
 
-  verify-desktop:
-    name: Desktop app verification
-    needs: smoke
+  verify-apps:
+    name: App verification
+    needs: platform-compat
     strategy:
       max-parallel: 4
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -131,14 +131,194 @@ jobs:
       run: |
         tox -e py
 
-  verify-desktop:
-    name: Desktop app verification
-    needs: smoke
+  verify-desktop-macOS:
+    name: macOS desktop app verification
+    needs: platform-compat
```
mhsmith, line 136: Is it necessary to have 5 configurations which have so much in common? Could the differences (e.g. cache directories, `--adhoc-sign` argument) and matrix limitations (e.g. building iOS on macOS only) have been dealt with using `include` variables and `if`?

freakboy3742: I definitely agree this is... verbose. The original version of this stage was a matrix build; however, adding iOS and Android as another axis, while only building Toga got prohibitively complex. However, thinking about it again, it might be easier to do this as a single OS x project matrix, and then exclude iOS and Android based on a single axis of the matrix. I'll have a fiddle and see what I can do.

.github/workflows/ci.yml
```diff
@@ -131,14 +131,194 @@ jobs:
...
+          ~/.cache/pip
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v3.1.2
+      with:
+        python-version: ${{ env.python_version }}
+    - name: Install system dependencies
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y python3-gi python3-gi-cairo gir1.2-gtk-3.0 python3-dev libgirepository1.0-dev libcairo2-dev pkg-config flatpak flatpak-builder
```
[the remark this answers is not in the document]
freakboy3742: I'm fairly sure we do - these packages aren't in the [list of apt packages in the 22.04 runner config](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#installed-apt-packages). However, I'll experiment and see.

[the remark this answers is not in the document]
freakboy3742: Oh - wait - now I see what you're saying: we don't need these because we're building *in* docker. You're completely correct - we only need the flatpak tools.

.github/workflows/ci.yml
freakboy3742: Ok - I've been able to clean this up quite a bit. It turns out it was the use of `tox` that was making the process complex; once I dropped Tox, the matrix and step definitions actually ended up quite elegant.

mhsmith (approved):

PR #851, opened 2022-09-05
Propagate `KeyboardInterrupt` out of output streamer

rmartin16:
When the output streamer is interrupted, it should (relatively quickly) stop streaming and `briefcase` should exit with the `Aborted by user` message. Currently, the streamer swallows the `KeyboardInterrupt` signal; consequently, upstream code usually detects this as a failure, raises `BriefcaseCommandError`, and creates a log file.

This change propagates the `KeyboardInterrupt` into upstream code to ensure "user abort" codeflows are followed.

## PR Checklist:

### DIFF
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -1,10 +1,10 @@
-import contextlib
 import operator
 import os
 import platform
 import re
 import sys
 import traceback
+from contextlib import contextmanager
 from datetime import datetime
 from pathlib import Path
 
@@ -255,8 +255,6 @@ class Log:
                 for func in self.log_file_extras:
                     try:
                         func()
-                    except KeyboardInterrupt:
-                        raise
                     except Exception:
                         self.error(traceback.format_exc())
 
@@ -329,7 +327,7 @@ class Console:
             console=self.print.console,
         )
 
-    @contextlib.contextmanager
+    @contextmanager
     def wait_bar(
         self,
         message="",
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -6,6 +6,7 @@ import subprocess
 import sys
 import threading
 import time
+from contextlib import suppress
 from pathlib import Path
 
 import psutil
@@ -276,7 +277,7 @@ class Subprocess:
             self.stream_output(args[0], process)
             if process.stderr and kwargs["stderr"] != subprocess.STDOUT:
                 stderr = process.stderr.read()
-            return_code = process.poll()
+        return_code = process.poll()
         self._log_return_code(return_code)
 
         if check and return_code:
@@ -398,6 +399,8 @@ class Subprocess:
             self.command.logger.info("Stopping...")
             # allow time for CTRL+C to propagate to the child process
             time.sleep(0.25)
+            # re-raise to exit as "Aborted by user"
+            raise
         finally:
             self.cleanup(label, popen_process)
             streamer_deadline = time.time() + 3
@@ -413,14 +416,18 @@ class Subprocess:
 
         :param popen_process: popen process to stream stdout
         """
-        while True:
-            # readline should always return at least a newline (ie \n)
-            # UNLESS the underlying process is exiting/gone; then "" is returned
-            output_line = ensure_str(popen_process.stdout.readline())
-            if output_line:
-                self.command.logger.info(output_line)
-            elif output_line == "":
-                return
+        # ValueError is raised if stdout is unexpectedly closed.
+        # This can happen if the user starts spamming CTRL+C, for instance.
+        # Silently exit to avoid Python printing the exception to the console.
+        with suppress(ValueError):
+            while True:
+                # readline should always return at least a newline (ie \n)
+                # UNLESS the underlying process is exiting/gone; then "" is returned
+                output_line = ensure_str(popen_process.stdout.readline())
+                if output_line:
+                    self.command.logger.info(output_line)
+                elif output_line == "":
+                    return
 
     def cleanup(self, label, popen_process):
         """Clean up after a Popen process, gracefully terminating if possible;
--- a/src/briefcase/platforms/iOS/xcode.py
+++ b/src/briefcase/platforms/iOS/xcode.py
@@ -376,7 +376,7 @@ class iOSXcodeRunCommand(iOSXcodeMixin, RunCommand):
         #  1. a log sender that matches that app binary; or,
         #  2. a log sender of that is a Python extension module,
         #     and a process that matches the app binary.
-        # Case (1) works when the standard libary is statically linked,
+        # Case (1) works when the standard library is statically linked,
         #   and for native NSLog() calls in the bootstrap binary
         # Case (2) works when the standard library is dynamically linked,
         #   and ctypes (which handles the NSLog integration) is an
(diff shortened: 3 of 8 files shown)


src/briefcase/integrations/subprocess.py
```diff
@@ -276,7 +276,7 @@ def _run_and_stream_output(self, args, check=False, **kwargs):
             self.stream_output(args[0], process)
             if process.stderr and kwargs["stderr"] != subprocess.STDOUT:
                 stderr = process.stderr.read()
-            return_code = process.poll()
```
rmartin16, line 279: It is safer to call `poll()` after the exiting the `with` since `__exit__()` waits for the process to end.

src/briefcase/integrations/subprocess.py
```diff
@@ -276,7 +276,7 @@ def _run_and_stream_output(self, args, check=False, **kwargs):
...
+        return_code = process.poll()
```
rmartin16, line 279: It is safer to call `poll()` after the exiting the `with` since `__exit__()` waits for the process to end.

src/briefcase/integrations/subprocess.py
```diff
@@ -413,14 +416,18 @@ def _stream_output_thread(self, popen_process):
...
-            # UNLESS the underlying process is exiting/gone; then "" is returned
-            output_line = ensure_str(popen_process.stdout.readline())
-            if output_line:
-                self.command.logger.info(output_line)
-            elif output_line == "":
-                return
+        # ValueError is raised if stdout is unexpectedly closed.
+        # This can happen if the user starts spamming CTRL+C, for instance.
+        # Silently exit to avoid Python printing the exception to the console.
+        with contextlib.suppress(ValueError):
```
rmartin16, line 422: I found this while experimenting with sending CTRL+C while building an AppImage in Docker. Different versions of Python manifest the error differently in the console but they all raise `ValueError`.

src/briefcase/platforms/iOS/xcode.py
```diff
@@ -410,21 +410,25 @@ def run_app(self, app: BaseConfig, udid=None, **kwargs):
...
-        self.subprocess.stream_output("log stream", simulator_log_popen)
+            # Start streaming logs for the app.
+            self.logger.info(
+                "Following simulator log output (type CTRL-C to stop log)...",
+                prefix=app.app_name,
+            )
+            self.logger.info("=" * 75)
+            self.subprocess.stream_output("log stream", simulator_log_popen)
+        finally:
+            self.subprocess.cleanup("log stream", simulator_log_popen)
```
rmartin16, line 431: Aligns iOS log streaming with handling for macOS.

src/briefcase/platforms/windows/__init__.py
```diff
@@ -88,6 +88,7 @@ def run_app(self, app: BaseConfig, **kwargs):
                 [os.fsdecode(self.binary_path(app))],
                 cwd=self.home_path,
                 check=True,
+                stream_output=True,
```
rmartin16, line 91: While this doesn't actually get the output in to the log for Windows apps (see https://github.com/beeware/briefcase/pull/782#discussion_r920621894), the consistency seems appropriate.

src/briefcase/console.py
```diff
@@ -255,8 +255,6 @@ def _build_log(self, command):
                 for func in self.log_file_extras:
                     try:
                         func()
-                    except KeyboardInterrupt:
-                        raise
                     except Exception:
                         self.error(traceback.format_exc())
```
rmartin16, line 259: Unless I'm confused, there is no need to re-raise; `KeyboardInterrupt` will propagate naturally.

src/briefcase/integrations/flatpak.py
```diff
@@ -252,6 +252,7 @@ def run(self, bundle, app_name):
                     f"{bundle}.{app_name}",
                 ],
                 check=True,
+                stream_output=True,
```
rmartin16, line 255: I can't build a Flatpak right now....still trying to figure out the underlying problem. Something with `libffi.so` not being available.

rmartin16: Found the source of the issue:
- #858

Confirmed log is captured in the log.

src/briefcase/integrations/subprocess.py
```diff
@@ -413,14 +416,18 @@ def _stream_output_thread(self, popen_process):
...
-            # UNLESS the underlying process is exiting/gone; then "" is returned
-            output_line = ensure_str(popen_process.stdout.readline())
-            if output_line:
-                self.command.logger.info(output_line)
-            elif output_line == "":
-                return
+        # ValueError is raised if stdout is unexpectedly closed.
+        # This can happen if the user starts spamming CTRL+C, for instance.
+        # Silently exit to avoid Python printing the exception to the console.
+        with suppress(ValueError):
```
rmartin16, line 422: I found this while experimenting with sending CTRL+C while building an AppImage in Docker. Different versions of Python manifest the error differently in the console but they all raise `ValueError`.

freakboy3742: Can we add a test case for this as well?

freakboy3742 (changes requested):
Looks good from visual inspection; unfortunately, this is one of those PRs which only fixes a class of problems that occur when you mash CTRL-C in a *very* specific way, so it's difficult to reliably test. I've flagged 2 areas that need specific additional tests (and in one case, a possible reversion), but otherwise I'm happy with this.

src/briefcase/console.py
rmartin16: AFAICT, that's the whole reason they derived both `Exception` and [`KeyboardInterrupt`](https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt) from `BaseException`....so you didn't have to worry about `except Exception` catching `KeyboardInterrupt`. IIRC, this was a Python 2->3 change.

I can definitely add a test for it, though.

freakboy3742: Huh... you are absolutely correct. TIL another thing that I need to forget after the Py2->3 transition :-)

Regardless, a test can't hurt.

rmartin16: Looks like [test_save_log_to_file_extra_interrupted](https://github.com/rmartin16/briefcase/blob/main/tests/console/test_Log.py#L147) already exists and ensures the correct behavior.

src/briefcase/integrations/subprocess.py
rmartin16: Let me know if you think this test goes too far...or isn't testing the right thing. I didn't just want to mock `readline()` to raise `ValueError` and instead verify that using `readline()` on closed IO actually led to a `ValueError`.

Both `io.StringIO` and `io.TextIOWrapper` (which is used for `stdout`) both inherit from `io.TextIOBase`....which is where I'm assuming this `ValueError` comes from....if not deeper.

freakboy3742: Definitely agreed that it's better to try and cause a ValueError rather than just raise one as a mock; the problem is that in my testing, this tests still passes when the `suppress(ValueError)` is disabled - I suspect this is because some interaction of pytest and threading causes the `ValueError` to be hidden from the console. However, after poking around, I can't come up with a better test for this, so I'm happy to accept it as is.

freakboy3742 (approved):
Those extra tests looks great - thanks for the PR!

src/briefcase/integrations/subprocess.py
rmartin16: Oh yeah; that's a good point....the error only surfaces in the console when Python is exiting.

I came up with this change to the test....its a little janky but seems to work:

```python
def test_stdout_closes_unexpectedly(mock_sub, popen_process, monkeypatch, capsys):
    """Streamer silently exits from ValueError because stdout was closed."""

    class suppress:
        is_valueerror_exception_raised = []

        def __init__(self, *exceptions):
            self._exceptions = exceptions

        def __enter__(self):
            pass

        def __exit__(self, exctype, excinst, exctb):
            self.is_valueerror_exception_raised.append(exctype is ValueError)
            return exctype is not None and issubclass(exctype, self._exceptions)

    def monkeypatch_ensure_str(value):
        """Close stdout when ensure_str() runs on output from readline()."""
        popen_process.stdout.close()
        return value

    popen_process.stdout = StringIO(initial_value="output line 1\noutput line 2")
    monkeypatch.setattr(subprocess, "ensure_str", monkeypatch_ensure_str)
    monkeypatch.setattr(subprocess, "suppress", suppress)

    mock_sub.stream_output("testing", popen_process)

    assert capsys.readouterr().out == "output line 1\n"
    assert suppress.is_valueerror_exception_raised == [True]
```

I haven't figured out why `is_valueerror_exception_raised` needs to be a mutable type...but that was the only thing I could quickly get working.

freakboy3742: Yeah - that's... janky :-) I'm not sure it's worth it as a test improvement; it's monkey patching the object that we're trying to verify is working.

PR #852, opened 2022-09-06
Only Instantiate `Log` and `Console` Once

rmartin16:
Currently, when `briefcase` executes multiple Commands (e.g. `update`/`build` on behalf of `run`), a new `Log` and `Console` are created for each Command. So far, this hasn't proven problematic since each copy is effectively identical and once a new one is created, the previous one is never referenced again. In principle, though, only one copy of these objects should exist.

Future changes (via #801) will (effectively) allow multiple Commands to exist at the same time. Given this, all Commands must reference the same `Log` and `Console` to ensure agreement of the state of the program.

This change ensures the existing `Log` and `Console` are passed to any Command created as a result of running the user-requested Command.

## PR Checklist:

### DIFF
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -14,7 +14,7 @@ from .exceptions import (
 )
 
 
-def parse_cmdline(args):
+def parse_cmdline(args, logger=None):
     parser = argparse.ArgumentParser(
         prog="briefcase",
         description="Package Python code for distribution.",
@@ -78,15 +78,15 @@ def parse_cmdline(args):
     if options.command is None:
         raise NoCommandError(parser.format_help())
     elif options.command == "new":
-        command = NewCommand(base_path=Path.cwd())
+        command = NewCommand(base_path=Path.cwd(), logger=logger)
         options = command.parse_options(extra=extra)
         return command, options
     elif options.command == "dev":
-        command = DevCommand(base_path=Path.cwd())
+        command = DevCommand(base_path=Path.cwd(), logger=logger)
         options = command.parse_options(extra=extra)
         return command, options
     elif options.command == "upgrade":
-        command = UpgradeCommand(base_path=Path.cwd())
+        command = UpgradeCommand(base_path=Path.cwd(), logger=logger)
         options = command.parse_options(extra=extra)
         return command, options
 
@@ -160,6 +160,6 @@ def parse_cmdline(args):
         )
 
     # Construct a command, and parse the remaining arguments.
-    command = Command(base_path=Path.cwd())
+    command = Command(base_path=Path.cwd(), logger=logger)
     options = command.parse_options(extra=extra)
     return command, options
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -77,87 +77,85 @@ def parse_cmdline(args, logger=None):
     # If no command has been provided, display top-level help.
     if options.command is None:
         raise NoCommandError(parser.format_help())
-    elif options.command == "new":
-        command = NewCommand(base_path=Path.cwd(), logger=logger)
-        options = command.parse_options(extra=extra)
-        return command, options
+
+    # Commands agnostic to the platform and format
+    if options.command == "new":
+        Command = NewCommand
     elif options.command == "dev":
-        command = DevCommand(base_path=Path.cwd(), logger=logger)
-        options = command.parse_options(extra=extra)
-        return command, options
+        Command = DevCommand
     elif options.command == "upgrade":
-        command = UpgradeCommand(base_path=Path.cwd(), logger=logger)
-        options = command.parse_options(extra=extra)
-        return command, options
+        Command = UpgradeCommand
 
-    parser.add_argument(
-        "platform",
-        choices=list(platforms.keys()),
-        default={
-            "darwin": "macOS",
-            "linux": "linux",
-            "win32": "windows",
-        }[sys.platform],
-        metavar="platform",
-        nargs="?",
-        type=normalize,
-        help="The platform to target (one of %(choices)s; default: %(default)s",
-    )
-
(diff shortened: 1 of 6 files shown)


src/briefcase/cmdline.py
```diff
@@ -78,15 +78,15 @@ def normalize(name):
     if options.command is None:
         raise NoCommandError(parser.format_help())
     elif options.command == "new":
-        command = NewCommand(base_path=Path.cwd())
+        command = NewCommand(base_path=Path.cwd(), logger=logger)
         options = command.parse_options(extra=extra)
         return command, options
```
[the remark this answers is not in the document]
rmartin16: Good point. Refactored to create a single `return`.

src/briefcase/commands/base.py
```diff
@@ -294,7 +296,8 @@ def create_command(self):
         command = format_module.create(
             base_path=self.base_path,
             apps=self.apps,
-            input_enabled=self.input.enabled,
+            logger=self.logger,
+            console=self.input,
```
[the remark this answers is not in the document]
rmartin16: Fair. Tunnel vision on the goal :)

src/briefcase/cmdline.py
```diff
@@ -14,7 +14,7 @@
 )
 
 
-def parse_cmdline(args):
+def parse_cmdline(args, logger=None):
```
freakboy3742, line 17: Is the default value for `logger` needed here?

src/briefcase/commands/base.py
```diff
@@ -192,7 +194,7 @@ def __init__(
         # These are abstracted to enable testing without patching.
         self.cookiecutter = cookiecutter
         self.requests = requests
-        self.input = Console(enabled=input_enabled)
+        self.input = console or Console(enabled=input_enabled)
```
freakboy3742, line 197: Is there any value in having this fallback? Given that the purpose of this PR is to ensure we only ever have a single Console and Logger, preserving a mechanism in the constructor to create a new Console seems like a potential footgun. Should `logger` and `console` be promoted to required arguments (and, I guess, input_enabled removed as an argument)?

src/briefcase/commands/base.py
```diff
@@ -202,8 +204,8 @@ def __init__(
         # The internal Briefcase integrations API.
         self.integrations = integrations
 
-        # Initialize default logger (replaced when options are parsed).
-        self.logger = Log()
+        # Initialize logging.
+        self.logger = logger or Log()
```
freakboy3742, line 208: As above - is this desirable?

src/briefcase/commands/base.py
```diff
@@ -546,9 +522,8 @@ def clone_options(self, command):
 
         :param command: The command whose options are to be cloned
         """
-        self.input.enabled = command.input.enabled
-        self.logger = command.logger
         self.is_clone = True
+        self.save_log = command.save_log
```
freakboy3742, line 526: Should these properties be passed in as `**kwargs` to the constructor, rather than "fixed" after the construction of the main command object?

freakboy3742 (changes requested):
On the whole, this looks good; I've flagged a couple of relatively minor things inline.

src/briefcase/commands/base.py
mhsmith: If they became required arguments, wouldn't that require dozens of updates to all the places where commands are created in the unit tests?

I think a cleaner solution to this is for the Log and the Console to become attributes of a single "services" or "tools" object rather than the Command, as discussed at https://github.com/beeware/briefcase/pull/801#pullrequestreview-1096865165. 
But this PR is still a good improvement in itself: the bigger picture can be left to #801.

freakboy3742: The impact on tests is definitely a concern; however, I'd prefer mildly more inconvenient tests over potentially dangerous public APIs.

I'm not sure I see how moving Log and Console to ToolsCache makes this substantially better though - we'd still need to modify every Command instantiation; the only difference would be providing a single tools/services object, rather than a Log and a Console.

rmartin16: ngl - part of my thinking was minimizing impact to tests....but that isn't a great reason by any means.

Skimming the use of command attributes from within tools, the use of `input` and `logger` are common. So, if `input` and `logger` are not moved in to the new `ToolCache`, they would need to be part of the initialization. Given that, it may be better to move everything over to reference `input` and `logger` from `command.tools`...

freakboy3742: That's a really good point - I hadn't considered that. I guess that seals the case for input and logger being on the tools object.

The downside is that we'll need to use `self.tools.logger.info()` everywhere that we want to log inside a command... which isn't great API. I guess we could make `self.log_info()` a member method on BaseCommand?

mhsmith: I don't think there's any need for that, the Command constructor can just set `self.logger = self.tools.logger`, and all the existing logging code can remain unchanged.

rmartin16:
Based on the discussion in https://github.com/beeware/briefcase/pull/852#discussion_r965376858, the changes in this PR will be moved back in to #801 since the changes become dependent again.

src/briefcase/commands/base.py
rmartin16: In light of this discussion, I am:
 - moving these changes back to #801
 - removing footgun defaults for `logger` and `console`

src/briefcase/commands/base.py
rmartin16: No; removed.

src/briefcase/commands/base.py
rmartin16: Moving to keyword arguments for `BaseCommand`.

rmartin16: Actually, I moved `save_log` to `Log` so it's handled similarly to `verbosity` and `is_clone` is now set by the consolidated command factory.

PR #854, opened 2022-09-09
Fixes #550 - Add ability to remove files from the bundled app

### LINKED ISSUE #550: Allow excluding files and folders from the build
I'd like for there to be a way to exclude files and folders from the final application build. I have a number of files in the src folder which are only needed for development and not on clients actually running the application.

A simple `excludes` parameter in the pyproject.toml would be ideal. something like this:

```toml
[tool.briefcase.app.myapp]
formal_name = "My App"
description = "My App"
icon = "src/myapp/client/resources/client"
sources = ['src/myapp']
sources_exclude = [
    '*.ui',
    '*.xcf',
    '*.qrc',
    'src/myapp/server',
    'src/myapp/tests',
    '**/__pycache__']
```

I need to keep these files in the source tree for development, but packaging them up with the application doesn't really do any good, and just means extra data to send over the network and store on everyone's hard drives.

As an extra bonus, it would also be nice to have a minify option for all the Python source files as well, but that's a separate issue.

freakboy3742:
The template, sources and support package all add files to the bundled app; however, sometimes files are added that aren't required. 

For example:

* the macOS support package includes an XCframework that isn't needed in a runtime app
* the python standard library contains many modules that a specific app may not require
* An application may contain files in its `sources` folder that aren't needed at runtime.

With this PR, an app can declare a `cleanup_paths` attribute that describes the paths to be removed after generating the template, and installing app sources, app dependencies, and the support package.

A similar attribute can be declared by the app template.

## PR Checklist:


tests/commands/create/test_cleanup_app_content.py
```diff
@@ -0,0 +1,134 @@
...
+    """A directory can be cleaned up."""
+    myapp_unrolled.cleanup_paths = ["path/to/support/dir1"]
+
+    # Cleanup app content
+    create_command.cleanup_app_content(myapp_unrolled)
+
+    # Confirm the files in dir1 have been removed
+    assert not (support_path / "dir1" / "a_file1.txt").exists()
+    assert not (support_path / "dir1" / "a_file2.doc").exists()
+    assert not (support_path / "dir1" / "b_file.txt").exists()
```
mhsmith, line 42: This should also check whether the directory itself has been removed. And if it the directory doesn't exist, then the checks for the individual files are unnecessary.

We should also test a configuration that removes every file in a directory, but leaves the directory itself.

tests/commands/create/test_cleanup_app_content.py
```diff
@@ -0,0 +1,134 @@
...
+    myapp_unrolled.cleanup_paths = ["path/to/support/dir*"]
+
+    # Cleanup app content
+    create_command.cleanup_app_content(myapp_unrolled)
+
+    # Confirm the matching directories have been removed
+    assert not (support_path / "dir1" / "a_file1.txt").exists()
+    assert not (support_path / "dir1" / "a_file2.doc").exists()
+    assert not (support_path / "dir1" / "b_file.txt").exists()
+    assert not (support_path / "dir2" / "b_file.txt").exists()
```
mhsmith, line 73: As above.

tests/commands/create/test_cleanup_app_content.py
```diff
@@ -0,0 +1,134 @@
...
+    assert (support_path / "other" / "b_file.txt").exists()
+
+
+def test_file_glob_cleanup(create_command, myapp_unrolled, support_path):
+    """A glob of files can be cleaned up."""
+    myapp_unrolled.cleanup_paths = ["path/to/support/dir1/*.txt"]
```
mhsmith, line 79: The `**` syntax should also be tested.

mhsmith (commented):
> the macOS support package includes an XCframework that isn't needed in a runtime app

Is this because it's used by the "xcode" template but not by the "app" template?

> An application may contain files in its sources folder that aren't needed at runtime.

In that case, shouldn't this setting also be implemented by `briefcase update`?

freakboy3742:
> > the macOS support package includes an XCframework that isn't needed in a runtime app
> 
> Is this because it's used by the "xcode" template but not by the "app" template?

Correct. The Xcode template uses the XCframework as an input, but the support library is copied into the final location by the Xcode project, so the XCframework isn't duplicated.

> > An application may contain files in its sources folder that aren't needed at runtime.
> 
> In that case, shouldn't this setting also be implemented by `briefcase update`?

Good catch - I'll add this.

mhsmith (approved):

PR #855, opened 2022-09-09
Add FAQ entry about binary packages.

freakboy3742:
Adds a FAQ entry about binary package support, and the process for requesting additional iOS/Android binary packages.

## PR Checklist:

### DIFF
--- a/changes/471.doc.rst
+++ b/changes/471.doc.rst
@@ -1,0 +1,1 @@
+Added FAQ entries on the state of binary package support on mobile.
--- a/docs/background/faq.rst
+++ b/docs/background/faq.rst
@@ -4,22 +4,21 @@ Frequently Asked Questions
 What version of Python does Briefcase support?
 ----------------------------------------------
 
-Python 3.5 or higher.
+Python 3.8 or higher.
 
 What platforms does Briefcase support?
 --------------------------------------
 
 Briefcase currently has support for:
 
-  * macOS (producing DMG files, or raw .app files)
-  * Linux (producing AppImage files)
-  * Windows (producing MSI installers)
-  * iOS (producing Xcode projects)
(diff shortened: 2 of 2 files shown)


freakboy3742:
@mhsmith I've added a stub entry for the error message that is generated by Android; I had some difficulty getting an Android project going with the current state of the Chaquopy template PR. On iOS, there's *no* error generated - a macOS wheel is transparently generated. I probably need to add something similar to the shim you've put in place for Chaquopy to prevent source builds on iOS.

mhsmith (approved):
> I had some difficulty getting an Android project going with the current state of the Chaquopy template PR

This was probably because Chaquopy 12.0.1 requires a local.properties file at the root of the project. This has been fixed in the development version (https://github.com/chaquo/chaquopy/issues/672); meanwhile, you can work around it by creating an empty file.

I've added the Chaquopy error message along with a link to its own FAQ. In hindsight, the wording "Chaquopy cannot compile native code" is misleading, since we *can* compile native code, just not in the context of a Gradle build. Maybe you can come up with something better for iOS.

> I probably need to add something similar to the shim you've put in place for Chaquopy to prevent source builds on iOS.

I don't know if GitHub has a way to compare a specific directory between two branches, but you can see how Chaquopy does it with this command: 
```
git diff upstream-pip master product/gradle-plugin/src/main/python/setuptools
```
As discussed, Briefcase isn't running a fixed version of pip or setuptools like Chaquopy, and newer versions of pip which support PEP 517/518 may build sdists in separate virtual environments with arbitrary versions of setuptools if requested by the package's pyproject.toml. So in that case, the monkey-patch will have to be applied using sitecustomize, and will have to work across a reasonably wide range of setuptools versions. Hopefully the patch is simple enough that that will be feasible.

PR #857, opened 2022-09-12
Updated FAQ to include text produced by iOS when installing binary packages

freakboy3742:
[This commit](https://github.com/beeware/Python-Apple-support/pull/161/commits/92089b7f6ec01fab7c752fe0b2675765f6a1be20) adds a monkeypatch to distutils to disable the C compiler. This PR updates the FAQ to reference the message that will be generated.

Also adds the `--prefer-binary` flag when installing packages so that an older binary will be preferred to a more recent source package.

## PR Checklist:

### DIFF
--- a/changes/857.misc.rst
+++ b/changes/857.misc.rst
@@ -1,0 +1,1 @@
+Updated FAQ to include the iOS message when compiling binary modules.
--- a/docs/background/faq.rst
+++ b/docs/background/faq.rst
@@ -60,10 +60,11 @@ that is on PyPI. If you experience problems when building or running an app on a
 mobile platform that appear to be related to a missing dependency, check the
 build logs for your app. If you see:
 
+* On Android: the error `"Chaquopy cannot compile native code"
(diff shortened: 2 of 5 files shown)


mhsmith (approved):
`--prefer-binary` applies to all wheels, whether pure-Python or not. I've encountered some projects on PyPI which released pure-Python wheels for old versions, and then stopped for some reason and now only release sdists for new versions. In that case, `--prefer-binary` will prefer the old wheel over the newer sdist, which probably isn't what the user wants. 

I worked around this by adding [this patch](https://github.com/chaquo/chaquopy/blob/12.0.1/product/gradle-plugin/src/main/python/pip/_internal/index.py#L704) to Chaquopy's copy of pip, which causes it to prefer only binary wheels, not pure-Python ones.

Unfortunately I didn't make a note of what packages this affected, so I don't know whether this is still necessary. `--prefer-binary` will get us 99% of the way there, so let's just go with it, and if it causes any problems we'll find out soon enough.

PR #859, opened 2022-09-13
Capture app output in log

rmartin16:
Ensure output from running the app is captured in the `briefcase` log.

Windows apps are excepted because of [windows fun](https://github.com/beeware/briefcase/pull/782#discussion_r920621894).

## PR Checklist:

### DIFF
--- a/changes/859.misc.rst
+++ b/changes/859.misc.rst
@@ -1,0 +1,1 @@
+When an app runs, its output is captured is the ``briefcase`` log.
--- a/src/briefcase/commands/dev.py
+++ b/src/briefcase/commands/dev.py
@@ -99,6 +99,7 @@ class DevCommand(BaseCommand):
                 env=env,
                 check=True,
                 cwd=self.home_path,
+                stream_output=True,
             )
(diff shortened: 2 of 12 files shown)


freakboy3742 (approved):
Looks fairly straightforward... interestingly, this change doesn't appear to be needed for Android on macOS (it doesn't *hurt*, but it also doesn't seem necessary). Unfortunately, I'm not able to test Android logging behaviour on Linux or Windows, so I'll take your word for it that it is needed. However, it's definitely needed for Flatpak and AppImage in my testing, so I'm happy to accept on that basis.

rmartin16:
@freakboy3742 

Android logging via ADB was captured in the log when we implemented the logfile in #760. So, this PR shouldn't change anything for Android.

That said, I was thinking it may be more appropriate to default `stream_output=True` (similar to how we defaulted `text=True`). I may go through and identify the cases where `subprocess.run()` should not or cannot be in the log....Flatpak comes to mind due to the dynamic elements in its output that cannot be disabled.

PR #862, opened 2022-09-19
Fixes #824 - Use briefcase version rather than Python version for template versioning

### LINKED ISSUE #824: Change templates to one branch per Briefcase version
It's occurred to me that the Chaquopy template PR (https://github.com/beeware/briefcase-android-gradle-template/pull/52) won't be compatible with old versions of Briefcase, because its briefcase.toml omits `app_support_path`, and replaces `app_packages_path` with `app_requirements_path`. The Flatpak template did both of these things too, but that was no problem because old versions of Briefcase didn't support Flatpak anyway.

The issue of it being difficult to update the templates without breaking old versions of Briefcase has come up several other times in the last couple of months, most recently at https://github.com/beeware/briefcase/issues/804#issuecomment-1198778788. So I'd like to revisit the idea of a closer coupling between the Briefcase version and the template version,, which we previously discussed at https://github.com/beeware/briefcase/pull/775#issuecomment-1169359557:

> there are significant differences in the iOS and macOS templates between versions. [...] The macOS App template has an embedded binary which is, by definition, Python version dependent. [...] On top of that, I'm working on #629 right now which will (hopefully) add a stub binary for Windows

Here are the differences between the 3.9 and 3.10 branches of each current template:

* android-gradle: none
* ios-Xcode: one string in a .m file
* linux-appimage: none
* linux-flatpak: none
* macOS-app: one binary
* macOS-Xcode: one string in a .m file
* windows-app: one binary
* windows-VisualStudio: one string in a .cpp file

The strings are obviously easy to template. For the binaries, we can make the template contain them all side by side, with Python version numbers in their names, and provide a [cookiecutter "post" hook](https://cookiecutter.readthedocs.io/en/stable/advanced/hooks.html) which renames one of them to the target name and removes all the others. From the perspective of Briefcase, the cookiecutter output would be exactly the same as it is now. 

> the existence of a 3.X branch is a way for us to gate whether we support [Python] 3.X in apps

This could be done by a cookiecutter "pre" hook.

> Embedding the template hash/tag for a specific version in the Briefcase code would be one way to address this; but that would limit our ability to make smaller hot fixes to the template (or, at least, it would make hot fixes more difficult).

In general I think we should avoid making hot fixes which affect already-released versions of Briefcase, as it makes app builds non-reproducible. Even if a hot fix doesn't actually break anything, if users notice that their app is constantly changing underneath them despite them staying on the same Briefcase version, it'll make them unwilling to trust us as the foundation of a serious project.

However, if we used a branch rather than a tag, then that would still leave us the option of making hot fixes, as well as making development more convenient:

* Increment the Briefcase version number in the repository immediately after a release, e.g. it should currently be 0.3.9.
* Rename the "dev" branch on each template to "0.3.9", and make it compatible with all Python versions as described above.
* When Briefcase 0.3.9 is released, we create new template branches for the next version (initially pointing at the same commit as the previous version), and make them the default branch on GitHub. We'd still have the option of pushing bug fixes to the old template branches, but I don't think we'd need to do that very often, and then probably only to the current stable version. 

The top-level briefcase-template doesn't currently have multiple branches, but since it also receives updates to support new Briefcase features (e.g. https://github.com/beeware/briefcase-template/pull/31, which accidentally broke AppImage builds), it would make sense to handle that in exactly the same way.

Advantages of this scheme:

* There would be no need to constantly merge between 5 parallel branches, so the commit history of each template would mostly be linear and easy to follow.
* It would avoid the need to ask whether every template update is compatible with every old version of Briefcase. Even if it is compatible, convincing ourselves of that is still a drag on our time, because we can't feasibly test every version of Briefcase every time we update a template. And if it's not compatible, we waste even more time coming up with an alternative approach. Either way, as time goes by and more Briefcase versions are in use in the wild, template updates will become progressively more difficult. Whereas if each version of the template only needs to support one version of Briefcase, we have a lot more flexibility to update it.

freakboy3742:
Templates currently use the Python version; this means there is a 3.7, 3.8, 3.9, 3.10 (and soon to be 3.11) template for every template we use. Most of these templates are identical, except for minor changes. 

However, there can be very different needs for templates between *Briefcase* versions - for example, the introduction of Chaquopy into the Android backend introduces requirements on templates that aren't supported with Briefcase 0.3.9. 

This PR modifies template usage to use the Briefcase version as the template branch for the wizard and the application templates. 

To support this:

* The PR drops Python 3.7 support. This is because the template differences between 3.7 and 3.8 are significant; also, back porting iOS binary module support and Android Chaquopy support to 3.7 will be onerous. Python 3.7 will be EOL in about 12 months, so this isn't especially aggressive.
* The PR introduces setuptools_scm. As a result, `briefcase.__version__` will automatically return a git-hash-based version identifier. 
* As introducing setuptools_scm required moving more configuration into pyproject.toml, I've moved a couple of the setup.cfg configurations into pyproject.toml. `setuptools` still lists pyproject.toml support as beta, so we can't completely move to pyproject.toml (yet).
* When running the project wizard, or creating an app, Cookiecutter will clone a template that is the `v{basic_version}` - so, Briefcase `0.3.10` will use the `v0.3.10` template; `v0.3.10a1`, `v0.3.10b2`, and `v0.3.10rc3` would also use the `0.3.10` template.
* Dev versions have additional behaviour: If you are in a git checkout, the `briefcase.__version__` will automatically include a git distance and hash (e.g., `0.3.10.dev73+gad61a29.d20220919`). If you are in a dev branch, briefcase will try to use the basic version as the template branch; but if that branch doesn't exist, it will fall back to the `main` branch of the repo.
* When we start pre-release testing for the `v0.3.10` release, we will branch the repository to create a `v0.3.10` branch in each of the Briefcase template repos (the wizard template, plus the template for each backend).
* Templates that specify a `support_path` are now also expected to specify a `support_revision`, which pins the support package revision that will be used with that template. End users can override this with the `support_revision` configuration options in `pyproject.toml`. 
* The automated addition of a support revision query argument when specifying support packages with a custom support URL has been removed; if you need to include a revision when specifying a custom support URL query, include the query argument in the URL.

## PR Checklist:


mhsmith:
> If you are in a git checkout, the `briefcase.__version__` will automatically include a git distance and hash (e.g., `0.3.10.dev73+gad61a29.d20220919`

For future reference: setuptools_scm generates a version like this when the most recent tag in the repository was `0.3.9`, and it was 73 commits ago. It increments the final part of the version number automatically.

.github/workflows/ci.yml
```diff
@@ -92,7 +92,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11.0-alpha - 3.11.0']
+        python-version: ['3.9', '3.10', '3.11.0-alpha - 3.11.0']
```
mhsmith, line 95: Should this now change to "3.11.0-rc"?

rmartin16:
> The PR drops Python 3.7 support

I'm gonna take this to mean we can use the walrus operator now :)

freakboy3742:
> I'm gonna take this to mean we can use the walrus operator now :)

It does! RELEASE THE WALRUS! 🤣

.github/workflows/ci.yml
freakboy3742: It could, but it doesn't have to. The build step will pull in the most recent tag that matches; if you check the build logs, the most recent build was against 3.11.0rc2. Once 3.11 is finalised, we should be able to add 3.11 without any adornment, and add "3.12.0-alpha - 3.12.0"

pyproject.toml
```diff
@@ -1,10 +1,29 @@
...
+multi_line_output = 3
+
+[tool.pytest.ini_options]
+testpaths = ["tests"]
+
+# need to ensure build directories aren't excluded from recursion
+norecursedirs = []
+
+[tool.setuptools_scm]
+# To enable SCM versioning, we need an empty tool configuration for setuptools_scm
```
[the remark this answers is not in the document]
freakboy3742: I originally thought it was still needed; but after a conversation with @jezdez and some experimentation, it looks like it might not be.

src/briefcase/__init__.py
```diff
@@ -1,13 +1,8 @@
...
-# __version__ = '1.2.3.dev1'  Development release 1
-# __version__ = '1.2.3a1'     Alpha Release 1
-# __version__ = '1.2.3b1'     Beta Release 1
-# __version__ = '1.2.3rc1'    RC Release 1
-# __version__ = '1.2.3'       Final Release
-# __version__ = '1.2.3.post1' Post Release 1
-
-__version__ = "0.3.9"
+# Read version from SCM metadata
+__version__ = version("briefcase")
```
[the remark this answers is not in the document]
freakboy3742: That's a good point. I've had a long chat with @jezdez about this, and the tl;dr is that there isn't a good best-practice here. The tools are definitely optimised for the "generate a dist-info file with a version"; and there are a variety of commonly used, but not 100% endorsed uses/hacks that let you use setuptools_scm at runtime (one of which is exactly what you've described here). 

I'll put up a change that uses the import fallback approach, and add setuptools_scm to the development requirements.

.github/workflows/ci.yml
mhsmith: Oh, I see, that hyphen is actually [a version range](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-input).

src/briefcase/commands/create.py
```diff
@@ -367,28 +326,27 @@ def _download_support_package(self, app):
...
-                try:
-                    self.logger.info(f"... pinned to revision {app.support_revision}")
-                    # If a revision has been specified, add the revision
-                    # as a query argument in the support package URL.
-                    # This is a lot more painful than "add arg to query" should
-                    # be because (a) url splits aren't appendable, and
-                    # (b) Python 3.5 doesn't guarantee dictionary order.
-                    url_parts = list(urlsplit(support_package_url))
-                    query = list(parse_qsl(url_parts[3]))
-                    query.append(("revision", app.support_revision))
```
mhsmith, line 352: If this URL query argument functionality is being removed, then it should also be removed from `support_revision` in configuration.rst.

tests/commands/create/test_install_app_support_package.py
```diff
@@ -323,9 +268,6 @@ def test_install_pinned_custom_app_support_package_url_with_args(
 ):
     """A custom support package can be specified as URL with args, and pinned
     to a revision."""
```
mhsmith, line 270: If I understand correctly, it can no longer be pinned to a revision: the URL is just used as-is. In which case, this whole test function isn't really necessary anymore: it was only added because we knew we were modifying the URL.

tests/commands/new/test_new_app.py
```diff
@@ -46,18 +53,79 @@ def test_new_app(monkeypatch, new_command, tmp_path, briefcase_version):
...
         output_dir=os.fsdecode(tmp_path),
-        checkout="v37.42.7",
+        checkout=expected_branch,
         extra_context=app_context,
     )
 
 
+def test_new_app_dev(monkeypatch, new_command, tmp_path):
+    """In a dev version, template will fall back to the 'main' branch if a
+    versioned template doesn't exist."""
```
mhsmith, line 70: There should also be a test that it *doesn't* fall back to the main branch in a non-dev version, but throws an error. And likewise if the user explicitly requests a branch that doesn't exist.

tests/commands/create/test_install_app_support_package.py
freakboy3742: It *can* be pinned, as long as you're not specifying a custom template as well. The use case here is someone actively opting into a new version of an official support package, or explicitly *not* opting into a new version of a support package - e.g., if we published an update to the 3.10.0 branch to use a new support revision, a user could pin the older support revision.

freakboy3742: I've added some extra tests around this (and found a bug in the process - around explicit branches in the development version).

src/briefcase/commands/create.py
```diff
@@ -324,10 +278,15 @@ def generate_app_template(self, app: BaseConfig):
 
             # Development branches can use the main template.
             self.logger.info(
-                f"Branch template {app.template_branch} not found; falling back to development template"
+                f"Template branch {app.template_branch} not found; falling back to development template"
             )
             app.template_branch = "main"
```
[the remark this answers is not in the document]
freakboy3742: I've added some extra tests around this; and found a bug in the process - the case of an explicit, invalid branch on the development version.

tests/commands/create/test_install_app_support_package.py
freakboy3742: Looking closer, the problem here is that the test name and description is misleading. Testing a support URL that *has* args is still valid (although it's less of an issue than before, because weren't actually manipulating those args).

I've also added a warning (plus tests) for the user if they specify a custom template and a template revision.

freakboy3742:
@mhsmith CI builds for the support packages completed, so the test suite can now complete on the test apps. Pending final review, I think this is ready to land.

mhsmith (approved):

PR #867, opened 2022-09-26
Dependency Management

rmartin16:
## Changes
 - Update project dependencies to align with their most recent major release
 - Use `dependabot` to update python requirements and CI actions
 - Update `pre-commit` dependencies
 - Replace requirements files with `options.extras_require` in setup.cfg
 - Promote `pytest` warnings to errors

## Noteworthy/Controversial
- Removing `requirements.txt` files
  - Every time i set up a new environment, I have to run `pip` multiple times.
  - This allows `pip install -e .[dev,test,docs]` to set up a dev environment.
- Promote all warnings to errors
  - Useful to find sneaky problems...but can be overly noisy. I fixed all the current issues.
- Dependency versions use any "[compatible release](https://peps.python.org/pep-0440/#compatible-release)" 
  - I updated the minimum version for all dependencies to their most recent major version using `~=` so any bumps in their minor versions are automatically used....but major version bumps will not be used.
  - `dependabot` should let us know when major version releases happen.
  - A few notes:
    - `cookiecutter`
       - briefcase requires at least 1.5 to run
       - 1.7.0 was a revive in 2019; 2.1 released May 2022
     - `GitPython`
       - 10 days after 3.0.8 was released, they moved to 3.1 where they are today
     - `dmgbuild`
       - 1.4 released one year after 1.3.3
       - they've been looking for a new maintainer since the 1.4 release

## Related:
- #868

## PR Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,16 +1,16 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 22.8.0
     hooks:
     - id: black
       language_version: python3
   - repo: https://github.com/PyCQA/flake8
-    rev: 4.0.1
+    rev: 5.0.4
     hooks:
       - id: flake8
-        additional_dependencies: [flake8-eradicate==1.2.0]
+        additional_dependencies: [flake8-eradicate==1.4.0]
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.2.0
+    rev: v4.3.0
     hooks:
       - id: check-toml
       - id: check-yaml
@@ -24,12 +24,12 @@ repos:
       - id: isort
         additional_dependencies: [toml]
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.32.1
+    rev: v2.38.2
     hooks:
       - id: pyupgrade
         args: [--py37-plus]
   - repo: https://github.com/myint/docformatter
-    rev: v1.4
+    rev: v1.5.0
     hooks:
     - id: docformatter
       args: [--in-place]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,14 +1,4 @@
 repos:
-  - repo: https://github.com/psf/black
-    rev: 22.8.0
-    hooks:
-    - id: black
-      language_version: python3
-  - repo: https://github.com/PyCQA/flake8
-    rev: 5.0.4
-    hooks:
-      - id: flake8
-        additional_dependencies: [flake8-eradicate==1.4.0]
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.3.0
     hooks:
@@ -33,3 +23,13 @@ repos:
     hooks:
     - id: docformatter
       args: [--in-place]
+  - repo: https://github.com/psf/black
+    rev: 22.8.0
+    hooks:
+    - id: black
+      language_version: python3
+  - repo: https://github.com/PyCQA/flake8
+    rev: 5.0.4
+    hooks:
+      - id: flake8
+        additional_dependencies: [flake8-eradicate==1.4.0]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
     rev: v2.38.2
     hooks:
       - id: pyupgrade
-        args: [--py37-plus]
+        args: [--py38-plus]
   - repo: https://github.com/myint/docformatter
     rev: v1.5.0
     hooks:
--- a/tests/console/test_Log.py
+++ b/tests/console/test_Log.py
@@ -69,7 +69,8 @@ def test_save_log_to_file_no_exception(tmp_path, now):
     log_filepath = tmp_path / "briefcase.2022_06_25-16_12_29.dev.log"
 
     assert log_filepath.exists()
-    log_contents = open(log_filepath, encoding="utf-8").read()
+    with open(log_filepath, encoding="utf-8") as log:
+        log_contents = log.read()
 
     assert log_contents.startswith("Date/Time:       2022-06-25 16:12:29")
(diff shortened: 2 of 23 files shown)


tests/console/test_Log.py
```diff
@@ -69,7 +69,8 @@ def test_save_log_to_file_no_exception(tmp_path, now):
     log_filepath = tmp_path / "briefcase.2022_06_25-16_12_29.dev.log"
 
     assert log_filepath.exists()
-    log_contents = open(log_filepath, encoding="utf-8").read()
+    with open(log_filepath, encoding="utf-8") as log:
+        log_contents = log.read()
```
rmartin16, line 73: Needed to close these files to prevent `ResourceWarning` errors.

tox.ini
```diff
@@ -26,13 +28,13 @@ commands =
 [testenv:towncrier]
 skip_install = True
 deps =
-    towncrier == 21.9.0
+    towncrier ~= 22.8
```
rmartin16, line 31: This upgrade seems the most capricious...however, I ran `tox -e towncrier` and it deleted the news fragments and updated release docs.

tox.ini
```diff
@@ -9,9 +9,11 @@ skip_missing_interpreters = true
...
 [testenv]
 setenv = PYTHONPATH = {toxinidir}/src
-deps =
-    -r{toxinidir}/requirements.test.txt
-
+extras =
+    test
+allowlist_externals =
+    pytest
+    coverage
```
rmartin16, line 16: I added these because I get the error below if aren't there when I run `tox -e py310`:

```
WARNING: test command found but not installed in testenv
  cmd: /home/user/github/rmartin16/briefcase/venv-3.10.7-rmartin16-briefcase/bin/coverage
  env: /home/user/github/rmartin16/briefcase/.tox/py310
Maybe you forgot to specify a dependency? See also the allowlist_externals envconfig setting.

DEPRECATION WARNING: this will be an error in tox 4 and above!
```

freakboy3742 (commented):
These changes look great. There's one complication, though: while the switch to `extra_requires` is very nice, it breaks the ReadTheDocs configuration. 

As it happens, there's been a related breakage tied to the introduction of importlib_metadata;  I suspect we will need to introduce a RTD configuration file. I'm looking into this now.

.pre-commit-config.yaml
```diff
@@ -24,12 +24,12 @@ repos:
       - id: isort
         additional_dependencies: [toml]
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.32.1
+    rev: v2.38.2
     hooks:
       - id: pyupgrade
         args: [--py37-plus]
```
rmartin16, line 30: Just noticed `pyupgrade` is still referencing Python 3.7.  Looks like it'll replace `FsPathMock` in tests if this is updated to `--py38-plus`.

freakboy3742: IIRC, there's a bunch of stuff that is flagged for Py3.6/3.7 deprecation that we could clean up - mostly related to fsencode/decode and places where Pathlib objects are being used.

rmartin16: yeah....we can save it for later.

btw, this is one of the automated change it made...although, it left the imports for flake then to get made at:

```
@@ -194,11 +189,8 @@ def test_unpack_fail(mock_tools, tmp_path):
     wix_zip_path = os.fsdecode(tmp_path / "tools" / "wix.zip")
     # Consider to remove if block when we drop py3.7 support, only keep statements from else.
     # MagicMock below py3.8 doesn't have __fspath__ attribute.
-    if sys.version_info < (3, 8):
-        wix_zip = FsPathMock(wix_zip_path)
-    else:
-        wix_zip = MagicMock()
-        wix_zip.__fspath__.return_value = wix_zip_path
+    wix_zip = MagicMock()
+    wix_zip.__fspath__.return_value = wix_zip_path
 ```

freakboy3742 (approved):
I've been able to resolve the RTD problem; with the added bonus that we're actually getting PR previews of docs from RTD itself now. I've also integrated the `pyupgrade` flag update (and the changes that follow from that).

mhsmith:
> dependabot should let us know when major version releases happen

I'm not sure how often this will happen, but I think we should be cautious about picking up new versions just for their own sake. We'd be giving ourselves busy-work, and potentially introducing new bugs, in exchange for no clear benefit.

rmartin16:
> I think we should be cautious about picking up new versions just for their own sake.

I don't necessarily disagree; however, I decided against pinning specific versions in this PR because we're still sharing the briefcase install env with the user's dev env. While version conflicts may be unlikely, they are at least possible...perhaps especially if we don't support recent releases. So, I do think there's still room for improvement here.

mhsmith:
Yes, to avoid conflicts with other things installed in the same environment, we should be declaring requirements on a fairly wide range of versions, like this:
* If a new major version of a requirement comes out and is confirmed to be compatible for our purposes, we can extend the upper end of the range and leave the lower end alone.
* Or it's too difficult to maintain compatibility with both the old and the new versions, extend the upper end and contract the lower end simultaneously. But in this case, wait until the new version has been adopted widely enough that this move will solve more conflicts than it creates.

freakboy3742:
To be a slight contrarian here - there *is* a tangible benefit to doing lots of small incremental changes: it's much easier to schedule that work as a series of 10 minute fixes done as part of general housekeeping. If you use the "only upgrade when the LTS changes" approach, you can easily end up in a situation where you need to schedule an open ended block of time to upgrade because you have no idea how complex the upgrade is - and that upgrade is often required because you're backed into a corner by an incompatibility with some other factor in your environment.

I agree that we don't want to convert our day job into "keeping Dependabot quiet", but with a combination of good CI (which we have), automated PRs (which we mostly have - we just need to work out how to work around the changenote issue), and the ability to tell Dependabot to be quiet as required (which the bot provides through its chat interface), I think we can keep this managed.

rmartin16:
I'm working through the details of automating news fragments for Dependabot now; I have a workable plan...just have to implement it.

PR #877, opened 2022-09-27
Disable Wait Bar for Windows batch scripts (Fixes #811)

### LINKED ISSUE #811: CMD.exe's Terminate Batch Script Prompt is Swallowed When Using a Wait Bar While Running a `bat` Script
This originated from and is made significantly worse by:
 - #809

However, all Python versions are affected by the aspect described here.

**Describe the bug**
`CMD.exe` emits a `Terminate batch job (Y/N)?` prompt to the user when a CTRL+C is sent during execution of a batch script. AFAICT, this is to allow the execution of the rest of batch script while canceling its current action (perhaps similar to bash's default behavior of continuing on error without `set -e`). This behavior will be ubiquitous regardless of the shell the user started briefcase in; all shells in Windows (e.g. powershell, terminal, msys2) will all defer batch script interpretation to `cmd`.

Many of the Android tools are invoked from `bat` scripts provided from Google. For instance, sending a CTRL+C while building an Android APK results in:
- `cmd` emitting its terminate prompt
- the prompt is swallowed by Rich's Wait Bar
- the user can't tell anything really happened since the Wait Bar is still pulsing
- the whole process remains stagnate until the user sends another CTRL+C
- (On Python 3.7/3.8, this also leads to Python hanging entirely, see # 809)

**To Reproduce**
1. Run `briefcase build android gradle`
2. Once the `Building...` Wait Bar starts, send CTRL+C
3. Witness issue

**Expected behavior**
CTRL+C should allow the user to exit or provide clear instructions (such as the terminate batch prompt) on how to exit.

**Environment:**
 - Python version: 3.7, 3.8, 3.9, 3.10
 - Software versions:
   - Briefcase: https://github.com/beeware/briefcase/commit/8df2d112eba97175c740931cf168c89c370307d1

rmartin16:
Finally getting around to this:
 - #811 

## Changes
Removed the Wait Bar when running batch scripts on Windows.

## Design
I tried to generalize this without going too overboard. In theory, though, this should make it pretty trivial to achieve the same outcome for other reasons.

## PR Checklist:


src/briefcase/integrations/subprocess.py
```diff
@@ -214,7 +255,8 @@ def verify(cls, tools):
         tools.subprocess = Subprocess(tools)
         return tools.subprocess
 
-    def run(self, args, stream_output=False, **kwargs):
+    @pause_console_control_if
+    def run(self, args, /, stream_output=False, **kwargs):
```
freakboy3742, line 259: Is there any particular reason for introducing the positional-only marker here? We're trying to proxy the API from `subprocess`, and `run()` et al don't have this requirement, so there will be an inconsistency in the ways these methods can be invoked. This probably won't matter in practice, as anything after the `args` argument will usually be a kwarg - but it's an added complexity in the codebase that doesn't really improve anything.

src/briefcase/integrations/subprocess.py
```diff
@@ -88,6 +89,46 @@ def get_process_id_by_command(
...
+        if not sub.tools.input.is_console_controlled:
+            return sub_method(sub, cmdline, **kwargs)
+
+        remove_dynamic_elements = False
+
+        # Batch (.bat) scripts on Windows.
+        # If cmd.exe is interrupted with CTRL+C while running a bat script,
+        # it may prompt the user to abort the script and dynamic elements
+        # such as the Wait Bar can hide this message from the user.
+        if not remove_dynamic_elements:
```
freakboy3742, line 118: Is this if redundant? I can't see any way that `remove_dynamic_elements` could be `true` here, given that `remove_dynamic_elements` is a local that was set on L112.

src/briefcase/integrations/subprocess.py
```diff
@@ -88,6 +89,46 @@ def get_process_id_by_command(
...
+            if sub.tools.host_os == "Windows":
+                executable = str(cmdline[0]).strip() if cmdline else ""
+                remove_dynamic_elements = executable.lower().endswith(".bat")
+
+        # Run subprocess command with or without console control
+        if remove_dynamic_elements:
+            with sub.tools.input.release_console_control():
+                return sub_method(sub, cmdline, **kwargs)
+        return sub_method(sub, cmdline, **kwargs)
```
freakboy3742, line 127: This is a really minor stylistic thing - but I would prefer this `return` to be in an `else` clause. Functionally, there's no difference; but on a casual inspection, it wasn't immediately obvious to me that `sub_method` wasn't going to be invoked twice in the case of pre-existing dynamic elements. It won't, of course, because the `return` shortcuts the method flow - but if you don't see that *first*, it's a source of confusion. Adding an `else` clause removes that potential for confusion.

src/briefcase/integrations/subprocess.py
```diff
@@ -88,6 +89,46 @@ def get_process_id_by_command(
     return None
 
 
+def pause_console_control_if(sub_method):
```
freakboy3742, line 92: Not sure I like the naming here - this "reads" like a dangling clause: Pause console control if... what? The dangling "if" part would make sense if the decorator took an argument that was evaluated (e.g., `@pause_console_control_if(is_windows_batch_file)`), but I'm not sure that sort of refactoring is worth it.

Not sure I have an obvious alternative - maybe something like `@ensure_console_is_safe`?

freakboy3742 (changes requested):
Makes sense; a couple of mostly cosmetic questions/clarifications.

src/briefcase/integrations/subprocess.py
rmartin16: This check is redundant right now. I added it for forward safety. If multiple checks are done here in the future, we don't want a check to reverse the decision of a previous check to turn off the Wait Bar. This is intended to help ensure that's apparent.

I can remove it if that isn't a particularly compelling rationale.

freakboy3742: I figured this was probably the intended reason; but given the logic is fairly straightforward, easy to add back in if we need it, and there's no obvious candidate for another "suspend waitbar" use case (that I can think of), I'm inclined to say remove it for now and add it back in if we need it later.

src/briefcase/integrations/subprocess.py
rmartin16: Added the `else`.

src/briefcase/integrations/subprocess.py
rmartin16: I added the positional-only marker so I could simplify inferring where the `cmdline` is in the call parameters of `run()` and `check_output()`. Without this, devs can technically call `run(args=['ls'])` and it works.

Furthermore, though, both `run()` and `check_output()` appear to have always deviated from the the `subprocess` module.

In the `subprocess` module, these functions support an arbitrary number of positional args that are passed to `Popen`; if one isn't careful, these pos args could easily collide with those in `kwargs`.

For the wrapper `Subprocess.check_output()`, it only takes 1 positional argument (or a kwarg called `args`) that must be a an iterable of stringlikes to execute and additional pos args are not supported at all.

The `Subprocess.run()` wrapper has the same issue but goes further because we added the `stream_output` param.

So, while I agree these are proxies for `subprocess`, I don't think they ever actually had compatible signatures. Ultimately, I felt this positional-only marker mostly enforced an already used convention within `briefcase` and removed a little complexity from this decorator.

Alternatively, we might could take a step back and try to better align these wrappers back with `subprocess`.

freakboy3742: The divergence from `subprocess` is a fair point; I guess it's more "aspirational" than a firm requirement.

However, unless I'm missing something, the `args` problem you describe doesn't exist AFAIK:
```
In [1]: def foo(args, **kwargs):
   ...:     print(f"{args=}, {kwargs=}")
   ...: 

In [2]: foo('a', b=3)
args='a', kwargs={'b': 3}

In [3]: foo(args='a', b=3)
args='a', kwargs={'b': 3}

In [4]: foo(b=3, args='a')
args='a', kwargs={'b': 3}
```

If you declare a positional argument, that key won't ever appear in the kwargs, no matter how it's passed in.

rmartin16: That's true; it just changes the assumptions at bit. Instead of assuming the `cmdline` iterable is the first param of the decorated method, this will assume it always has the name `args`. But that isn't necessarily an inherent problem here.

I can switch to that.

src/briefcase/integrations/subprocess.py
rmartin16: That's fair. This name was wonky. Updated it to `ensure_console_is_safe`.

freakboy3742 (approved):
I've cleaned up a couple of extra stray positional markers; but otherwise, looks good!

PR #878, opened 2022-09-27
PackageCommand: run `verify_app_tools()` after other pre-requisite commands

rmartin16:
Got this wrong even after spending too much time thinking about it....or because I spent too much time.

## Changes
The call to `verify_app_tools()` for `PackageCommand` now happens after any per-requisite commands are run.

## PR Checklist:


freakboy3742 (approved):
Makes sense. From a quick check, `publish` has the same problem; `publish` is a no-op, but for the sake of consistency (and future sanity) it's worth cleaning this up now. 

I've pushed the update; once CI passes I'm happy for this to land.

PR #879, opened 2022-09-28
Avoid ResourceWarnings during tests for /dev/null handle

rmartin16:
After #867, `pytest` will raise for all Python errors including those normally suppressed. This change is also causing the errors/warnings below to be printed _sometimes_ when pytest is run.

```
(venv3.10) rkm@eunectes briefcase % pytest tests/commands/new/test_call.py 
..
Ran 2 tests in 0.01s

OK
Exception ignored in: <_io.FileIO name='/dev/null' mode='wb' closefd=True>
ResourceWarning: unclosed file <_io.TextIOWrapper name='/dev/null' mode='w' encoding='utf-8'>
```

These changes ensure the file handle for `/dev/null/ that's opened for the logfile Rich Console is closed before `pytest` exits so this error is showed to developers and avoids subsequent confusion.

In practice, these errors are innocuous. Furthermore, end-users cannot see these errors unless they run a debug version of Python or use the `-Werror` (or perhaps the `-Wdefault`) flag.

## PR Checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,11 +8,11 @@ from git import exc as git_exceptions
 from briefcase.config import AppConfig
 from briefcase.console import Printer
 
-# Stop Rich from inserting line breaks in to long lines of text.
-# Rich does this to prevent individual words being split between
-# two lines in the terminal...however, these additional line breaks
-# cause some tests to fail unexpectedly.
-Printer.console.soft_wrap = True
+
+def pytest_sessionfinish(session, exitstatus):
+    """When pytest is wrapping up, close the /dev/null file handle for the
+    logfile Rich Console to avoid spurious ResourceWarning errors."""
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -8,11 +8,11 @@
 from briefcase.config import AppConfig
 from briefcase.console import Printer
 
-# Stop Rich from inserting line breaks in to long lines of text.
-# Rich does this to prevent individual words being split between
-# two lines in the terminal...however, these additional line breaks
-# cause some tests to fail unexpectedly.
-Printer.console.soft_wrap = True
```
rmartin16, line 15: Setting `soft_wrap=True` is no longer necessary since we always set it to `True` in `console.py`.

freakboy3742 (approved):
👍 Seems reasonable to me!

PR #885, opened 2022-09-29
Add workflow to add a change note to Dependabot PRs

rmartin16:
## Changes
Adds a change note to PRs from Dependabot.

I tested this in an isolated repo:
- Dependabot PR: https://github.com/rmartin16/test-briefcase-5/pull/5
- Non-Dependabot PR: https://github.com/rmartin16/test-briefcase-5/pull/6

## Strategy
Uses the `push` action to catch all commits but filters to those on branches starting with `dependabot/`. The primary advantage of this is the Workflow only runs or shows up in checks for PRs if that branch name condition is met. (I could not find a way to do this filtering this early with other events.)

By default, any commits made to a repo from within a Workflow do not trigger Workflows to run for those commits. To change this behavior, the repo must be checked out using a PAT for a user with write access to the repo. The @brutusthebee bot user can serve this purpose well.

**The PAT must be set as a secret for dependabot and named `BRUTUS_PAT_TOKEN`.** 
 - The token "only" needs `public_repo` permissions.
 - This token is not loaded into the Workflow Run if someone were to create a PR from a fork on a `dependabot/` branch.

## Trade-offs
Using the `pull` event makes other tasks slightly more difficult as opposed to using the `pull_request` event for instance.
- The relevant PR information isn't available from the `github` context so it must be fetched from the Github API.
- The `dependabot/fetch-metadata` action cannot be used to easily access details about what Dependabot did. So, instead, the commit message from Dependabot is used to create the change note.
- These aren't especially onerous, though, IMO.

## PR Checklist:

### DIFF
--- a/.github/workflows/dependabot-changenote.yml
+++ b/.github/workflows/dependabot-changenote.yml
@@ -1,0 +1,45 @@
+name: Dependabot Change Note
+on:
+  push:
+    branches:
+      - 'dependabot/**'
+
+permissions:
+  pull-requests: write
+
+jobs:
+  changenote:
+    name: Dependabot Change Note
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    if: github.actor == 'dependabot[bot]'
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v3.0.2
+        with:
+          fetch-depth: 1
+          token: ${{ secrets.BRUTUS_PAT_TOKEN }}
+
+      - name: Configure git
+        run: |
+          git config --local user.email "$(git log --pretty='%ae' -1)"
+          git config --local user.name "Dependabot[bot]"
+
+      - name: Commit Change Note
+        run: |
+          # Fetch PR number for commit from Github API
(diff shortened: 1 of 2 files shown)


rmartin16:
Ok....this was an fun journey. While Github did [stop allowing](https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/) Workflow runs initiated by Dependabot to have write permissions to the repo in Feb 2019, this action was met with a lot of [backlash](https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852549448) and Github eventually allowed users to configure Workflows such that even Runs for Dependabot could have [write permissions](https://github.blog/changelog/2021-10-06-github-actions-workflows-triggered-by-dependabot-prs-will-respect-permissions-key-in-workflows/) to the repo.

I've finalized the Workflow based on all this wonderful newfound information.

freakboy3742 (approved):
Can't argue with the success you've demonstrated, so lets go with it.

Approving, but not merging for now; I'm waiting for PyPI to start working again so I can push the 0.3.10 release...

freakboy3742:
I've added the `BRUTUS_PAT_TOKEN` secret so we're ready as soon as this lands.

PR #887, opened 2022-09-29
Turn on branch coverage checking.

### LINKED ISSUE #848: Every PR commit generates 2 Codecov email notifications
**Describe the bug**
Here are the first few emails I received for #1574:
* 10:50: PR created with commit f0d4f04
* 10:52: Codecov report for f0d4f04
* 11:03: Pushed commit c7533ef 
* 11:03: Another Codecov report for f0d4f04, with the warning: "Current head [f0d4f04](https://github.com/beeware/toga/commit/f0d4f04136462b2db9460ea7783e93c05c61cfea) differs from pull request most recent head [c7533ef](https://github.com/beeware/toga/commit/c7533ef84ea724a1bae90cd22a58f73dfa4f5551). Consider uploading reports for the commit [c7533ef](https://github.com/beeware/toga/commit/c7533ef84ea724a1bae90cd22a58f73dfa4f5551) to get more accurate results"
* 11:06: Codecov report for c7533ef 

The same pattern then repeats for each subsequent commit: an immediate redundant report for the *previous* commit, followed a few minutes later by a report for the new commit.

This isn't visible on the GitHub website because Codecov deletes its old comments and leaves only the most recent one. But it clutters my email notification threads so badly that I basically have to come and read everything on the GitHub website instead.

**Expected behavior**
Every commit, or group of simultaneously pushed commits, should generate only one Codecov notification.

freakboy3742:
I wasn't aware that (a) coverage had an option for branch coverage, and (b) that it wasn't turned on by default.

This turns on branch coverage, and moves the coverage configuration to pyproject.toml. This reveals a handful of edge cases that aren't currently covered.

This also addresses the noisy emails caused by CodeCov... by removing codecov.

Fixes #848.

## PR Checklist:


freakboy3742:
@mhsmith @rmartin16 This is still in draft form, but I've got it producing a coverage report as an alternative to CodeCov. We'll likely need to get to 100% coverage for this report to be useful; we'd be losing the ability to report on "coverage of this PR", and the ability to ratchet the minimum allowed coverage as tests are added.

I'd be interested in your thoughts on how this output compares.

rmartin16:
I'm mostly indifferent regarding exactly how coverage is tracked....but I don't think I'm convinced removing CodeCov creates a better situation necessarily. I think the benefits outweighs the negatives....and we can probably address the negatives.

**CodeCov notification noise**
Given this PR does all the work to combine the coverage reports, couldn't we just upload that single combined report to CodeCov?
- Even if we cant, turning off the CodeCov PR comments still leaves the [CodeCov check](https://github.com/beeware/briefcase/runs/8585808660) that contains all the same information and is as accessible as this approach.

**"Coverage decrease" false positives**
CodeCov supports a [threshold parameter](https://docs.codecov.com/docs/commit-status#threshold) to help avoid this....albeit in a `codecov.yml` file.

**Final coverage report**
While I really like the native html report from `coverage`, this strategy of providing it as zip file is particularly clunky.
If we keep this strategy:
- It might be better to target something like a MHTML file where we're just dealing with a single file to show the report. (although, Ive done zero legwork to determine feasibility...)
- Ensure a top-level directory in the zip. This avoids potentially dumping all these files in to whatever directory it's in.

freakboy3742:
Uploading a single combined report is a good thought - I hadn't considered that, but it would seem to address the biggest complaint described by #848.

I'm not sure the threshold approach fixes the problem; the issue is with the algorithm used to compute coverage. If a file has 100 lines and 95% coverage, deleting 10 lines drops coverage to ~85%. The only way to address this is to fix this is to have 100% coverage to start with. The good news is that from looking at the current coverage misses, we're actually not *that* far from getting 100% coverage. The missing cases are mostly things like "checking that the Windows filename actually has .bat", or "actually running validate_tools() on a build step" - things that are easy to add tests for, if we have a focussed effort.

Looking at the HTML coverage report - unfortunately, there aren't a whole lot of options there. `coverage html` doesn't provide any customisation options (that I'm aware of) that would provide MHTML output or include a directory. Those all sound like good suggestions, but they'd be upstream contributions to Coverage (and almost certainly substantial projects in themselves).

That said - I'm actually wondering whether the HTML report is valuable at all. The CI failure includes a summary report which details the specific lines that are missing; those lines are easy enough to find locally in your editor. The printed code context is pretty... but I'm not sure it's necessarily helpful. If we can get to 100% coverage as a baseline, then a new PR that misses coverage will only be reporting lines that are a problem in that PR.

rmartin16:
> the issue is with the algorithm used to compute coverage

Yeah; that makes sense. Although, I think this issue is moot if we need to get to 100% coverage either way.

Another feature of CodeCov I really like is the lines without coverage are called out directly in the github diff. Although, that information is just as available from the `coverage` report.

pyproject.toml
```diff
@@ -2,6 +2,26 @@
 requires = ["setuptools>=60", "setuptools_scm[toml]>=7.0"]
 build-backend = "setuptools.build_meta"
 
+[tool.coverage.run]
+parallel = true
+branch = true
+relative_files = true
+source = ["src/briefcase"]
+omit = [
+    "src/briefcase/__init__.py",
```
freakboy3742, line 11: Added this because it only contains version identification code, and it's impossible to branch check both branches; you either have setuptools_scm or you don't in the test environment; both are possible situations, so `#pragma` won't help here.

src/briefcase/__main__.py
```diff
@@ -16,7 +17,7 @@ def main():
         command = Command(logger=logger, console=console)
         options = command.parse_options(extra=extra_cmdline)
         command.check_obsolete_data_dir()
-        command.parse_config("pyproject.toml")
+        command.parse_config(Path.cwd() / "pyproject.toml")
```
freakboy3742, line 20: This makes the path absolute, which means it can be tested reliably by monkeypatching cwd.

src/briefcase/__main__.py
```diff
@@ -41,8 +42,8 @@ def main():
         with suppress(KeyboardInterrupt):
             logger.save_log_to_file(command)
 
-    sys.exit(result)
+    return result
```
freakboy3742, line 45: This makes it easier to test the result, as you can test the return value, rather than catching the `sys.exit()`

src/briefcase/commands/base.py
```diff
@@ -117,7 +117,7 @@ def __init__(
         console: Console,
         tools: ToolCache = None,
         apps: dict = None,
-        base_path: Path = Path.cwd(),
+        base_path: Path = None,
```
freakboy3742, line 120: This ensures that `Path.cwd()` is evaluated when the instance is instantiated, rather than when the code is imported, which means it's testable for the default case.

src/briefcase/commands/base.py
```diff
@@ -479,17 +482,17 @@ def app_module_path(self, app):
...
-        try:
-            if len(app_home) == 1:
-                path = Path(str(self.base_path), *app_home[0])
-            else:
-                raise BriefcaseCommandError(
-                    f"Multiple paths in sources found for application '{app.app_name}'"
-                )
-        except IndexError as e:
+
+        if len(app_home) == 0:
```
freakboy3742, line 486: This was a bug; previously assumed to be caught as an IndexError, it was being returned as a `BriefcaseCommandError` with the wrong message.

src/briefcase/commands/create.py
```diff
@@ -519,12 +519,10 @@ def install_app_code(self, app: BaseConfig):
 
         :param app: The config object for the app
         """
-
         # Remove existing app folder
         app_path = self.app_path(app)
-        if app_path.is_dir():
-            self.tools.shutil.rmtree(app_path)
-            self.tools.os.mkdir(app_path)
+        self.tools.shutil.rmtree(app_path)
```
freakboy3742, line 524: The `if is_dir` wasn't needed, as the app_path is *always* a directory, generated by the template.

src/briefcase/commands/publish.py
```diff
@@ -33,7 +33,9 @@ def publish_app(self, app: BaseConfig, channel: str, **options):
         :param app: The application to publish
         :param channel: The publication channel to use
         """
-        self.logger.info(f"TODO: Publish {app.app_name} to {channel}")
+        self.logger.info(
+            f"TODO: Publish {app.app_name} to {channel}"
+        )  # pragma: no cover
```
freakboy3742, line 38: This is effectively a `raise NotImplementedError()`, but with nicer optics if it's invoked.

src/briefcase/platforms/macOS/app.py
```diff
@@ -42,32 +38,6 @@ def entitlements_path(self, app):
...
-                self.tools.shutil.move(os.fsdecode(lib_path), os.fsdecode(tmpdir))
-                self.tools.shutil.rmtree(self.support_path(app))
-
-                self.tools.os.makedirs(Path(lib_path).parent)
-                # TODO: Py3.8 compatibility; os.fsdecode not required in Py3.9
-                self.tools.shutil.move(
-                    os.fsdecode(Path(tmpdir) / "lib"),
-                    os.fsdecode(lib_path),
-                )
-
```
freakboy3742, line 70: This was only needed to support old-style support packages; now that we're version locked on support packages, this isn't needed.

freakboy3742:
Down to just 5 uncovered branches - but it's not obvious what those 5 are even *describing*...

mhsmith:
> Down to just 5 uncovered branches - but it's not obvious what those 5 are even describing...

```
src/briefcase/console.py                      202      0     68      1    99.6%   166->170
```
This means the tests don't cover the case where the `if` expression is false and the `if` block is skipped.

```
src/briefcase/integrations/android_sdk.py     448      0    161      1    99.8%   298->264
```
As above, but in this case the `if` block is the last statement in a `with` block, so if the expression is false then the next code to execute will be the `exit` method of the context manager.

```
src/briefcase/integrations/subprocess.py      225      0     87      1    99.7%   496->490
```
As above, but in this case the `if` block is the last statement in a `while` block, so if the expression is false then the next code to execute will be the `while` expression.

```
src/briefcase/platforms/macOS/__init__.py     222      0     88      2    99.4%   107->exit, 468->exit
```
Not sure about these ones, but based on the discussion [here](https://stackoverflow.com/questions/37304078) it might have something to do with the code throwing an exception and therefore never completing the function normally.

freakboy3742:
> > Down to just 5 uncovered branches - but it's not obvious what those 5 are even describing...
> 
> ```
> src/briefcase/console.py                      202      0     68      1    99.6%   166->170
> ```
> 
> This means the tests don't cover the case where the `if` expression is false and the `if` block is skipped.

🤦 That one was obvious when I took a second look. I was missing the "prefix" requirement for that for that line to be (not) executed.
 
> ```
> src/briefcase/integrations/android_sdk.py     448      0    161      1    99.8%   298->264
> ```
> 
> As above, but in this case the `if` block is the last statement in a `with` block, so if the expression is false then the next code to execute will be the `exit` method of the context manager.

That's what I thought too... but we run the test suite on Windows, which should (and does, based on my testing) trigger that branch. Looking closer, I'm wondering if it might be a problem with the merging of the coverage data.

> ```
> src/briefcase/integrations/subprocess.py      225      0     87      1    99.7%   496->490
> ```
> 
> As above, but in this case the `if` block is the last statement in a `while` block, so if the expression is false then the next code to execute will be the `while` expression.

Ah - now I see what was tripping me up. It's a case that can't actually happen (`readline() returning None`).

> ```
> src/briefcase/platforms/macOS/__init__.py     222      0     88      2    99.4%   107->exit, 468->exit
> ```
> 
> Not sure about these ones, but based on the discussion [here](https://stackoverflow.com/questions/37304078) it might have something to do with the code throwing an exception and therefore never completing the function normally.

Hrm... that's as good a theory as any, I suppose... I'll keep poking.

src/briefcase/integrations/subprocess.py
```diff
@@ -493,7 +493,7 @@ def _stream_output_thread(self, popen_process):
                 output_line = ensure_str(popen_process.stdout.readline())
                 if output_line:
                     self.tools.logger.info(output_line)
-                elif output_line == "":
+                else:
```
freakboy3742, line 496: The old version of this left an open coverage case for `readline()` returning None. This shouldn't ever happen; if it does, it's probably safe to assume that the process has ended, so I've made this a bucket else case.

freakboy3742:
This is really weird - it looks like there's a discrepancy between 3.9 and 3.10 in how they report exiting a context manager.

Looking at the missing branch on `android_sdk.py`:
* [Python 3.9 on Windows](https://github.com/beeware/briefcase/actions/runs/3170671946/jobs/5163441395) reports coverage of that branch, only missing 299-301 (as expected)
* [Python 3.9 on macOS](https://github.com/beeware/briefcase/actions/runs/3170671946/jobs/5163441368) reports 298-304 as the missing branch
* [Python 3.9 on ubuntu](https://github.com/beeware/briefcase/actions/runs/3170671946/jobs/5163432967) reports 298-304 as the missing branch
* [Python 3.10 on ubuntu](https://github.com/beeware/briefcase/actions/runs/3170671946/jobs/5163441310) reports 298-264 as the missing branch

I'm guessing that since we're not reporting 3.10 on Windows, there's no match for the 298-264 missing branch, which is then reported as a miss in the overall results.

rmartin16:
💯

freakboy3742:
Now that we're at 100% coverage, the only question that remains is whether we stick with "self-hosted" coverage testing, or revert to CodeCov reporting.

As I see it, CodeCov is an external dependency which has been flaky from time to time, and I'm always in favour of having less external dependencies if we can. It's also very noisy, although only uploading a single CI report may mitigate that somewhat. The major advantage is that it provides an in-browser way to view coverage results.

Self-hosted coverage isn't as pretty, but it does present the specific lines in the test failure, rather than needing a couple of clicks to get there. It also matches the report you'll get if you run coverage locally. We *can* generate HTML reports, but they're not easy to browse (you have to download to view them). 

On that basis, I have a mild preference to drop CodeCov and go self-hosted, dropping the upload of the coverage HTML result. However, if anyone has a strong preference to retain CodeCov, I won't put up much of a fight.

Opinions?

mhsmith (approved):

PR #892, opened 2022-10-03
Run Python with unbuffered output via subprocess (Fixes #891)

### LINKED ISSUE #891: Output streaming doesn't work on `briefcase dev`
`print()` (and other stdout content) isn't reflected in the console when running `briefcase dev`.

**To Reproduce**
1. Run toga tutorial0 with `briefcase dev`
2. Click the button several times

There's no error, but:
* On Windows and Linux, the output isn't flushed until you quit the app
* On macOS, the output isn't visible at all.

stderr content *is* printed to the console.

**Expected behavior**

`print()` statements should be reflected in the console.

**Environment:**
 - Operating System: Linux, macOS, Windows
 - Python version: 3.10
 - Software versions:
   - Briefcase: 0.3.10
   - Toga: 0.3.0.dev38

**Additional context**

Possibly introduced by 390ccb36?
 
If you disable `stream_output=True` on the invocation that starts the dev process, console output is restored; however, you lose the contents of stdout in the log file.

This doesn't appear to impact `briefcase run`, as full apps have very different processes for managing logs.

rmartin16:
Python buffers its output if `stdout` isn't connected to a tty. This is the case when using the output streamer.

This change adds `-u` to all invocations of Python. Alternatively, we could inject `PYTHONUNBUFFERED=1` in to the environment for subprocess calls.

- Fixes #891 

## PR Checklist:

### DIFF
--- a/changes/891.bugfix.rst
+++ b/changes/891.bugfix.rst
@@ -1,0 +1,1 @@
+Output from spawned Python processes, such as when running ``briefcase dev``, is no longer buffered and displays in the console immediately.
--- a/src/briefcase/commands/create.py
+++ b/src/briefcase/commands/create.py
@@ -470,6 +470,7 @@ class CreateCommand(BaseCommand):
                     self.tools[app].app_context.run(
                         [
                             sys.executable,
+                            "-u",
                             "-m",
                             "pip",
                             "install",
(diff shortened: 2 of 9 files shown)


freakboy3742 (approved):
Awesome - thanks!

PR #898, opened 2022-10-05
Added a web backend for deploying apps as a static web server

### LINKED ISSUE #3: Add browser/web support (via brython? pypy.js? batavia? all of the above?)
It would be really awesome if briefcase could build a minified, combined JS bundle for deploying into a browser, as well.

freakboy3742:
Adds a static web deployment backend.

This uses PyScript to provide client-side Python support.

Removes the placeholder `Django` backend. The app can be build as an entirely static site.

Also requires:
1. The new [briefcase-web-static-template](https://github.com/beeware/briefcase-web-static-template)
2. The [static-web branch of briefcase-template](https://github.com/beeware/briefcase-template/pull/34)
3. The [pyscript branch of Toga](https://github.com/beeware/toga/pull/1475)

CI will fail on the web build until (2) is merged.

In the meantime, you can test this by adding:
```
[tool.briefcase.app.myapp.web]
requires = [
    "../path/to/toga/src/web"
]
style_framework = "Bootstrap v4.6"
```
to a project that is using a source checkout of Toga on the `pyscript` branch mentioned above.

Fixes #3.

## PR Checklist:


src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,347 @@
...
+            for line in f:
+                if line.rstrip("\n") == sentinel:
+                    content.append(line)
+                    break
+                else:
+                    content.append(line)
+
+        with path.open("w", encoding="utf-8") as f:
+            for line in content:
+                f.write(line)
```
rmartin16, line 80: mostly curiosity perhaps...any reason not to use `f.writelines(content)`?

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,347 @@
...
+        with a *lot* of content before the sentinel.
+
+        :param path: The path to the file to be trimmed
+        :param sentinel: The content of the sentinel line. This will
+            become the last line in the trimmed file.
+        """
+        content = []
+        with path.open("r", encoding="utf-8") as f:
+            for line in f:
+                if line.rstrip("\n") == sentinel:
```
rmartin16, line 72: Unless something really interesting happens with file iteration, I would expect `rstrip('\n')` to ever actually find a line break to strip.

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,347 @@
...
+            required=False,
+        )
+
+    def run_app(self, app: AppConfig, host, port, open_browser, **kwargs):
+        """Start the application.
+
+        :param app: The config object for the app
+        :param host: The host on which to run the server
+        :param port: The device UDID to target. If ``None``, the user will
+            be asked to select a device at runtime.
```
rmartin16, line 251: copy paste typo.

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,347 @@
...
+        )
+
+        httpd = None
+        try:
+            # Create a local HTTP server
+            httpd = LocalHTTPServer(
+                self.project_path(app),
+                host=host,
+                port=port,
+            )
```
rmartin16, line 267: curious if this was intentional design or not. it is pretty easy to get this to fail; for instance, just specify a privileged port like 80.

Should we catch these exceptions or just give them to the user?

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,347 @@
...
+                "Web server log output (type CTRL-C to stop log)...",
+                prefix=app.app_name,
+            )
+            self.logger.info("=" * 75)
+
+            # Run the server.
+            httpd.serve_forever()
+        except KeyboardInterrupt:
+            # CTRL-C is the accepted way to stop the server.
+            pass
```
rmartin16, line 283: This brings up a point I considered when I updated the output streamer to re-raise `KeyboardInterrupt`; that is, when we tell the user to press `CTRL+C`, should we also show them the `Aborted by User` message when Briefcase exits?

This suggests so; I can create a separate PR to imitate this for Android and iOS.

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,347 @@
...
+        finally:
+            if httpd:
+                with self.input.wait_bar("Shutting down server..."):
+                    httpd.server_close()
+
+            # Not sure why, but this is needed to mollify coverage for the
+            # "test_cleanup_server_error" case. Without this pass, a missing branch
+            # is reported for the "if httpd: -> exit" branch
+            pass
```
rmartin16, line 292: Alternatively, `coverage` will respect `# pragma: no branch`...not sure which is considered the better approach here.

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,347 @@
...
+            "Packaging web app for distribution...",
+            prefix=app.app_name,
+        )
+
+        with self.input.wait_bar("Building archive..."):
+            with ZipFile(
+                self.distribution_path(app, packaging_format=packaging_format), "w"
+            ) as archive:
+
+                for filename in self.project_path(app).glob("**/*"):
```
rmartin16, line 325: Good opportunity for a progress bar....but using the glob generator is obviously more efficient.

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,347 @@
...
+            ) as f:
+                f.write(f"name = '{app.formal_name}'\n")
+                f.write(f"description = '{app.description}'\n")
+                f.write(f"version = '{app.version}'\n")
+                f.write("\n")
+                f.write("autoclose_loader = true\n")
+                f.write("packages = [\n")
+                f.write(
+                    "\n".join(
+                        f'    "/{wheel.relative_to(self.project_path(app))}",'
```
rmartin16, line 176: The toml library doesn't appear to always like these double quotes. On Linux, it doesn't seem to matter; on Windows, pyscript throws an error for `pyscript.toml`.

```
config supplied: name = 'mywebapp' description = 'My first application' version = '0.0.1' autoclose_loader = true packages = [ "/static\wheels\mywebapp-0.0.1-py3-none-any.whl", "/static\wheels\toga_core-0.3.0.dev36-py3-none-any.whl", "/static\wheels\toga_dummy-0.3.0.dev36-py3-none-any.whl", "/static\wheels\toga_web-0.3.0.dev36-py3-none-any.whl", "/static\wheels\travertino-0.1.3-py3-none-any.whl", ] is an invalid TOML and cannot be parsed: SyntaxError: w 7 | "/static\wheels\mywebapp-0.0.1-py3-none-any.whl", ^
```

rmartin16, line 176: Additionally, looks like pyscript et al are expecting the Unix path separator.

From browser console:
```
[pyscript/pyodide] micropip install /static\wheels\mywebapp-0.0.1-py3-none-any.whl,/static/wheels/toga_core-0.3.0.dev36-py3-none-any.whl,/static/wheels/toga_dummy-0.3.0.dev36-py3-none-any.whl,/static/wheels/toga_web-0.3.0.dev36-py3-none-any.whl,/static/wheels/travertino-0.1.3-py3-none-any.whl [logger.ts:39:19](https://pyscript.net/src/logger.ts)
Uncaught (in promise) PythonError: Traceback (most recent call last):
  File "/lib/python3.10/asyncio/futures.py", line 201, in result
    raise self._exception
  File "/lib/python3.10/asyncio/tasks.py", line 234, in __step
    result = coro.throw(exc)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 548, in install
    await transaction.gather_requirements(requirements)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 305, in gather_requirements
    await gather(*requirement_promises)
  File "/lib/python3.10/asyncio/futures.py", line 284, in __await__
    yield self  # This tells Task to wait for completion.
  File "/lib/python3.10/asyncio/tasks.py", line 304, in __wakeup
    future.result()
  File "/lib/python3.10/asyncio/futures.py", line 201, in result
    raise self._exception
  File "/lib/python3.10/asyncio/tasks.py", line 232, in __step
    result = coro.send(None)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 315, in add_requirement
    wheel = WheelInfo.from_url(req)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 75, in from_url
    name, version, build, tags = parse_wheel_filename(file_name)
  File "/lib/python3.10/site-packages/packaging/utils.py", line 100, in parse_wheel_filename
    raise InvalidWheelFilename(f"Invalid project name: {filename}")
packaging.utils.InvalidWheelFilename: Invalid project name: static\wheels\mywebapp-0.0.1-py3-none-any
    PythonError https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.asm.js:10
    new_error https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.asm.js:10
    method_call_trampoline https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.asm.js:10
    callPyObjectKwargs https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.asm.js:10
    callPyObject https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.asm.js:10
    wrapper https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.asm.js:10
    setTimeout handler*hiwire_call_bound https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.asm.js:10
    callPyObjectKwargs https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.asm.js:10
    callPyObject https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.asm.js:10
    wrapper https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.asm.js:10
    setTimeout handler*hiwire_call_bound https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.asm.js:10
    ```

**Review by rmartin16 — CHANGES_REQUESTED** [review]
This is awesome. I've only spun pyscript up before just to see it working but being able to run toga makes it a lot more attractive.

A few thoughts:
- Should the docs be more forthcoming that this is still early days for all this?
  - Mostly thinking of the case where someone imports a package that uses something in the stdlib that stubbed for wasm.
  - Maybe linking [here](https://github.com/python/cpython/tree/v3.11.0b1/Tools/wasm#wasm32-emscripten-limitations-and-issues) or similar.
- Python's simple HTTP server and Chrome on Windows
  - CTLC+C isn't always working for me....Chrome must be holding open a connection and the web server must be blocking on it. When I close the tab and reopen it, Chome seemingly resets the connection and the web server is unblocked and sees the interrupt and exits.
  - Firefox seems to play much nicer with this web server.
  - Running the web server like macOS's log steamer would probably avoid this but is a lot more overhead.
- Toga Tutorials 0 and 1
  - I can't get these to actually render anything. (I've verified i'm using your `pyscript` toga branch.)
  - Not sure what's missing but I'll keep messing with it.

**Comment by freakboy3742** [comment]
> This is awesome. I've only spun pyscript up before just to see it working but being able to run toga makes it a lot more attractive.
> 
> A few thoughts:
> 
> * Should the docs be more forthcoming that this is still early days for all this?
>   * Mostly thinking of the case where someone imports a package that uses something in the stdlib that stubbed for wasm.
>   * Maybe linking [here](https://github.com/python/cpython/tree/v3.11.0b1/Tools/wasm#wasm32-emscripten-limitations-and-issues) or similar.

At the very least, it worth mentioning that there are limitations on the binary modules that are available, and some runtime complications. I'll add a note.

> * Python's simple HTTP server and Chrome on Windows
>   
>   * CTLC+C isn't always working for me....Chrome must be holding open a connection and the web server must be blocking on it. When I close the tab and reopen it, Chome seemingly resets the connection and the web server is unblocked and sees the interrupt and exits.
>   * Firefox seems to play much nicer with this web server.
>   * Running the web server like macOS's log streamer would probably avoid this but is a lot more overhead.

Hrm... That's weird. I'm not seeing any problems like that with Chrome on macOS. 

As taking a different approach - the comparison with the macOS log streamer isn't really fair, because a macOS streams it's logs to a separate process. In this case, the logs are the stderr output from invoking `httpd.serve_forever()`. Unless you're suggesting spawning a process to run the web server, and then waiting for that web server process to terminate? Which... I guess it could work, but it seems... overkill.

> * Toga Tutorials 0 and 1
>   
>   * I can't get these to actually render anything. (I've verified i'm using your `pyscript` toga branch.)
>   * Not sure what's missing but I'll keep messing with it.

Yeah - I'm aware of those problems. I'm still tinkering with the Toga branch, and I don't think I've pushed all the modifications I've made to date. There's been some changes in PyScript since I started that branch, and I'm trying to get on top of what has changed. I'm hoping to get those issues sorted out today.

**Comment by rmartin16** [comment]
> Hrm... That's weird. I'm not seeing any problems like that with Chrome on macOS.

I'm pretty sure I was only seeing this when pyscript entered an error state; such as when wheel installation was failing because it didn't like the names. Normally, the problem doesn't occur.

> unless you're suggesting spawning a process to run the web server, and then waiting for that web server process to terminate? Which... I guess it could work, but it seems... overkill.

That's what I am talking about...but agreed; this should be a last resort or something.

**Comment by freakboy3742** [comment]
I've now pushed updates to the pyscript branch.

I can confirm I'm seeing the log lockup issue on Windows - no idea what is causing that.

There's one other edge case that I've observed in testing - if you declare the mainline app as `MyApp()` (and rely on package metadata), rather than `MyApp("Formal name", "org.beeware.bundle")`, there's a Heisenbug at app startup - sometimes it works fine, sometimes it doesn't (and it complains about the app needing a formal name). I'm still digging into this one.

**Comment by freakboy3742** [comment]
@rmartin16 I think I've resolved the CTRL-C issue - it appears that on Windows, an explicit server shutdown is required, in addition to the close. It doesn't hurt to do the same on macOS, so we might as well.

File: src/briefcase/platforms/web/static.py (hunk 1 above)
Comment (freakboy3742) [reply src/briefcase/platforms/web/static.py:80]: Didn't know it existed :-)

File: src/briefcase/platforms/web/static.py (hunk 2 above)
Comment (freakboy3742) [reply src/briefcase/platforms/web/static.py:72]: `for line in f` includes the `\n` on the line being read, so it should match on *every* line.

File: src/briefcase/platforms/web/static.py [hunk 9]
```diff
@@ -0,0 +1,347 @@
...
+        self.logger.info("Compile static web content from wheels...")
+        with self.input.wait_bar("Compiling static web content from wheels..."):
+            # Trim previously compiled content out of briefcase.css
+            briefcase_css_path = (
+                self.project_path(app) / "static" / "css" / "briefcase.css"
+            )
+            self._trim_file(
+                briefcase_css_path,
+                sentinel=" ******************* Wheel contributed styles **********************/",
+            )
```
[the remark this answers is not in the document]
freakboy3742: Good catch - I've updated the template to match.

src/briefcase/platforms/web/static.py
freakboy3742: It was only intentional in the sense of "don't try and be too smart about it". I've added explicit handling and an error message for the permission error raised by low ports.

src/briefcase/platforms/web/static.py
freakboy3742: Agreed - we should only log a message when something is unexpected. In this (and the Android/iOS case), Ctrl-C is normal shutdown behavior.

src/briefcase/platforms/web/static.py
freakboy3742: Much of a muchness, really. Both approaches work  :-)

src/briefcase/platforms/web/static.py
freakboy3742: Good point about possibly using a progress bar; however, my experience is that this archiving is sufficiently fast that progress won't add *that* much. Given the complexity of needing to pre-process the glob, I'm inclined to keep this as a waitbar for now.

src/briefcase/platforms/web/static.py
freakboy3742: I've reversed the quoting so that we're using `"` consistently in output.

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,347 @@
...
+            # Extract static resources from packaged wheels
+            for wheelfile in self.wheel_path(app).glob("*.whl"):
+                self.logger.info(f"Processing {wheelfile.name}...")
+                with briefcase_css_path.open("a", encoding="utf-8") as css_file:
+                    self._process_wheel(wheelfile, css_file=css_file)
+
+        return {}
+
+
+class HTTPHandler(SimpleHTTPRequestHandler):
```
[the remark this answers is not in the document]
freakboy3742: Not a bad idea; although it might be worth treating that as a separate feature request if it turns out to be needed.

src/briefcase/platforms/web/static.py
freakboy3742: ack - fixed.

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,352 @@
...
+                prefix=app.app_name,
+            )
+            self.logger.info("=" * 75)
+
+            # Run the server.
+            httpd.serve_forever()
+        except PermissionError:
+            raise BriefcaseCommandError(
+                "Unable to start web server. Are you sure you specified a valid host and port?"
+            )
```
rmartin16, line 285: Without the `from e`, we don't get the inner exception in the log.

src/briefcase/platforms/web/static.py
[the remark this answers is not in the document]
rmartin16: As a demonstration, attempting to use the web server while switching between two apps in the same project fails because the browser uses the cached `pyscript.toml` and tries to pull the wheel for the other app.

Of course, holding SHIFT while clicking refresh is enough to force the browser to flush the cache.

Given that, this is only mildly convincing to me we should introduce complications to address this now...especially with a trivial (and normal web dev) workaround. Example for posterity, nonetheless.

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,352 @@
...
+                        self.bundle_path(app) / "requirements.txt",
+                    ],
+                    check=True,
+                )
+            except subprocess.CalledProcessError as e:
+                raise BriefcaseCommandError(
+                    f"Unable to install dependencies for app {app.app_name!r}"
+                ) from e
+
+        with self.input.wait_bar("Writing Pyscript configuration file..."):
```
rmartin16, line 164: disclosure: i consider this at least borderline nitpicking.

This section of logging is difficult for me to easily parse:
```
Writing Pyscript configuration file... done
Compile static web content from wheels...
Processing mywebapp-0.0.1-py3-none-any.whl...
Processing toga_core-0.3.0.dev39-py3-none-any.whl...
Processing toga_dummy-0.3.0.dev39-py3-none-any.whl...
Processing toga_web-0.3.0.dev39-py3-none-any.whl...
Found toga_web/static/toga.css
Processing travertino-0.1.3-py3-none-any.whl...
Compiling static web content from wheels... done
```
1) There are three important things happening here but it's hard to separate them out.
2) There are a lot of ellipsis....but only some say `done` after them.

Two possible improvements.

1) Consider the output from `pip` that directly proceeds this:
```
Processing /home/user/github/freakboy3742/toga/src/core
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Processing /home/user/github/freakboy3742/toga/src/dummy
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Processing /home/user/github/freakboy3742/toga/src/web
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Collecting travertino>=0.1.3
  Using cached travertino-0.1.3-py3-none-any.whl (15 kB)
Saved ./web/static/mywebapp/www/static/wheels/travertino-0.1.3-py3-none-any.whl
Building wheels for collected packages: toga-core, toga-dummy, toga-web
  Building wheel for toga-core (setup.py): started
  Building wheel for toga-core (setup.py): finished with status 'done'
...
```
This use of the indent makes it dramatically easier to quickly skim the output. Using an indent for the `Processing <wheel>` output really improves readability IMO:
```
Writing Pyscript configuration file... done
Compile static web content from wheels...
  Processing mywebapp-0.0.1-py3-none-any.whl...
  Processing toga_core-0.3.0.dev39-py3-none-any.whl...
  Processing toga_dummy-0.3.0.dev39-py3-none-any.whl...
  Processing toga_web-0.3.0.dev39-py3-none-any.whl...
Found toga_web/static/toga.css
  Processing travertino-0.1.3-py3-none-any.whl...
Compiling static web content from wheels... done
```
I guess if we really wanted to, we could add an `indent` argument to `Log` to manage this...

2) As for all the ellipsis...

I've been wondering for a while if ellipsis should be exclusively reserved for the Wait Bar. If the indent is added, then I definitely think it makes sense to at least avoid the ellipsis for the `Compile static web content from wheels` log.

As for the ellipsis for the `Processing <wheel>` log, I was initially thinking they should be Wait bars....but that would mean the Wait bar would be changing very quickly and that isn't great. So, probably best to get rid of the ellipsis there too and just let the single Wait Bar convey that tasks are ongoing.

This is where this would ultimately end up:
```
[mywebapp] Building web project...
Removing old wheels... done
Repacking wheel as /home/user/tmp/beeware/mywebapp/web/static/mywebapp/www/static/wheels/mywebapp-0.0.1-py3-none-any.whl...OK
Building app wheel... done
Processing /home/user/github/freakboy3742/toga/src/core
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Processing /home/user/github/freakboy3742/toga/src/dummy
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Processing /home/user/github/freakboy3742/toga/src/web
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Collecting travertino>=0.1.3
  Using cached travertino-0.1.3-py3-none-any.whl (15 kB)
Saved ./web/static/mywebapp/www/static/wheels/travertino-0.1.3-py3-none-any.whl
Building wheels for collected packages: toga-core, toga-dummy, toga-web
  Building wheel for toga-core (setup.py): started
  Building wheel for toga-core (setup.py): finished with status 'done'
  Created wheel for toga-core: filename=toga_core-0.3.0.dev39-py3-none-any.whl size=496221 sha256=7d63aaeae0a49ac0b40b5ee57541b8d4f3df935c3e8c5832dae548df32cd7cae
  Stored in directory: /tmp/pip-ephem-wheel-cache-ox82hom4/wheels/d6/c5/94/ed3559e7d1e227a880e8f68277a009ba9faa83254cdcb5fbce
  Building wheel for toga-dummy (setup.py): started
  Building wheel for toga-dummy (setup.py): finished with status 'done'
  Created wheel for toga-dummy: filename=toga_dummy-0.3.0.dev39-py3-none-any.whl size=27249 sha256=e28701baadd3879cc8530082f8be391c54982dbae3584db35257960c7484ae65
  Stored in directory: /tmp/pip-ephem-wheel-cache-ox82hom4/wheels/c0/a2/58/b2501a1be28389106869c459b744fec9e6f5631ada886e2e74
  Building wheel for toga-web (setup.py): started
  Building wheel for toga-web (setup.py): finished with status 'done'
  Created wheel for toga-web: filename=toga_web-0.3.0.dev39-py3-none-any.whl size=13068 sha256=d369f6eac3c796b1579e223f65f5c0efc17c332aaf868f1071281c979b580f53
  Stored in directory: /tmp/pip-ephem-wheel-cache-ox82hom4/wheels/19/e5/bf/617886587698943f953cf16ea7e3623985fedffbc8b187e98c
Successfully built toga-core toga-dummy toga-web
Installing wheels for dependencies... done
Writing Pyscript configuration file... done
Compile static web content from wheels
  Processing mywebapp-0.0.1-py3-none-any.whl
  Processing toga_core-0.3.0.dev39-py3-none-any.whl
  Processing toga_dummy-0.3.0.dev39-py3-none-any.whl
  Processing toga_web-0.3.0.dev39-py3-none-any.whl
  Found toga_web/static/toga.css
  Processing travertino-0.1.3-py3-none-any.whl
Compiling static web content from wheels... done
```

Note: `pip` has a whole infrastructure to manage the current level of indent and incrementing and decrementing it. I'd say limiting ourselves to a single indent level is reasonable.

rmartin16 (changes requested):
> I think I've resolved the CTRL-C issue - it appears that on Windows, an explicit server shutdown is required, in addition to the close. It doesn't hurt to do the same on macOS, so we might as well.

I can still create the situation where the web server doesn't respond to CTRL+C.

1. Add `1/0` to `app.py` to create an exception.
2. Start the web server and load the app in chrome on windows.
3. CTRL+C (most likely) will not work.

This can recreate the situation probably 9 in 10 times for me (refreshing twice before a CTRL-C may help). Refreshing the page in chrome allows the web server to receive the interrupt and exit.

I cannot replicate this with Firefox but I can with Edge (although its chromium now so not surprised). I cannot replicate this on Linux at all; this perhaps suggests this is more a function of Python's interaction with the OS via `select` on Windows. The stack below seems to reinforce this.

<details>
<summary>Python stack when stuck on Windows</summary>

```
0x0000000000000000
ntdll.dll!NtWaitForSingleObject+0x14
mswsock.dll+0x80fc
mswsock.dll!Tcpip4_WSHAddressToString+0xa8e
WS2_32.dll!select+0x137
select.pyd!PyInit_select+0x125e
select.pyd!PyInit_select+0x10ab
python310.dll!PyEval_EvalFrameDefault+0x346b
python310.dll!PyEval_EvalFrameDefault+0xf27
python310.dll!PyObject_Hash+0xd25
python310.dll!PyVectorcall_Call+0xb8
python310.dll!PyObject_Call+0x14f
python310.dll!PyEval_EvalFrameDefault+0x5efe
python310.dll!PyFunction_Vectorcall+0x87
python310.dll!PyObject_FastCallDictTstate+0xd3
python310.dll!PyObject_Call_Prepend+0x7f
python310.dll!PyWeakref_NewProxy+0x1e14
python310.dll!PyObject_Call+0x1cb
python310.dll!PyEval_EvalFrameDefault+0x5efe
python310.dll!PyFunction_Vectorcall+0x87
python310.dll!PyEval_EvalFrameDefault+0x5f4
python310.dll!PyMapping_Check+0x1d5
python310.dll!PyEval_EvalCode+0x82
python310.dll!PyArena_Free+0x3f3
python310.dll!PyArena_Free+0x2f3
python310.dll!PyEval_EvalFrameDefault+0x1875
python310.dll!PyFunction_Vectorcall+0x87
python310.dll!PyEval_EvalFrameDefault+0x5f4
python310.dll!PyFunction_Vectorcall+0x87
python310.dll!PyVectorcall_Call+0x5c
python310.dll!PyObject_Call+0x4f
python310.dll!PyMem_SetupAllocators+0x1aa
python310.dll!Py_RunMain+0x13d
python310.dll!Py_RunMain+0x15
python310.dll!Py_Main+0x25
python.exe+0x1230
KERNEL32.DLL!BaseThreadInitThunk+0x14
ntdll.dll!RtlUserThreadStart+0x21
```

</details>

Poking around google, this definitely doesn't look like an unknown problem. It seems to boil down to `sockets` inherently blocking in python. The underlying abstractions support turning blocking off. 

When I added `self.socket.setblocking(False)` to `LocalHTTPServer.__init__()`, the server would always respond to CTRL+C...as well as working normally.

Admittedly, I haven't quite sussed out the implications of turning blocking off, though, in this use case. The suggestion seems to be that exceptions may be raised....but I haven't seen this happen.

Do you have any insight here?

[edit]
After further testing, I'm seeing a lot of network errors in the browser console suggesting that the server is abandoning the connection before data is sent or the connection is accepted. idk

src/briefcase/platforms/web/static.py
freakboy3742: Completely agreed that the "wall of text" from that stage isn't great; I went through a couple of iterations of waitbar use before I settled where I did, and I still wasn't happy with the result.

I think the indentation and removing the ellipsis is a good improvement, though; I'll add those in.

More generally, I see 3 generic problems with the output as rendered in the build command:

Firstly, there's no vertical breaks. There's a lot being done, and being done quickly, but no vertical breaks to indicate "phases" in the build. We've been leaning away from empty calls to `info()`; but the stages are sufficiently short lived (and minor) that they don't really warrant a 'prefix' block. Not sure if there's a good solution here other than those two options.

Secondly, the need to have a "prefix" statement when using a waitbar that produces output. This is easy enough to do with a call to "info", but the "grammar" of those statements isn't clear. I broadly agree with your comment about keeping ellipsis exclusive to waitbars - an ellipsis should always be "closed" with "done" - but if you do that, the grammar of the "What I'm about to do" statement doesn't flow well to me. In this example, we need something to explain what "processing" means, but "Compile static web content" and "Compiling static web content" both feel like the wrong tense. "Compile" might make more sense if it was a prefixed marker, but as currently rendered, it feels like the best of a bad bunch of options, rather than a good option in itself.

Thirdly, there's the indentation issue. I agree indentation is a nice touch; although I do wonder if using indentation here is a crutch for getting over the first two issues. If we're going to start adopting indentation, then I agree having some sort of baked-in API would be a good idea - but the fact that we've been able to survive this long without indentation makes me wonder whether coming up with a good fixes for the first two might be the better (and easier) approach.

freakboy3742:
> > I think I've resolved the CTRL-C issue - it appears that on Windows, an explicit server shutdown is required, in addition to the close. It doesn't hurt to do the same on macOS, so we might as well.
> 
> I can still create the situation where the web server doesn't respond to CTRL+C.

Ok - that's reproducible for me as well.

The good news - I think I've found a fix. I found an old Python bug https://bugs.python.org/issue31639 that made reference to using ThreadingHTTPServer for `http.server` for the same reason we're seeing here. The threaded server gives better server performance anyway, so it seems like a good move regardless - and at least in my testing, it addresses the Windows locking bug.

freakboy3742:
@rhmartin Also - if I could get a review of beeware/briefcase-template#34, the CI failures on the app tests should be resolved.

rmartin16:
> if I could get a review of https://github.com/beeware/briefcase-template/pull/34, the CI failures on the app tests should be resolved.

Oh, good point; I'll review and merge that PR.

Also, did you have any feedback on the additional [http server exceptions](https://github.com/beeware/briefcase/pull/898#discussion_r989054802) or the [toml thoughts](https://github.com/beeware/briefcase/pull/898#discussion_r989114414)?

freakboy3742:
> Also, did you have any feedback on the additional [http server exceptions](https://github.com/beeware/briefcase/pull/898#discussion_r989054802) or the [toml thoughts](https://github.com/beeware/briefcase/pull/898#discussion_r989114414)?

I swear Github is hiding things from me... I completely missed those. I'll address them now.

src/briefcase/platforms/web/static.py
freakboy3742: Ok - I think we might need to move to a proper TOML serialiser rather than writing TOML as strings. Update incoming.

src/briefcase/platforms/web/static.py
freakboy3742: > I don't actually know what this one does... 🙃

Oh no... you've opened a connection to the NETHERWORLD!!! :-)

I've added more common error cases.

rmartin16: Another path to the netherworld.... 👻 

```
❯ briefcase run web -p 0

[webtemplate] Starting web server on http://localhost:0/

[webtemplate] Web server log output (type CTRL-C to stop log)...
```

Oh....interesting....this actually asks the OS to use a random available port....cool.

src/briefcase/platforms/web/static.py
rmartin16: I created #905 to reflect the need to ensure `pyproject.toml` is a proper TOML document.

Of note, while using `toml_w` here is nice....it only supports writing an entire TOML document; so, it wouldn't be useful for sending strings in to `cookiecutter` to create `pyproject.toml`.

src/briefcase/platforms/web/static.py
rmartin16: Yeah...this is quite reminiscent of my ruminations in #755. I'm hesitant to go too much back to the drawing board given I didn't really find good answers then.

With that, I went through the output of the `package` command for all the platforms and formats before the apps were created so I could get a good sense of Briefcase output. I couldn't really identify anywhere that a lot of information generated by Briefcase is presented to the user. This may be the first case where particularly relevant information is being shown to the user and quickly. So, this may be the first use-case for something like indentation to increase readability.

In several of the outputs from tools, indentation use was certainly popular (especially with google).

I think indentation is a reasonable addition to logging. It would also be fairly trivial to add an "indent level" param to the APIs; this approach would also inherently limit how broadly this could be used since you wouldn't be able to call arbitrary functions that are "indent-unaware" (for lack of a better term).

setup.cfg
```diff
@@ -91,19 +94,16 @@ console_scripts =
...
     # tvOS = briefcase.platforms.tvOS
     # watchOS = briefcase.platforms.watchOS
     # wearos = briefcase.platforms.wearos
+    web = briefcase.platforms.web
     windows = briefcase.platforms.windows
 briefcase.formats.android =
     gradle = briefcase.platforms.android.gradle
-# briefcase.formats.django =
-#     project = briefcase.platforms.django.project
-#     app = briefcase.platforms.django.app
```
rmartin16, line 106: Throwing this out there while we're here...

Do you want to remove `homebrew` as well for now?

rmartin16 (approved):

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,371 @@
...
+
+            # If requested, open a brower tab on the newly opened server.
+            if open_browser:
+                webbrowser.open_new_tab(f"http://{host}:{port}")
+
+            self.logger.info(
+                "Web server log output (type CTRL-C to stop log)...",
+                prefix=app.app_name,
+            )
+            self.logger.info("=" * 75)
```
rmartin16, line 277: In lieu of disabling browser caching, I was thinking it might be nice to add a line like this here for folks that aren't perhaps steeped in web dev:

```
Tip: holding SHIFT while clicking refresh will bypass browser caching
```

setup.cfg
freakboy3742: Oddly, the homebrew case is slightly different. The Django backend never really existed; there was a code module, but it wasn't loaded. The homebrew *does* exit, and is actually being used by tests. It probably shouldn't be, as it's a bit of a red herring... but it's a little difficult to extract.

src/briefcase/platforms/web/static.py
freakboy3742: Ah - that one is legal - and points out that we shouldn't be just reflecting the port from the input arguments, we should be using `getsockname()`.

src/briefcase/platforms/web/static.py
freakboy3742: If it's going to cause enough problems to include that message, it's worth including the disable-cache headers. I'll add them.

setup.cfg
freakboy3742: Ok - you've successfully nerdsniped me :-)

src/briefcase/platforms/web/static.py
```diff
@@ -0,0 +1,380 @@
...
+    folder."""
+
+    def translate_path(self, path):
+        return str(self.server.base_path / path[1:])
+
+    def end_headers(self):
+        self.send_header("Cache-Control", "no-cache, no-store, must-revalidate")
+        self.send_header("Pragma", "no-cache")
+        self.send_header("Expires", "0")
+        super().end_headers(self)
```
rmartin16, line 214: `self` should be implicit here.

freakboy3742: Copy/paste fail.

src/briefcase/platforms/web/static.py
rmartin16: :) It just kept coming to mind because I was constantly running in to unexpected caching. even if i shutdown the server, made code changes, and start everything back up again.

PR #911, opened 2022-10-08
Save log to `logs` directory and only when log is needed (Fixes #883)

### LINKED ISSUE #883: Write logs to a subdirectory
As discussed in #760, we currently put log files directly in the top-level project folder. However, after a few failed runs this starts to get cluttered and obscure the view of the other files. So the user will probably delete the logs to get them out of the way, potentially losing useful information.

How about writing the logs to a "logs" subdirectory of the project folder? Since this is a directory the user won't constantly be navigating through, they'll be under no pressure to delete the logs prematurely.

This directory should also be added to the default .gitignore file. `briefcase.*.log` is currently in the *top-level* .gitignore file in https://github.com/beeware/briefcase-template, but not in the .gitignore file within the template itself, which I guess is a mistake.

rmartin16:
The issue is marked `first-timers-only` but I think we all really want this QoL.

## Changes
- Write log files to the `logs` subdirectory of the Briefcase project
  - Considered calling the directory `briefcase_logs` to avoid any conflicts with user directories...but ultimately that didn't seem particularly necessary.
- Only write log files when useful
  - First...if a user specifies `--log`, then a log file will be written (as long as a `Command` is also created).
  - I baked in a flag to `BriefcaseError` to control whether a log file is written.
    - The flag is not _currently_ exposed to devs wanting to control whether a specific error should write a file.
    - That's a relatively trivial change to do that though for an individual error....didn't seem necessary now though.
  - I disabled log files for these exceptions:
    - `HelpText`
    - `NoCommandError`
    - `ShowOutputFormats`
    - `InvalidFormatError`
    - `UnsupportedCommandError`
    - `BriefcaseConfigError`
  - Notably, all but `BriefcaseConfigError` already wouldn't create a log file because of the requirement that  a`Command` object is created and one won't be if those exceptions are thrown. I explicitly set the flag for completeness, though.
  - **Let me know if you think other situations should omit a log file.**

## Fixes
 - #883 

## PR Checklist:

### DIFF
--- a/changes/883.feature.rst
+++ b/changes/883.feature.rst
@@ -1,0 +1,1 @@
+Briefcase log files are now stored in the ``logs`` subdirectory and only when the current directory is a Briefcase project.
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -619,7 +619,9 @@ a custom location for Briefcase's tools.
 
         except FileNotFoundError as e:
             raise BriefcaseConfigError(
-                f"""Configuration file not found. Did you run briefcase in the directory that contains {filename!r}?"""
+                f"""Configuration file not found.
+
+Did you run Briefcase in a project directory that contains {filename.name!r}?"""
             ) from e
 
     def update_cookiecutter_cache(self, template: str, branch="master"):
@@ -696,7 +698,7 @@ a custom location for Briefcase's tools.
         return cached_template
 
     def generate_template(self, template, branch, output_path, extra_context):
-        """Ensure the named template is up to date for the given branch, and
+        """Ensure the named template is up-to-date for the given branch, and
         roll out that template.
 
         :param template: The template URL or path to generate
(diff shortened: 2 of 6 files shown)


freakboy3742 (approved):
One minor tweak around attribute handling - but otherwise, looks great 👍 . From my pass through the errors, I can't see any other obvious candidates that should be omitted from generating logs.

Regarding the "First timers only" flag - TBH, I was on the border of implementing this myself :-) 

FWIW, I consider that flag to be general guidance about a change whose level of difficulty is low, rather than a hard requirement preventing anyone else from contributing.

mhsmith:
> FWIW, I consider that flag to be general guidance about a change whose level of difficulty is low, rather than a hard requirement preventing anyone else from contributing.

The word "only" is pretty clear, so maybe we should change it to something like "good first issue"?

mhsmith:
@rmartin16: Thanks for doing this. As mentioned in #883, the .gitignore files in briefcase-template should be updated to match: do you want to do that as well?

rmartin16:
> @rmartin16: Thanks for doing this. As mentioned in #883, the .gitignore files in briefcase-template should be updated to match: do you want to do that as well?

Ah yeah; thanks for the reminder. I'll do that.

PR #914, opened 2022-10-11
Print Android Emulator Output for CTRL+C During Startup

rmartin16:
Some of the comments in #903 are demonstrating that potentially useful emulator output is not being printed if the emulator seemingly goes rogue and doesn't terminate....or at least doesn't work as expected.

This stems from the fact that calling `communicate()` on the emulator process with a `timeout` is all or nothing. So, emulator output is only printed if the emulator process actually terminated.

These changes allow (as much as reasonable) output to be read from the emulator regardless of the state of the process and shows it to the user and includes it in the log.

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -3,7 +3,6 @@ import re
 import shutil
 import subprocess
 import time
-from contextlib import suppress
 from pathlib import Path
 
 from briefcase.config import PEP508_NAME_RE
@@ -1069,11 +1068,16 @@ find this page helpful in diagnosing emulator problems.
 
                         # Try again in 2 seconds...
                         self.sleep(2)
-        except Exception:
-            # if the emulator exited, this should return its output immediately;
-            # if it is still running, this will quickly time out and print nothing.
-            with suppress(subprocess.TimeoutExpired):
-                self.tools.logger.info(emulator_popen.communicate(timeout=1)[0])
+        except BaseException:
+            # if the emulator exited or the user sent CTRL+C, show the
+            # available output from the emulator to the user before exiting.
+            output_deadline = time.time() + 1
+            self.tools.subprocess.stream_output(
+                label="Android emulator",
+                popen_process=emulator_popen,
+                stop_func=lambda: time.time() > output_deadline,
+                skip_cleanup=True,
+            )
             raise
 
         # Return the device ID and full name.
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -436,7 +436,13 @@ class Subprocess:
             [str(arg) for arg in args], **self.final_kwargs(**kwargs)
         )
 
-    def stream_output(self, label, popen_process, stop_func=lambda: False):
+    def stream_output(
+        self,
+        label,
+        popen_process,
+        stop_func=lambda: False,
+        skip_cleanup=False,
+    ):
         """Stream the output of a Popen process until the process exits. If the
         user sends CTRL+C, the process will be terminated.
 
@@ -449,11 +455,17 @@ class Subprocess:
         :param popen_process: a running Popen process with output to print
         :param stop_func: a Callable that returns True when output streaming
             should stop and the popen_process should be terminated.
+        :param skip_cleanup: True to skip terminating the process. Defaults False.
+            This leaves the process running in the background and instructs the
+            output streamer to stop printing its output as soon as it can. Due
+            to the implementation of readline(), the output streamer thread may
+            persist until Python exits.
         """
+        output_streamer_stop_flag = threading.Event()
         output_streamer = threading.Thread(
             name=f"{label} output streamer",
             target=self._stream_output_thread,
-            args=(popen_process,),
+            args=(popen_process, output_streamer_stop_flag),
             daemon=True,
         )
         try:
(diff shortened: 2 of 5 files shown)


src/briefcase/integrations/android_sdk.py
```diff
@@ -1069,11 +1068,16 @@ def start_emulator(self, avd):
...
                         self.sleep(2)
-        except Exception:
-            # if the emulator exited, this should return its output immediately;
-            # if it is still running, this will quickly time out and print nothing.
-            with suppress(subprocess.TimeoutExpired):
-                self.tools.logger.info(emulator_popen.communicate(timeout=1)[0])
+        except BaseException:
+            # if the emulator exited or the user sent CTRL+C, show the
+            # available output from the emulator to the user before exiting.
+            output_deadline = time.time() + 1
```
rmartin16, line 1074: Given a lot more is being done now....this seems like a potentially low deadline...especially on slow machines.

src/briefcase/integrations/subprocess.py
```diff
@@ -469,19 +481,23 @@ def stream_output(self, label, popen_process, stop_func=lambda: False):
...
+                output_streamer_stop_flag.set()
+            else:
+                self.cleanup(label, popen_process)
+                streamer_deadline = time.time() + 3
+                while output_streamer.is_alive() and time.time() < streamer_deadline:
+                    time.sleep(0.1)
+                if output_streamer.is_alive():
+                    self.tools.logger.error(
+                        "Log stream hasn't terminated; log output may be corrupted."
+                    )
```
rmartin16, line 494: With the introduction of a "stop flag" for the output streamer, I think it begs the question if it should be used as well as an alternative to printing "log stream hasn't terminated".

Given we don't have intended use-cases for process termination to fail, I'm leaning towards not setting the stop flag to get as much output as possible from the process.

rmartin16:
@freakboy3742, @mhsmith:

Thoughts?

This adds yet more complication to output streaming but there really doesn't seem to be an easier way to capture output for a running process.

freakboy3742:
> @freakboy3742, @mhsmith:
> 
> Thoughts?
> 
> This adds yet more complication to output streaming but there really doesn't seem to be an easier way to capture output for a running process.

I think I understand the approach you're taking here; the extra complication doesn't fill me with excitement though.

An alternative approach I can think of is to use `communicate` [with the fallback described in the Python docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate). Increase the initial timeout (3s, maybe?); if that timeout occurs, catch the error, warn the user, and try again *without* a timeout. That would:
1. Increase the likelihood that the process will terminate without impeding the "normal" fast shutdown case 
2. Provide an indication to the user (and in the logs for our benefit) if there's a slow shutdown causing potential problems
3. Will wait a *long* time for an actual shutdown.   

As long as the user can CTRL-C in the final shutdown state (and is prompted that they might need to do so, and that CTRL-C is logged for our benefit), that should capture all the shutdown logs AFAICT - at least, as much as is possible if the process is behaving badly.

> [edit] branch coverage is a little wild...[line 1075]

FWIW: That one I can explain from experience - macOS had a similar weird report.

The issue is the lambda. You're defining a lambda on L1078, but it's never exercised. It reports as L1075 because functionally that's the line of code in effect when the lambda is executed; the exit is the exit of the lambda.

rmartin16:
> > @freakboy3742, @mhsmith:
> > Thoughts?
> > This adds yet more complication to output streaming but there really doesn't seem to be an easier way to capture output for a running process.
> 
> I think I understand the approach you're taking here; the extra complication doesn't fill me with excitement though.
> 
> An alternative approach I can think of is to use `communicate` [with the fallback described in the Python docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate). Increase the initial timeout (3s, maybe?); if that timeout occurs, catch the error, warn the user, and try again _without_ a timeout. That would:
> 
>     1. Increase the likelihood that the process will terminate without impeding the "normal" fast shutdown case
> 
>     2. Provide an indication to the user (and in the logs for our benefit) if there's a slow shutdown causing potential problems
> 
>     3. Will wait a _long_ time for an actual shutdown.
> 
> 
> As long as the user can CTRL-C in the final shutdown state (and is prompted that they might need to do so, and that CTRL-C is logged for our benefit), that should capture all the shutdown logs AFAICT - at least, as much as is possible if the process is behaving badly.

I actually tried something similar in the beginning; unfortunately, the Android emulator didn't respond to receiving `SIGKILL`.

I can only test the Android emulator on Windows, though; I'd be curious if this behavior if different elsewhere.

> > [edit] branch coverage is a little wild...[line 1075]
> 
> FWIW: That one I can explain from experience - macOS had a similar weird report.
> 
> The issue is the lambda. You're defining a lambda on L1078, but it's never exercised. It reports as L1075 because functionally that's the line of code in effect when the lambda is executed; the exit is the exit of the lambda.

hmm.....that's interesting. Thanks, though; I'll see what can be done.

rmartin16:
Also, given our recent/previous issues with IO `read` calls on Windows end up hard blocking, even for CTRL+C, I think I'd be more hesitant to add another `read` call without a timeout in the main thread 😕

rmartin16:
I did some more testing today.

Calling `emulator_popen.kill()` on linux does actually shut down the emulator and allows `subprocess` to call `read()` on the `pipe` to get the output.

On Windows, though, things go much differently. Calling `kill()` does actually terminate the process....but only that process. The rest of the process tree seems to carry on. Additionally, though, the rest of the process tree seems to keep the `pipe` open so `communicate()` never returns....and as we've seen before, CTRL+C cannot interrupt it.

I was able to create an organic emulator failure in a VM earlier to demonstrate this. While in my case I got a window to close for the emulator, in the reported case, they didn't seem to have any visible failure.

This uninterruptible `read()` also happens during communicate if you just call `emulator_popen.kill()` after starting but before booting because the actual emulator will keep going while seemingly holding open the `pipe`.

https://user-images.githubusercontent.com/387848/195705541-66d7c351-9990-411e-8ab9-3fc6cc73d09a.mp4

rmartin16:
for the record, I'm not 100% convinced introducing this complication in to code that has already had several corner case bugs is necessarily juice worth the squeeze.

A few other ideas:
 - After a set amount of time passes waiting for the emulator to start, inform users something is probably wrong.
 - Catch `KeyboardInterrupt` during emulator startup and provide users with instructions on how to manually start the emulator if doesn't seem to be working for them.

freakboy3742:
@rmartin16 Yeah - I think I agree mitigation with documentation is likely our best solution here. Adding complication on complication to the output logger is only going to make things more fragile; and ultimately, for edge cases like the ones that are going to be the cause of problem there, getting a clean read of what is actually going on, independent of Briefcase, will probably be desirable.

rmartin16:
Good deal. I'll probably pursue an additional message to users if they send CTRL+C during emulator startup.

PR #918, opened 2022-10-14
Minor cleanups stemming from v0.3.11 release.

freakboy3742:
The v0.3.11 release highlighted a couple of minor problems.

1. The push instructions for the detached branch weren't quite right
2. The gitignore file for the repo doesn't ignore `logs` directories.

## PR Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ venv*/
 *.wp[ur]
 
 tests/apps/verify-*/
+logs/
--- a/changes/918.misc.rst
+++ b/changes/918.misc.rst
@@ -1,0 +1,1 @@
+Minor problems with the release instructions were corrected.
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -34,3 +34,4 @@ venv*/
 *.wp[ur]
 
 tests/apps/verify-*/
+logs/
```
mhsmith, line 37: What would create a logs directory that isn't already ignored by one of the other rules? I tried running the unit tests, but they didn't do it.

freakboy3742: I did a manual build of the verify-toga app, to replicate one of the failures I saw during the release process. That generated a logs folder in tests/apps.

mhsmith: But the tests/apps subdirectories are already ignored by the line above.

freakboy3742: Yes, but the logs folder was generated in `test/apps`, because it was `briefcase new` that failed, rather than `create` etc. I guess we could narrow it to that specific known failure; but I figured we might as well have a blanked rule to avoid any stray logs folders.

mhsmith (approved):

PR #921, opened 2022-10-16
Converge `run` Command Behavior

### LINKED ISSUE #904: Don't Log "Aborted by User" When User is Instructed to Use CTRL+C to Exit
**Describe the bug**
"Aborted by User" should not be logged when the user is instructed to use CTRL+C to exit Briefcase.

**To Reproduce**
Start the Android emulator or iOS simulator via `briefcase run` and send CTRL+C.

**Expected behavior**
Briefcase should exit normally.

**Screenshots**
```
src/briefcase/integrations/android_sdk.py:            # If the user sends CTRL+C:
src/briefcase/integrations/android_sdk.py:            # Exit normally since the user was instructed to use CTRL+C.
src/briefcase/platforms/android/gradle.py:            "Following device log output (type CTRL-C to stop log)...",
src/briefcase/platforms/iOS/xcode.py:                "Following simulator log output (type CTRL-C to stop log)...",
src/briefcase/platforms/macOS/__init__.py:                    "Following system log output (type CTRL-C to stop log)...",
src/briefcase/platforms/web/static.py:                "Web server log output (type CTRL-C to stop log)...",
src/briefcase/platforms/web/static.py:            # CTRL-C is the accepted way to stop the server.
```

rmartin16:
## Changes
macOS
- Removed CTRL-C instruction
- Catch `KeyboardInterrupt` to exit normally
- Send `SIGTERM` to App when Briefcase exits

Windows
- Catch `KeyboardInterrupt` to exit normally

Linux
- Catch `KeyboardInterrupt` to exit normally
- Add `====` App log header

iOS:
- Catch `KeyboardInterrupt` to exit normally

Android:
- Catch `KeyboardInterrupt` to exit normally

## Issues
- Fixes #904

## PR Checklist:

### DIFF
--- a/changes/904.bugfix.rst
+++ b/changes/904.bugfix.rst
@@ -1,0 +1,1 @@
+Briefcase now exits normally when CTRL-C is sent while tailing logs for the App when using ``briefcase run``.
--- a/src/briefcase/platforms/android/gradle.py
+++ b/src/briefcase/platforms/android/gradle.py
@@ -238,12 +238,15 @@ class GradleRunCommand(GradleMixin, RunCommand):
                 if not pid:
                     time.sleep(0.5)
 
-        self.logger.info(
-            "Following device log output (type CTRL-C to stop log)...",
-            prefix=app.app_name,
-        )
-        self.logger.info("=" * 75)
-        adb.logcat(pid)
+        try:
+            self.logger.info(
+                "Following device log output (type CTRL-C to stop log)...",
+                prefix=app.app_name,
+            )
+            self.logger.info("=" * 75)
+            adb.logcat(pid)
+        except KeyboardInterrupt:
+            pass  # catch CTRL-C to exit normally
 
 
 class GradlePackageCommand(GradleMixin, PackageCommand):
--- a/src/briefcase/platforms/android/gradle.py
+++ b/src/briefcase/platforms/android/gradle.py
@@ -189,9 +189,7 @@ class GradleRunCommand(GradleMixin, RunCommand):
         :param device_or_avd: The device to target. If ``None``, the user will
             be asked to re-run the command selecting a specific device.
         """
-        device, name, avd = self.tools.android_sdk.select_target_device(
-            device_or_avd=device_or_avd
-        )
+        device, name, avd = self.tools.android_sdk.select_target_device(device_or_avd)
 
         # If there's no device ID, that means the emulator isn't running.
         # If there's no AVD either, it means the user has chosen to create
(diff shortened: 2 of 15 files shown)


rmartin16:
My macOS VM keeps throwing kernel panic when I start the simulator....not sure if its recent updates are the problem...

@freakboy3742, can you please check this works for the iOS simulator?

rmartin16:
Finally got macOS to stay up long enough to test the iOS simulator.

This did give me a moment, though, to reflect on this nature of the `run` command. Namely, this behavior is not uniform among the platforms; users aren't asked to send CTRL-C for Windows and Linux apps. Moreover, though, what happens when users send CTRL-C is different.

When sending CTRL-C to Windows, Linux, and web apps, the app stops running. For macOS, the app continues running. For Android and iOS, the emulator/simulator keeps running.

Conversely, all platforms (except web) automatically detect when the app is closed and stop tailing the log.

**Proposal:**
- Remove instructions to send CTRL-C (except for web)
- When users send CTRL-C, shut down the app (except for Android and iOS)
- Catch CTRL-C on all platforms and do not report "Aborted by User"

Thoughts?

mhsmith:
I agree with all of that, except I think it makes sense to keep the instruction to send Ctrl-C for Android and iOS, because:
* In those cases Ctrl-C is the *only* way to terminate Briefcase (closing the app won't do it), so it's worth letting the user know that in case they're unfamiliar with terminals.
* If the user *is* familiar with terminals, it's useful to make it clear that Ctrl-C will terminate the log but not the app, which is something they would otherwise be unable to guess without trying it.

rmartin16:
> In those cases Ctrl-C is the only way to terminate Briefcase (closing the app won't do it), so it's worth letting the user know that in case they're unfamiliar with terminals.

Are you saying that closing the App _within_ the simulator does not stop Briefcase? I guess I hadn't thought of it that way. Nonetheless, leaving the CTRL-C instructions makes sense.

mhsmith:
Yes, I think closing the app within the emulator is the most analogous action to closing the app window on desktop. Or you might be working on multiple apps on the same emulator, in which case you wouldn't want to close the emulator down when moving between them.

Also, the user might not even be using an emulator, they might be using a physical device.

rmartin16:
My macOS VM completely s**t the bed now :( Attempting to create one in qemu now since Virtualbox on Windows seems to be a no-go now...

rmartin16:
Got macOS running on qemu...although, BeeWare apps always trigger a seg fault on the iOS simulator...

Nonetheless, I was able to test the changes here.

freakboy3742 (approved):
Looks good; FWIW, testing works fine for me on iOS and macOS.

PR #927, opened 2022-10-16
Treat `dev0` versions of Briefcase as development versions

### LINKED ISSUE #926: Briefcase `v0.3.12.dev0+xxx` is not Treated as a Dev Version
**Describe the bug**
The `dev` attribute of a `packaging.version.Version` object is the number from `dev#` in the version. When the number is `0`, the version is not treated as a dev version.

`create.py`
```python
            # If we're on a development branch, and the template branch was *not*
            # provided explicity, we can use a fallback development template.
            # Otherwise, re-raise the exception about the unsupported template version.
            if version.dev and app.template_branch is None:
```

`new.py`
```python
            # If we're *not* on a development branch, raise an error about
            # the missing template branch.
            if not version.dev:
                raise
```

```python
❯ python
Python 3.10.7 (main, Sep 28 2022, 17:55:46) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from packaging.version import Version
>>> Version('0.3.12.dev1+gf0e822f.d20221016').dev
1
>>> Version('0.3.12.dev0+gf0e822f.d20221016').dev
0
```

**To Reproduce**
Run `briefcase create` or `briefcase new` with a `dev0` version.

**Expected behavior**
`dev0` versions are dev versions.

**Environment:**
 - Operating System: Ubuntu 22
 - Python version: 3.10
 - Software versions:
   - Briefcase: `0.3.12.dev0+gf0e822f.d20221016`

rmartin16:
Fixes #926.

## PR Checklist:

### DIFF
--- a/changes/926.misc.rst
+++ b/changes/926.misc.rst
@@ -1,0 +1,1 @@
+Briefcase versions containing ``dev0`` are now considered development versions for fetching templates.
--- a/src/briefcase/commands/create.py
+++ b/src/briefcase/commands/create.py
@@ -291,9 +291,9 @@ class CreateCommand(BaseCommand):
             )
         except TemplateUnsupportedVersion:
             # If we're on a development branch, and the template branch was *not*
-            # provided explicity, we can use a fallback development template.
+            # provided explicitly, we can use a fallback development template.
             # Otherwise, re-raise the exception about the unsupported template version.
-            if version.dev and app.template_branch is None:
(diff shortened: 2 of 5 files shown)


mhsmith (changes requested):
The fix looks good, but the tests should be updated to check the `.dev0` case. Search the tests for the string `.dev` to find the relevant places.

PR #929, opened 2022-10-18
Show Android emulator troubleshooting steps for CTRL-C

rmartin16:
As discussed in #914, this is a less intrusive approach to provide more information when the user interrupts the process to start the Android emulator.

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -812,7 +812,7 @@ Use the -d/--device option to explicitly specify the device to use.
                 f"""
 In future, you can specify this device by running:
 
-    $ briefcase run android -d @{avd}
+    $ briefcase run android -d "@{avd}"
 """
             )
         elif device:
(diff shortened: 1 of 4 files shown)


src/briefcase/integrations/android_sdk.py
```diff
@@ -999,6 +999,27 @@ def start_emulator(self, avd):
             start_new_session=True,
         )
 
+        # wrap AVD name in quotes since '@' is a special char on Windows
+        emulator_command = " ".join(
+            f'"{arg}"' if arg.startswith("@") else arg
```
mhsmith, line 1004: Is it special in cmd, or only in PowerShell? If it's only PowerShell, then the comment should say that.

rmartin16: It is reserved in PowerShell; cmd treats it as a normal character. I'll update the comment.

FWIW, while confirming this, I realized the `-avd` option can be used in lieu of `@`.

mhsmith (approved):
Works fine for me on Windows.

Issue #737, opened 2022-05-11
Improve flexibility of Android emulator system image


freakboy3742:
The Android toolchain currently hard-codes the system image that is used on emulator images built by Briefcase (at time of writing, the default Android-31 image is used). This system image is downloaded when the emulator binary is downloaded, and is never directly updated.

This has some unfortunate side effects:
1. If a user has an emulator created by a different set of tools (e.g., Android Studio), the emulator will appear in the list of available emulators, but won't be able to start 
2. A user can't intentionally choose a different base image - e.g., to test older/newer devices.
3. A user can't intentionally choose a base image other than the Default image (e.g., a Google Play enabled image)

**Describe the solution you'd like**

It would be desirable for Briefcase to:
1. provide the option of selecting a base image from the list provided by the SDK manager
2. check that the system image required by an emulator is available before trying to start the image

**Describe alternatives you've considered**

There aren't really any alternatives, other than manually invoking sdkmanager to create images and download system images that meet your requirements.

**Additional context**

`sdkmanager` is a tool provided by the Android toolchain to do package management. Run `briefcase run -v` to see examples of how it can be run, including the environment variables that must be set in order to invoke it. 

`sdkmangager --list` will give you a list of available images; any package starting with `system_images;` is, as the name suggests, a system image. The default system image used by briefcase on M1 hardware is `system-images;android-31;default;arm64-v8a` (system images are architecture dependent; only architecture-appropriate options should be provided). 

`sdkmangager --list_installed` will give you a list of images that are currently installed.

`sdkmanager <package name>` will download and install a package.

Android emulators (called AVDs) are stored in `~/.briefcase/avd`. There is a `.ini` config file and a `.avd` directory for each emulator, with the name of these files matching the emulator name. The `.avd` folder contains a `config.ini` that details the system image required by the AVD under the `image.sysdir.1` key (e.g., `image.sysdir.1=system-images/android-31/default/arm64-v8a/`)

freakboy3742:
For the benefit of anyone interested in this ticket - it should now be significantly easier, as point (2) on the "desired solution" list has been implemented as part of #766. The problem is effectively reduced to:

1. Wrapping `sdkmanager --list` to generate a list of *all* available packages
2. Filtering that list to only include system images that correspond to supported Android versions for the current architecture (e.g., no point displaying arm64 images on an x86_64 laptop; no point displaying any image prior to Android-26, as that's the minimum supported version).
[...]

PR #934, opened 2022-10-21
Provide list of system images and change method to check for system image existence

cryptext:
- Created a way to download all available system images through the `sdkmanager` that correspond to currently supported Android versions and architecture
- Presented those images for user's choice
- Updated `verify_system_image()` to check for `system.img` instead of the existence of its parent directory in order to correct problem of incomplete downloads
- Fixes issue [#737](https://github.com/beeware/briefcase/issues/737)

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -500,7 +500,6 @@ connection.
             image (e.g., ``"system-images;android-31;default;x86_64"``)
         """
         # Look for the directory named as a system image.
-        # If it exists, we already have the system image.
         system_image_parts = system_image.split(";")
 
         if len(system_image_parts) < 4 or system_image_parts[0] != "system-images":
@@ -527,13 +526,15 @@ connection.
             )
 
         # Convert the system image into a path where that system image
-        # would be expected, and see if the location exists.
+        # files would be expected, and see if the `system.img` exists.
         system_image_path = self.root_path
         for part in system_image_parts:
             system_image_path = system_image_path / part
 
-        if system_image_path.exists():
-            # Found the system image.
+        system_image_file = system_image_path / "system.img"
+
+        if system_image_file.exists():
+            # Found the system image file.
             return
 
         # System image not found; download it.
@@ -885,8 +886,31 @@ An emulator named '{avd}' already exists.
         # Ensure the required skin is available.
         self.verify_emulator_skin(skin)
 
-        # TODO: Provide a list of options for system images.
-        system_image = f"system-images;android-31;default;{self.emulator_abi}"
+        # Provide a list of options for system images.
+        with self.tools.input.wait_bar("Retrieving list of available system images..."):
+            try:
+                download_options = self.tools.subprocess.check_output(
+                    [os.fsdecode(self.sdkmanager_path), "--list"],
+                    env=self.env,
+                    stderr=subprocess.STDOUT,
+                )
+                options = download_options.splitlines()
+                images = []
+                for line in options:
+                    if re.match(
+                        rf"\s\ssystem-images;android-([2-9][6-9]|[3-9][0-9]|\d{3,});default;{self.emulator_abi}",
+                        line,
+                    ):
+                        strip_line = line.strip().split(" ")[0]
+                        if (strip_line, strip_line) not in images:
+                            images.append((strip_line, strip_line))
+            except subprocess.CalledProcessError as e:
+                raise BriefcaseCommandError("Unable to retrieve system images") from e
+        # Show image options to the user.
+        self.tools.input.prompt()
+        self.tools.input.prompt("Select system image:")
+        self.tools.input.prompt()
+        system_image = select_option(images, input=self.tools.input)
 
         # Ensure the required system image is available.
         self.verify_system_image(system_image)


freakboy3742 (changes requested):
Thanks for the contribution! You may have noticed that CI is currently failing; the current failure is because you're [missing a changenote](https://briefcase.readthedocs.io/en/latest/how-to/contribute-code.html#add-change-information-for-release-notes). Once that test passes, you'll likely find that you'll get another failure because the tests will no longer have 100% coverage.

If you're looking for specific advice, let us know; otherwise, we tend not to review PRs until CI is passing.

cryptext:
Thank you. Was able to add the appropriate file for the release notes and updated `test_verify_system_image.py` in order to pass the tests, as those two where simple enough. But, having never written tests before, if you have any guidance on how to update `test_create_emulator.py` in order to test for the applicable commit updates that would certainly be appreciated. Will start studying how to use `pytest` to figure it out. (Previously wrote some comments about the commit in the [issue tracker](https://github.com/beeware/briefcase/issues/737#issuecomment-1286460173)).

freakboy3742:
> Thank you. Was able to add the appropriate file for the release notes and updated `test_verify_system_image.py` in order to pass the tests, as those two where simple enough. But, having never written tests before, if you have any guidance on how to update `test_create_emulator.py` in order to test for the applicable commit updates that would certainly be appreciated. 

I'm not sure exactly what sort of guidance you're looking for. Broadly speaking, the problem you're seeing in the tests at present is that the code you've added is now asking for user input. However, under test conditions, you can't ask for user inputs, as the tests need to be fully automated. As a result, you need to mock the behavior of the input method so that the code can continue as if the end user had provided specific inputs.

The test cases for `command/create/test_create_app.py` contain a number of examples of this - the creation process *can* ask for user input in a number of cases, and we need to mock what happens when the user provides various input to those prompts. That's very similar to what is happening in your case., 

The other major area where tests are required are around the detection of system images. The tests required in this case will be similar to the tests in `integrations\android_sdk\AndroidSDK\test_devices.py` - examples of parsing various possible outputs from a command line tool. To support this, you may find it helpful to factor out the specific "detect system images" code into a utility method - that way, your tests will be of a single utility "get a list of system images" method, and the tests of the `create_emulator` method will be simpler code that can assume a list of system images can be obtained.

tests/integrations/android_sdk/AndroidSDK/test_detect_system_images.py
```diff
@@ -0,0 +1,35 @@
...
+        "  system-images;android-31;google_apis_playstore;arm64-v8a | 9  | Google Play ARM 64 v8a System Image\n"
+    )
+
+    mock_tools.subprocess.check_output.return_value = system_images
+
+    select_option = MagicMock()
+    select_option.return_value = "system-images;android-26;default;x86_64"
+    system_image = select_option.return_value
+
+    assert system_image == "system-images;android-26;default;x86_64"
```
freakboy3742, line 35: This test doesn't appear to be actually testing anything - it never actually invokes `detect_system_images`. There's also multiple code paths through this method (which means there's a need for multiple tests), and multiple options for success and failure (Nothing to match, nothing valid to match, a single option matches, multiple options match...)

freakboy3742 (changes requested):
Getting closer - The refactoring of `detect_system_image` is really good, and you've got the test suite is passing; now the problem is coverage. The test you've added for `test_detect_system_images` doesn't actually exercise the code - and this is then picked up in the coverage report, which shows that the test suite is not ever running lines 832-856 (which is effectively the entire body of the new method.

cryptext:
Thank you for the encouragements. Learning Python testing while doing this has not been an easy process. Hopefully it is getting better.

Was finally able to figure out how to start testing the code and have almost 100% coverage, except for one line where an error is raised (more below). Though the coverage is close to 100%, perhaps more can be done to cover all cases beyond what is there. While doing that though, found a bigger bug...

The code is commented to reflect the following, but `sdkmanager --list` seems to have multiple processes and, if they don't *all* fail, the `subprocess.CalledProcessError` exception won't be called. If, for instance, the user loses their internet connection, the `sdkmanager` might not download the list of available system messages, but will still continue and show what is already installed. (It also seems to cache what it already downloaded for up to ten minutes and will display that too, even if it didn't currently download anything.) If the list of available system images is not downloaded, the user is not presented with any list. There is only the prompt to choose from a non-existent list. Don't know right now how to fix this other than to check if the `images` list is empty and then raise an error if so. (This is the one line that is not tested yet and don't know if it can be.) So, right now, all seems to be working as intended, but checking if `images == []` and raising an error does not feel like a proper way to fix the problem. Do you think this needs to be refractored again into at least two different methods? Perhaps, another possible try-except would help? At the moment, at a loss for ideas.

freakboy3742:
> So, right now, all seems to be working as intended, but checking if `images == []` and raising an error does not feel like a proper way to fix the problem. Do you think this needs to be refractored again into at least two different methods? Perhaps, another possible try-except would help? At the moment, at a loss for ideas.

That's a tricky one. I completely agree this is a failure mode that we should be handling in some way - but if the tool doesn't provide any way to identify that failure mode, then there's a limit to what we can do. We can only do as much as the tools allow us to do; if the tools don't provide a useful interface for us to use, the best we can do is fail as gracefully as we can.

I agree that this may require splitting the command into two pieces, as you describe - but possibly for a different reason than you're suggesting. I think if we go back to look at the problem we're trying to solve, we can make raising an error on the "empty list" case make a lot more sense.

While the original problem statement is "allow the user to use any system image", most Briefcase users will only ever need one system image (i.e., the current behavior). You only need the full list if you're doing specific device testing that you need more than one image. So, this "full list" behavior is a "power user" feature. It's absolutely a feature that is *needed*, but it doesn't need to be the behavior the user sees by default.

With that in mind, here's a slightly reworked approach to this feature:

`create_emulator` invokes `sdk_manager --list_installed` to check the list of *locally* available images first, and present that as the candidate list to the user. However, this list will also include:
1. the "briefcase default" image (regardless of whether the image is available locally); and
2. a fallback option to "obtain a new system image"

The first time you run Briefcase and create an emulator, you'll see a menu of 2 options: "default image", and "obtain new image", with "default image" being the default choice for the menu. *Most* users will only ever see this menu of 2 options, as the default option on this selection will always be the "briefcase default" image, and they won't have a reason to seek out a different image.

However, if they select "obtain a new system image", we move to phase 2; invoke `sdkmanager --list` to get the *complete* list of available images. In that second phase, getting an empty list is an indication of network failure, so we can raise that error to the user reporting that fact. If the command succeeds, they will select a new image from the list of available options.

If they *do* select a custom image, the next time the user creates an emulator, they'll have an extra option on the "phase 1" list; the "briefcase default" will always be there (and be the default), no matter whether it's been downloaded or installed.

This workflow makes the simple case (a new user creating their first emulator) simpler, as it doesn't involve the network round trip to download and select an image. However, it also provides the configurability option for the advanced case, and provides a clear interpretation of the "empty list" case. 

In terms of refactoring, it effectively means there's a wrapper method around the two possible sdkmanager invocations ("get local images" and "get all images"); and then a utility method that manages the process of allowing the user to select an image. This also provides 3 clear interfaces that can be tested independently.

Does that make sense?

cryptext:
It all makes sense. Thank you. Will have time to work on it this coming week.

tests/integrations/android_sdk/AndroidSDK/test_detect_system_images.py
```diff
@@ -0,0 +1,63 @@
...
+
+@patch("briefcase.integrations.android_sdk.AndroidSDK.select_system_images")
+def test_detect_system_images_obtain_new(
+    mock_select_system_images, android_sdk, mock_tools
+):
+    mock_tools.input.values = ["3"]
+    mock_select_system_images.return_value = "system-images;android-31;default;x86_64"
+    system_image = android_sdk.detect_system_images()
+    assert system_image == "system-images;android-31;default;x86_64"
+    assert len(mock_tools.input.prompts) == 1
```
freakboy3742, line 52: As currently constructed, this is more of a test of `input`, than a test of the `detect_system_images` method. The part of the logic that is most complex is the processing of the return value from  `check_output`; what is needed are more extensive tests that that filtering and processing will behave as expected, with a wide range of inputs. This probably means testing that the inputs to `input()` are what you expect for a wide range of potential configurations.

tests/integrations/android_sdk/AndroidSDK/test_detect_system_images.py
```diff
@@ -0,0 +1,63 @@
...
+import pytest
+
+from briefcase.exceptions import BriefcaseCommandError
+from briefcase.integrations.android_sdk import AndroidSDK
+
+
+@pytest.fixture
+def android_sdk(android_sdk, tmp_path) -> AndroidSDK:
+    android_sdk.tools.subprocess.check_output = MagicMock(
+        return_value=(
```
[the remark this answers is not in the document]
freakboy3742: This test is also slightly misleading - it contains the "useful" part of the `sdkmanager` output, but not the pre-postamble. Since one of the purposes of this function is to strip the pre/postamble, we need to have a test of that capability.

tests/integrations/android_sdk/AndroidSDK/test_select_system_images.py
```diff
@@ -0,0 +1,72 @@
...
+            "  system-images;android-30;android-tv;arm64-v8a     | 3    | Android TV ARM 64 v8a System Image\n"
+            "  system-images;android-30;android-tv;x86           | 3    | Android TV Intel x86 Atom System Image\n"
+            "  system-images;android-30;default;arm64-v8a        | 4    | ARM 64 v8a System Image\n"
+            "  system-images;android-30;default;x86_64           | 4    | Intel x86 Atom_64 System Image\n"
+            "  system-images;android-30;google-tv;arm64-v8a      | 3    | Google TV ARM 64 v8a System Image\n"
+            "  system-images;android-30;google-tv;x86            | 3    | Google TV Intel x86 Atom System Image\n"
+            "  system-images;android-30;google_apis;arm64-v8a    | 10   | Google APIs ARM 64 v8a System Image\n"
+            "  system-images;android-30;google_apis;x86_64       | 10   | Google APIs Intel x86 Atom_64 System Image\n"
+            "  system-images;android-30;google_apis_playstore;arm64-v8a | Google Play ARM 64 v8a System Image\n"
+        )
```
freakboy3742, line 32: Most of the comments I've made about `detect_system_images` also apply to these tests.

freakboy3742 (changes requested):
The implementation now looks fairly good; and you've got to 100% coverage 🎉 - but there are a lot of tests still required. Coverage is a good guide for what you've forgotten to test when you've added new functionality - but coverage only tells you that every line of code has been run. It won't tell you whether you've covered the full range of potential inputs - which is especially important when you're integrating with an external system like `sdkmanager` outputs.

PR #943, opened 2022-10-25
Stream Subprocess Command Output by Default

### LINKED ISSUE #917: Disabling the Wait Bar for Batch Scripts Prevents Including Script Output in the Logfile
**Describe the bug**
#877 disabled the Wait Bar for batch scripts. Since an active Wait Bar is the signal to stream output for the command, the output for batch scripts is not streamed and not included in the logfile.

For example, building an Android APK is not captured in the logfile.

**To Reproduce**
Run `briefcase run android --log` and see the build output is absent in the logfile.

**Expected behavior**
Batch script output should be captured in the logfile.

Passing `stream_output=True` for calls to `Subprocess.run` is probably the needed change.

Alternatively, defaulting to `stream_output=True` and requiring devs to intentionally set `stream_output=False` would help prevent this broader case of issues. For instance, the output for flatpak builds cannot be streamed.

**Environment:**
 - Python version: 3.10.7
 - Software versions:
   - Briefcase: `0.3.11rc6.dev3+g0a0f3a36.d20221013`

rmartin16:
## Changes
Switches the default for streaming `subprocess.run()` output from `False` to `True`. The primary advantage of this is avoiding inadvertently using `subprocess.run()` in a way that the output is not streamed and is therefore excluded from the logfile.

Given output streaming is the more desirable situation, updating this default seems most appropriate....especially since we keep finding missing output in the logfile.

## Assessment
The large majority of `subprocess.run()` calls are not affected because they run in the context of the Wait Bar when output streaming is implicitly used.

I have reviewed every use of `subprocess.run()`; they fall in to one of these camps:
- `stream_output=True` was being explicitly set
- The Wait Bar is active
- `stream_output=False` was being implicitly depended on
  - For instance, `flatpak install` cannot use output streaming.
  - This is also true of some uses of the Android tools.

I have been able to test each workflow except for macOS notarization since I do not have an Apple developer account.

## Issues
- Fixes #917 

## PR Checklist:


rmartin16:
@freakboy3742, can you please comment on whether you agree with these changes?

It is going to take some effort to get the tests working again....just want to make sure we're on at least roughly the same page first.

freakboy3742 (commented):
I guess this makes sense. All told, I'd rather not need this streaming at all... however, that's clearly not an option that we have :-) 

The only general comment I'd have is one about documentation; the docstring describing how output streaming works and why it's needed is now out of date. Given that this is one of the more complex internals, it would be good to have a thorough teardown of why it is needed, how it works, and when it *won't* work. The (unavoidable) complexity of the implementation is reaching the point where having a solid guide for when/how to use this option that doesn't require a deep understanding of the actual implementation is becoming increasingly necessary.

If that's too much for a docstring, an internal implementation document (linked from the docstring) would be another option.

rmartin16:
That's reasonable. Towards that end, I think this change simplifies the situation since the default will now be to stream all output....instead of only (implicitly) when the Wait Bar was active. And in a lot of ways, the tl;dr for `stream_output=False` is "use it if the output looks bad and 'plain output' cant be enforced." I can flesh that out.

OTOH, I realized that output streaming isn't currently required when a progress bar is active....such as when an Apple app is being signed. That'll probably need some non-trivial changes if it's addressed...

Note: the CI will currently run forever since several of the tests get stuck in `Popen` now.

rmartin16:
The tests are running now. However, the testing of `Subprocess.run()` is a little wonky now....since all the testing was orchestrated around the real `subprocess.run()` getting called....and that's the uncommon case now.

I'm gonna spend some more time considering if testing for `stream_output=True/False` and "run in wait bar" could be better organized at this point.

freakboy3742 (approved):
I've made 2 cosmetic tweaks:
1. Docstrings should be in ReST format, so I've made a couple of small formatting changes to the primer
2. While the filenames including `=True` and `=False` are 100% legal... I'm amazed that is actually true :-) To avoid any possible surprises in future that might happen because a code module isn't importable, I've renamed those tests.

Other than that, this looks great. The primer on output streaming is *really* good; and the rest of the test changes all make sense. 

Now, of course, all our output streaming problems are resolved, and we'll never need to look at this code again....

... right??

:-)

rmartin16:
"and they lived happily ever after” ;)

Issue #400, opened 2020-05-21
Document how to use Briefcase in CI


freakboy3742:
We should add a topic guide for how to incorporate Briefcase into a CI setup.
This includes 
 * On a pull request, running:
    - code style checks
    - checking documentation

saroad2:
As you probably remember, I incorporate Briefcase in my CI/CD process of Eddington for about a few months now, but I'm using CircleCI as my platform for that.

If anyone who works on this issue needs some reference, he/she can take a look [here](https://github.com/EddLabs/eddington_gui/blob/dev/.circleci/config.yml) to see how I did it there.

jgirardet:
is `briefcase create` runnable now on github action ?
There are problems with docker.
For now it fails  on linux.
- Github action doesn't provide ubuntu 16.04 so you can't use `--no-docker` (I'm not sure about it)
- Github runner 18.04 provide an outdated version of docker (see #429)
- Github runner 20.04 fails for the following reason : `the input device is not a TTY` (https://pipelines.actions.githubusercontent.com/ppe4ZSLak9UTOCP3VTfZBea8jfANvOikJ7KH4UBWg38usmFQSU/_apis/pipelines/1/runs/477/signedlogcontent/4?urlExpires=2020-06-28T17%3A52%3A59.8504242Z&urlSigningMethod=HMACV1&urlSignature=YKVqhxtQ6%2BOi7iPmhgJF%2FYM8sSlepBu04mpPz%2B4BcWM%3D)
maybe a fix there (https://serverfault.com/questions/897847/what-does-the-input-device-is-not-a-tty-exactly-mean-in-docker-run-output). I have no knowledge on docker, I can't help on  it, just report.

freakboy3742:
@jgirardet Not sure where you got the idea that Github Actions doesn't have an Ubuntu 16.04 environment - [it's definitely on their official list](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources)

As for the "input device is not a TTY" error - I'm not sure what *specifically* caused that (and the log has expired); but that will almost certainly be because the console is asking for input. There's a `--no-input` flag to explicitly turn that off (and fall back to safe defaults).

jgirardet:
@freakboy3742  ok sorry it's fine on 16.04 (how I missed it ?) but it's not working on ubuntu 20.04 even with `--no-input`` : https://pastebin.com/FyvENBip
[...]

Issue #399, opened 2020-05-21
Add documentation for how to test a Briefcase-packaged app


freakboy3742:
The following exchange occurred in our Gitter chat room. A formalized version of this should be incorporated into our documentation as a topic guide on testing:

@tmontes: Given that briefcase leans towards a `src`-based layout -- which I like, for many reasons -- how are people running tests against their application code? My previous experience, as is the case with briefcase 0.2, is mostly based on `setuptools`+`setup.py` where I'd install the application in editable mode (with `pip install -e .`) and from there, running tests would be "fine" because the application modules/packages would be importable.

@freakboy3742: There’s two options. The first is to change the `sources` definition. Although the default Briefcase project uses the `src` prefix for all it’s code, it doesn’t have to; you can specify any folder you want as a source. If you drop the `src` prefix from the default sources definition, you can use a “prefixless” source.

The other option is to use a test runner that supports different source prefixes. `pytest`, for example, has a `pytest-pythonpath` plugin that lets you add a python_paths option in the `[tool:pytest]` section of `setup.cfg` so you can add other prefixes to your source path.

---

We should include details for how to use `unittest`, `pytest` and as a stretch goal - `trial`.

freakboy3742:
The docs should also mention [pytest-briefcase](https://pypi.org/project/pytest-briefcase)

saroad2:
> The docs should also mention [pytest-briefcase](https://pypi.org/project/pytest-briefcase)

The link is broken. Is this package exists yet?

freakboy3742:
You've beaten me to it :-)

I posted that comment thinking I was moments away from pushing to pypi... and then realized I needed new Briefcase release before I can push pytest-briefcase.

Hopefully that can happen in the next day or two; in the meantime, the WIP link is https://github.com/beeware/pytest-briefcase

GenevieveBuckley:
Does `pytest-briefcase` allow you to run the test suite on the bundled app/environment? Or is it equivalent to pytest with a `testpaths` argument (i.e. same as regular development testing)?

Right now I'm using a `pytest.ini`to specify that pytest should only look for tests in the source code.
```
[pytest]
testpaths = myprojectname/
```
This ignores any packaged bundles (macOS, windows, linux) that might also exist in the directory. This is good, now our CI pipeline won't break.

But it'd also be good if we could run tests on the actual bundled app (or run tests on the source from an environment that exactly matches all the dependency versions that were bundled). Is that what `pytest-briefcase` is about?

[...]

PR #949, opened 2022-11-02
Add a `briefcase test` command

freakboy3742:
Adds a `briefcase test` command that can be used to run test suites in project-based deployments.

Implemented for:
* iOS (requires the beeware/briefcase-iOS-Xcode-template#12 template updates)
* macOS Xcode (requires the beeware/briefcase-macOS-Xcode-template#14 template updates

Adds 2 new settings to projects:
* `test_sources` - a list of folders whose contents will be copied into the `tests` folder
* `test_requires` - a list of requirements needed during testing.

The most common example of usage will be a pytest suite, where all the test code is contained in a `tests` folder. This would be configured as:
```
test_sources = ['tests']
test_requires = ['pytest']
```
In the `tests` folder, in addition to any test code, there should be a file named `run_tests.py` that will start the test suite; this script should contain:
```
import pytest
import sys

if __name__ == "__main__":
    sys.exit(pytest.main(["-vv", "--color=no"]))
```

Other notable changes:
* Factors some the requirements installation code in to the base command, since it can now be re-used between create/update and test
* Simplifies the build process for iOS, dropping the need to define a target device when compiling.

Refs #699
Refs #400 
Refs #399 

## PR Checklist:


mhsmith:
We agreed that it's useful to have a `briefcase test` command, or a `--test` option for `briefcase run` and `briefcase dev`, which can parse a unit test pass/fail status from the stdout of either of the following types of app:
* A "test-only" app, such as Python-support-testbed, or what we're planning do for Toga UI tests.
* A "test-optional" app, such as a more typical Toga app.

For the "test-optional" case, we came up with two options for how to enable test mode:
* Putting something in the platform-specific metadata which causes the Python startup to launch a different entry point. This would have to be implemented separately on each platform.
* Putting something in the Python metadata which indicates that test mode is enabled, and allowing the Python entry point to do whatever it wants in response to that information.

But I'm not sure if we discussed this question: Since we have no plans to create a "test-optional" app in the near future, do we need to implement either of these options just now? Maybe it's good enough to just implement a Briefcase test mode which will be able to accommodate either of them, and come back to it later when we actually need it. By then, we'll have a lot more experience with this kind of testing, so it'll be easier to decide what to do.

freakboy3742:
Closed in favour of #962

PR #955, opened 2022-11-05
Use Python development mode for `briefcase dev` (fixes #806)

### LINKED ISSUE #806: `briefcase dev` should enable all warnings
As discussed in beeware/toga#1521, any changes to the Toga API should preserve the old API and give it a DeprecationWarning. However, DeprecationWarning is disabled by default to avoid displaying irrelevant messages to a program's end user. 

Unit test runners are only used during development, so they usually enable all warnings, but this won't be much help for Toga because user interface code isn't often unit tested. So as it stands, there isn't much chance of these warnings ever being seen.

Therefore, I think `briefcase dev` should enable all warnings, This can be done using Python's `-W` option.

rmartin16:
## Changes
`briefcase dev` now uses Python [development mode](https://docs.python.org/3/library/devmode.html) to show all warnings from the App and its dependencies.

While performance is definitely impacted by this setting, it should not be intolerable. I have been regularly using debug versions of Python to do Briefcase development; while it noticeably impacts Python's performance, even it isn't untolerable for dev.

In way of example, here are `pytest` timings (averaged over 10 consecutive runs) for Briefcase:
- Without `-X dev`:
```
real 19.79
user 10.60
sys 0.76
```
- With `-X dev`:
```
real 21.19
user 12.01
sys 0.78
```
- Debug Python (ooof)
```
real 71.80
user 62.45
sys 0.90
```

Also, as Victor [points out](https://vstinner.github.io/python37-dev-mode.html), (at least one) impetus for dev mode was apparently a more performant debug mode.

## Issues
- Fixes #806

## PR Checklist:

### DIFF
--- a/changes/806.feature.rst
+++ b/changes/806.feature.rst
@@ -1,0 +1,1 @@
+All warnings from the App and its dependencies are now shown when running ``briefcase dev`` by invoking Python in `development mode <https://docs.python.org/3/library/devmode.html>`_.
--- a/src/briefcase/commands/dev.py
+++ b/src/briefcase/commands/dev.py
@@ -91,6 +91,8 @@ class DevCommand(BaseCommand):
                 [
                     sys.executable,
                     "-u",
+                    "-X",
+                    "dev",
                     "-c",
                     (
                         "import runpy, sys;"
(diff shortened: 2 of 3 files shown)


freakboy3742 (approved):
Good idea!

Issue #753, opened 2022-06-03
Improve verification and recovery for corrupted downloads


freakboy3742:
It is possible for support packages and tool downloads to be corrupted during download; this can lead to a confusing situation where a user has a corrupted support file (or other resource), but the file *exists*, so Briefcase doesn't try to re-download the file. (e.g., beeware/beeware#157)

**Describe the solution you'd like**

Briefcase should have a more robust recovery mechanism for dealing with corrupted downloads. This could include:

* Verifiying checksums at time of download. This may not be easy to implement, as there's no consistent standard for how to publish the checksum of a file for automated verification.
* When an attempt to unpack an archive fails due to an apparent corrupted file, the original file is automatically deleted (or moved aside) so that a retry of the same command will cause a re-download.
* A user-accessible mechanism to manually flush the download cache (e.g., an option on `briefcase upgrade` to clear the support file cache). This could be combined with a "purge old support packages" command (e.g., no need to keep b1, b2 and b3 if b4 is available).
* A flag on commands that forces files to be re-downloaded (e.g., `briefcase create --always-download` will *always* re-download the support file, even if it already exists. I'm not in love with `--always-download` as a name; other suggestions welcome)

**Describe alternatives you've considered**

Status quo is to manually delete the problem files from `~/.briefcase/support`, but that folder isn't obvious to the end user.

mhsmith:
I think by far the most likely cause of this problem is the download being interrupted before it finished. That could be fixed by downloading to a temporary filename and then renaming if it completed without error.

freakboy3742:
Yes - using a temp name wouldn't be bulletproof, but it should definitely resolve a lot of problems. 

The only thing to be wary of is the location of the temp file - if we just append ".tmp" to the filename in the final location, we could end up with stray partial download files that are difficult for users to find and are never cleaned up - or, even worse, we could end up with collisions on the *temp* file name. We either need to be very careful about doing pre/post download cleanup, or make use of temporary directories to ensure that there won't be collisions.

mhsmith:
A common way of dealing with this is to include the process ID in the filename. That way, it's guaranteed to be unique at any point in time, so in the unlikely event that there's an existing file with the same ID, it can safely be overwritten.  

It also accommodates two Briefcase processes downloading the same file at the same time. But the second process to finish downloading must be able to deal with the target filename already existing, which would cause the rename to fail on Windows if the first process currently had the file open.

rmartin16:
[...]

PR #956, opened 2022-11-05
Download Failure Robustness/Recovery

rmartin16:
## Changes
- Download to a temporary file
  - Instead of downloading directly to the Briefcase cache, files are downloaded to the platform's designated temporary directory.
  - If the download succeeds, the temporary file is moved in to the Briefcase cache.
  - In most cases, the temporary file will be deleted is the download fails; one such exception may be Briefcase receiving SIGKILL.

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/download.py
+++ b/src/briefcase/integrations/download.py
@@ -1,3 +1,5 @@
+import tempfile
+from contextlib import suppress
 from email.message import Message
 from urllib.parse import urlparse
 
@@ -77,17 +79,30 @@ class Download:
                 # We have meaningful content, and it hasn't been cached previously,
                 # so save it in the requested location
                 self.tools.logger.info(f"Downloading {cache_name}...")
-                with filename.open("wb") as f:
+
+                # Using a temporary file avoids a partially downloaded file from
+                # masquerading as a complete file in later Briefcase runs
+                temp_file = tempfile.NamedTemporaryFile(delete=False)
+
+                try:
                     total = response.headers.get("content-length")
                     if total is None:
-                        f.write(response.content)
+                        temp_file.write(response.content)
                     else:
                         progress_bar = self.tools.input.progress_bar()
                         task_id = progress_bar.add_task("Downloader", total=int(total))
                         with progress_bar:
                             for data in response.iter_content(chunk_size=1024 * 1024):
-                                f.write(data)
+                                temp_file.write(data)
                                 progress_bar.update(task_id, advance=len(data))
+
+                    self.tools.shutil.move(temp_file.name, filename)
+                    self.tools.os.chmod(filename, 0o664)
+                finally:
+                    # Ensure the temporary file is deleted; this file may still
+                    # exist if the download fails or the user sends CTRL+C.
+                    with suppress(OSError):
+                        self.tools.os.remove(temp_file.name)
 
         except requests_exceptions.ConnectionError as e:
             if role:
--- a/tests/integrations/download/test_Download__file.py
+++ b/tests/integrations/download/test_Download__file.py
@@ -1,3 +1,4 @@
+import shutil
 from unittest import mock
 
 import pytest
@@ -16,6 +17,8 @@ from briefcase.integrations.base import ToolCache
 @pytest.fixture
 def mock_tools(mock_tools) -> ToolCache:
     mock_tools.requests = mock.MagicMock(spec_set=requests)
+    # Restore move so the temporary file can be moved after downloaded
+    mock_tools.shutil.move = mock.MagicMock(wraps=shutil.move)
     return mock_tools
 
 
@@ -86,6 +89,11 @@ def test_new_download_oneshot(mock_tools, url, content_disposition):
     # The filename is derived from the URL or header
     assert filename == mock_tools.base_path / "downloads" / "something.zip"
 
+    # Temporary file was upgraded to intended destination
+    mock_tools.shutil.move.assert_called_with(mock.ANY, filename)
+    mock_tools.os.chmod.assert_called_with(filename, 0o664)
+    mock_tools.os.remove.assert_called_with(mock.ANY)
+
     # File content is as expected
     with (mock_tools.base_path / "downloads" / "something.zip").open() as f:
         assert f.read() == "all content"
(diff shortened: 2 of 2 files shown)


src/briefcase/integrations/download.py
```diff
@@ -97,3 +87,38 @@ def file(self, url, download_path, role=None):
...
+                    task_id = progress_bar.add_task("Downloader", total=int(total))
+                    with progress_bar:
+                        for data in response.iter_content(chunk_size=1024 * 1024):
+                            temp_file.write(data)
+                            progress_bar.update(task_id, advance=len(data))
+
+            self.tools.shutil.move(temp_file.name, filename)
+            # retrieving the current umask requires changing it...
+            os.umask(current_umask := os.umask(0o022))
+            self.tools.os.chmod(filename, 0o664 & ~current_umask)
```
rmartin16, line 119: The `NamedTemporaryFile` file is only created with `600` permissions. So, all this `umask` stuff restores the previous behavior from `filename.open()`.

src/briefcase/integrations/download.py
```diff
@@ -97,3 +87,38 @@ def file(self, url, download_path, role=None):
...
+                    temp_file.write(response.content)
+                else:
+                    progress_bar = self.tools.input.progress_bar()
+                    task_id = progress_bar.add_task("Downloader", total=int(total))
+                    with progress_bar:
+                        for data in response.iter_content(chunk_size=1024 * 1024):
+                            temp_file.write(data)
+                            progress_bar.update(task_id, advance=len(data))
+
+            self.tools.shutil.move(temp_file.name, filename)
```
rmartin16, line 116: I just realized this `move` may not be a simple inode update if the system's temporary directory is on a filesystem different than that of Briefcase's cache directory. In that case, Python will fall back to a `copy` and `rm`.

Therefore, the underlying problem still exists if a user in this situation sends CTRL+C during the move potentially leaving an incomplete file in the Briefcase cache. (There's also the issue of the `move` potentially taking a relatively long time.)

Given this, I'll pause for feedback:
- Should I move forward with this approach? Or re-implement this solution always creating a temp file within the Briefcase cache?

The primary advantage IMO of the `NamedTemporaryFile` file is if anything goes wrong, the file already exists in an OS-designated temporary directory with (potential) OS faculties to delete it whenever.

I can also wrap the file move in a transient Wait Bar in case it is slow.

rmartin16: hmm....another thought is if the Briefcase cache is on a different filesystem, that is likely because the default filesystem is low on space....and that same default filesystem is likely to be used by Python's `NamedTemporaryFile`. Therefore, using it for this purpose here would probably betray some users' configuration....guess I have my answer...

rmartin16: on third thought....`NamedTemporaryFile` supports specifying a directory to create the temporary file in....so, that can mitigate the cross-filesystem concerns and still delegate creating a unique filename to Python's `tempfile`. We could even use the support to add a `suffix` or `prefix` to the filename to detect if one of these temporary files persists between Briefcase runs.

rmartin16: ## tl;dr
Updated the creation of the temporary file to occur in the directory of the Briefcase cache that the file will ultimately be moved to. This mitigates the concerns here

src/briefcase/integrations/download.py
```diff
@@ -97,3 +87,43 @@ def file(self, url, download_path, role=None):
...
+
+        :param response: ``requests.Response``
+        :param filename: full filesystem path to save data
+        """
+        temp_file = tempfile.NamedTemporaryFile(
+            dir=filename.parent,
+            prefix=f"{filename.name}.",
+            suffix=".download",
+            delete=False,
+        )
```
rmartin16, line 107: By default, `tempfile` uses random chars for the filename; prefixing with the intended filename makes it possible to know what the temporary data is and the suffix makes it easy to identity it as a temporary file.

Here's an example filename:
```
Python-3.10-linux-x86_64-support.b4.tar.gz.vdqfwdt2.download
```

rmartin16:
As mentioned in #753:

> The only thing to be wary of is the location of the temp file - if we just append ".tmp" to the filename in the final location, we could end up with stray partial download files that are difficult for users to find and are never cleaned up - or, even worse, we could end up with collisions on the temp file name. We either need to be very careful about doing pre/post download cleanup, or make use of temporary directories to ensure that there won't be collisions.

Right now, this PR doesn't address the possibility of `*.download` files accumulating in the Briefcase cache because of some bizarre unaccounted for problem.

So, should I also make updates to glob the Briefcase cache directory for `*.download` files at startup and delete them? Seems less than ideal...

Alternatively, perhaps this type of cleanup would be more appropriate as part of a "[a] user-accessible mechanism to manually flush the download cache".

tests/integrations/download/test_Download__file.py
```diff
@@ -86,6 +89,11 @@ def test_new_download_oneshot(mock_tools, url, content_disposition):
     # The filename is derived from the URL or header
     assert filename == mock_tools.base_path / "downloads" / "something.zip"
 
+    # Temporary file was upgraded to intended destination
+    mock_tools.shutil.move.assert_called_with(mock.ANY, filename)
```
freakboy3742, line 93: While this test passes, it's not especially resilient - `mock.ANY` really does allow *any* content. This initial "mock.ANY" call is definitely required to verify that *a* filename was used, but it would be desirable to add checks that the same temporary filename was used for all the subsequent calls.

src/briefcase/integrations/download.py
freakboy3742: A longer explanatory comment would be helpful here. The interplay between the 0o022 mask and the `and not` with 0o644 wan't clear to me.

There's also no test that the resulting file mask is actually followed; there's code coverage, but no real check of the logic.

freakboy3742 (changes requested):
I'm sure there are edge cases where this will still fail, but they're a lot smaller than the current edge case. 2 testing related comments, and one request for clarification in a docstring, but otherwise this looks really solid.

tests/integrations/download/test_Download__file.py
rmartin16: That's fair; I was initially struggling with how to introspect information about the temporary file and I went too far towards "assume it's working if everything else is working." I think my changes now verify the logic is working as designed.

src/briefcase/integrations/download.py
rmartin16: Added (some probably overly verbose) comments about umask. I also tried to implement file permission verification in the tests.....things are really wonky on windows, though...

freakboy3742 (approved):
Awesome - nice work.

Also - Kudos on the first walrus in the codebase :-P

PR #961, opened 2022-11-07
Add a `briefcase test` command (simplified)

freakboy3742:
A revised approach to #949, using a simpler approach based on repurposing the main application.

Instead of adding a native project test suite as is done by #949, the test runner in this PR modifies the execution of the normal app by re-writing app metadata. This changes the entry point module of the app from from `<app_name>` to `tests.<app_name>`. The log runner then filters the output stream of the running app, looking for a sentinel value indicating test success/failure.

This approach has the the following advantages:
* It is *much* faster on iOS. I don't know what Xcode is doing when it starts a test suite, but this approach is up to 2 minutes faster to start the test suite.
* It requires minimal platform-specific modifications. We don't need to work out how to drive native unit test frameworks on every platform; we can just use "normal" app execution
* It can be used on "app" targets, not just project targets.

The downside is that:
* We lose access to any platform-specific test automation tooling. e.g., Xcode has frameworks for taking screenshots etc during a test; we won't be able to use those (at least, not easily)
* We need to manually manage the process of starting an application in the background, if we want to programatically drive an application.

Implemented for:
* iOS (requires the https://github.com/beeware/briefcase-iOS-Xcode-template/pull/13 template updates)
* macOS Xcode (requires the https://github.com/beeware/briefcase-macOS-Xcode-template/pull/15 template updates

Adds 2 new settings to projects:

* `test_sources` - a list of folders whose contents will be copied into the tests folder
* `test_requires` - a list of requirements needed during testing.

The most common example of usage will be a pytest suite, where all the test code is contained in a tests folder. This would be configured as:

```
test_sources = ['tests']
test_requires = ['pytest']
```

the code copied in as part of the `test_sources` should contain an `app_name.py` that will start the test suite. This script should contain:
```
import os
from pathlib import Path

import pytest


if __name__ == "__main__":
    os.chdir(Path(__file__).parent.parent)
    result = pytest.main(["-vv", "--color=no"])
    if result:
        print(">>>>>>>>>> Test Suite Failed <<<<<<<<<<")
    else:
        print(">>>>>>>>>> Test Suite Passed <<<<<<<<<<")
```

The print statements are important; those are required by the implementation as sentinels. The log streamer will look for this sentinels in output, and exit (with overall command success/failure) if it sees one of them. If one of the sentinels is *not* seen, behavior varies by platform. 

On macOS, where we can identify the process ID, the process will terminate, with an error reporting that no test result was observed. 
On iOS, the log streamer will wait indefinitely for a success/failure result.

## PR Checklist:


freakboy3742:
Closed in favour of #962

PR #964, opened 2022-11-08
Add `ToolCache` typing

rmartin16:
## Changes 
Added typing support for `ToolCache` for QoL.
Specifically:
- Enumerate each Tool and its type in `ToolCache`
- Type `tools` as `ToolCache` each place it is introduced in a Tool

## Rationale
Personally, I really like my IDE's hints and auto-completion. The current implementation of `ToolCache` prevents any of these luxuries because its attributes cannot be inferred.
- This is most noticeable for tools in `briefcase.integrations` because `self.tools` cannot be inferred as a `ToolCache`.
- Within commands, `self.tools` is at least inferred to be a `ToolCache` (at least for the subclasses that aren't mix-ins) but since none of the Tools are actually defined within `ToolCache`, none of them can be inferred.

## Tradeoff
Unfortunately, this typing requires jumping through a few hoops and a bit more maintenance due to new dependencies within `ToolCache` itself....but I think its ultimately a win.

Also interested to know if this is just a PyCharm problem or if VS is in the same boat.

## PR Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,8 @@ precision = 1
 exclude_lines = [
     "pragma: no cover",
     "@(abc\\.)?abstractmethod",
-    "NotImplementedError\\(\\)"
+    "NotImplementedError\\(\\)",
+    "if TYPE_CHECKING:",
 ]
 
 [tool.isort]
--- a/src/briefcase/integrations/base.py
+++ b/src/briefcase/integrations/base.py
@@ -8,8 +8,7 @@ from abc import ABC
 from collections import defaultdict
 from collections.abc import Mapping
 from pathlib import Path
-from types import ModuleType
-from typing import TYPE_CHECKING, DefaultDict, TypeVar, Union
+from typing import TYPE_CHECKING, DefaultDict
 
 import requests
 from cookiecutter.main import cookiecutter
@@ -18,7 +17,9 @@ from briefcase.config import AppConfig
 from briefcase.console import Console, Log
 
 if TYPE_CHECKING:
-    import git as _git
+    # Tools are imported only for type checking
+    # to avoid circular import errors.
+    import git as git_
 
     from briefcase.integrations.android_sdk import AndroidSDK
     from briefcase.integrations.docker import Docker, DockerAppContext
@@ -37,9 +38,6 @@ class Tool(ABC):
     """Tool Base."""  # pragma: no cover
 
 
-TOOL_T = TypeVar("TOOL_T", bound=Union[Tool, ModuleType])
-
-
 class ToolCache(Mapping):
     # Briefcase tools
     android_sdk: AndroidSDK
@@ -47,7 +45,7 @@ class ToolCache(Mapping):
     docker: Docker
     download: Download
     flatpak: Flatpak
-    git: _git
+    git: git_
     java: JDK
     linuxdeploy: LinuxDeploy
     rcedit: RCEdit
@@ -104,17 +102,6 @@ class ToolCache(Mapping):
                 home_path=self.home_path,
             )
         )
-
-    def add_tool(self, name: str, tool: TOOL_T) -> TOOL_T:
-        """Add a tool to the cache.
-
-        :param name: Name of the tool to add to the cache; The tool will be
-            accessible as an attribute of the ToolCache using this name.
-        :param tool: The instantiated Tool to make available.
-        :return: The added Tool.
-        """
-        setattr(self, name, tool)
-        return tool
 
     def __getitem__(self, app: AppConfig) -> ToolCache:
         return self.app_tools[app]
--- a/src/briefcase/integrations/base.py
+++ b/src/briefcase/integrations/base.py
@@ -8,7 +8,7 @@ from abc import ABC
 from collections import defaultdict
 from collections.abc import Mapping
 from pathlib import Path
-from typing import DefaultDict
+from typing import TYPE_CHECKING, DefaultDict
 
 import requests
 from cookiecutter.main import cookiecutter
@@ -18,6 +18,18 @@ from briefcase.console import Console, Log
 from briefcase.integrations.download import Download
 from briefcase.integrations.subprocess import Subprocess
 
+if TYPE_CHECKING:  # pragma: no cover
+    import git as _git
+
+    from briefcase.integrations.android_sdk import AndroidSDK
+    from briefcase.integrations.docker import Docker, DockerAppContext
+    from briefcase.integrations.flatpak import Flatpak
+    from briefcase.integrations.java import JDK
+    from briefcase.integrations.linuxdeploy import LinuxDeploy
+    from briefcase.integrations.rcedit import RCEdit
+    from briefcase.integrations.visualstudio import VisualStudio
+    from briefcase.integrations.wix import Wix
+
 
 # TODO: Implement Tool base class
 class Tool(ABC):
@@ -25,11 +37,29 @@ class Tool(ABC):
 
 
 class ToolCache(Mapping):
+    # Briefcase tools
+    android_sdk: AndroidSDK
+    app_context: Subprocess | DockerAppContext
+    docker: Docker
+    download: Download
+    flatpak: Flatpak
+    git: _git
+    java: JDK
+    linuxdeploy: LinuxDeploy
+    rcedit: RCEdit
+    subprocess: Subprocess
+    visualstudio: VisualStudio
+    wix: Wix
+    xcode: bool
+    xcode_cli: bool
+
+    # Python stdlib tools
     platform = platform
     os = os
     shutil = shutil
     sys = sys
 
+    # Third party tools
     cookiecutter = staticmethod(cookiecutter)
     requests = requests
 
--- a/src/briefcase/integrations/base.py
+++ b/src/briefcase/integrations/base.py
@@ -8,32 +8,36 @@ from abc import ABC
 from collections import defaultdict
 from collections.abc import Mapping
 from pathlib import Path
-from typing import TYPE_CHECKING, DefaultDict
+from types import ModuleType
+from typing import TYPE_CHECKING, DefaultDict, TypeVar, Union
 
 import requests
 from cookiecutter.main import cookiecutter
 
 from briefcase.config import AppConfig
 from briefcase.console import Console, Log
-from briefcase.integrations.download import Download
-from briefcase.integrations.subprocess import Subprocess
 
-if TYPE_CHECKING:  # pragma: no cover
+if TYPE_CHECKING:
     import git as _git
 
     from briefcase.integrations.android_sdk import AndroidSDK
     from briefcase.integrations.docker import Docker, DockerAppContext
+    from briefcase.integrations.download import Download
     from briefcase.integrations.flatpak import Flatpak
     from briefcase.integrations.java import JDK
     from briefcase.integrations.linuxdeploy import LinuxDeploy
     from briefcase.integrations.rcedit import RCEdit
+    from briefcase.integrations.subprocess import Subprocess
     from briefcase.integrations.visualstudio import VisualStudio
-    from briefcase.integrations.wix import Wix
+    from briefcase.integrations.wix import WiX
 
 
 # TODO: Implement Tool base class
 class Tool(ABC):
     """Tool Base."""  # pragma: no cover
+
+
+TOOL_T = TypeVar("TOOL_T", bound=Union[Tool, ModuleType])
 
 
 class ToolCache(Mapping):
@@ -49,7 +53,7 @@ class ToolCache(Mapping):
     rcedit: RCEdit
     subprocess: Subprocess
     visualstudio: VisualStudio
-    wix: Wix
+    wix: WiX
     xcode: bool
     xcode_cli: bool
 
@@ -101,9 +105,16 @@ class ToolCache(Mapping):
             )
         )
 
-        # Built-in tools without any external dependencies
-        Subprocess.verify(tools=self)
-        Download.verify(tools=self)
+    def add_tool(self, name: str, tool: TOOL_T) -> TOOL_T:
+        """Add a tool to the cache.
+
+        :param name: Name of the tool to add to the cache; The tool will be
+            accessible as an attribute of the ToolCache using this name.
+        :param tool: The instantiated Tool to make available.
+        :return: The added Tool.
+        """
+        setattr(self, name, tool)
+        return tool
 
     def __getitem__(self, app: AppConfig) -> ToolCache:
         return self.app_tools[app]
--- a/tests/commands/base/test_verify_tools.py
+++ b/tests/commands/base/test_verify_tools.py
@@ -1,3 +1,10 @@
+from briefcase.integrations.download import Download
+from briefcase.integrations.subprocess import Subprocess
+
+
 def test_base_tools_exist(base_command):
-    """No verification required for base command."""
+    """Ensure default tools are always available."""
+    assert isinstance(base_command.tools.subprocess, Subprocess)
+    assert isinstance(base_command.tools.download, Download)
+
     base_command.verify_tools()
--- a/tests/integrations/base/test_ToolCache.py
+++ b/tests/integrations/base/test_ToolCache.py
@@ -10,8 +10,6 @@ from cookiecutter.main import cookiecutter
 
 from briefcase.console import Console, Log
 from briefcase.integrations.base import ToolCache
-from briefcase.integrations.download import Download
-from briefcase.integrations.subprocess import Subprocess
 
 
 @pytest.fixture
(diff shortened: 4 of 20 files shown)


rmartin16:
## Explanation of the Hoop Jumping
The primary problem with typing for `ToolCache` is circular imports. To avoid this, we have to make sure none of the typing is evaluated at runtime.

The `from __future__ import annotations` import makes all typing annotations be interpreted as strings. So `android_sdk: AndroidSDK` is interpreted as `android_sdk: "AndroidSDK"`, for example. This avoids the requirement to actually import these definitions during runtime just for typing.

The `if TYPE_CHECKING: <do imports>` prevents actually doing the imports when Briefcase runs but allows type checkers and IDE's know what the types are. 

Feedback is welcome.

freakboy3742 (commented):
FWIW: Visual Studio will make use of type annotations if they're available, but doesn't get bent out of shape if they're missing. Anecdotally, my observation is that PyCharm is a lot more enthusiastic (or noisy, your pick 😄 ) about them.

I'm weakly in favour of adding type annotations - they can be good documentation, and as the codebase grows, they're another piece reinforcing correctness. My concern is the extent to which they become onerous to work with - adding a simple type declaration is mostly harmless; but requiring `Optional[List[Dict[Tuple(...` makes code harder to write, and I'm not sure it really improves readability, either.

The other thing to consider is whether we're treating this as "light documentation assistance", or actual type correctness. If we're going to pursue type annotation much further, I'd like to see some CI validation of the type declarations - there's not much point having them if they're wrong :-) My concern on this front is whether adding mypy on a partially typed codebase is going to cause more headaches than benefits; and whether the type of dynamic attribute management that we're doing with `AppConfig` is going to lead to contributors playing mypy whack-a-mole with every contribution.

rmartin16:
@freakboy3742,
My primary impetus for this change is to avoid the common workflow I have now. For instance, I want to use a method from `os` but I can't remember its name. So I literally have to write `import os; os.<blah>` to find it, and then I can use it with `self.tools.os`. This extrapolates to everything in `ToolCache` though because nothing can have it types inferred.

This is purely efficiency for me.

As for actually enforcing typing.....I'm less interested in that goal....or in achieving some sort of typing coverage. To that end, I have been experimenting with typing for `Command` and `AppConfig`. I haven't landed on a good strategy yet though that isn't over-onerous to maintain.

freakboy3742 (commented):
That makes sense; and it's certainly a lot less onerous to add typing in a few places where it's helpful than to do a wholesale type strictness change.

However, the question about long term maintenance still holds. What (if anything) can we do to ensure these type declarations are accurate, and stay that way? In particular, it makes me a little twitchy that we need to add chunks of code that are marked `#pragma nocover` just to keep an IDE happy; looking forward, if a new tool is added, there's nothing other than goodwill to ensure that any additional type declarations are added.

That's not necessarily a blocker from my perspective; I just want to make sure we've exhausted any possible safety mechanisms first.

pyproject.toml
```diff
@@ -22,7 +22,8 @@ precision = 1
 exclude_lines = [
     "pragma: no cover",
     "@(abc\\.)?abstractmethod",
-    "NotImplementedError\\(\\)"
+    "NotImplementedError\\(\\)",
+    "if TYPE_CHECKING:",
```
rmartin16, line 26: I added this after discovering Ned [said](https://github.com/nedbat/coveragepy/issues/831#issuecomment-1020479593) it should be part of the default `exclude`.

I'll revert back to `# pragma: no cover` if you're not a fan.

rmartin16:
At first I was thinking testing these annotations is probably overkill....but then I discovered I spelled `WiX` wrong and quickly rethought my position 🤦🏼 

This sent me down quite a few rabbit holes before getting here.
- I was able to reduce `TYPE_CHECKING` use to only `integrations.base` to prevent circular import errors.
- There are now tests to check if the types listed in `ToolCache` are correct.
- Tests will also catch if a tool is added to `ToolCache` without it being typed.
- I took the first step towards implementing the `Tool` base class....which is something I'd like to finish in the soonish future.
- To avoid a lot of hassle with `Download` and `Subprocess`, I moved their initialization out of `ToolCache` and in to `BaseCommand`.

src/briefcase/integrations/base.py
```diff
@@ -8,28 +8,62 @@
...
+    from briefcase.integrations.visualstudio import VisualStudio
+    from briefcase.integrations.wix import WiX
 
 
 # TODO: Implement Tool base class
 class Tool(ABC):
     """Tool Base."""  # pragma: no cover
 
 
+TOOL_T = TypeVar("TOOL_T", bound=Union[Tool, ModuleType])
```
freakboy3742, line 40: I presume this is an interim measure until we can convert tools like Xcode into full Tool classes?

src/briefcase/integrations/base.py
```diff
@@ -71,9 +105,16 @@ def __init__(
...
+    def add_tool(self, name: str, tool: TOOL_T) -> TOOL_T:
+        """Add a tool to the cache.
+
+        :param name: Name of the tool to add to the cache; The tool will be
+            accessible as an attribute of the ToolCache using this name.
+        :param tool: The instantiated Tool to make available.
+        :return: The added Tool.
+        """
+        setattr(self, name, tool)
+        return tool
```
freakboy3742, line 117: This strikes me as an odd construction - is there a reason for using this instead of the previous `tools.mytool = mytool; return mytool`? Unless we're planning to do additional verification on tools when they're added, I'm not sure I see what we gain over simple attribute assignment.

tests/commands/base/test_verify_tools.py
```diff
@@ -1,3 +1,10 @@
+from briefcase.integrations.download import Download
+from briefcase.integrations.subprocess import Subprocess
+
+
 def test_base_tools_exist(base_command):
-    """No verification required for base command."""
+    """Ensure default tools are always available."""
+    assert isinstance(base_command.tools.subprocess, Subprocess)
+    assert isinstance(base_command.tools.download, Download)
```
freakboy3742, line 8: This test has lost the check of `app` binding in the port from test_ToolCache.py - it looks like that might also be a side effect of moving the registration to base. Was this intentional? Is it a source of concern that the app subclasses don't inherit these tools any more?

freakboy3742 (changes requested):
These mostly look like good improvements; a couple of questions/clarifications inline.

I'm not sure the tests are really testing that much, though. They're all tests of strings - so all we're really validating is that we're making the same typo in multiple locations :-) The only way I can think of improving this would be do walk over `briefcase.integrations.__dict__` - that would be a list of actual module names, so `__name__` would be reliably accurate.

src/briefcase/integrations/base.py
rmartin16: Not necessarily. This idiom allows IDEs (and type checkers) to track a specific type being used in a higher level function that may operate on many types. In this case, it explicitly hints that the return value of `ToolCache.add_tool(name=..., tool=...)` will be the same as the type passed as `tool`.

Furthermore, though, `ModuleType` is necessary because `tools.git` is not a class; it is an actual module. Right now, `TOOL_T` isn't even accommodating `xcode` and `xcode_cli`.

src/briefcase/integrations/base.py
rmartin16: This is just for evaluating things added as tools during testing so every tool could be confirmed as typed in `ToolCache`. It doesn't serve a functional purpose otherwise.

rmartin16:
> These mostly look like good improvements; a couple of questions/clarifications inline.
> 
> I'm not sure the tests are really testing that much, though. They're all tests of strings - so all we're really validating is that we're making the same typo in multiple locations :-) The only way I can think of improving this would be do walk over `briefcase.integrations.__dict__` - that would be a list of actual module names, so `__name__` would be reliably accurate.

We're a bit on the verge of rolling our own type checker... It may be possible to iterate though the namespace of `briefcase.integrations` but there are going to be a lot of caveats to which classes are relevant and which are not.

I think given the low volume and low churn of tools, we can confirm this typing is accurate; if not, I know PyCharm will definitely let people know.

While `ToolCache.add_tool()` is basically superfluous....I couldn't think of another way to 100% identify tools that aren't typed in `ToolCache` when they should be. I'm concerned that introspection of classes in `briefcase.integrations` could lead to false positives and largely irrelevant noise.

I'm not quite sure how far unaligned we may be...but I'm definitely willing to get on a call if it'd be at all useful to hash out these specifics as well as larger perspectives about typing. Like I mentioned, I would also like to institute a bit more typing for `AppConfig` and `Command`....but only if it fits with the within a reasonable maintenance lift.

src/briefcase/integrations/base.py
freakboy3742: I'm not sure I follow how that validation works (or how it's a guarantee) - could you elaborate? 

On top of that - my concern would be that `add_tool` is sufficiently un-idiomatic that the next tool added will inadvertently use `tool.mytool = ...`, and as a result, won't get picked up in testing.

freakboy3742:
> > These mostly look like good improvements; a couple of questions/clarifications inline.
> > I'm not sure the tests are really testing that much, though. They're all tests of strings - so all we're really validating is that we're making the same typo in multiple locations :-) The only way I can think of improving this would be do walk over `briefcase.integrations.__dict__` - that would be a list of actual module names, so `__name__` would be reliably accurate.
> 
> We're a bit on the verge of rolling our own type checker... It may be possible to iterate though the namespace of `briefcase.integrations` but there are going to be a lot of caveats to which classes are relevant and which are not.

To be clear - I definitely don't want to go down the path of rolling our own. The `__dict__` suggestion was mostly spitballing on my part; if there was an easy way to use "actual types" in the tests, rather than strings, then it seemed like an easy win. However, if there's no easy win, then I'm definitely not going to advocate for building massive amounts of tooling to validate a list of 10 strings that won't change that often.

> I think given the low volume and low churn of tools, we can confirm this typing is accurate; if not, I know PyCharm will definitely let people know.

My concern is that something is committed to main incorrect, but it won't get picked up until someone with a specific editor tries to use the code. Given how often that list will change, maybe eyeballing it is enough.

> I'm not quite sure how far unaligned we may be...but I'm definitely willing to get on a call if it'd be at all useful to hash out these specifics as well as larger perspectives about typing. Like I mentioned, I would also like to institute a bit more typing for `AppConfig` and `Command`....but only if it fits with the within a reasonable maintenance lift.

I don't think we're *that* far misaligned - I'm mostly pushing around the edges of this because (a) I don't use PyCharm, so I'm not experiencing the underlying problem that is motivating this change, and (b) generally speaking, I'm not a fan of changes that can't be (or aren't) validated by automated testing. 

That said - practicality also beats purity. At this point, the only things that really stand out to me are the big "ignored" import block in `integrations/base.py`, and the architectural oddity of `add_tools()`. I can live with the former; and if the latter is a case of accepting an architectural oddity and getting weak testing, or not getting testing, then honestly, if having some additional types will improve the ergonomics for contributors using PyCharm, and it won't become an active land mine for everyone else, then I think I can live without the testing aspect.

However, if you want to catch up to talk through your plans about typing, I'm happy to jump on a call - grab me on Discord whenever you see me; or, if you want to document them on a ticket, I'm happy to respond there too.

src/briefcase/integrations/base.py
rmartin16: I created `ProxyToolCache` in `tests.integrations.conftest` to inspect if tools added through `ToolCache.add_tool()` are annotated in `ToolCache`.

I agree it is straightforward to skip using this wrapper altogether; it would ultimately depend on devs and the reviewers enforcing it.

For the record, I'm not a big fan either of this either given the architecture of `ToolCache`....I just couldn't think of anything else to try to meet your reasonable comment below:

> looking forward, if a new tool is added, there's nothing other than goodwill to ensure that any additional type declarations are added.

rmartin16:
> I don't use PyCharm, so I'm not experiencing the underlying problem that is motivating this change

Out of curiosity, does VS provide typing for tools in `ToolCache`? For instance, does VS know what methods (and their args) are available on `self.tools.subprocess`?

At the end of the day, that's the problem I'd really like to solve here.

> the big "ignored" import block in `integrations/base.py`

Agree; I went in to a lot of rabbit holes today trying to find a way to do these imports _with_ coverage. (Although, while the imports don't have coverage, the code also never actually executes either.) Unfortunately, it proved impossible to get coverage without circular import errors....unless I guess we put all the tools and the `ToolCache` in the same file....or something.

> the architectural oddity of add_tools()

Also agree; I can remove `add_tools()`.

Finally, do you see any value in any of the other "testing" I added? I think that's just limited to `test_toolcache_typing` which only verifies we made the same typo in both locations ;)

rmartin16:
I removed `ToolCache.add_tools()` and updated the type testing to be much more dynamic. Feel free to say its too heavy headed and we're probably better off without it. It further reinforces that the name of the tool in the `ToolCache` must be the value of `Tool.name` as well as the name of the module. It's intended to ensure all Tools are annotated and any annotations are accurate insofar as they can be resolved to existing classes.

tests/integrations/base/test_ToolCache.py
```diff
@@ -23,6 +24,62 @@ def simple_tools(tmp_path):
...
+    # Tool names to exclude from the dynamic annotation checks; they are manually checked.
+    tool_names_skip_dynamic_check = {"app_context", "git", "xcode", "xcode_cli"}
+    # Tool classes to exclude from dynamic annotation checks.
+    tool_klasses_skip_dynamic_checks = {"DockerAppContext", "NativeAppContext"}
+
+    # Ensure defined Tool modules/classes are annotated in ToolCache.
+    for tool_module_name in briefcase.integrations.__all__:
+        if tool_module_name not in tools_unannotated:
+            assert tool_module_name in ToolCache.__annotations__.keys()
+        for tool_name in tools_for_module(tool_module_name):
```
freakboy3742, line 54: It's not a huge deal, but `tools_for_module` is a moderately complex method with an answer that shouldn't change; both of the invocations are occurring inside for loops. Is there any reason we shouldn't pre-compute once and store, rather than re-computing every time?

src/briefcase/integrations/base.py
```diff
@@ -8,15 +8,29 @@
...
 
 from briefcase.config import AppConfig
 from briefcase.console import Console, Log
-from briefcase.integrations.download import Download
-from briefcase.integrations.subprocess import Subprocess
+
+if TYPE_CHECKING:
+    # Tools are imported only for type checking
+    # to avoid circular import errors.
+    import git as git_
```
freakboy3742, line 22: I was almost ready to approve this PR... and then I [remembered something](https://github.com/beeware/briefcase/pull/964/files#diff-d705b8400e278985802862242e3ab69651e42f2317d9ef7f4351428ccac4a751R8-R17).

That opens up a bunch of questions:
1. On what condition will `TYPE_CHECKING` be true?
2. Is it possible for the user to have `TYPE_CHECKING` be true and *not* have a working git install?
3. Is there any meaningful fallback/error recovery to use in this case?

freakboy3742 (changes requested):
I think we're getting close; one minor cleanup and one ponderous question inline.

tests/integrations/base/test_ToolCache.py
rmartin16: I avoided caching mostly for simplicity. Although, `tools_for_module()` should only be called twice for any particular module.

I can add the `functools.lru_cache` decorator to `tools_for_module()` to get caching for "free" though.

freakboy3742: That's a bit more complexity than is warranted here. This isn't a big deal; I won't go to the mat over 2 calls to a test function that isn't *that* complex.

src/briefcase/integrations/base.py
rmartin16: > I was almost ready to approve this PR... and then I [remembered something](https://github.com/beeware/briefcase/pull/964/files#diff-d705b8400e278985802862242e3ab69651e42f2317d9ef7f4351428ccac4a751R8-R17).
> 
> That opens up a bunch of questions:
> 
> 1. On what condition will `TYPE_CHECKING` be true?

`TYPE_CHECKING` probably shouldn't ever be `True`; even for type checkers like mypy, they should only [assume](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) `TYPE_CHECKING` is the same as `True`.

> 2. Is it possible for the user to have `TYPE_CHECKING` be true and _not_ have a working git install?

It's probably possible if something monkeypatched `typing.TYPE_CHECKING` at interpreter startup. But that definitely isn't the intended way to "assume" `TYPE_CHECKING` is `True` when relevant. I really hope the stronger promise of the design of `TYPE_CHECKING` is that it is always `False` at runtime.

> 3. Is there any meaningful fallback/error recovery to use in this case?

Not currently. But if we wanted to ensure this was avoided, we could change `if TYPE_CHECKING:` to simply `if False:`. Although, for instance, `pytest` explicitly [moved](https://github.com/pytest-dev/pytest/pull/6435) from `if False:` once Python 3.5 could use `typing.TYPE_CHECKING`.

freakboy3742 (approved):
If someone is monkey patching an internal variable like `TYPE_CHECKING`, I think we're firmly in "you get to keep all the shiny pieces" territory.

On that basis, I think I'm happy with this; if we ever enable mypy as part of the CI checks, we might need to revisit that, but I'm happy to call that a problem for Future Apiarists.

PR #965, opened 2022-11-09
Add Python 3.12 support

freakboy3742:
With the release of 3.12.0a1, we can start testing for 3.12 support.

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,8 +68,13 @@ jobs:
     strategy:
       matrix:
         platform: ["macos", "ubuntu", "windows"]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12.0-alpha - 3.12.0"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12-dev"]
+        experimental: [false]
+        include:
+          - python-version: "3.12-dev"
+            experimental: true
     runs-on: ${{ matrix.platform }}-latest
+    continue-on-error: ${{ matrix.experimental }}
     steps:
     - uses: actions/checkout@v3.1.0
       with:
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -20,6 +20,7 @@ classifiers =
     Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
     Programming Language :: Python :: 3.11
+    Programming Language :: Python :: 3.12
```
mhsmith, line 23: By declaring this. we're committing to fixing any Python 3.12 compatibility issues before each Briefcase release from now on, which should be manageable. But by adding it to CI the same way as the other versions, we would be forced to fix any Python 3.12 issues *immediately*, which could be inconvenient, and isn't necessary so early in the 3.12 release cycle.

Can we use the `continue-on-error` setting as shown [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-preventing-a-specific-failing-matrix-job-from-failing-a-workflow-run) to make Python 3.12 failures non-fatal for now? Then we can make them fatal next year when it reaches RC status.

.github/workflows/ci.yml
```diff
@@ -67,8 +67,8 @@ jobs:
     needs: [pre-commit, towncrier, package]
     strategy:
       matrix:
-        platform: ['macos', 'ubuntu', 'windows']
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        platform: ["macos", "ubuntu", "windows"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12.0-alpha - 3.12.0"]
```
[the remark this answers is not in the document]
freakboy3742: `-dev` definitely makes sense for 3.12; I wasn't aware that syntax was an option. 

I'm not convinced we want to adopt it for established releases, though. Looking at the tagged releases of Python, RCs of Python micro releases aren't *that* common; using `3.8-dev` could communicate the impression that we're doing some sort of pre-release branch testing, which would only be the case if someone pushed a PR during the very limited window that an RC was available prior to the next final release.

setup.cfg
freakboy3742: Good call. Fixed.

mhsmith (approved):

PR #966, opened 2022-11-10
Convert Android to use log streaming, and handle fast fail

### LINKED ISSUE #936: Fast-failing Android app hangs at "Launching app..."
If an Android app fails extremely quickly, Briefcase will hang at the `Launching app...` waitbar.

**To Reproduce**
1. Create a Hello World Toga app.
2. Modify pyproject.toml to *not* have `toga-android` as a requirement.
3. Build and run the app on Android

The app will build and attempt to start, but will hang at the `Launching app...` waitbar.

If you hit CTRL-C and start the app a second time, you get the stack trace that indicates that `import toga` failed.

**Expected behavior**

The `Launching app...` waitbar shouldn't hang. Ideally, you'd get the partial log to highlight the problem; but at the very least it should indicate that the launch failed.

**Environment:**
 - Operating System: macOS
 - Python version: All
 - Software versions:
   - Briefcase: 0.3.11+
   - Toga: 0.3.0dev39+

**Additional context**

I found this when running a Briefcase app that was using Toga installed from a source checkout; for some reason, the wheels were built, but they didn't appear in the `app` folder. Not sure why that was happening; I'm still investigating that problem.

freakboy3742:
Previously, Android used a blocking call to `logcat` with a PID to show logs to the user. This works well for the normal case of app operation; however, it has 2 notable problems:

1. If the Android process crashes or exits, the log stream continues.
2. If the PID can't be identified, a log stream can't be started at all.

(2) is a particular problem if the app fails/exits quickly. The current implementation tries to use `pidof()` immediately, but this can fail because the process hasn't started yet; for this reason, Briefcase implements a polling loop looking for the PID. However if the process starts, but exits before the next call to `pidof()`, Briefcase hangs, eternally waiting for a PID that only existed during a brief window when Briefcase wasn't looking for it. 

The original report for #936 saw this because the app itself had a syntax error in the code; if the app didn't *immediately* return a PID (a reasonably common, but unpredictable occurrence), it would crash before the next check for the PID. However, it's also a common case for test suites, which are (by definition) programs that run and exit quickly. Based on the current implementation, if the test suite takes <0.5s to execute, the test suite app will appear to crash on app startup.

This PR makes several changes to address this:

1. It reduces the delay period before the next PID check from 0.5s to 0.01s.
2. It caps the waiting period for a PID to 5 seconds. If a PID isn't available 5 seconds after the call to start the app, the app probably isn't going to start.
3. If a PID can't be found, a time-based logcat is displayed (showing all log messages from 10s before the start time of the app - this window is to allow for clock discrepancies between the emulator and the machine), and Briefcase exits with an error.
4. If a PID is found, the call to logcat is passed through a streamer. This allows us to install a `stop_func` that interrogates ADB to see if the PID is still active. If the PID exits, the log stream is terminated.
 
(4) is also required to support #963 `--test` mode. 

The streaming stop function and the increased frequency of the pidof call result in a lot of calls to subprocess being made to implement polling behaviour, which results in a lot of noise in the logs. To prevent this noise, a `stealth` mode has been added to `subprocess.check_output()` - this silences the logging of the call itself.

Fixes #936.

## PR Checklist:


src/briefcase/platforms/android/gradle.py
```diff
@@ -229,22 +230,51 @@ def run_app(self, app: BaseConfig, device_or_avd=None, **kwargs):
...
+            # Any log after this point must be associated with the new instance
+            # Pad by a few seconds because the android emulator's clock and the
+            # local system clock may not be perfectly aligned.
+            start_time = datetime.datetime.now() - datetime.timedelta(seconds=10)
             adb.start_app(package, "org.beeware.android.MainActivity")
             pid = None
-            while not pid:
-                pid = adb.pidof(package)
+            attempts = 0
+            delay = 0.01
```
mhsmith, line 240: > However, it's also a common case for test suites, which are (by definition) programs that run and exit quickly. Based on the current implementation, if the test suite takes <0.5s to execute, the test suite app will appear to crash on app startup.

But on Android, the app process won't actually exit when the test suite ends, unless we explicitly make it do so. And if the app does crash during startup, Briefcase will timeout looking for the PID and the user will see the crash report anyway. So maybe reducing the pidof delay is unnecessary, in which case stealth mode is unnecessary here as well.

src/briefcase/platforms/android/gradle.py
```diff
@@ -229,22 +230,51 @@ def run_app(self, app: BaseConfig, device_or_avd=None, **kwargs):
...
+            while not pid and attempts < 500:
```
mhsmith, line 241: This should check the elapsed time rather than count the attempts, because running adb might take significantly longer than 0.01 seconds.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1144,14 +1144,19 @@ def has_booted(self):
                 f"Unable to determine if emulator {self.device} has booted."
             ) from e
 
-    def run(self, *arguments):
+    def run(self, *arguments, stealth=False):
```
mhsmith, line 1147: I think `quiet` would be a clearer name than `stealth`.

And Instead of repeating its default value 3 times in this file, how about making `pidof` and `pid_exists` take a `**kwargs` which they pass to `run`, and the same between `run` and `subprocess.check_output`?

src/briefcase/platforms/android/gradle.py
freakboy3742: So, in my example, the test *was* exiting, but that may be a problem of my own making. I've been using a test script that calls `sys.exit(result)` on completion of the test suite for the benefit of Linux/Windows, which don't currently do any streaming; however, if we adopt the filtering+regex approach calling `sys.exit()` may not be necessary. FWIW, here's what I was seeing:

<details>
<subject>Log of a running test suite</subject>

```
[togatest] Following device log output (type CTRL-C to stop log)...
--------- beginning of main
11-11 11:59:54.327 10523 10523 I xample.togates: Late-enabling -Xcheck:jni
11-11 11:59:54.389 10523 10523 D CompatibilityChangeReporter: Compat change id reported: 171979766; UID 10124; state: ENABLED
11-11 11:59:54.416 10523 10523 V GraphicsEnvironment: ANGLE Developer option for 'com.example.togatest' set to: 'default'
11-11 11:59:54.416 10523 10523 V GraphicsEnvironment: Neither updatable production driver nor prerelease driver is supported.
11-11 11:59:54.417 10523 10523 D NetworkSecurityConfig: Using Network Security Config from resource network_security_config debugBuild: true
11-11 11:59:54.418 10523 10523 D NetworkSecurityConfig: Using Network Security Config from resource network_security_config debugBuild: true
11-11 11:59:54.422 10523 10549 D libEGL  : loaded /vendor/lib64/egl/libEGL_emulation.so
11-11 11:59:54.422 10523 10549 D libEGL  : loaded /vendor/lib64/egl/libGLESv1_CM_emulation.so
11-11 11:59:54.423 10523 10549 D libEGL  : loaded /vendor/lib64/egl/libGLESv2_emulation.so
11-11 11:59:54.432 10523 10523 D MainActivity: onCreate() start
11-11 11:59:54.435 10523 10523 D MainActivity: onCreate(): captured stdout/stderr
11-11 11:59:54.435 10523 10550 D stdio   : Starting to capture stdout/stderr to Android log.
11-11 11:59:54.435 10523 10550 D stdio   : stdout now successfully routes into the stdio logger.
11-11 11:59:54.458 10523 10523 W xample.togates: Accessing hidden method Landroid/view/View;->computeFitSystemWindows(Landroid/graphics/Rect;Landroid/graphics/Rect;)Z (unsupported, reflection, allowed)
11-11 11:59:54.458 10523 10523 W xample.togates: Accessing hidden method Landroid/view/ViewGroup;->makeOptionalFitsSystemWindows()V (unsupported, reflection, allowed)
11-11 11:59:54.463 10523 10523 D MainActivity: Starting Python
11-11 11:59:54.501 10523 10550 D stdio   : Could not find platform independent libraries <prefix>
11-11 11:59:54.501 10523 10550 D stdio   : Could not find platform dependent libraries <exec_prefix>
11-11 11:59:54.501 10523 10550 D stdio   : Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
11-11 11:59:54.596 10523 10523 W asset   : seek out of range: want -20, end=22
11-11 11:59:54.625 10523 10523 D MainActivity: Running main module tests.togatest
11-11 11:59:54.741 10523 10523 I python.stdout: ============================= test session starts ==============================
11-11 11:59:54.741 10523 10523 I python.stdout:  
11-11 11:59:54.741 10523 10523 I python.stdout: platform linux -- Python 3.10.6, pytest-7.2.0, pluggy-1.0.0 -- /system/bin/app_process64
11-11 11:59:54.741 10523 10523 I python.stdout:  
11-11 11:59:54.741 10523 10523 I python.stdout: cachedir: .pytest_cache
11-11 11:59:54.741 10523 10523 I python.stdout:  
11-11 11:59:54.741 10523 10523 I python.stdout: rootdir: /data/data/com.example.togatest/files/chaquopy/AssetFinder/app
11-11 11:59:54.741 10523 10523 I python.stdout:  
11-11 11:59:54.741 10523 10523 I python.stdout: collecting ... 
11-11 11:59:54.741 10523 10523 I python.stdout: collected 0 items
11-11 11:59:54.741 10523 10523 I python.stdout:  
11-11 11:59:54.742 10523 10523 I python.stdout:  
11-11 11:59:54.742 10523 10523 I python.stdout: ============================ no tests ran in 0.00s =============================
11-11 11:59:54.742 10523 10523 I python.stdout:  
11-11 11:59:54.742 10523 10523 W python.stderr: ERROR: file or directory not found: tests
11-11 11:59:54.742 10523 10523 W python.stderr:  
11-11 11:59:54.742 10523 10523 I python.stdout: >>>>>>>>>> Test Suite Failed <<<<<<<<<<
11-11 11:59:54.742 10523 10523 I python.stdout:  
11-11 11:59:54.745 10523 10523 D AndroidRuntime: Shutting down VM
--------- beginning of crash
11-11 11:59:54.747 10523 10523 E AndroidRuntime: FATAL EXCEPTION: main
11-11 11:59:54.747 10523 10523 E AndroidRuntime: Process: com.example.togatest, PID: 10523
11-11 11:59:54.747 10523 10523 E AndroidRuntime: java.lang.RuntimeException: Unable to start activity ComponentInfo{com.example.togatest/org.beeware.android.MainActivity}: com.chaquo.python.PyException: SystemExit: ExitCode.USAGE_ERROR
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3635)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3792)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:103)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2210)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.os.Handler.dispatchMessage(Handler.java:106)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.os.Looper.loopOnce(Looper.java:201)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.os.Looper.loop(Looper.java:288)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.app.ActivityThread.main(ActivityThread.java:7839)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at java.lang.reflect.Method.invoke(Native Method)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
11-11 11:59:54.747 10523 10523 E AndroidRuntime: Caused by: com.chaquo.python.PyException: SystemExit: ExitCode.USAGE_ERROR
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at <python>.__main__.<module>(togatest.py:15)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at <python>.runpy._run_code(runpy.py:86)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at <python>.runpy._run_module_code(runpy.py:96)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at <python>.runpy.run_module(runpy.py:224)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at <python>.chaquopy_java.call(chaquopy_java.pyx:379)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at <python>.chaquopy_java.Java_com_chaquo_python_PyObject_callAttrThrowsNative(chaquopy_java.pyx:351)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at com.chaquo.python.PyObject.callAttrThrowsNative(Native Method)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at com.chaquo.python.PyObject.callAttrThrows(PyObject.java:232)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at com.chaquo.python.PyObject.callAttr(PyObject.java:221)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at org.beeware.android.MainActivity.onCreate(MainActivity.java:64)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.app.Activity.performCreate(Activity.java:8051)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.app.Activity.performCreate(Activity.java:8031)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1329)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3608)
11-11 11:59:54.747 10523 10523 E AndroidRuntime:        ... 12 more
11-11 11:59:54.770 10523 10523 I Process : Sending signal. PID: 10523 SIG: 9
```

</details>

The test code is printing the sentinel, then invoking `sys.exit(result)`; the tests themselves are failing because of path issues, but the test harness is running, failing, then exiting. This is crashing the main process, and it's taking less than 0.5s to do it. 

However, more generally, I think the stealth+fast polling is still worth it. 

If we revert to 0.5s polling, then the 5s timeout becomes the expected behavior for having a SyntaxError in your code. By polling faster, we substantially increase the likelihood that we can proceed in log streaming mode - which means the 5 second delay waiting for polling to fail *doesn't* happen - which is a much nicer development experience IMHO.

From my testing, there's only ever a handful of polling attempts, even at the 0.01s repeat rate. The 5s limit is only required as a safety mechanism to prevent the previous behavior of "infinite timeout waiting for PID". With faster polling, it's highly unlikely that you'll ever hit this limit, unless something is badly broken. I suspect that if you actually *hit* the 5s limit, it's entirely likely there's nothing in the app log anyway, unless you're on a *pathologically* slow machine/emulator.

src/briefcase/integrations/android_sdk.py
freakboy3742: /me facepalms. I cycled through a bunch of names trying to come up with one that felt right, and `quiet` was sitting right there.

The reorganisation of the kwargs is also a good call; done.

src/briefcase/platforms/android/gradle.py
mhsmith: > which is a much nicer development experience IMHO

Good point, and it also means that the stack trace will be the last thing in the log output, rather than random stuff from other apps.

mhsmith (approved):
Works great on Windows with both an emulator and a physical device.

Briefcase exiting automatically when the app process terminates is a nice touch. Since starting processes is relatively expensive on Windows, I was concerned that checking for termination 10 times a second could increase CPU usage, but it only goes up by about 2%, most of which seems to be in the anti-virus system.

PR #967, opened 2022-11-11
Convert Linux backends to use app log streaming

freakboy3742:
In order to support test mode (Refs #962), we need to convert app runners to use streamers, rather than just outputting logs directly to stdout. To make the code easier to review, I've split the introduction of streaming to the linux backends as a standalone PR, similar to what has been done with #966 for Android.

## PR Checklist:

### DIFF
--- a/src/briefcase/platforms/linux/appimage.py
+++ b/src/briefcase/platforms/linux/appimage.py
@@ -278,6 +278,9 @@ class LinuxAppImageRunCommand(LinuxAppImagePassiveMixin, RunCommand):
             finally:
                 self.tools.subprocess.cleanup(app.app_name, log_popen)
 
+            # If the process didn't exit cleanly, raise an error.
+            if log_popen.returncode != 0:
+                raise BriefcaseCommandError(f"Problem running app {app.app_name}.")
         except KeyboardInterrupt:
             pass  # Catch CTRL-C to exit normally
         except OSError as e:
--- a/src/briefcase/platforms/linux/appimage.py
+++ b/src/briefcase/platforms/linux/appimage.py
@@ -270,15 +270,10 @@ class LinuxAppImageRunCommand(LinuxAppImagePassiveMixin, RunCommand):
 
             try:
                 # Start streaming logs for the app.
-                self.logger.info(
-                    "Following log output (type CTRL-C to stop log)...",
-                    prefix=app.app_name,
-                )
                 self.logger.info("=" * 75)
                 self.tools.subprocess.stream_output(
                     app.app_name,
                     log_popen,
-                    stop_func=lambda: log_popen.poll() is not None,
                 )
             finally:
                 self.tools.subprocess.cleanup(app.app_name, log_popen)
--- a/src/briefcase/platforms/linux/appimage.py
+++ b/src/briefcase/platforms/linux/appimage.py
@@ -276,22 +276,12 @@ class LinuxAppImageRunCommand(LinuxAppImagePassiveMixin, RunCommand):
                 )
                 self.logger.info("=" * 75)
                 self.tools.subprocess.stream_output(
-                    "log stream",
+                    app.app_name,
                     log_popen,
                     stop_func=lambda: log_popen.poll() is not None,
                 )
             finally:
-                # If the app hasn't exited, ensure it is terminated
-                # This is unlikely to happen unless the streamer stops
-                # for some reason, but clean up just in case.
-                return_code = log_popen.poll()
-                if return_code is None:
-                    log_popen.terminate()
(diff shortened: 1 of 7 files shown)


src/briefcase/platforms/linux/appimage.py
```diff
@@ -259,16 +259,33 @@ def run_app(self, app: AppConfig, **kwargs):
...
+                self.logger.info(
+                    "Following log output (type CTRL-C to stop log)...",
+                    prefix=app.app_name,
+                )
+                self.logger.info("=" * 75)
+                self.tools.subprocess.stream_output(
+                    app.app_name,
+                    log_popen,
+                    stop_func=lambda: log_popen.poll() is not None,
+                )
```
rmartin16, line 282: curious....did you find the `stop_func` was necessary for log streaming to exit properly?

I'd expect checking `poll()` is functionally equivalent to the log streamer exiting when `readline()` returns an empty string....since that'll happen once the PIPE is closed...and that'll happen when the underlying process exits.

src/briefcase/platforms/linux/appimage.py
```diff
@@ -259,16 +259,33 @@ def run_app(self, app: AppConfig, **kwargs):
...
                 cwd=self.tools.home_path,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                bufsize=1,
             )
+
+            try:
+                # Start streaming logs for the app.
+                self.logger.info(
+                    "Following log output (type CTRL-C to stop log)...",
```
rmartin16, line 274: This also appears to be adding back in the "send CTRL-C to exit" messages that were [removed](https://github.com/beeware/briefcase/pull/921#issuecomment-1281159124) in #921.

src/briefcase/platforms/linux/appimage.py
freakboy3742: Good point - this was mostly done for consistency with the iOS/macOS/etc backends, but you're right - as we're streaming the actual process, and the actual process will exit, there's no need to poll here.

src/briefcase/platforms/linux/appimage.py
freakboy3742: Oh yeah - I forgot we made a conscious decision to pull those out.

mhsmith (approved):

rmartin16 (commented):
In general, I think I have concerns about how much of work that's typically been encapsulated inside `Subprocess` is now being implemented in platform code.

More or less, this re-implements `Subprocess._run_and_stream_output()` in multiple places....but really `_run_and_stream_output()` is just an implementation of `Subprocess.run()`.

**Would it be possible to augment the `Subprocess.run()` signature to achieve the same outcome?**

I've reviewed #962 and AFAICT, the primary goal is to monitor the output for sentinels. I think that would be achievable by exposing `filter_func` in `Subprocess.run()` (although, I think it should renamed since it isn't really filtering anything but we can bikeshed later).

All that said, macOS and iOS are fundamentally different since the App process and the process making the App output available are independent; so, they require a different architecture from the Linux backends here.

Let me know if I haven't taken in the whole picture properly.

freakboy3742:
> **Would it be possible to augment the `Subprocess.run()` signature to achieve the same outcome?**
> 
> I've reviewed #962 and AFAICT, the primary goal is to monitor the output for sentinels. I think that would be achievable by exposing `filter_func` in `Subprocess.run()` (although, I think it should renamed since it isn't really filtering anything but we can bikeshed later).
> 
> All that said, macOS and iOS are fundamentally different since the App process and the process making the App output available are independent; so, they require a different architecture from the Linux backends here.
> 
> Let me know if I haven't taken in the whole picture properly.

You're not wrong - I've had very similar thoughts.

The only reason I didn't do it as part of this PR is that I wanted to confirm that the run sequence in fact *was* identical on every platform, and could be abstracted consistently, especially in the light of the changes we're going to need to make to stream filters for #962. As you note, macOS has a quite different approach; I wasn't 100% sure if we were going to end up with similar differences once we resolved the issues like the Android startup problem (#966) or Windows logging problem (#970).

Think of it as an incremental approach - get everything into a stream-based approach on a platform-by-platform basis (and fix some of the other log-related issues); then accomodate the testing changes; then refactor to pull out the common "streaming execution" code. Smaller focussed PRs tend to be easier to review.

Based on where I am right now, it seems very likely we'll be able to factor out some common code; the "run this app" code workflow is very similar across every platform. I'm not completely sure if it will just be adding a `filter_func` argument to `run()`, or a new method that encompasses a "testable execution", I should get a better feel for that once I've got testing working on a few platforms.

Issue #787, opened 2022-07-14
App output isn't captured on Windows


freakboy3742:
#782 added a stub app for Windows, which means that stdout/stderr output is now visible. However, that output is not captured in the log file for diagnostic purposes.

**To Reproduce**
1. Create and run a helloworld app on Windows using `briefcase run --log`
2. Open the log file that was created.

The log will will contain the debug showing how the app was started, but will not contain any of the app's logging output displayed to the user.

**Expected behavior**

The log should contain the contents of stdout/stderr from the app's run.

**Screenshots**

**Environment:**
 - Operating System: Windows
 - Python version: All
 - Software versions:
   - Briefcase: [fbab44d](https://github.com/beeware/briefcase/commit/fbab44d4d192d81eb3c0e6a5a790e78639d3c670) or later.

**Additional context**

As noted in the discussion around #782, this is a function of how the Windows app handles output.

On Linux, app output is written to stdout/err, which can be captured by `run(stream_output=True)`.

On macOS, we install the std-nslog shim to redirect stdout/err to the system log, and we tail the system log to stdout, capturing that output in the log.

On Windows, things are more complex. Windows GUI apps don't have a stdout/stderr, so we can't just redirect stdout/err; and there's no analog of the macOS "System log". 

When you see the output in the terminal in Briefcase, you're not seeing stdout/stderr - you're seeing the output of a Windows specific device `CONOUT$`... and there will only be output on CONOUT$ if you're explicitly writing to it. The stub app sets up this redirection: if there's a parent console, it attaches to the console and redirects stdout/err to that console; if there isn't, it re-opens the stdout/err file descriptors as a file, redirecting output to those logs.

Therefore, you're not seeing stdout/err in the briefcase terminal window - you're seeing `CONOUT$`. Unfortunately, subprocess.run() doesn't have any handling for CONOUT$. It defines [subprocess.STD_OUTPUT_HANDLE](https://docs.python.org/3/library/subprocess.html#subprocess.STD_OUTPUT_HANDLE), but there doesn't appear to be any obvious way to capture this content. From an initial search, [this stack overflow answer](https://stackoverflow.com/a/38749458) looks promising... if a little baroque.

PR #970, opened 2022-11-14
Convert Windows app runners to use log streaming

freakboy3742:
In order to support test mode (Refs #962), we need to convert app runners to use streamers, rather than just outputting logs directly to stdout. To make the code easier to review, I've split the introduction of streaming to the Windows backends as a standalone PR, similar to what has been done with #966 for Android and #967 for Linux.

Requires a small change to the app templates (Refs beeware/briefcase-windows-VisualStudio-template#5); that template change corrects #787. That change will need to be propagated into the windows-app template.

## PR Checklist:

### DIFF
--- a/src/briefcase/platforms/windows/__init__.py
+++ b/src/briefcase/platforms/windows/__init__.py
@@ -82,15 +82,27 @@ class WindowsRunCommand(RunCommand):
         self.logger.info("Starting app...", prefix=app.app_name)
         try:
             # Start streaming logs for the app.
-            self.logger.info("=" * 75)
-            self.tools.subprocess.run(
+            log_popen = self.tools.subprocess.Popen(
                 [os.fsdecode(self.binary_path(app))],
                 cwd=self.tools.home_path,
-                check=True,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                bufsize=1,
             )
+
+            try:
+                # Start streaming logs for the app.
+                self.logger.info("=" * 75)
+                self.tools.subprocess.stream_output(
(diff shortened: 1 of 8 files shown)


src/briefcase/platforms/windows/__init__.py
```diff
@@ -82,15 +82,30 @@ def run_app(self, app: BaseConfig, **kwargs):
...
+                self.tools.subprocess.cleanup(app.app_name, log_popen)
+
+            # If the process didn't exit cleanly, raise an error.
+            if log_popen.returncode != 0:
+                raise BriefcaseCommandError(f"Problem running app {app.app_name}.")
         except KeyboardInterrupt:
             pass  # Catch CTRL-C to exit normally
-        except subprocess.CalledProcessError as e:
+        except OSError as e:
             raise BriefcaseCommandError(f"Unable to start app {app.app_name}.") from e
```
rmartin16, line 109: I was wondering this about the Linux updates as well....why is `subprocess.cleanup()` being called here? The call to `stream_output()` will handle this, right? or is this for particular corner cases?

rmartin16 (commented):
One clarifying question about the call to `cleanup()`.

src/briefcase/platforms/windows/__init__.py
freakboy3742: Nope - it's oversight. I was copying the general pattern from macOS where there *is* a cleanup, but forgot that because we're streaming the actual process we care about, the cleanup is done automatically by `stream_output()`.

rmartin16 (approved):

PR #971, opened 2022-11-14
Sign frameworks in the canonical macOS app bundle location

samschott:
The code-signing step currently signs mach-O binaries and Frameworks in the "Contents/Resources" folder of a macOS app bundle. Frameworks however are often (typically) found in the "Contents/Frameworks" folder.

The default Xcode and app templates used by briefcase currently don't bundle any Frameworks there. I'd argue that we should still code-sign anything found here because:

1. We might want to include Frameworks by default in the app bundle at some point.
2. Custom templates may include Frameworks and briefcase's code-signing step should still work with those, as long as they don't go against platform conventions.

## PR Checklist:

### DIFF
--- a/changes/971.bugfix.rst
+++ b/changes/971.bugfix.rst
@@ -1,0 +1,1 @@
+Code-sign any frameworks contained in the macOS app bundle inside "Contents/Frameworks".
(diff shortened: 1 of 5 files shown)


freakboy3742 (approved):
Can't argue with this - it makes complete sense.

The only comment on the PR itself is the changenote - I'd argue this is a full feature, not just a bug fix, and the tense needs to be tweaked a little; I've made that change. Thanks for the fix!

PR #982, opened 2022-11-29
Convert log filters into generators.

### LINKED ISSUE #981: On Windows, test mode requires an extra line after the summary line
**Describe the bug**
The app in https://github.com/beeware/toga/pull/1687 contains a [call to exit the app](https://github.com/beeware/toga/blob/06dc8acaf81cbeb89565b18ac0b1942a73440256/test/tests/toga_test.py#L33) once the tests are complete. This line should not be necessary, because `briefcase run --test` should exit once the pytest summary line is detected. But if you comment it out, Briefcase prints the summary line:
```
======================== 7 passed, 1 skipped in 2.44s =========================
```
and then hangs until you manually close the window. Only then does it print:
```
[toga-test] Test suite passed!
```
and then Briefcase exits.

I suspect this may be something to do with the way the LogFilter sets its `success` attribute but doesn't check it until the following iteration. Sure enough, if you add a `print` after `pytest.main`, the extra line does not appear, but `briefcase run --test` now exits correctly, and kills the app as well.

**Environment:**
 - Operating System: Windows
 - Python version: 3.8
 - Software versions:
   - Briefcase: 4cac510392478

freakboy3742:
Modifies log filtering so that the filter is required to be a generator, rather than a filter function.

This allows greater flexibility over how input is handled (1 line of input can cause 0-N lines of output) without complicating the interface of the filter function itself.

It also removes the need for an extra line of content to cause the termination of the log stream.

Fixes #981

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -22,6 +22,10 @@ class ParseError(Exception):
     """Raised by parser functions to signal parsing was unsuccessful."""
 
 
+class StopStreaming(Exception):
+    """Raised by streaming filters to terminate the stream."""
+
+
 def ensure_str(text):
     """Returns input text as a string."""
     return text.decode() if isinstance(text, bytes) else str(text)
@@ -566,9 +570,9 @@ class Subprocess(Tool):
             should stop and the popen_process should be terminated.
         :param filter_func: a callable that will be invoked on every line
             of output that is streamed. The function accepts the "raw" line
-            of input (stripped of any trailing newline); it returns the
-            filtered output that should be displayed to the user, or ``None``
-            if the line of input should be ignored by the output stream.
+            of input (stripped of any trailing newline); it returns a generator
+            that yields the filtered output that should be displayed to the user.
+            Can raise StopStreaming to terminate the output stream.
         """
         output_streamer = threading.Thread(
             name=f"{label} output streamer",
@@ -608,17 +612,16 @@ class Subprocess(Tool):
         # ValueError is raised if stdout is unexpectedly closed.
         # This can happen if the user starts spamming CTRL+C, for instance.
         # Silently exit to avoid Python printing the exception to the console.
-        # StopIteration can be raised by a filter function to indicate that
+        # StopStreaming can be raised by a filter function to indicate that
         # there's no need to stream any more.
-        with suppress(ValueError, StopIteration):
+        with suppress(ValueError, StopStreaming):
             while True:
                 # readline should always return at least a newline (ie \n)
                 # UNLESS the underlying process is exiting/gone; then "" is returned
                 output_line = ensure_str(popen_process.stdout.readline())
                 if output_line:
                     if filter_func:
-                        filtered_output = filter_func(output_line.rstrip("\n"))
-                        if filtered_output is not None:
+                        for filtered_output in filter_func(output_line.rstrip("\n")):
                             self.tools.logger.info(filtered_output)
                     else:
                         self.tools.logger.info(output_line)
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -6,7 +6,6 @@ import subprocess
 import sys
 import threading
 import time
-from contextlib import suppress
 from functools import wraps
 from pathlib import Path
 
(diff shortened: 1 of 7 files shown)


src/briefcase/integrations/subprocess.py
```diff
@@ -608,17 +612,16 @@ def _stream_output_thread(self, popen_process, filter_func):
         # ValueError is raised if stdout is unexpectedly closed.
         # This can happen if the user starts spamming CTRL+C, for instance.
         # Silently exit to avoid Python printing the exception to the console.
-        # StopIteration can be raised by a filter function to indicate that
+        # StopStreaming can be raised by a filter function to indicate that
         # there's no need to stream any more.
-        with suppress(ValueError, StopIteration):
+        with suppress(ValueError, StopStreaming):
```
mhsmith, line 617: There's quite a large amount of code being run inside this `suppress` block. So to avoid hiding `ValueErrors` that are raised for other reasons, it would be safer to remove the suppression of `ValueError`, and instead put an exception handler around `readline` alone, which catches `ValueError` and raises `StopStreaming`.

rmartin16: That's a good point. Furthermore, the output streamer never implemented any robust exception handling since it was pretty simple; however, any exceptions will simply spill out in to the console as the thread rolls up. We may want to consider handling that at least involves including such exceptions in the log file....perhaps via `Log.add_log_file_extra()`.

freakboy3742: I've restructured the streaming function to have more specific exception handling, and added exception handling for safety while streaming.

mhsmith (approved):

src/briefcase/integrations/subprocess.py
```diff
@@ -605,25 +608,36 @@ def _stream_output_thread(self, popen_process, filter_func):
...
+                            return
                     else:
                         self.tools.logger.info(output_line)
                 else:
                     return
+        except Exception as e:
+            self.tools.logger.error(
+                f"Error while streaming output: {e.__class__.__name__}: {e}"
+            )
+            self.tools.logger.capture_stacktrace()
```
rmartin16, line 640: Using `capture_stacktrace()` here may create confusing outcomes.
- In perhaps the best outcome, the Rich stacktrace in the log is simply from the streamer's exception....assuming the main thread doesn't raise another elsewhere.
- If the main thread raises, it becomes a bit of a race condition which stacktrace is included in the log.

Even in the current usecase for testing, if the streamer raises, then Briefcase will later raise `BriefcaseCommandError: Test suite didn't report a result`....but only one of the stracktraces will make it in the log.

I think that `capture_stacktrace()` should be reserved for main thread exceptions since they will probably be more interesting/useful.

This is a "least effort" implementation to simply add the streamer's Rich stacktrace to the bottom of the log:
```python
        except Exception as e:
            self.tools.logger.error(
                f"Error while streaming output: {e.__class__.__name__}: {e}"
            )
            from rich.traceback import Traceback
            stracktrace = [
                "Output Streamer Stacktrace:",
                Traceback(
                    width=self.tools.logger.print.LOG_FILE_WIDTH,
                    show_locals=True,
                ),
            ]
            self.tools.logger.add_log_file_extra(
                lambda: self.tools.logger.print.log.print(*stracktrace)
            )
```

Alternatively, of course, refactoring stracktrace management in `Console` to accommodate an arbitrary number of stracktraces is possible as well.

rmartin16 (changes requested):

src/briefcase/integrations/subprocess.py
freakboy3742: Good call; and it wasn't *that* hard to make the stacktrace management generic. It also now allows for capturing stack traces that don't result in the app crashing - not sure how useful that will be, but it's an option at least.

tests/console/test_Log.py
```diff
@@ -145,8 +152,41 @@ def test_save_log_to_file_with_exception(tmp_path, now):
...
+        try:
+            1 / 0
+        except ZeroDivisionError:
+            logger.capture_stacktrace(f"Thread {i}")
+
+    logger.save_log_to_file(command=command)
+
+    log_filepath = tmp_path / logger.LOG_DIR / "briefcase.2022_06_25-16_12_29.dev.log"
+
+    log_filepath.exists()
```
rmartin16, line 181: Should this be an `assert`? same in `test_save_log_to_file_with_exception()` it seems

rmartin16 (approved):
Small typo in tests. Good to go otherwise.

tests/console/test_Log.py
freakboy3742: Yup - looks like a weird copy-and-paste error.

PR #983, opened 2022-11-29
Fully expose update flags to run and build commands

freakboy3742:
Although the run and build commands are dependent on app requirements and resources, those commands did not expose a mechanism for updating requirements or resources. 

This update consistently exposes options to update code, requirements and resources, as well as exposing options to explicitly *disable* updates when in test mode. This has the added benefit of providing faster test paths for some use cases (e.g., rapid testing cycles on mobile platforms, where you need to update code, but don't want requirements updates because they're slow).

This change necessitated changing the flag for updating dependencies from `-d`, because it clashed with `--device` used by the iOS and Android run commands. This, in turn, caused a clash with `--update-resources`. I've opted to converge on:

* `-d` / `--device` to specify a mobile emulator/simulator
* `-r` / `--update-requirements` to update requirements
* `--update-resources` to update resources.

I figured that app resources change infrequently, so a short option probably isn't required.

## PR Checklist:


src/briefcase/commands/run.py
```diff
@@ -159,15 +159,53 @@ def add_options(self, parser):
...
-            help="Update and rebuild the app before execution",
+            help="Update and rebuild the app before running",
         )
         parser.add_argument(
             "--no-update",
             dest="update",
             action="store_const",
             const=False,
-            help="Prevent any automated update and rebuild before running.",
+            help="Prevent any automated update of requirements before running.",
```
mhsmith, line 169: "of requirements" should be removed.

More generally, since several options are now shared between the update, build and run commands, could they be factored out? The only difference between the three copies is the last few words of the help text, which is either "before building", "before running", or nothing at all. But think those words are unnecessary, since the user already knows what command they're looking at. And removing them would allow each argument description to fit on a single line, making the help text easier to read.

docs/reference/commands/run.rst
```diff
@@ -67,19 +67,51 @@ name specified should be the machine-readable package name for the app.
...
+    $ briefcase run
+
+``-r`` / ``--update-requirements``
+----------------------------------
+
+Update application requirements before running. Equivalent to running::
+
+    $ briefcase update -r
+    $ briefcase build
+    $ briefcase run
```
mhsmith, line 80: It isn't entirely equivalent, because `build -r` and `run -r` will update the requirements without updating the sources. This is a combination that `update` itself doesn't support, and I don't see much use in adding it.

So how about this:
* The update phase would always update the sources, regardless of how it was triggered: `-u`, `-r` or `--update-resources`. 
* `--no-update` would prevent the update phase entirely, with no need to pass `--no-update-requirements` and `--no-update-resources` as well. If `--no-update` was used along with `-r` or `--update-resources`, we would give an error message.

That way, the update phase really would be equivalent to a standalone `update` command, and it would be easier for the user to understand and control.

src/briefcase/commands/run.py
freakboy3742: Yeah - I agree there's lots of duplication; I was hesitant to pull out a common definition because there's no obvious "common" base class; however, I guess adding a utility method to the base command that can be invoked as needed is an option. I've made this change.

I'm not sure I agree about truncating the help text, though. It never hurts to be explicit, and help text is the very definition of "you're only looking at it if you don't know what you're doing", so assuming that the context is obvious is problematic. I'm not overly worried about running onto multiple lines, either - that only gets enforced if you're literally at 80 characters. From the look of it, if your terminal is only slightly wider, the wrapping stops; and there's much bigger formatting problems caused by the length of `--no-update-requirements` (and similar) options).

docs/reference/commands/run.rst
freakboy3742: The equivalence issue in documentation is a good point - that's a copy-and-paste that I clearly didn't fully think through.

As for `--no-update`; I went back and forth on whether `--no-update` should be a "disable all updates" or a "disable only code updates". I ended up landing on the "full parity between update and no-update pairs" because I figured these were all power options, so flexibility would be useful; however, on reflection I think you're probably right that it can be simplified (plus, it addresses the "long options" problem I mentioned in the other comment).

To confirm we're on the same page, we're talking about the following behavior:
1. `update` - updates code, but nothing else
2. `update -r` updates code and requirements
3. `update --update-resources` updates code and resources, but not requirements
4. `build` - does no updates
5. `build -u` updates code before building
6. `build -r` updates requirements before building (but *not* code)
7.  `build --test` updates code into test mode before building; no requirements/resources changes
8.  `build --test -r` updates code and requirements into test mode before building; no resources changes
9. `build --test --no-update` updates nothing; just does the build
10. `build --test -r --no-update ` is an error.

4-10 would then have the same behaviour for `run`.

PR #984, opened 2022-11-30
Make pytest regexes accept content after the time

mhsmith:
[This CI run](https://github.com/beeware/Python-support-testbed/actions/runs/3580373832/jobs/6023434680) shows that when the run takes longer than a minute, pytest will display the elapsed time in two different formats. Our regexes don't currently accept this.

Fortunately the first format is the one we already accept, so I've just made it accept any content at all between the time and the closing line of `=`s. I think the rest of the regex is selective enough that this won't cause any false positives. 

## PR Checklist:

### DIFF
--- a/changes/984.misc.rst
+++ b/changes/984.misc.rst
@@ -1,0 +1,1 @@
+Made pytest regexes accept content after the time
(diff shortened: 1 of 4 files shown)


freakboy3742 (approved):
👍

PR #994, opened 2022-12-07
Pin CI requirement versions

mhsmith:
A new version of tox was released today which broke our CI workflow (see [here](https://github.com/beeware/briefcase/actions/runs/3642076566/jobs/6148779467)). This PR fixes that by pinning the versions of most requirements used by the workflow. However, it doesn't pin pip and setuptools, because those will be provided by the Briefcase user, so if our packages don't work with a new version we'd want to be notified about that ASAP.

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
     - uses: actions/setup-python@v4.3.0
       with:
         python-version: ${{ env.python_version }}
-    - run: pip install tox==3.27.1
+    - run: pip install tox~=3.27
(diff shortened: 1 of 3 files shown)


freakboy3742 (approved):
Makes sense; I've made a small tweak to (a) use "compatible" rather than hard version pins, and reflected those pins in the setup.cfg as well.

Longer term, it would be desirable for the versions in the CI config to be in a place where dependabot will pick them up. Pinned versions are good for stability (and preventing breakages like the one we've just had; but it also runs the risk of calcifying if there isn't a regular reminder to go through update cycles.

PR #995, opened 2022-12-08
Upgrade tox to v4

freakboy3742:
Upgraded tox to v4, including changes to the configuration file format required by that upgrade (Refs #994).

Took the opportunity to pin all remaining packages, upgrade everything that is pinned, and correct some minor metadata problems.

## PR Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,21 +73,21 @@ install_requires =
 
 [options.extras_require]
 dev =
-    pre-commit
+    pre-commit~=2.20
     setuptools_scm[toml] ~= 7.0
     tox ~= 3.27
     coverage[toml] ~= 6.5
 test =
-    pytest
-    pytest-tldr
-    pytest-cov
+    pytest~=7.2
+    pytest-tldr~=0.2
+    pytest-cov~=4.0
 docs =
-    sphinx
-    sphinxcontrib-spelling
-    pyenchant
-    sphinx-autobuild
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -57,36 +58,36 @@ package_dir =
 install_requires =
     pip >= 22
     setuptools >= 60
-    wheel ~= 0.34
+    wheel ~= 0.37
     cookiecutter ~= 2.1
     tomli ~= 2.0; python_version <= "3.10"
-    importlib_metadata ~= 4.4; python_version <= "3.9"
-    requests ~= 2.22
+    importlib_metadata >= 4.4; python_version <= "3.9"
```
freakboy3742, line 64: This has been deliberately made a minimum version pin; importlib_metadata will track the implementation in the current version of Python, so we should ensure that we're tracking that version.

mhsmith (approved):

webknjaz:
@freakboy3742 FYI I saw this in your `tox.ini`:
> python -m build --sdist --wheel

It's best if you remove `--sdist --wheel` because it builds both sdist and wheel from the Git checkout independently now. Without these args, it'll create an sdist and then make a wheel from that sdist — this helps make sure that the files included in sdist are enough to make a wheel.

freakboy3742:
@webknjaz Thanks for the heads up!

PR #998, opened 2022-12-12
Allow test suite completion and test suite exit to be different.

freakboy3742:
At present, under `briefcase run --test`, the test suite exits as soon as the success or failure criterion is found.

However, if there is output in the test log that is emitted *after* the test suite, that output is not displayed. The most obvious use case for this is coverage; the coverage report is output after the test suite completes.

This PR adds a `test_exit_regex` setting that allows a project to define a criterion for exiting the test suite that is independent of the success or failure of the test suite.

## PR Checklist:

### DIFF
--- a/changes/998.misc.rst
+++ b/changes/998.misc.rst
@@ -1,0 +1,1 @@
+Test suites can now define an independent criterion to match for test suite exit.
--- a/docs/reference/configuration.rst
+++ b/docs/reference/configuration.rst
@@ -406,6 +406,18 @@ Briefcase defaults to regular expressions that are able to match the output of
 ``unittest`` or ``pytest``; if you use a different test framework, you will need
 to customize these settings.
 
+``test_exit_regex``
+~~~~~~~~~~~~~~~~~~~
+
+A regular expression that will be used to determine when the test suite has
+finished. The ``re.MULTILINE`` flag will be used.
+
+This option is only required if there is test suite content generated *after* the
+success/failure criteria has been output that you wish to preserve in the log
+(e.g., a coverage report). If ``test_exit_regex`` is *not* specified, Briefcase
+will assume that the test suite has completed as soon as a success or failure
+pattern has matched.
+
 ``url``
 ~~~~~~~
 
--- a/src/briefcase/commands/run.py
+++ b/src/briefcase/commands/run.py
@@ -47,6 +47,7 @@ class LogFilter:
         clean_output,
         success_filter,
         failure_filter,
+        exit_filter,
     ):
         """Create a filter for a log stream.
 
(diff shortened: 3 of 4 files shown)


mhsmith:
As discussed, a better approach might be to make `briefcase run` and `briefcase dev` pass through the app's exit status, so the app itself can determine arbitrarily complex pass/fail conditions. I'll look into how this could be done on Android and post some suggestions.

mhsmith:
On Android, it's possible to exit the app process using [`os._exit`](https://docs.python.org/3/library/os.html#os._exit). This generates the following log message, although it's not included in Briefcase's current logcat filter because it comes from a different process.
```
2022-12-12 20:38:50.786 368-368/? I/Zygote: Process 5667 exited cleanly (42)
```
I've tested this on our min and max API levels (24 and 33).

Another option, more cross-platform but with a higher chance of false positives, is to use the pattern generated by a SystemExit traceback:
```
SystemExit: \d+$
```

This would also match the Java traceback if `sys.exit` is called on the main thread, as in the Python-support-testbed:
```
12-12 19:32:48.356  2643  2643 E AndroidRuntime: com.chaquo.python.PyException: SystemExit: 42
```
However, if `sys.exit` was called on any other thread, as in the Toga testbed app, that's dealt with by [`threading.excepthook`](https://docs.python.org/3/library/threading.html#threading.excepthook), which norrmally prints the traceback to stderrr, but has a special case to silently ignore SystemExit. Removing that special case would be a reasonable default behavior to add to Chaquopy.

freakboy3742:
A quick survey of the state of other platforms:

Windows directly launches the app binary; exit status can be determined from the Popen object used to run the app.

Linux has the same basic setup as Windows on both AppImage and Flatpak.

macOS invokes `open <formal name>.app`, which returns immediately; the exit code only indicates if the app was able to start. Briefcase then attempts to detect the PID of the process that is the 'actual' app; this PID is polled to check whether it has exited. Unfortunately, I don't see a reliable way to get the exit status based purely on PID. `psutils.Process.wait()` can be used to block on a PID and return the status code, but that's a blocking mechanism; we'd need to get into a situation where we're running a background thread to block and monitor a specific process, and having that thread report the exit status.

The situation on iOS is similar to macOS, except that it uses iOS simulator tooling to launch the app. The PID returned is a simulator process. We don't current support on-device execution; but if we were able to in future, I suspect we'd be in a similar situation to Android where there isn't a OS-level mechanism to observe the exist status.

So - it appears we're going to need to use an output log filtering approach for Android regardless; and will likely need a similar approach for iOS-on-device in the future. macOS and iOS-on-simulator will need a significant change to move to an exit-status approach. On that basis, it seems to me that a filtering approach based on a consistent output sentinel that reports the exit status will be the easiest path forward in the short term. Longer term, we can investigate moving to a PID-based approach, but I don't think that needs to block progress right now.

mhsmith:
Agreed: closing in favor of #1002.

PR #1001, opened 2022-12-13
Simplify package build.

freakboy3742:
Although the `--sdist --wheel` arguments to `build` cause an sdist and wheel to be packaged, they are completely isolated builds. If you omit these two arguments, an sdist and wheel are still built; but the wheel is built *using* the sdist, so there's an extra layer of security that the sdist contains everything it needs.

Thanks to @webknjaz for [this suggestion](https://github.com/beeware/briefcase/pull/995#issuecomment-1347563630)

## PR Checklist:

### DIFF
--- a/changes/1001.misc.rst
+++ b/changes/1001.misc.rst
@@ -1,0 +1,1 @@
+The packaging of sdist and wheel artefacts was simplified.
--- a/tox.ini
+++ b/tox.ini
@@ -36,5 +36,5 @@ deps =
     twine
 commands =
     check-manifest -v
-    python -m build --sdist --wheel --outdir dist/ .
+    python -m build --outdir dist/ .
     python -m twine check dist/*


changes/1001.misc.rst
```diff
@@ -0,0 +1 @@
+The packaging of sdist and wheel artefacts was simplified.
```
webknjaz, line 1:
```suggestion
The packaging of sdist and wheel artifacts was simplified.
```

freakboy3742: Artefact is also a valid spelling, and is the preferred spelling (in my style guide, anyway); it's from the Latin "arte factum".

webknjaz: Ah, I didn't know that English has it in this form.

mhsmith (approved):

PR #1005, opened 2022-12-13
Help text includes command descriptions, platforms, and output formats

### LINKED ISSUE #906: Add Option to Display Supported Platforms
Briefcase help text does not include a clear method to show supported platforms.

This should be added as a proper option alongside `-f/--formats`.

Add `--platforms` option to enumerate supported platforms.

If you run, for example, `briefcase run asdf`, a list of supported platforms are in fact printed.

### LINKED ISSUE #907: Add command summary to `briefcase --help`

When I was first learning how to use Briefcase, I had trouble remembering what all the commands did. A brief overview in one place would have been very useful.

`briefcase --help` should give a 1-line summary of what each command does, in the same style as `git --help`:
```
   add        Add file contents to the index
   mv         Move or rename a file, a directory, or a symlink
   reset      Reset current HEAD to the specified state
   rm         Remove files from the working tree and from the index
   ...
```

It would also be useful to have this in the documentation, e.g. [here](https://briefcase.readthedocs.io/en/latest/reference/index.html).

rmartin16:
## Changes
Help text is updated to include:
- Commands and their descriptions for `briefcase` or `briefcase -h`
- Supported platforms for `briefcase` or `briefcase -h`
- Support formats for `briefcase <command> -h`

## Issues
- Fixes #906 
- Fixes #907

## PR Checklist:

### DIFF
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -42,7 +42,7 @@ def parse_cmdline(args):
     :return: Command and command-specific arguments
     """
     platforms = get_platforms()
-    width = min(shutil.get_terminal_size().columns, 80) - 2
+    width = max(min(shutil.get_terminal_size().columns, 80) - 2, 20)
 
     briefcase_description = textwrap.fill(
         "Briefcase is a tool for converting a Python project "
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -56,7 +56,7 @@ def parse_cmdline(args):
         for cmd in COMMANDS
     )
 
-    platform_list = ", ".join([p.title() if p.islower() else p for p in platforms])
+    platform_list = ", ".join(sorted(platforms, key=str.lower))
 
     additional_instruction = textwrap.fill(
         "Each command, platform, and format has additional options. "
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -1,5 +1,7 @@
 import argparse
+import shutil
 import sys
+import textwrap
 from argparse import RawDescriptionHelpFormatter
 
 from briefcase import __version__
@@ -40,31 +42,44 @@ def parse_cmdline(args):
     :return: Command and command-specific arguments
     """
     platforms = get_platforms()
+    width = min(shutil.get_terminal_size().columns, 80) - 2
+
+    briefcase_description = textwrap.fill(
+        "Briefcase is a tool for converting a Python project "
+        "into a standalone native application for distribution.",
+        width=width,
+    )
 
     description_max_pad_len = max(len(cmd.command) for cmd in COMMANDS) + 2
-    cmd_helptext = "\n".join(
+    command_description_list = "\n".join(
         f"  {cmd.command}{' ' * (description_max_pad_len - len(cmd.command))}{cmd.description}"
         for cmd in COMMANDS
+    )
+
+    platform_list = ", ".join([p.title() if p.islower() else p for p in platforms])
+
+    additional_instruction = textwrap.fill(
+        "Each command, platform, and format has additional options. "
+        "Use the -h option on a specific command for more details.",
+        width=width,
     )
 
     parser = argparse.ArgumentParser(
         prog="briefcase",
         description=(
-            "Briefcase is a tool for converting a Python project into a standalone native\n"
-            "application for distribution.\n"
+            f"{briefcase_description}\n"
             "\n"
             "Commands:\n"
-            f"{cmd_helptext}\n"
+            f"{command_description_list}\n"
             "\n"
             "Platforms:\n"
-            f"  {', '.join([p.title() if p.islower() else p for p in platforms])}\n"
+            f"  {platform_list}\n"
             "\n"
-            "Each command, platform, and format has additional options. Use the -h option on\n"
-            "a specific command for more details."
+            f"{additional_instruction}"
         ),
         usage="briefcase [-h] <command> [<platform>] [<format>] ...",
         add_help=False,
-        formatter_class=lambda prog: RawDescriptionHelpFormatter(prog, width=80),
+        formatter_class=lambda prog: RawDescriptionHelpFormatter(prog, width=width),
     )
     parser.add_argument("-V", "--version", action="version", version=__version__)
 
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -27,14 +27,14 @@ from .exceptions import (
 COMMANDS = [
     NewCommand,
     DevCommand,
-    UpgradeCommand,
     CreateCommand,
-    UpdateCommand,
     OpenCommand,
     BuildCommand,
+    UpdateCommand,
     RunCommand,
     PackageCommand,
     PublishCommand,
+    UpgradeCommand,
 ]
 
 
@@ -46,9 +46,9 @@ def parse_cmdline(args):
     """
     platforms = get_platforms()
 
-    max_cmd_name_len = max(len(cmd.command) for cmd in COMMANDS)
+    description_max_pad_len = max(len(cmd.command) for cmd in COMMANDS) + 2
     cmd_helptext = "\n".join(
-        f"  {cmd.command}{' ' * (max_cmd_name_len - len(cmd.command) + 2)}{cmd.description}"
+        f"  {cmd.command}{' ' * (description_max_pad_len - len(cmd.command))}{cmd.description}"
         for cmd in COMMANDS
     )
 
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -17,12 +17,7 @@ from briefcase.commands import (
 )
 from briefcase.platforms import get_output_formats, get_platforms
 
-from .exceptions import (
-    InvalidFormatError,
-    NoCommandError,
-    ShowOutputFormats,
-    UnsupportedCommandError,
-)
+from .exceptions import InvalidFormatError, NoCommandError, UnsupportedCommandError
 
 COMMANDS = [
     NewCommand,
@@ -70,13 +65,6 @@ def parse_cmdline(args):
         usage="briefcase [-h] <command> [<platform>] [<format>] ...",
         add_help=False,
         formatter_class=lambda prog: RawDescriptionHelpFormatter(prog, width=80),
-    )
-    parser.add_argument(
-        "-f",
-        "--formats",
-        action="store_true",
-        dest="show_output_formats",
-        help="show the available output formats and exit (specify a platform for more details).",
     )
     parser.add_argument("-V", "--version", action="version", version=__version__)
 
@@ -150,21 +138,14 @@ def parse_cmdline(args):
         # Import the platform module
         platform_module = platforms[options.platform]
 
-        output_formats = get_output_formats(options.platform)
-        # If the user requested a list of available output formats, output them.
-        if options.show_output_formats:
-            raise ShowOutputFormats(
-                platform=options.platform,
-                default=platform_module.DEFAULT_OUTPUT_FORMAT,
-                choices=list(output_formats.keys()),
-            )
-
         # If the output format wasn't explicitly specified, check to see
         # Otherwise, extract and use the default output_format for the platform.
         if options.output_format is None:
             output_format = platform_module.DEFAULT_OUTPUT_FORMAT
         else:
             output_format = options.output_format
+
+        output_formats = get_output_formats(options.platform)
 
         # Normalise casing of output_format to be more forgiving.
         output_format = {n.lower(): n for n in output_formats}.get(
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -1,8 +1,20 @@
 import argparse
 import sys
+from argparse import RawDescriptionHelpFormatter
 
 from briefcase import __version__
-from briefcase.commands import DevCommand, NewCommand, UpgradeCommand
+from briefcase.commands import (
+    BuildCommand,
+    CreateCommand,
+    DevCommand,
+    NewCommand,
+    OpenCommand,
+    PackageCommand,
+    PublishCommand,
+    RunCommand,
+    UpdateCommand,
+    UpgradeCommand,
+)
 from briefcase.platforms import get_output_formats, get_platforms
 
 from .exceptions import (
@@ -12,6 +24,19 @@ from .exceptions import (
     UnsupportedCommandError,
 )
 
+COMMANDS = [
+    NewCommand,
+    DevCommand,
+    UpgradeCommand,
+    CreateCommand,
+    UpdateCommand,
+    OpenCommand,
+    BuildCommand,
+    RunCommand,
+    PackageCommand,
+    PublishCommand,
+]
+
 
 def parse_cmdline(args):
     """Parses the command line to determine the Command and its arguments.
@@ -19,15 +44,32 @@ def parse_cmdline(args):
     :param args: the arguments provided at the command line
     :return: Command and command-specific arguments
     """
+    platforms = get_platforms()
+
+    max_cmd_name_len = max(len(cmd.command) for cmd in COMMANDS)
+    cmd_helptext = "\n".join(
+        f"  {cmd.command}{' ' * (max_cmd_name_len - len(cmd.command) + 2)}{cmd.description}"
+        for cmd in COMMANDS
+    )
+
     parser = argparse.ArgumentParser(
         prog="briefcase",
-        description="Package Python code for distribution.",
-        usage="briefcase [-h] <command> [<platform>] [<format>] ...",
-        epilog=(
-            "Each command, platform and format has additional options. "
-            "Use the -h option on a specific command for more details."
+        description=(
+            "Briefcase is a tool for converting a Python project into a standalone native\n"
+            "application for distribution.\n"
+            "\n"
+            "Commands:\n"
+            f"{cmd_helptext}\n"
+            "\n"
+            "Platforms:\n"
+            f"  {', '.join([p.title() if p.islower() else p for p in platforms])}\n"
+            "\n"
+            "Each command, platform, and format has additional options. Use the -h option on\n"
+            "a specific command for more details."
         ),
+        usage="briefcase [-h] <command> [<platform>] [<format>] ...",
         add_help=False,
+        formatter_class=lambda prog: RawDescriptionHelpFormatter(prog, width=80),
     )
     parser.add_argument(
         "-f",
@@ -45,26 +87,11 @@ def parse_cmdline(args):
     # usage string so that the instructions displayed are correct
     parser.add_argument(
         "command",
-        choices=[
-            "new",
-            "dev",
-            "upgrade",
-            "create",
-            "update",
-            "open",
-            "build",
-            "run",
-            "package",
-            "publish",
-        ],
+        choices=list(cmd.command for cmd in COMMANDS),
         metavar="command",
         nargs="?",
-        help="the command to execute (one of: %(choices)s)",
+        help=argparse.SUPPRESS,
     )
-
-    # <platform> *is* optional, with the default value based on the platform
-    # that you're on.
-    platforms = get_platforms()
 
     # To make the UX a little forgiving, we normalize *any* case to the case
     # actually used to register the platform. This function maps the lower-case
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -531,23 +531,31 @@ a custom location for Briefcase's tools.
         pass
 
     def parse_options(self, extra):
+        """Parse the command line arguments for the Command.
+
+        After the initial ArgumentParser runs to choose the Command for the
+        selected platform and format, a new ArgumentParser is created here to
+        parse the remaining command line arguments specific to the Command.
+        Additionally, the default options for disabling input, log verbosity,
+        and log saving are parsed out and saved to the Command.
+
+        :param extra: the remaining command line arguments after the initial
+            ArgumentParser runs over the command line.
+        :return: dictionary of parsed arguments for Command
+        """
         default_format = getattr(
-            get_platforms().get(self.platform, None), "DEFAULT_OUTPUT_FORMAT", None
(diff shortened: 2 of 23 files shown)


rmartin16:
New help text examples:

```
❯ briefcase -h

usage: briefcase [-h] <command> [<platform>] [<format>] ...

Briefcase is a tool for converting a Python project into a standalone native
application for distribution.

Commands:
  new      Create a new briefcase project
  dev      Run a briefcase project in the dev environment
  upgrade  Upgrade Briefcase-managed tools
  create   Create a new App for a target platform
  update   Update the source, dependencies, etc. for an App
  open     Open an App for manual introspection
  build    Build an App for a target platform
  run      Run a fully built App
  package  Package an App for distribution
  publish  Publish an App to a deployment

Platforms:
  Android, iOS, Linux, macOS, Web, Windows

Each command, platform, and format has additional options. Use the -h option on
a specific command for more details.

options:
  -f, --formats  show the available output formats and exit (specify a platform
                 for more details).
  -V, --version  show program's version number and exit
```

```
❯ briefcase run -h
usage: briefcase run linux appimage [-h] [-v] [-V] [--no-input] [--log]
                                    [-a APPNAME] [-u] [-r] [--update-resources]
                                    [--no-update] [--test] [--no-docker]

Run a Linux AppImage.

Supported formats:
  Appimage (default), Flatpak

options:
  -h, --help            show this help message and exit
  -v, --verbosity       set the verbosity of output
  -V, --version         show program's version number and exit
  --no-input            Don't ask for user input. If any action would be
                        destructive, an error will be raised; otherwise, default
                        answers will be assumed
  --log                 Save a detailed log to file. By default, this log file
                        is only created for critical errors
  -a APPNAME, --app APPNAME
                        The app to run
  -u, --update          Update the app before running
  -r, --update-requirements
                        Update requirements for the app before running
  --update-resources    Update app resources (icons, splash screens, etc) before
                        running
  --no-update           Prevent any automated update before running
  --test                Run the app in test mode
  --no-docker           Don't use Docker for building the AppImage
```

```
❯ briefcase run windows visualstudio -h
usage: briefcase run windows VisualStudio [-h] [-v] [-V] [--no-input] [--log]
                                          [-a APPNAME] [-u] [-r]
                                          [--update-resources] [--no-update]
                                          [--test]

Run a Visual Studio project.

Supported formats:
  App (default), VisualStudio

options:
  -h, --help            show this help message and exit
  -v, --verbosity       set the verbosity of output
  -V, --version         show program's version number and exit
  --no-input            Don't ask for user input. If any action would be
                        destructive, an error will be raised; otherwise, default
                        answers will be assumed
  --log                 Save a detailed log to file. By default, this log file
                        is only created for critical errors
  -a APPNAME, --app APPNAME
                        The app to run
  -u, --update          Update the app before running
  -r, --update-requirements
                        Update requirements for the app before running
  --update-resources    Update app resources (icons, splash screens, etc) before
                        running
  --no-update           Prevent any automated update before running
  --test                Run the app in test mode
```

rmartin16:
A lot of room for opinion here....sooo...

There's (at least) two levels of review: presentation and code. I'm mostly interested in presentation right now....some of the code is dubious; I'll call these out in code comments.

As for the presentation:
- What can be improved? I think this aligns with what I'd expect in CLI helptext...open to critique.
- Does the `-f` option serve a purpose any longer? or just remove it?
- I limited the display width to 80 characters.
  - This is partially accomplished with `RawDescriptionHelpFormatter`....but manually elsewhere.

src/briefcase/cmdline.py
```diff
@@ -12,22 +24,52 @@
...
-            "Use the -h option on a specific command for more details."
+        description=(
+            "Briefcase is a tool for converting a Python project into a standalone native\n"
+            "application for distribution.\n"
+            "\n"
+            "Commands:\n"
+            f"{cmd_helptext}\n"
+            "\n"
+            "Platforms:\n"
+            f"  {', '.join([p.title() if p.islower() else p for p in platforms])}\n"
```
rmartin16, line 65: I think the list of Platforms look better when they are capitalized. However, this avoids capitalizing words like `macOS` and `iOS`.

freakboy3742: I think we have a big decision to make here. 

The command line interface is forgiving about case, but the capitalisation matters from the perspective of configuration files - you actually need to use `tool.briefcase.app.myapp.windows.VisualStudio`. There's a "preferred" capitalisation for every platform and backend... which doesn't necessarily match the capitalisation of the underlying code modules that implement those modules. There's an attribute on each platform/backend module that provides the preferred capitalisation.

The actual capitalisation scheme is... inconsistent. It's *mostly* all lower case; except for backends and platforms that have very specific capitalisation styles (iOS, macOS, watchOS, Xcode, VisualStudio). Being lower case makes it easier to type; captialized makes it more 'humane'. 

If we're going to display something to the user, I'd argue it should be consistent with the "preferred" capitalisation. Two separate issues to consider are:
1. Are our current preferred capitalisations correct?
2. Should modify code to be case insensitive wherever possible?

src/briefcase/commands/update.py
```diff
@@ -8,6 +8,7 @@
 
 class UpdateCommand(CreateCommand):
     command = "update"
+    description = "Update the source, dependencies, etc. for an App"
```
freakboy3742, line 11:
```suggestion
    description = "Update the source, dependencies, and resources for an app"
```

src/briefcase/commands/run.py
```diff
@@ -142,6 +142,7 @@ def filter_func(recent):
 
 class RunCommand(BaseCommand):
     command = "run"
+    description = "Run a fully built App"
```
freakboy3742, line 145:
```suggestion
    description = "Run an app"
```

src/briefcase/commands/publish.py
```diff
@@ -6,6 +6,7 @@
 
 class PublishCommand(BaseCommand):
     command = "publish"
+    description = "Publish an App to a deployment"
```
freakboy3742, line 9:
```suggestion
    description = "Publish an app to a distribution channel"
```

src/briefcase/commands/package.py
```diff
@@ -7,6 +7,7 @@
 
 class PackageCommand(BaseCommand):
     command = "package"
+    description = "Package an App for distribution"
```
freakboy3742, line 10:
```suggestion
    description = "Package an app for distribution"
```

src/briefcase/commands/open.py
```diff
@@ -7,6 +7,7 @@
 
 class OpenCommand(BaseCommand):
     command = "open"
+    description = "Open an App for manual introspection"
```
freakboy3742, line 10:
```suggestion
    description = "Open an app in the build tool for the target platform"
```

src/briefcase/commands/create.py
```diff
@@ -118,6 +118,7 @@ def write_dist_info(app: BaseConfig, dist_info_path: Path):
 
 class CreateCommand(BaseCommand):
     command = "create"
+    description = "Create a new App for a target platform"
```
freakboy3742, line 121:
```suggestion
    description = "Create a new app for a target platform"
```

src/briefcase/commands/build.py
```diff
@@ -8,6 +8,7 @@
 
 class BuildCommand(BaseCommand):
     command = "build"
+    description = "Build an App for a target platform"
```
freakboy3742, line 11:
```suggestion
    description = "Build an app for a target platform"
```

src/briefcase/commands/base.py
```diff
@@ -526,13 +529,32 @@ def verify_app_tools(self, app: BaseConfig):
...
+            except ValueError:
+                formats[
+                    formats.index(default_format.capitalize())
+                ] = f"{default_format} (default)"
+            formats_helptext = (
+                f"Supported formats:\n"
+                f"  {', '.join(f.capitalize() if f.islower() else f for f in formats)}\n"
+            )
+        else:
+            formats_helptext = ""
```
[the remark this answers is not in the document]
freakboy3742: Yeah - see above for a discussion about capitalisation.

The Upgrade platform dependency is definitely an unintended side effect; I'm intrigued why that happens for Upgrade, but not for Dev (unless it does...?)

A lot of the complexity here stems from the way we are bending argparse to the limit so we can defer the construction of subcommand help until we know the actual platform. If there's a better solution to argparse (or a better way to use argparse), I'm not opposed to a major rework of the argument handling code.

src/briefcase/cmdline.py
```diff
@@ -12,22 +24,52 @@
...
+    DevCommand,
+    UpgradeCommand,
+    CreateCommand,
+    UpdateCommand,
+    OpenCommand,
+    BuildCommand,
+    RunCommand,
+    PackageCommand,
+    PublishCommand,
+]
```
freakboy3742, line 38:
```suggestion
COMMANDS = [
    NewCommand,
    DevCommand,
    CreateCommand,
    OpenCommand,
    BuildCommand,
    UpdateCommand,
    RunCommand,
    PackageCommand,
    PublishCommand,
    UpgradeCommand,
]
```

This is a minor re-ordering to get commands into "normal use order" - i.e., the order in which you'll normally encounter them.

freakboy3742 (changes requested):
The broad strokes of this look good to me; A few micro-bike shed suggestions for individual descriptions, and one tarpit around capitalisation.

freakboy3742:
> * Does the `-f` option serve a purpose any longer? or just remove it?

With this additional detail in the default help, I think we can probably remove the `-f` option. The only reason I can think to retain the option is if it were a way to generate machine readable output... but I'm not sure I've actually got a use case for this.

> * I limited the display width to 80 characters.
>   
>   * This is partially accomplished with `RawDescriptionHelpFormatter`....but manually elsewhere.

That makes sense to me.

src/briefcase/cmdline.py
mhsmith: I'm in favor of making the preferred capitalizations lower-case everywhere. This is a programming tool, and making things easy to type and to remember is more important than matching each platform's branding.

If we did this, then the pyproject.toml section name would have to become case insensitive for backward compatibility, if it isn't already.

Although it doesn't directly affect the Briefcase user, I would change the Python module names to be lower-case as well, for the same reasons as above, plus the fact that capital letters in Python module names are very unusual.

mhsmith: > then the pyproject.toml section name would have to become case insensitive for backward compatibility

And so would the app subdirectories.

src/briefcase/commands/update.py
rmartin16: Updated.

src/briefcase/commands/run.py
rmartin16: Updated

src/briefcase/commands/publish.py
rmartin16: Updated

src/briefcase/commands/package.py
rmartin16: Updated

src/briefcase/commands/open.py
rmartin16: Updated

src/briefcase/commands/create.py
rmartin16: Updated

src/briefcase/commands/build.py
rmartin16: Updated

src/briefcase/cmdline.py
rmartin16: Good call. Updated.

rmartin16:
Note on help text wrapping.....bit of a rabbit hole.

The text wrapping functionality from `argparse` is limited. Most notably, all whitespace is collapsed to a single space. To overcome this, `RawDescriptionHelpFormatter` and `RawTextHelpFormatter` can help maintain whitespace....but then all other formatting, including text wrapping to a width, is lost. So, the user is basically forced to perform all formatting themselves.

python/cpython#57015 has long documented this. python/cpython#22129 has even implemented a much more useful formatter for `argparse`....but it remains unmerged. [`argparse-formatter`](https://pypi.org/project/argparse-formatter/) was introduced to make this functionality available from a third-party. However, IMO, it's a bit fragile because it's overriding methods of `argparse.HelpFormatter` that all have this message:
```
    Only the name of this class is considered a public API. All the methods
    provided by the class are considered an implementation detail.
```

Sooo....I tried to find somewhat of a middle-ground in 68fce1d473981df9e7eebcd730a048e7e9db9ed4 without overdoing it...

src/briefcase/commands/base.py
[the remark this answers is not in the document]
rmartin16: > The Upgrade platform dependency is definitely an unintended side effect; I'm intrigued why that happens for Upgrade, but not for Dev (unless it does...?)

The same happens for `dev`...so, it'll need to be handled similar to `new` and `upgrade`.

> A lot of the complexity here stems from the way we are bending argparse to the limit so we can defer the construction of subcommand help until we know the actual platform. If there's a better solution to argparse (or a better way to use argparse), I'm not opposed to a major rework of the argument handling code.

We're definitely stretching `argparse`....although, I'm hesitant to revamp/replace it now...

src/briefcase/cmdline.py
freakboy3742: Ugh. Writing `macos` and `ios` is going to make my teeth itch... but you're probably right...

rmartin16: I support lowercase as well...but also:
- allowing case insensitive user input for the purposes of platform/format matching
- presenting the brand's capitalization in general information to the user
  - such as here when the list of supported platforms is shown; `iOS` is much more recognizable that `ios`.

(Of note, this is also an issue for specifying a specific tool in `briefcase upgrade [tool name]` and what's displayed from `briefcase upgrade --list`.)

Moving to case insensitivity for filepaths and TOML names for backwards compatibility is likely to be fraught with issues, IMO.

As pointed out in #1011, the TOML names are expected to be ASCII....but that enforcement isn't bulletproof. Nonetheless, if names supported in TOML remain only ASCII, then case folding for matching is much more reliable.

As for supporting case insensitivity for filepaths....this sounds like a ticking time bomb....for instance, considering how often these filepaths are passed to other programs. I think it would be better to force users to rename the directory or delete it to continue using Briefcase. This would be similar to how the transition from `~/.briefcase` to `~/.cache/briefcase` was managed.

Thoughts?

I can start assessing this in the meantime.

freakboy3742: Based on what you're describing, it sounds like you're more in the "liberal in what you accept, conservative in what you output" camp. If we display to the user that "Android" and "iOS" are two platform options, then should either be enforcing that "Android" and "iOS" are the literal required keys, or accept "Android", "android", "ANDROID", "AnDrOiD", and anything else you might imagine are valid, either at the command line, or in a configuration file.

I think that's an equally valid approach; my only concern is the UX confusion associated with showing "Android" as a supported platform, but the default tutorial using "android". It's a little thing, but it's a potential source of confusion when we usually need to teach new users to be paranoid about exact capitalisation.

Beyond the confusion factor, the only place this policy would make a substantial difference to end users is in the directories that are generated. My tooth-itch notwithstanding, I agree with @mhsmith that being a programmer tool gives us significant license to go all-in on lower case; but whatever we do, I think we should be consistent between the "preferred" capitalisation as displayed to end users, and the "generated" capitalisation that manifests on disk.

The one thing that might make our life a *little* easier is that macOS is case insensitive on HFS and APFS:
```
(venv3.10) rkm@eunectes tmp % touch hello
(venv3.10) rkm@eunectes tmp % touch Hello
(venv3.10) rkm@eunectes tmp % touch HELLO
(venv3.10) rkm@eunectes tmp % ls
hello
```

Based on a hacked-together test: if I manually move a current `macOS` folder to `macos`, then re-run `briefcase run`, the app runs without regenerating. So - we could switch to lower case for generated filenames for iOS and macOS almost transparently.

The only fly in this ointment is `windows/VisualStudio`. It's the only platform that currently generates upper case in its output folders. However, I suspect the VisualStudio backend isn't extensively used (it's mostly for our own internal benefit). On top of that, we broadly encourage thinking of the generated directories as emphemeral artefacts that shouldn't be user-modified or committed to version control, so having a new folder generated in this one case isn't the end of the world, IMHO.

rmartin16: > The one thing that might make our life a little easier is that macOS is case insensitive on HFS and APFS

This is a configuration option, though; APFS and HFS filesystems can be configured as case-sensitive when the disk is formatted as such. Additionally, APFS supports volumes that transparently behave as directories in the host filesystem but can have different case sensitivity.

> The only fly in this ointment is windows/VisualStudio

Given case insensitivity on mac isn't reliable...and for posterity...the top level build directories for `iOS` and `macOS` contain capitalization. And for `iOS`, the xcode directory ends up `Xcode`.

It wouldn't really be a microsoft product if something like case sensitivity for NTFS wasn't a true cluster....while the filesystem is technically case sensitive....many interactions with NTFS via Windows are case _insensitive_ as the semantics are controlled by Windows APIs which use devtime-controlled configuration... Throw in a ext4 network drive or WSL and it may be difficult to say what'll happen.

That said...

I'm really only advocating case insensitivity at the command line; everywhere else, including "names" in configuration files, should only support lowercase. Case insensitive TOML really seems like a nightmare.

This approach allows backwards compatibility for any scripted commands or muscle memory...and it's already implemented and released.

As for showing branded names to users, I'm fine if that isn't done when showing the literal expected values to the user. in the help text.

Another take on presenting the platforms:
```
Google mobile OS     Android (android)
Apple mobile OS      iOS (ios)
Apple desktop OS     macOS (macos)
Linux desktop OS     Linux (linux)
Browser WebAssembly  Web (web)
Windows desktop OS   Windows (windows)
```

While this definitely seems ridiculously verbose right now.....when `tvOS`, `watchOS`, or `wearOS` is (eventually) supported, it really may not be clear to someone what those platforms actually are.

freakboy3742: > > The one thing that might make our life a little easier is that macOS is case insensitive on HFS and APFS
> 
> This is a configuration option, though; APFS and HFS filesystems can be configured as case-sensitive when the disk is formatted as such. Additionally, APFS supports volumes that transparently behave as directories in the host filesystem but can have different case sensitivity.

True - but case-insensitive is the default, and won't be changed by most users.

> It wouldn't really be a microsoft product if something like case sensitivity for NTFS wasn't a true cluster....while the filesystem is technically case sensitive....many interactions with NTFS via Windows are case _insensitive_ as the semantics are controlled by Windows APIs which use devtime-controlled configuration... Throw in a ext4 network drive or WSL and it may be difficult to say what'll happen.
> 
> That said...
> 
> I'm really only advocating case insensitivity at the command line; everywhere else, including "names" in configuration files, should only support lowercase. Case insensitive TOML really seems like a nightmare.

I'm not convinced it would be *that* hard - when we import the TOML, we convert any key to lowercase; and when we look up the key, we cast to lowercase. It's not *nothing*, but it's not hideously complex either.

However, even if *did* adopt case insensitive TOML, that doesn't address the filesystem problem. The issue is that are existing projects with "iOS", "macOS", "Xcode" and "VisualStudio" directories.
 
> This approach allows backwards compatibility for any scripted commands or muscle memory...and it's already implemented and released.
> 
> As for showing branded names to users, I'm fine if that isn't done when showing the literal expected values to the user. in the help text.

I think we're getting tied in knots here. 

AFAICT, we've got 3 options:

1. We adopt lower case universally, and make the code adaptive to different case options (both in configuration files and with existing directories on disk)
2. We adopt lower case universally, but accept the backwards incompatibility caused by existing configuration files
3. We keep the existing "preferred case" handling, and use the preferred case wherever it's displayed to the user. 

(3) requires the least work (read: no work outside making the help text display the preferred case); but it means the list of platforms in help text are `android, iOS, macOS, windows, ...`. 

There's possibly a continuum between (1) and (2), if a partial backwards compatibility solution is acceptable (e.g., case insensitive configuration files, but no migration path for existing directories on disk). 

Honestly, at this point, I'd advocate for (3). Making a bigger push on normalising to lower case might be worthwhile, but I'd argue it's a separate issue that should be tackled as a separate ticket. We can dramatically improve the help options without also requiring a radical change to case handling.

> Another take on presenting the platforms:
> 
> ```
> Google mobile OS     Android (android)
> Apple mobile OS      iOS (iOS)
> Apple desktop OS     macOS (macOS)
> Linux desktop OS     Linux (linux)
> Browser WebAssembly  Web (web)
> Windows desktop OS   Windows (windows)
> ```
> 
> While this definitely seems ridiculously verbose right now.....when `tvOS`, `watchOS`, or `wearOS` is (eventually) supported, it really may not be clear to someone what those platforms actually are.

I'm not sure this would be an improvement. Nobody calls it "Google mobile OS", and it's not clear to me what "Android (android)" is trying to convey.

rmartin16: > > I'm really only advocating case insensitivity at the command line; everywhere else, including "names" in configuration files, should only support lowercase. Case insensitive TOML really seems like a nightmare.
> 
> I'm not convinced it would be _that_ hard - when we import the TOML, we convert any key to lowercase; and when we look up the key, we cast to lowercase. It's not _nothing_, but it's not hideously complex either.

My concern here could arguably be considered hyperbole or overblown but it stems from corner cases:

1) Without context, Python's `lower()` is locale-unaware; therefore, it can produce the wrong lowercase letter in certain cases. An recent example, `'Isparta'.lower()` is `'isparta'` instead of `'ısparta'`. This can be managed relatively easily for the purposes of lookup as long as the same case function is used on both sides...but user edits to TOML can break that symmetry.

2) Collapsing non-unique keys in the TOML can create collisions. Unless detected, this could easily lead to confusion.

> Honestly, at this point, I'd advocate for (3). Making a bigger push on normalising to lower case might be worthwhile, but I'd argue it's a separate issue that should be tackled as a separate ticket. We can dramatically improve the help options without also requiring a radical change to case handling.

Good deal; I'm on board.

rmartin16:
To avoid the shenanigans with identifying the default format for a platform, I updated `DEFAULT_OUTPUT_FORMAT` for iOS to match the casing of `iOSXcodePassiveMixin.output_format`. Other than that, I revised some of the descriptions so they are more similar to each other.

freakboy3742 (approved):
Some gnarly argparse wrestling, but ultimately fairly straightforward - nice work!

PR #1007, opened 2022-12-14
Install Rosetta if necessary

### LINKED ISSUE #1000: Android build on ARM Mac without Rosetta gives error "Bad CPU type in executable"
### Describe the bug

Briefcase installs an x86 version of Java, because there's no ARM build available for Java 8 (see #827 for why we're still on that version). But the x86 build won't run unless Rosetta is installed.

### Steps to reproduce

On an ARM Mac which doesn't have Rosetta, follow the Android section of the BeeWare tutorial.

### Expected behavior

Build succeeds.

### Screenshots

_No response_

### Environment

- Operating System: macOS 12.6
- Python version: 3.11
- Software versions:
  - Briefcase: 0.3.11

### Logs

```
% briefcase create android

[java] The Java JDK was not found; downloading and installing...
Downloading OpenJDK8U-jdk_x64_mac_hotspot_8u242b08.tar.gz...
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.0% • 00:17
Installing AdoptOpenJDK... done

[android_sdk] The Android SDK was not found; downloading and installing...
Downloading commandlinetools-mac-8092744_latest.zip...
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.0% • 00:20
Installing Android SDK Command-Line Tools... done

The Android tools provided by Google have license terms that you must accept
before you may use those tools.
/Users/msmith/Library/Caches/org.beeware.briefcase/tools/android_sdk/cmdline-tools/latest/bin/sdkmanager: line 172: /Users/msmith/Library/Caches/org.beeware.briefcase/tools/java/Contents/Home/bin/java: Bad CPU type in executable
/Users/msmith/Library/Caches/org.beeware.briefcase/tools/android_sdk/cmdline-tools/latest/bin/sdkmanager: line 172: /Users/msmith/Library/Caches/org.beeware.briefcase/tools/java/Contents/Home/bin/java: Undefined error: 0

Error while reviewing Android SDK licenses. Please run this command and examine
its output for errors.

    $ /Users/msmith/Library/Caches/org.beeware.briefcase/tools/android_sdk/cmdline-tools/latest/bin/sdkmanager --licenses

Log saved to /Users/msmith/git/beeware/helloworld/logs/briefcase.2022_12_12-17_06_16.create.log
```
[briefcase.2022_12_12-17_06_16.create.log](https://github.com/beeware/briefcase/files/10210494/briefcase.2022_12_12-17_06_16.create.log)

### Additional context

As shown [here](https://apple.stackexchange.com/questions/408375), this can be fixed by running `softwareupdate --install-rosetta`. Briefcase should do this automatically.

mhsmith:
Fixes #1000.

## PR Checklist:

### DIFF
--- a/changes/1000.bugfix.rst
+++ b/changes/1000.bugfix.rst
@@ -1,0 +1,1 @@
+On MacOS, Rosetta is now installed automatically if needed.
--- a/changes/1000.bugfix.rst
+++ b/changes/1000.bugfix.rst
@@ -1,1 +1,1 @@
-On MacOS, Rosetta is now installed automatically if needed.
+On macOS, Rosetta is now installed automatically if needed.
--- a/src/briefcase/integrations/java.py
+++ b/src/briefcase/integrations/java.py
@@ -70,7 +70,7 @@ class JDK(Tool):
         install_message = None
 
         if tools.host_arch == "arm64" and tools.host_os == "Darwin":
-            # Java 8 is not available for MacOS on ARM64, so we will require Rosetta.
+            # Java 8 is not available for macOS on ARM64, so we will require Rosetta.
             cls.verify_rosetta(tools)
 
         # macOS has a helpful system utility to determine JAVA_HOME. Try it.
--- a/src/briefcase/integrations/java.py
+++ b/src/briefcase/integrations/java.py
@@ -69,6 +69,10 @@ class JDK(Tool):
         java_home = tools.os.environ.get("JAVA_HOME", "")
         install_message = None
 
+        if tools.host_arch == "arm64" and tools.host_os == "Darwin":
+            # Java 8 is not available for MacOS on ARM64, so we will require Rosetta.
+            cls.verify_rosetta(tools)
+
         # macOS has a helpful system utility to determine JAVA_HOME. Try it.
         if not java_home and tools.host_os == "Darwin":
             try:
@@ -284,3 +288,24 @@ Delete {jdk_zip_path} and run briefcase again.
 
         self.uninstall()
         self.install()
+
+    @classmethod
+    def verify_rosetta(cls, tools):
+        try:
+            tools.subprocess.check_output(
+                ["arch", "-x86_64", "true"], stderr=subprocess.STDOUT
+            )
+        except subprocess.CalledProcessError:
+            tools.logger.info(
+                """\
+This command requires Rosetta, but it does not appear to be installed. Briefcase will
+attempt to install it now.
+"""
+            )
+            try:
+                tools.subprocess.run(
+                    ["softwareupdate", "--install-rosetta", "--agree-to-license"],
+                    check=True,
+                )
+            except subprocess.CalledProcessError as e:
+                raise BriefcaseCommandError("Failed to install Rosetta") from e
--- a/src/briefcase/platforms/android/gradle.py
+++ b/src/briefcase/platforms/android/gradle.py
@@ -112,9 +112,30 @@ class GradleMixin:
             / "app-release.aab"
         )
 
-    def gradlew_path(self, app):
+    def run_gradle(self, app, args):
+        # Gradle may install the emulator via the dependency chain build-tools > tools >
+        # emulator. (The `tools` package only shows up in sdkmanager if you pass
+        # `--include_obsolete`.) However, the old sdkmanager built into Android Gradle
+        # plugin 4.2 doesn't know about macOS on ARM, so it'll install an x86_64 emulator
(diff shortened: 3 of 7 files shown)


mhsmith:
@freakboy3742: I still need to adjust the Android emulator install timing, but the Rosetta installation is now complete and tested, so comments are welcome on that.

For testing, I was able to uninstall Rosetta by following the instructions [here](https://iboysoft.com/news/uninstall-rosetta-2.html), except:
* One of the paths to delete should be /Library/Apple/usr/libexec/orh, not /Library/Apple/usr/libexec.
* Emptying the trash was not necessary.

src/briefcase/integrations/java.py
```diff
@@ -284,3 +288,24 @@ def upgrade(self):
 
         self.uninstall()
         self.install()
+
+    @classmethod
+    def verify_rosetta(cls, tools):
+        try:
+            tools.subprocess.check_output(
+                ["arch", "-x86_64", "true"], stderr=subprocess.STDOUT
```
freakboy3742, line 296: Does this reliably test CPU architecture? `true` is a shell builtin, so I wouldn't have thought it will trigger the CPU check. Running `/bin/bash -version` might be a better option - with the added benefit that as debug, it leaves a paper trail:
```
(venv3.10) rkm@eunectes testbed % arch -arm64 /bin/bash -version 
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin21)
Copyright (C) 2007 Free Software Foundation, Inc.
(venv3.10) rkm@eunectes testbed % arch -x86_64 /bin/bash -version 
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin21)
Copyright (C) 2007 Free Software Foundation, Inc.
```

changes/1000.bugfix.rst
```diff
@@ -0,0 +1 @@
+On MacOS, Rosetta is now installed automatically if needed.
```
freakboy3742, line 1: Nitpick:
```suggestion
On macOS, Rosetta is now installed automatically if needed.
```

src/briefcase/integrations/java.py
```diff
@@ -69,6 +69,10 @@ def verify(cls, tools: ToolCache, install=True):
         java_home = tools.os.environ.get("JAVA_HOME", "")
         install_message = None
 
+        if tools.host_arch == "arm64" and tools.host_os == "Darwin":
+            # Java 8 is not available for MacOS on ARM64, so we will require Rosetta.
```
freakboy3742, line 73: Nitpick:
```suggestion
            # Java 8 is not available for macOS on ARM64, so we will require Rosetta.
```

freakboy3742 (changes requested):
Looks good; a couple of minor capitalisation nitpicks and a question about the "is Rosetta installed" check.

src/briefcase/integrations/java.py
mhsmith: Most shell builtins that can exist as separate executables, actually do: 
```
(bw-311) msmith@mbp briefcase % where echo
echo: shell built-in command
/bin/echo
(bw-311) msmith@mbp briefcase % where true
true: shell built-in command
/usr/bin/true
```

Since `arch` doesn't run a shell, it'll try to run /usr/bin/true. And with Rosetta uninstalled, the check fails correctly:
```
(bw-311) msmith@mbp briefcase % arch -x86_64 true
arch: posix_spawnp: true: Bad CPU type in executable
(bw-311) msmith@mbp briefcase % echo $?
1
```

freakboy3742 (approved):
Looks good. 

The only nitpick/design alternative I can think of is whether `verify_emulator()` should be invoked as part of `AndroidSDK.verify()` instead of being a gradle prerequisite. There might be a (minor) UX improvement by getting all the big downloads out of the way right at the start; but (a) the gradle build involves a big NDK download anyway (at least for now), and (b) I guess there's value in being explicit about the emulator before invoking gradle, just in case something has modified the SDK after initial installation has completed.

I'm happy for this to land as is; just throwing this out there in case you were having similar thoughts.

mhsmith:
As discussed, I considered putting it in `AndroidSDK.verify`, but decided that since the issue is caused by Gradle, it made more sense to deal with it immediately before running Gradle. This means that the change won't affect commands that never run Gradle (e.g. `create`), or commands that normally do run Gradle (e.g. `build`), but fail at an earlier stage for any reason.

PR #1041, opened 2023-01-10
Dependency pins

freakboy3742:
Modify the usage of dependency pins:

* Runtime dependencies are all specified as open ended, with a minimum version. Packages using semver also have an upper version. Packages that are part of the "core python toolchain" (e.g., wheel) do not have an upper pin as we always want the most recent versions those tools. Packages that use calver have no upper pin on the basis that there is no rationale for an upper pin as a basis for API stability.
* Development dependencies are are hard-pinned to a specific version. This ensures development environment consistency.

Updating to the most recent version of Sphinx caused a version incompatibility with the RTD theme; since we've planning to move to Furo on Toga anyway, I figured this is as good a time as any to make that change.

Refs #1026.
Refs #1027.
Refs #1037.

## PR Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -76,7 +76,6 @@ dev =
     pre-commit == 2.21.0
     setuptools_scm[toml] == 7.0.5
     tox == 4.0.1
-test =
     pytest == 7.2.0
     pytest-tldr == 0.2.5
     coverage[toml] == 7.0.4
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,45 +56,50 @@ include_package_data = True
 package_dir =
     = src
 install_requires =
+    # Dependencies required at runtime are set as ranges to ensure maximum
+    # compatibility with the end-user's development environment.
+    #
     # Any dependency that is part of the "core python toolchain" (e.g. pip,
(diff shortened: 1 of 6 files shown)


setup.cfg
```diff
@@ -58,36 +58,35 @@ package_dir =
...
+    pre-commit == 2.21.0
+    setuptools_scm[toml] == 7.0.5
+    tox == 4.0.1
 test =
-    pytest~=7.2
-    pytest-tldr~=0.2
-    pytest-cov~=4.0
+    pytest == 7.2.0
+    pytest-tldr == 0.2.5
+    coverage[toml] == 7.0.4
```
mhsmith, line 82: A development environment should probably have all the test tools installed, so Toga will merge`[test]` into `[dev]`as soon as the setuptools-scm PR is merged (https://github.com/beeware/toga/pull/1675#pullrequestreview-1178871290). Would it make sense to do the same for Briefcase?

freakboy3742: Yeah - the distinction is pretty narrow, so I guess that makes sense, if only for simplicity.

mhsmith (approved):

PR #1052, opened 2023-01-19
Remove transitory obsolete cache directory check

rmartin16:
Removes the check for the old Briefcase cache directory at `~/.briefcase` introduced in #777.

~~This also appears to have removed the last usage of `briefcase.exceptions.InfoHelpText` but I didn't remove the exception since it may be useful in the future as it has been in the past.~~

Well...we're failing coverage with `InfoHelptText`....so, I guess it'll have to be recreated if it's needed again.

## PR Checklist:


freakboy3742 (approved):
Can't argue with the TODO comment :-)

PR #1054, opened 2023-01-21
Ensure cache directory exists

### LINKED ISSUE #922: Accommodate AppData Redirection for Windows Store Python
**Describe the bug**
Windows will redirect interaction with `%LOCALAPPDATA%` when Python is installed from the Windows Store. This means that the `briefcase` data directory will be created at somewhere like `%LOCALAPPDATA%\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\Local\BeeWare` instead of `%LOCALAPPDATA%\BeeWare`.

**To Reproduce**
1. Ensure `%LOCALAPPDATA%\BeeWare` directory does not exist.
2. Run `briefcase build android` using Python from the Windows Store
3. See error on trying to verify licenses

**Expected behavior**
Briefcase should prevent this redirection or incorporate it in to sourcing the `briefcase` data directory.

Of note, if the `%LOCALAPPDATA%\BeeWare` exists, this redirection will not happen.

**Environment:**
 - Operating System: Windows 10 1903 and later
 - Python version: 3.10
 - Software versions:
   - Briefcase: 0.3.11

**Additional context**
https://github.com/python/cpython/issues/84557
https://docs.microsoft.com/en-us/windows/msix/desktop/desktop-to-uwp-behind-the-scenes#file-system

## Workarounds
- Use the public [Python installer](https://www.python.org/downloads/)
- Move the `%LOCALAPPDATA%\Packages\<Python>\LocalCache\Local\BeeWare` directory to `%LOCALAPPDATA%\BeeWare`

rmartin16:
## Changes
- On Windows, create the cache directory with `cmd` to bypass the filesystem redirection of `%LOCALAPPDATA%` enforced for Python installed from the Windows Store.
- For Linux and macOS, the directory is created via `os.makedirs()`.
- Run the unit tests against the latest Windows Store Python.

## Related Issues
- Fixes #922

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
   unit-tests:
     name: Unit tests
     needs: [pre-commit, towncrier, package]
-    runs-on: ${{ fromJSON('{"winstore":"windows"}')[matrix.platform] || matrix.platform }}-latest
+    runs-on: ${{ matrix.platform }}-latest
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
@@ -86,23 +86,23 @@ jobs:
         - python-version: "3.12-dev"
           experimental: true
         # Run tests against the latest Windows Store Python
-        - platform: "winstore"
-          python-version: "3.11"
+        - platform: "windows"
+          python-version: "winstore"
           experimental: false
     steps:
     - uses: actions/checkout@v3.3.0
       with:
         fetch-depth: 0
     - name: Set up Python
-      if: matrix.platform != 'winstore'
+      if: matrix.python-version != 'winstore'
       uses: actions/setup-python@v4.5.0
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install Windows Store Python
-      if: matrix.platform == 'winstore'
+      if: matrix.python-version == 'winstore'
       uses: beeware/.github/.github/actions/install-win-store-python@main
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: 3.11
     - name: Get packages
       uses: actions/download-artifact@v3
       with:
@@ -132,7 +132,7 @@ jobs:
   coverage:
     name: Combine & check coverage.
     runs-on: ubuntu-latest
-    needs: unit-tests
+    needs: [unit-tests]
     steps:
     - uses: actions/checkout@v3.3.0
       with:
@@ -168,7 +168,7 @@ jobs:
 
   verify-apps:
     name: Build apps
-    needs: unit-tests
+    needs: [unit-tests]
     runs-on: ${{ matrix.platform }}
     strategy:
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
   workflow_call:
 
 env:
-  python_version: '3.9'
+  python_version: "3.9"
 
 defaults:
   run:
@@ -75,23 +75,32 @@ jobs:
   unit-tests:
     name: Unit tests
     needs: [pre-commit, towncrier, package]
+    runs-on: ${{ fromJSON('{"winstore":"windows"}')[matrix.platform] || matrix.platform }}-latest
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
         platform: ["macos", "ubuntu", "windows"]
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12-dev"]
         include:
-          - experimental: false
-
-          - python-version: "3.12-dev"
-            experimental: true
-    runs-on: ${{ matrix.platform }}-latest
-    continue-on-error: ${{ matrix.experimental }}
+        - experimental: false
+        - python-version: "3.12-dev"
+          experimental: true
+        # Run tests against the latest Windows Store Python
+        - platform: "winstore"
+          python-version: "3.11"
+          experimental: false
     steps:
     - uses: actions/checkout@v3.3.0
       with:
         fetch-depth: 0
     - name: Set up Python
+      if: matrix.platform != 'winstore'
       uses: actions/setup-python@v4.5.0
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install Windows Store Python
+      if: matrix.platform == 'winstore'
+      uses: beeware/.github/.github/actions/install-win-store-python@main
       with:
         python-version: ${{ matrix.python-version }}
     - name: Get packages
@@ -123,8 +132,7 @@ jobs:
   coverage:
     name: Combine & check coverage.
     runs-on: ubuntu-latest
-    needs:
-    - unit-tests
+    needs: unit-tests
     steps:
     - uses: actions/checkout@v3.3.0
       with:
@@ -132,7 +140,7 @@ jobs:
     - uses: actions/setup-python@v4.5.0
       with:
         # Use latest, so it understands all syntax.
-        python-version: "3.10"
+        python-version: '3.x'
     - name: Install dev dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -75,25 +75,35 @@ jobs:
   unit-tests:
     name: Unit tests
     needs: [pre-commit, towncrier, package]
+    runs-on: ${{ matrix.platform }}-latest
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
         platform: ["macos", "ubuntu", "windows"]
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12-dev"]
+        use-win-store-python: [ false ]
```
freakboy3742, line 84: Looking at an alternate way to structure this - rather than adding `use-win-store-python` to the build matrix, could we add `windows-winstore` as a platform (excluding 3.8, 3.9 etc from the matrix) The setup you have here works, but it renders the target platforms as "macOS, 3.8, false", with the winstore build being "windows, 3.11, true, false".

.github/workflows/ci.yml
```diff
@@ -75,25 +75,35 @@ jobs:
...
       with:
         fetch-depth: 0
     - name: Set up Python
+      if: matrix.use-win-store-python == false
       uses: actions/setup-python@v4.5.0
       with:
         python-version: ${{ matrix.python-version }}
+    - name: Install Windows Store Python
+      if: matrix.use-win-store-python == true
+      uses: ./.github/actions/install-win-store-python
```
freakboy3742, line 104: Presumably this would be updated to point at the `.github` repo version when beeware/.github#5 lands?

src/briefcase/commands/base.py
```diff
@@ -213,6 +215,39 @@ def validate_data_path(self, data_path):
...
+                if platform.system() == "Windows":
+                    subprocess.run(
+                        ["mkdir", data_path],
+                        shell=True,
+                        check=True,
+                        stdout=subprocess.DEVNULL,
+                        stderr=subprocess.DEVNULL,
+                    )
+                else:
+                    os.makedirs(data_path, exist_ok=True)
```
freakboy3742, line 235: Is this strictly needed? Presumably the data path will be created on first usage - this is more of a convenience/ensuring that all platforms are behaviourally consistent?

src/briefcase/commands/base.py
```diff
@@ -213,6 +215,39 @@ def validate_data_path(self, data_path):
...
+            except (subprocess.CalledProcessError, OSError):
+                raise BriefcaseCommandError(
+                    f"""
+Failed to create the Briefcase directory to store tools and support files:
+
+    {data_path}
+
+Create this directory manually and re-run Briefcase.
```
freakboy3742, line 243: Why this advice? Surely if Briefcase can't create the folder, manually creating it will just mask future permission-related problems?

tests/commands/base/test_paths.py
```diff
@@ -9,6 +10,39 @@
...
 
 
+def test_path_is_realpath(tmp_path):
+    """Briefcase's data path matches its realpath.
+
+    For the Windows Store Python, filesystem interaction with
+    `%LOCALAPPDATA%` can be redirected to a sandboxed location. However,
+    `os.path.realpath()` will reveal such a redirection.
+    """
+    # %LOCALAPPDATA% seemingly isn't available in GH CI...
```
freakboy3742, line 20: Is that an indicator of a difference between Windows 11 vs Windows Server? Is that something we need to accomodate more generally?

freakboy3742: Or - will it cause failures *outside* GH CI?

tests/commands/base/test_paths.py
```diff
@@ -9,6 +10,39 @@
...
+    if user_dir := os.environ.get("USERPROFILE"):
+        tmp_path = Path(user_dir) / "AppData/Local"
+    data_path = tmp_path / "realpathbase"
+    command = DummyCommand(data_path=data_path)
+
+    assert command.data_path == Path(os.path.realpath(data_path))
```
freakboy3742, line 26: This test seems a bit broad/weak; if you're not on Windows, I t isn't needed, and if you *are* on Windows, it's dependent on the existence of the `USERPROFILE` variable.

Could we:
1. lock this down to a Windows-only test;
2. Add an explicit failure on Windows if the expected environment variable isn't present
3. Use try/finally to do cleanup, so we don't leave dirty state in the user's home folder

Also - is LOCALAPPDATA a complete substitute for USERPROFILE?

freakboy3742 (changes requested):
Nice work on the CI config - it looks like we might be able to use a similar technique to factor out a number of other common CI tasks (like towncrier, pre-check). 

A couple of questions/comments inline.

.github/workflows/ci.yml
rmartin16: Agreed; I've been doing a lot of fighting with github's matrix configuration....and landed here for testing. Using a new platform does make the most sense....but we're driving the runner OS off of it and string manipulation isn't really supported in the yaml. I'll see what I can do.

src/briefcase/commands/base.py
rmartin16: That's correct; I only added this to keep behavior the same on all platforms.

.github/workflows/ci.yml
rmartin16: The absence of string manipulation support in github action expressions is bit annoying. But I found actions/runner#409 and it helped arbitrate the `runs-on` value. Open to critique.

```
runs-on: ${{ fromJSON('{"winstore":"windows"}')[matrix.platform] || matrix.platform }}-latest
```

.github/workflows/ci.yml
rmartin16: Correct.

src/briefcase/commands/base.py
rmartin16: Yeah, it's hard to know why this directory creation would fail in the real world. Right now, if this directory couldn't be created, it would just bomb out. However, I'd expect a knowledgeable user to be able to diagnose and resolve the issue without further problems.

I'd be fine updating the message just to state the problem though and leave it at that. (Although, leaving the note about `BRIEFCASE_HOME` I think makes sense.)

freakboy3742: Yeah - that would be my inclination as well. Unless we have a reason to believe that the advice will help, it's better to have people come to us with a specific problem, rather than something they've tried to ... fix  :-)

tests/commands/base/test_paths.py
```diff
@@ -9,6 +10,43 @@
...
+def test_path_is_realpath(tmp_path):
+    """Briefcase's data path matches its realpath.
+
+    For the Windows Store Python, filesystem interaction with
+    `%LOCALAPPDATA%` can be redirected to a sandboxed location. However,
+    `os.path.realpath()` will reveal such a redirection.
+    """
+    from pprint import pprint
+
+    pprint(dict(os.environ))
```
freakboy3742, line 23: Looks like stray debugging?

tests/commands/base/test_paths.py
rmartin16: Tracked it down; `tox` strips many environment variables from the environment the tests run in. I explicitly passed it through for the unit tests.

src/briefcase/commands/base.py
rmartin16: Makes sense. Removed remediation recommendations.

tests/commands/base/test_paths.py
rmartin16: Removed.

tests/commands/base/test_paths.py
rmartin16: Addressed. The test only runs on Windows now, will throw `KeyError` is `LOCALAPPDATA` isn't available, and deletes the directory.

.github/workflows/ci.yml
rmartin16: Updated now.

freakboy3742 (approved):
This looks great. I've got one idea about the CI config that I think will clean it up even more (and make the winstore target more obvious - I misread "winstore" as "windows" in my first pass). I'll have a quick poke to see if I can make it work; but otherwise I think this is good to go!

PR #1058, opened 2023-01-23
Use sdists rather than filesystem mounts of Appimage local package references

### LINKED ISSUE #1051: Appimage: can't install local `requires` directories which require extra context
https://github.com/beeware/briefcase/pull/993 allowed AppImage to support most local `requires` directories by adding each one as a volume mount when running pip within Docker. However, the build will still fail if the directory's build script needs to read anything outside of the directory.

For example, see [this job](https://github.com/beeware/toga/actions/runs/3961028678/jobs/6786354567) in the [Toga setuptools-scm PR](https://github.com/beeware/toga/pull/1675). Setuptools-scm needs to run within a Git repository in order to detect the version number. But the volume mount doesn't include the whole repository, only a subdirectory of it. This produces the error:
```
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [26 lines of output]
        File "/home/brutus/.local/lib/python3.11/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 351, in <module>
          main()
        File "/home/brutus/.local/lib/python3.11/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 333, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/home/brutus/.local/lib/python3.11/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-kp9qyb0c/overlay/local/lib/python3.11/dist-packages/setuptools/build_meta.py", line 338, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-kp9qyb0c/overlay/local/lib/python3.11/dist-packages/setuptools/build_meta.py", line 320, in _get_build_requires
          self.run_setup()
        File "/tmp/pip-build-env-kp9qyb0c/overlay/local/lib/python3.11/dist-packages/setuptools/build_meta.py", line 335, in run_setup
          exec(code, locals())
        File "<string>", line 5, in <module>
        File "/tmp/pip-build-env-kp9qyb0c/overlay/local/lib/python3.11/dist-packages/setuptools_scm/__init__.py", line 148, in get_version
          _version_missing(config)
        File "/tmp/pip-build-env-kp9qyb0c/overlay/local/lib/python3.11/dist-packages/setuptools_scm/__init__.py", line 108, in _version_missing
          raise LookupError(
      LookupError: setuptools-scm was unable to detect version for /requirements.
      
      Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.
```
The simplest solution I can think of is to just mount the entire host filesystem into the Docker container, e.g. at the path /host, and use that instead of one mount per local requirement.

freakboy3742:
Modifies the AppImage backend to use sdists rather than volume mounts to support local packages.

The previous implementation created a Docker volume mount for each local package reference in the requires list. However as reported by #1051, if a package uses git checkout information, but the installable path doesn't include the git repository details (e.g., toga, where the installable packages are not in the project root), the Docker volume mount doesn't contain enough detail to install the package.

This approach inspects the list of requirements prior to installation. Any local file reference is built into an sdist (using `build`) in the *local* environment, and stored in a `_local` folder in the project path; the pip install list is modified to reference the sdist instead of a local filesystem path. 

sdists are platform independent; barring a platform-specific setup.py configuration, it should be possible to build an sdist on macOS and then install it inside Docker.

Some implementation notes:
1. Although this removes the need for configurable Docker volume mounts, I've retained the core functionality in case it is needed in future.
2. `build` *can* be invoked programatically, rather than over subprocess. However, I've opted to use subprocess because the implementation of build uses `logging`, which doesn't play nicely with our logging/streaming/wait_bar wrappers (at least, not without a lot of additional work). Using subprocess means we can just use the externally exposed stdout content, rather than needing to work with a logging configuration that isn't strictly part of the public `build` API.
3. I've swallowed the output of generating the sdist in the wait-bar. Build output is captured in the log, but isn't surfaced for the normal successful build use case.
4. Although sdists aren't actually required to build AppImages in `--no-docker` mode, I've used them anyway. This is partially for consistency; but also because a user who created the project with `--no-docker`, then ran build *with* docker, would end up in a situation where the requirements "weren't installed", because the sdists hadn't been compiled.

Fixes #1051.

## PR Checklist:


src/briefcase/platforms/linux/appimage.py
```diff
@@ -157,49 +140,77 @@ def support_package_url(self, support_revision):
...
-                final.append(f"/requirements/{source.name}")
-            else:
-                # Use the requirement as-specified
-                final.append(requirement)
-
-        return final
+                # Build an sdist for the local requirement
+                with self.input.wait_bar(f"Building sdist for {requirement}..."):
+                    self.tools.subprocess.check_output(
+                        [
```
mhsmith, line 174: > I've swallowed the output of generating the sdist in the wait-bar. Build output is captured in the log, but isn't surfaced for the normal successful build use case.

The first time I tried this with the testbed app, I forgot to install the new requirement for `build`. The correct error message did appear on screen, albeit on the same line as the wait bar:
```
[testbed] Updating requirements...
     ━━━━━━━━━━━━━━━━━━━━ Building sdist for ../core.../home/smith/.venv/bw-38/bin/python3.8: No module named build
Building sdist for ../core...
```
But it didn't appear anywhere in the log:
```
           >>> Running Command:                                                                                                                                    subprocess.py:660
           >>>     /home/smith/.venv/bw-38/bin/python3.8 -m build --sdist --outdir '/home/smith/git/beeware/toga/testbed/linux/appimage/Toga Testbed/_local'       subprocess.py:661
           ../core                                                                                                                                                                  
           >>> Working Directory:                                                                                                                                  subprocess.py:668
           >>>     /home/smith/git/beeware/toga/testbed                                                                                                            subprocess.py:669
           >>> Return code: 1                                                                                                                                      subprocess.py:696
           Building sdist for ../core...                                                                                                                             appimage.py:173

Main thread traceback:
...
```

Later, when I discovered the problem with existing sdists or wheels, that error message appeared in the log but not on the screen. Maybe the difference is down to stdout vs stderr?

src/briefcase/platforms/linux/appimage.py
```diff
@@ -33,6 +34,9 @@ def appdir_path(self, app):
     def project_path(self, app):
         return self.bundle_path(app)
 
+    def local_requirements_path(self, app):
+        return self.bundle_path(app) / "_local"
```
mhsmith, line 38: `_requirements` would be a more descriptive name.

src/briefcase/platforms/linux/appimage.py
```diff
@@ -157,49 +140,77 @@ def support_package_url(self, support_revision):
...
+        for requirement in requires:
             if _is_local_requirement(requirement):
-                source = Path(requirement).resolve()
-                final.append(f"/requirements/{source.name}")
-            else:
-                # Use the requirement as-specified
-                final.append(requirement)
-
-        return final
+                # Build an sdist for the local requirement
```
mhsmith, line 171: `_is_local_requirement` will also return true for existing sdist or wheel filenames. Attempting to pass these to `build` doesn't make any sense, and will fail with the error `<filename> is not a directory`.

The easiest way to pass these files into Docker is probably to copy them into the same directory as the sdists we build ourselves.

freakboy3742: Good catch. Copying non-folders to the requirements folder makes sense; I'll do that.

src/briefcase/platforms/linux/appimage.py
freakboy3742: stdout/err is definitely the cause here. @rmartin16 has gotten ahead of this and submitted #1059; but even with that change, I can improve the error reporting here.

src/briefcase/platforms/linux/appimage.py
freakboy3742: Yeah - I went round in circles a few times on this name. It's a bike shed; `_requirements` works for me.

freakboy3742:
@mhsmith I've merged in @rmartin16's stderr change, and addressed the other issues you've raised. I've also done a refactor of the create tests - adding extra tests for the pieces you flagged made the test setup duplication a little more acute. Ready for another review pass at your leisure.

src/briefcase/platforms/linux/appimage.py
```diff
@@ -157,49 +140,93 @@ def support_package_url(self, support_revision):
...
+            app_packages_path=app_packages_path,
+        )
 
-    def _pip_kwargs(self, app: AppConfig):
-        """Generate the kwargs to pass when invoking pip.
+    def _pip_requires(self, app: AppConfig, requires: List[str]):
+        """Convert the requirements list to an AppImage compatible format.
 
-        Adds any mounts needed to support local file requirements.
+        Any local file requirements are converted into a reference to the sdist
```
mhsmith, line 210:
```suggestion
        Any local file requirements are converted into a reference to the file
```

src/briefcase/platforms/linux/appimage.py
```diff
@@ -157,49 +140,93 @@ def support_package_url(self, support_revision):
...
         """
-        kwargs = super()._pip_kwargs(app)
+        # Copy all the requirements that are non-local
+        final = [
+            requirement
+            for requirement in super()._pip_requires(app, requires)
+            if not _is_local_requirement(requirement)
+        ]
+
+        # Add in the sdist versions of any local packages.
```
mhsmith, line 224:
```suggestion
        # Add in any local packages.
```

mhsmith (approved):

PR #1059, opened 2023-01-23
Default `stderr=subprocess.STDOUT` for `check_output`

rmartin16:
## Changes
Briefcase's use of `subprocess.check_output()` is practically exclusively intended to run a command without any output to the console. By default, though, `check_output()` will only return the value of `stdout` to the caller while `stderr` flows to the console.

This PR updates `check_output()` to default coalescing `stderr` into `stdout` to better align with how Briefcase uses it.

The prompt for pursuing this was @mhsmith's [comment](https://github.com/beeware/briefcase/pull/1058/files#r1084323585) on a recent PR showing console corruption from such an instance of `stderr` leaking in to the console while the console was controlled.

Of the 22 current uses of `check_output()`, all but the 5 below were already setting `stderr=subprocess.STDOUT` manually. The non-xcode calls are quite unlikely to produce output for `stderr`. I am less familiar with the `xcode` calls; though, I'm inclined to believe similarly.
```
src/briefcase/integrations/flatpak.py:            output = tools.subprocess.check_output(["flatpak", "--version"]).strip("\n")

src/briefcase/integrations/flatpak.py:            output = tools.subprocess.check_output(
src/briefcase/integrations/flatpak.py:                ["flatpak-builder", "--version"]
src/briefcase/integrations/flatpak.py:            ).strip("\n")

src/briefcase/integrations/xcode.py:        output = tools.subprocess.check_output(
src/briefcase/integrations/xcode.py:            ["security", "find-identity", "-v", "-p", policy],
src/briefcase/integrations/xcode.py:        )

src/briefcase/platforms/iOS/xcode.py:                output = self.tools.subprocess.check_output(
src/briefcase/platforms/iOS/xcode.py:                    ["xcrun", "simctl", "launch", udid, app_identifier]
src/briefcase/platforms/iOS/xcode.py:                )

src/briefcase/platforms/linux/appimage.py:            base_path = self.tools[app].app_context.check_output(echo_cmd).strip()
```

## Considerations
The nature of the handling of `stderr` within `subprocess.check_output()` means that this defaulting cannot be undone; that is, `stderr` cannot now be configured in such a way to show it in the console. (Although, on second thought, setting `stderr=None` may override this defaulting...) I think this is fine though; if a dev needs more nuanced control of the output, they should simply use `Subprocess.run()`. That said, if `stderr` is not useful, setting it to `subprocess.PIPE` or `subprocess.DEVNULL` effectively ignores it.

The primary basis for this PR (IMO) is integrating the intended usage behavior of `subprocess.check_output()` in to Briefcase beyond just calling convention. Along with this, though, we have a hole such that `subprocess.check_output()` can allow printing within a controlled console. So, in lieu of merging this PR, something should be implemented to avoid the child process writing to a controlled console....whether that's streaming it or perhaps throwing an error when the console is controlled and `stderr` isn't being redirected.

## PR Checklist:

### DIFF
--- a/changes/1059.misc.rst
+++ b/changes/1059.misc.rst
@@ -1,0 +1,1 @@
+For commands invoked with ``Subprocess.check_output()``, any output sent to ``stderr`` is now coalesced with ``stdout`` by default.
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -345,7 +345,6 @@ its output for errors.
             self.tools.subprocess.check_output(
                 [os.fsdecode(self.sdkmanager_path), "--list_installed"],
                 env=self.env,
-                stderr=subprocess.STDOUT,
             )
         except subprocess.CalledProcessError as e:
             raise BriefcaseCommandError(
@@ -622,8 +621,7 @@ connection.
             # Capture `stderr` so that if the process exits with failure, the
             # stderr data is in `e.output`.
             output = self.tools.subprocess.check_output(
-                [os.fsdecode(self.emulator_path), "-list-avds"],
-                stderr=subprocess.STDOUT,
+                [os.fsdecode(self.emulator_path), "-list-avds"]
             ).strip()
 
             # AVD names are returned one per line.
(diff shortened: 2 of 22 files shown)


freakboy3742 (approved):
Thanks for the fix. I think it definitely makes sense that if we're almost always using stderr redirection, then stderr redirection should be the default - the issue picked up in #1058 definitely suggests it's reasonable to have that as a default.

Looking at the cases that don't currently use stderr redirection - I can't see any particular reason those shouldn't be . I'm comfortable turning on redirection, and if we get any reports that indicate redirection is an issue, we can deal with those separately.

PR #1061, opened 2023-01-25
Pass test_sources to Gradle template for extractPackages

mhsmith:
See https://github.com/beeware/briefcase-android-gradle-template/pull/60#discussion_r1082002678.

## PR Checklist:

### DIFF
--- a/changes/1061.misc.rst
+++ b/changes/1061.misc.rst
@@ -1,0 +1,1 @@
+The list of packages that need to be manually extracted was added to the Android Gradle template context.
--- a/src/briefcase/platforms/android/gradle.py
+++ b/src/briefcase/platforms/android/gradle.py
@@ -2,6 +2,7 @@ import datetime
 import re
 import subprocess
 import time
+from os.path import basename, normpath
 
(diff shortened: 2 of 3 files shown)


freakboy3742 (approved):
I've made a couple of minor tweaks of API usage that are equivalent, but more consistent with the rest of the codebase (pathlib instead of os.path; sys.platform rather than os.name) and added a changenote; but otherwise, looks good.

Passing through `extract_packages` as a pre-rendered string bugs me a little - outside of literal template inclusions like `build_gradle_extra_content`, my usual preference is to pass through raw data, rather than pre-rendered template content, so that the template can decide what to do with it. However, cookiecutter makes it painful to pass a list, so we'd need to do something like pass through a dictionary of *all* sources; while that would work, it feels like overkill. As this won't ever be directly modified by the user, and it's a property that we (through chaquopy) are in control ok, I'm OK with practicality over purity in this case.

mhsmith:
Thanks: I meant to add a changenote but it looks like I forgot to save the file before committing.

PR #1066, opened 2023-01-28
Generalize enforcement of host operating system requirements

### LINKED ISSUE #1010: Prevent Running Commands that Target Windows When the Host OS is not Windows
### Describe the bug

Running Briefcase commands for the Windows platform only works when the host OS is Windows. However, Briefcase does not inform the user of this and instead attempts to perform the Command resulting in confusing errors.

Example:
- #1009 

### Steps to reproduce

Run `briefcase build windows` on macOS or Linux.

### Expected behavior

Emit a simple warning informing users.

### Screenshots

_No response_

### Environment

- Python version: 3.10.8
- Software versions:
  - Briefcase: `0.3.12.dev251+g6d29ece1`

### Logs

_No response_

### Additional context

This behavior changed in `v0.3.9`. Previously, these Commands would have aborted when `WiX` was verified. But `WiX` is only verified now for the Package Command.

rmartin16:
## Changes
Previously, all (or at least most) enforcement of host OS compatibility was done during tool verification. This seems to have grown organically over time and largely worked. However, once the (relatively) recent significant changes to tool management were implemented, a lot of these checks lost their veracity. For instance, all Windows commands (except `package`) attempt to run on all OSs; many macOS/iOS commands will attempt to run when not on macOS....etc.

This PR implements a more declarative framework for host OS compatibility for a Command:
- By default, a Command is supported on all platforms (i.e. Darwin, Linux, and Windows).
- A Command can limit the supported platforms by setting `Command.supported_host_os` to a `set` of the supported host OSs.
- This check is performed in `Command.verify_host()` right before `Command.verify_tools()`.

## Specific Command Deviations
- Linux AppImages can technically be built on macOS with Docker but cannot be run on macOS.

## Related Issues
- Fixes #1010 

## PR Checklist:


src/briefcase/platforms/iOS/__init__.py
```diff
@@ -6,12 +6,17 @@
...
 
     def verify_tools(self):
-        if self.tools.host_os != "Darwin":
-            raise BriefcaseCommandError(
-                "iOS applications require Xcode, which is only available on macOS."
-            )
+        # since Xcode must be verified first, explicitly check host OS here
+        # instead of letting super().verify_tools() verify the OS.
+        if self.tools.host_os not in self.supported_host_os:
+            raise BriefcaseCommandError(self.supported_host_os_reason)
```
freakboy3742, line 18: This suggests to me that what is needed is a more generic `verify_host()` check that precedes the `verify_tools()` check.

src/briefcase/commands/create.py
```diff
@@ -16,12 +16,18 @@
...
+
+class UnsupportedPlatform(BriefcaseCommandError):
+    def __init__(self, platform):
+        self.platform = platform
+        super().__init__(
+            f"""\
+App cannot be deployed on {platform}. This is probably because one or more
+requirements (e.g., the GUI library) doesn't support {platform}.
+"""
+        )
```
freakboy3742, line 30: It's not clear why this exception is being moved to `create`; but If we're going to move this exception, I'd prefer to see it move to `briefcase.exceptions`.

src/briefcase/platforms/macOS/xcode.py
```diff
@@ -17,13 +17,15 @@
...
 
     def verify_tools(self):
-        if self.tools.host_os != "Darwin":
-            raise BriefcaseCommandError(
-                "macOS applications require the Xcode command line tools, "
-                "which are only available on macOS."
-            )
+        if self.tools.host_os not in self.supported_host_os:
+            raise BriefcaseCommandError(self.supported_host_os_reason)
+
```
freakboy3742, line 28: As above, I wonder if this indicates a need for a generic `verify_host()` check

src/briefcase/platforms/linux/appimage.py
```diff
@@ -27,6 +27,10 @@ class LinuxAppImagePassiveMixin(LinuxMixin):
     # docker exists. It is used by commands that are "passive" from the
     # perspective of the build system, like open and run.
     output_format = "appimage"
+    supported_host_os = {"Darwin", "Linux"}
+    supported_host_os_reason = (
+        "AppImages can only be built on Linux or in Docker on macOS."
```
freakboy3742, line 32: I found this message a little difficult to parse; I'd suggest: 
```suggestion
        "Linux Appimages can only be built on Linux, or on macOS using Docker."
```

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -382,10 +386,10 @@ def add_options(self, parser):
...
 
     def verify_tools(self):
-        if self.tools.host_os != "Darwin":
-            raise BriefcaseCommandError(
-                "Code signing and / or building a DMG requires running on macOS."
-            )
+        # since the cli tools must be verified first, explicitly check host
+        # OS here instead of letting super().verify_tools() verify the OS.
+        if self.tools.host_os not in self.supported_host_os:
+            raise BriefcaseCommandError(self.supported_host_os_reason)
```
freakboy3742, line 392: As above, I wonder if this indicates a need for a generic `verify_host()` check

src/briefcase/platforms/windows/__init__.py
```diff
@@ -12,6 +12,8 @@
 
 class WindowsMixin:
     platform = "windows"
+    supported_host_os = {"Windows"}
+    supported_host_os_reason = "Building a MSI requires running on Windows."
```
freakboy3742, line 16: MSI is strictly a *packaging* format, and technically only one option of many (even though it's the only one supported at present). The actual failure mode will be VisualStudio or rcedit failing to run. I'd suggest:
```suggestion
    supported_host_os_reason = "Windows applications can only be built on Windows."
```

tests/platforms/linux/appimage/test_mixin.py
```diff
@@ -102,13 +102,13 @@ def test_verify_non_linux_no_docker(tmp_path):
...
         data_path=tmp_path / "briefcase",
     )
-    command.tools.host_os = "WeirdOS"
+    command.tools.host_os = "Darwin"
     command.use_docker = False
 
     # Verify the Docker tool
     with pytest.raises(
         BriefcaseCommandError,
-        match="Linux AppImages can only be generated on Linux without Docker",
```
freakboy3742, line 111: It's not being picked up by coverage, but both Darwin and WeirdOS should be tested here.

tests/platforms/linux/appimage/test_mixin.py
```diff
@@ -159,7 +159,7 @@ def test_verify_non_linux_docker(tmp_path, first_app_config):
         base_path=tmp_path / "base_path",
         data_path=tmp_path / "briefcase",
     )
-    command.tools.host_os = "WeirdOS"
```
freakboy3742, line 162: As above; It's not being picked up by coverage, but both Darwin and WeirdOS should be tested here.

tests/platforms/linux/appimage/test_run.py
```diff
@@ -10,6 +10,19 @@
 from briefcase.platforms.linux.appimage import LinuxAppImageRunCommand
 
 
+@pytest.mark.parametrize("host_os", ["Darwin"])
```
freakboy3742, line 13: Worth checking Windows here as well, to be explicit about the check.

freakboy3742 (changes requested):
A couple of minor suggestions inline, but this mostly looks good.

src/briefcase/commands/create.py
rmartin16: I moved it because its use moved from `base` to `create` a little while ago. There are several Exceptions, though, outside of `briefcase.exceptions`; I can move all of them there.

src/briefcase/platforms/linux/appimage.py
rmartin16: Updated. Although, I capitalized the `i` in `AppImages` and omitted the comma since it's technically a comma splice :) but I see the distinction it's trying to make.

src/briefcase/platforms/windows/__init__.py
rmartin16: Yeah; I should've changed this one when I ripped it out of `wix.py`.

tests/platforms/linux/appimage/test_run.py
rmartin16: Added.

src/briefcase/platforms/iOS/__init__.py
rmartin16: Implemented `verify_host()`.

src/briefcase/platforms/macOS/xcode.py
rmartin16: Implemented `verify_host()`.

src/briefcase/platforms/macOS/__init__.py
rmartin16: Implemented `verify_host()`.

tests/platforms/linux/appimage/test_mixin.py
rmartin16: Discussed in [comment](https://github.com/beeware/briefcase/pull/1066#issuecomment-1411196292).

src/briefcase/commands/new.py
```diff
@@ -12,9 +12,10 @@
     is_valid_bundle_identifier,
     make_class_name,
 )
+from briefcase.exceptions import TemplateUnsupportedVersion
 from briefcase.integrations import git
 
-from .base import BaseCommand, BriefcaseCommandError, TemplateUnsupportedVersion
+from .base import BaseCommand, BriefcaseCommandError
```
freakboy3742, line 18: `BriefcaseCommandError` should be imported from it's original source (`briefcase.exceptions`)

src/briefcase/exceptions.py
```diff
@@ -92,13 +100,98 @@ def __init__(self, tool):
...
+
+
 class InvalidDeviceError(BriefcaseCommandError):
     def __init__(self, id_type, device):
         self.id_type = id_type
         self.device = device
         super().__init__(msg=f"Invalid device {id_type} '{device}'")
 
 
+class AndroidDeviceNotAuthorized(BriefcaseCommandError):
```
freakboy3742, line 177: I'd argue this is an exception that *shouldn't* be moved to the generic exception module; it's highly specific to one platform integration, rather than being a "generic" failure.

tests/platforms/linux/appimage/test_build.py
```diff
@@ -95,7 +95,7 @@ def test_verify_tools_wrong_platform(build_command):
     build_command.tools.download.file = mock.MagicMock()
 
     # Try to invoke the build
-    with pytest.raises(BriefcaseCommandError):
+    with pytest.raises(BriefcaseError):
```
freakboy3742, line 98: This isn't *wrong*, but we can be a bit more specific on why the test fails.

freakboy3742 (approved):
Woof... that's another monster alright - but it looks great. The fixture cleanups in the test suite are particularly appreciated - I've been doing those piecemeal, but it's good to have the band-aid ripped off in one go. I noticed that you caught a couple of tests that were missing `assert` statements as well - nice catch on those.

I picked up a couple of minor tweaks; flagged inline, but I've made the changes to save you another review iteration.

PR #1078, opened 2023-02-03
Add support for passthrough args on dev and run

### LINKED ISSUE #1077: Add ability to pass arguments to running apps

Briefcase doesn't currently allow users to pass command line arguments to the apps being executed. This would be especially useful in test mode, as it would allow passing in specific test targets (or other testing arguments) without making source code modifications to the 

`briefcase dev` and `briefcase run` should support "passthrough arg" syntax. i.e.:
```
briefcase dev --test -- --args --for test program
briefcase run iOS  -- --args --for test program
```
Anything before the `--` argument is consumed by Briefcase; anything after the `--` is ignored by Briefcase, and passed verbatim to the running app.

1. Do nothing, and expect the user to modify their program to introduce test arguments.
2. Expose specific arguments at the briefcase level for specific purposes (e.g., a `--target` argument to let the user specify the test target.

Passthrough arguments will be a lot more flexible.

### Additional context

When this change is made, the test harness in the briefcase template should be updated to honor command line arguments and pass them to the underlying test program.

freakboy3742:
Adds the ability to specify passthrough arguments when invoking `dev` and `run`.

    briefcase dev --test -- foo --bar

will pass run the app in test mode, passing `foo` and `--bar` as arguments to the test app.

At least for now, command line arguments will be ignored (with a warning) on web.

Disables Flake8 E203; this is in conflict with Black's style guide.

Fixes #1077 

beeware/briefcase-template#49 implements the app side of this.
beeware/briefcase-android-gradle-template#62 has a patch to support Android command line arguments
beeware/briefcase-windows-VisualStudio-template#9 has a patch to correct argv handling; the Windows app template will need to be retagged and rebuilt as a result of this change.

## PR Checklist:


mhsmith (commented):
Some documentation suggestions below.

I've started implementing this for Android: I'll finish that on Monday and add it to this PR, and finish my review at the same time.

mhsmith (commented):
I initially found that `briefcase run` *without* `--test` or any other Briefcase arguments was trying to parse the passthrough arguments itself, because the `--` marker was stripped by the second call to `parse_known_args` in cmdline.py. I've now fixed this, and added Android support, but there are some outstanding issues on the other platforms:

* On macOS, `briefcase run` isn't passing through arguments at all: instead, the `open` command is treating each one as an additional item to open. It looks like we should use the `--args` option.
* On Windows, the arguments are added to `sys.argv` starting at index 0 rather than 1.

I haven't tested Linux yet.

tests/test_cmdline.py
```diff
@@ -122,54 +159,43 @@ def test_dev_command(monkeypatch, logger, console):
...
+    dev_run_parameters("run")
+    + [
+        ("run -u", {"update": True}),
+        ("run --update", {"update": True}),
+        ("run --update-resources", {"update_resources": True}),
+        ("run --no-update", {"no_update": True}),
+    ],
+)
+def test_run_command(monkeypatch, logger, console, cmdline, expected_options):
+    """``briefcase run`` returns the Run command for the correct platform."""
```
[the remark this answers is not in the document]
freakboy3742: Yeah, there's overlap; but it doesn't really hurt to test the behaviour from multiple angles to be sure.

freakboy3742:
The Android modifications work well for me. 

I also took a closer look at how this could be handled on iOS... and it turned out to be trivial - the `simctl` tool used to start an app in the simulator passes through extra arguments. I've add that in.

I've updated macOS to use `--args` as you suggest (of all the platform combinations I tested, I missed *that* one... 🙄)

I totally missed the argv[0] issue on Windows; I guess I saw the arguments I was passing in, and ignored the fact that argv[0] was missing. I've modified the Windows VisualStudio template to inject the extra missing `argv[0]` (beeware/briefcase-windows-VisualStudio-template#9); this will require a rebuild of the windows app template once that PR lands.

Lastly I re-confirmed that flatpak and AppImage work, and pass through the expected arguments.

docs/reference/commands/dev.rst
```diff
@@ -40,3 +40,16 @@ Do not run the application; only install application requirements.
...
+Passthrough arguments
+---------------------
+
+If you want to pass any arguments to your app's command line, you can specify them
+using the ``--`` marker to separate Briefcase's arguments from your app's arguments.
+For example::
+
+    briefcase dev -- --wiggle --test
+
+will run the app normal mode, passing the ``--wiggle`` and ``--test`` flags to
```
mhsmith, line 53:
```suggestion
will run the app in normal mode, passing the ``--wiggle`` and ``--test`` flags to
```

docs/reference/commands/run.rst
```diff
@@ -102,3 +102,16 @@ contains the most recent test code. To prevent this update and build, use the
...
+Passthrough arguments
+---------------------
+
+If you want to pass any arguments to your app's command line, you can specify them
+using the ``--`` marker to separate Briefcase's arguments from your app's arguments.
+For example::
+
+    briefcase run -- --wiggle --test
+
+will run the app normal mode, passing the ``--wiggle`` and ``--test`` flags to
```
mhsmith, line 115:
```suggestion
will run the app in normal mode, passing the ``--wiggle`` and ``--test`` flags to
```

src/briefcase/platforms/web/static.py
```diff
@@ -272,6 +275,10 @@ def run_app(
 
         self.logger.info("Starting web server...", prefix=app.app_name)
 
+        # At least for now, there's no easy way to pass arguments to an Android app
```
mhsmith, line 278:
```suggestion
        # At least for now, there's no easy way to pass arguments to a web app.
```

mhsmith (approved):

Issue #1031, opened 2023-01-04
Add support for pygame


freakboy3742:
### What is the problem or limitation you are having?

[Pygame](https://www.pygame.org/news) is a cross-platform Python game programming library. It would be desirable to have native support for Pygame in Briefcase.

### Describe the solution you'd like

Add native Briefcase support for pygame.

### Describe alternatives you've considered

Pygame can be supported at present by selecting "No GUI toolkit", and manually building the stub app and dependencies. However, it would be desirable to have Pygame as an explicit option.

### Additional context

1. Adding a Pygame option to [briefcase-template](https://github.com/beeware/briefcase-template). This will involve changes to:
  * `cookiecutter.json` to declare the additional dependency,
  * `pyproject.toml` to define requirements, and disable Android/iOS/Web targets
  * `app.py` to add a stub app
2. Modifying the `new` command in Briefcase to include Pygame in the list of platform options. 
3. Adding Pygame to the CI configuration for Briefcase and briefcase-template so that pygame apps are built and tested.

It's worth noting that (2) is only required because of #392; if Briefcase read metadata options from the template, adding Pygame to briefcase-template would automatically result in the option becoming available.

freakboy3742:
This was resolved by #1125, released in Briefcase 0.3.14

PR #1125, opened 2023-03-08
Add support for pygame

nadi726:
Add native briefcase support to pygame
Solves #1031.
More information at the [briefcase-template PR](https://github.com/beeware/briefcase-template/pull/59), as most of the work is there. 
## PR Checklist:


freakboy3742:
Awesome - thanks for the contribution. I'll wait until CI passes and you flag this for review, but this looks like a great start!

freakboy3742 (changes requested):
I've merged this up with main, it looks great. 

I've found one problem, and there's one additional piece of work that needs to be done before it can be merged.

The problem is that when the app runs as a bundled app (i.e., briefcase run, not briefcase dev), the app exits with a status code of -1; I've flagged the cause of this on the briefcase-template PR.

The missing work is the inclusion of Pygame examples on the individual template repositories. As part of CI, we need to build every supported GUI framework; that means each of the templates have a sample project in them (linux-appimage, linux-system, linux-flatpak, windows-app, windows-VisualStudio, macOS-app and macOS-Xcode), and a corresponding line in their CI.

freakboy3742 (approved):
Awesome - thanks for the contribution!

PR #1134, opened 2023-03-14
misc: Replace flake8, isort, and pyupgrade with ruff

cclauss:
[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) including bandit, isort, pylint, pyupgrade, and flake8 plus its plugins, and is written in Rust for speed.

The `ruff` Action uses minimal steps to run in ~10 seconds, rapidly providing intuitive GitHub Annotations to contributors.

image

## PR Checklist:

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,0 +1,16 @@
+# https://beta.ruff.rs
+name: ruff
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+jobs:
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - run: pip install --user ruff
+    - run: ruff --format=github .
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,22 +8,17 @@ repos:
       - id: check-docstring-first
       - id: end-of-file-fixer
       - id: trailing-whitespace
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.12.0
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: v0.0.255
     hooks:
-      - id: isort
-        additional_dependencies: [toml]
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v3.3.1
-    hooks:
-      - id: pyupgrade
-        args: [--py38-plus]
+      - id: ruff
+        # args: [--fix, --exit-non-zero-on-fix]
   - repo: https://github.com/psf/black
     rev: 23.1.0
     hooks:
     - id: black
       language_version: python3
-  - repo: https://github.com/PyCQA/flake8
-    rev: 6.0.0
+  - repo: https://github.com/abravalheri/validate-pyproject
+    rev: v0.12.1
     hooks:
-      - id: flake8
+      - id: validate-pyproject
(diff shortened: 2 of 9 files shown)


freakboy3742:
Thanks for the contribution/suggestion. I've never heard of ruff before, so I'll need to do an assessment of whether this is a tool we want to adopt. Generally speaking, when we adopt a process tool like this, we adopt it across all the BeeWare projects, so we'd need to follow this up with commits on Toga, Rubicon, etc.

mhsmith:
Also, I'm not seeing any real advantages over our existing setup:
* pre-commit currently [takes about 40 seconds in CI](https://github.com/beeware/briefcase/actions/runs/4399807679), but it runs in parallel with other tasks that take longer, so reducing that to 10 seconds won't speed up the CI process. 
* On a developer's own machine, pre-commit only runs on the files that have changed on that commit, which normally takes only a couple of seconds. 
* And because developers should have pre-commit set up on their own machine, better failure annotations on the GitHub website don't really matter either, because any failures should have been caught before they were even pushed.

Does ruff have any other advantages we should consider?

cclauss:
> Does ruff have any other advantages we should consider?

From the first link in the commit message.

    ⚡️ 10-100x faster than existing linters
    🐍 Installable via pip
    🛠️ pyproject.toml support
    🤝 Python 3.11 compatibility
    📦 Built-in caching, to avoid re-analyzing unchanged files
    🔧 Autofix support, for automatic error correction (e.g., automatically remove unused imports)
    📏 Over [500 built-in rules](https://beta.ruff.rs/docs/rules/)
    ⚖️ [Near-parity](https://beta.ruff.rs/docs/faq/#how-does-ruff-compare-to-flake8) with the built-in Flake8 rule set
    🔌 Native re-implementations of dozens of Flake8 plugins, like flake8-bugbear
    ⌨️ First-party editor integrations for [VS Code](https://github.com/charliermarsh/ruff-vscode) and [more](https://github.com/charliermarsh/ruff-lsp)
    🌎 Monorepo-friendly, with [hierarchical and cascading configuration](https://beta.ruff.rs/docs/configuration/#pyprojecttoml-discovery)

Ruff aims to be orders of magnitude faster than alternative tools while integrating more functionality behind a single, common interface.

Ruff can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing tens or hundreds of times faster than any individual tool.

freakboy3742:
On further investigation, Ruff sounds interesting, but doesn't appear to offer us anything that we actually need at this time. The speed of execution of our pre-commit hooks (either in CI or locally) isn't enough of a problem that it warrants a major change in infrastructure; and the tools it aims to replace are all known, reliable quantities with a long history of maintenance and support. Beyond that, the features you've listed supporting the change are either "things we already have (Python 3.11 support)", or "things that aren't actually a benefit to us (monorepo support)".

On that basis, I'm going to close this PR.

cclauss:
I did not mean to offend anyone or to drive by.  I was merely trying to help with alignment.
* https://github.com/pyodide/pyodide/blob/main/pyproject.toml#L45
* https://github.com/pyscript/pyscript/blob/main/pyproject.toml#L11

freakboy3742:
No offence taken; however, I would advise that *asking* before before spending effort on a major project infrastructure change that hasn't been asked for is usually a good idea, regardless of the project.

Also, those two projects aren't part of BeeWare, so it's not really "alignment".

PR #1135, opened 2023-03-14
misc: PEP621: Migrate setup.cfg config into pyproject.toml

cclauss:
[PEP621](https://peps.python.org/pep-0621/) encourages the Python community to migrate away from `setup.cfg` files by putting config data in `pyproject.toml` instead.

Migrate `setup.cfg` to `pyproject.toml` using [ini2toml](https://pypi.org/project/ini2toml) to do the file conversion and running [validate-pyproject](https://github.com/abravalheri/validate-pyproject) to validate the results.
* https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html
* https://packaging.python.org/en/latest/tutorials/packaging-projects/#creating-pyproject-toml

Flake8 does not yet support `pyproject.toml` so move its config into a dedicated `.flake8` file.

## PR Checklist:


cclauss:
`FAILED (errors=28, failures=4, skipped=43)`

freakboy3742 (changes requested):
Thanks for the contribution; this is something that has been on our radar. However, as you can see from the test failures, it's not necessarily *that* simple. 

It's also a change we're not prepared to make quite yet. [setuptools has marked anything in the tools.setuptools section as beta](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration), and procedurally, we don't adopt tooling that is in beta status unless its necessary to achieve some other goal.

As I noted on #1134, it's also something we're not likely to do in isolation - we're going to want to adopt this across all our projects, not just Briefcase. This has the added complication that we need to make sure we *can* adopt this across all our projects - we're still locked into setup.py (for reasons) on Toga. 

On a meta-level - while we definitely appreciate the contributions, and we're enthusiastic to have you contribute more, having your first interactions with the project being "here's a rewrite of all your core infrastructure" means you haven't been able to take into account any historical discussion or thoughts that the core team might have on the topics you're addressing. Maybe were considering moving away from setuptools? Maybe someone has tried doing this in the past, and failed because there's something fundamentally missing? Maybe we've got a reason why we *haven't* done this yet?

We have a list of issues flagged for first time contributors that are problems that we *have* expressed an interest in having fixed; or, if you're set on this sort of project configuration task, might I suggest that you engage with us *first*, before you start working on large scale PRs that fundamentally change core aspects of the project? A quick thread on a discussion, or a conversation on Discord could save you a whole lot of effort.

freakboy3742:
I'm going to close this PR on the basis that on closer inspection, it has all the signs of an automated drive-by. The PR description frames this as an `ini2toml` conversion; however, as CI reveals, that conversion hasn't worked. In addition, as I flagged previously, it depends on beta features of setuptools, and as a project we don't adopt core project tooling that is beta unless there's a compelling reason to do so.

Issue #1137, opened 2023-03-15
system_requires aren't verified when building system packages locally


freakboy3742:
### Describe the bug

During the tutorial at Everything Open, a number of Linux users hit an error during `briefcase create` because pygobject couldn't be compiled. This was caused by missing system packages. This *should* be checked at runtime, when `system_requires` is verified - however, this clearly wasn't happening.

### Steps to reproduce

Run the BeeWare tutorial on a system that only has `python3` installed, not `python3-dev` and/or the Cairo etc dependencies mentioned during tutorial step 0.

### Expected behavior

Briefcase should warn that the python3-dev package, and the cairo/gobjectintrospection package have been installed as part of a `briefcase create`.

### Environment

- Operating System: Linux (any)
- Python version: Any
- Software versions:
  - Briefcase: 0.3.13

### Logs

```

```

### Additional context

This surfaces as an error during the installation of pygobject
* "<Python.h>" not found, if python3-dev (or platform equivalent) isn't installed
* pkg-config errors if cairo2-dev or gobject-introspection-dev libraries (or platform equivalent) aren't installed.

freakboy3742:
I think I've found the source of the problem - `briefcase dev` requires the verification of system tools to install pygobject et al into the local environment, but briefcase dev is considered a "platform agnostic" command.

We need a way for `briefcase dev` to run platform- and app-specific checks. The `finalize()` and `verify_app_tools()` entry point *exist*, but they don't use any of the logic in the system backend. This will require factoring out the logic that is currently in `LinuxSystemMostlyPassiveMixin.verify_system_packages()` into a location where it can be shared, and a `verify_app_tools()` implementation on `dev` that can call into that shared implementation (based on the currently active platform, and the default backend on that platform). 

This refactoring will need to take into account that `briefcase dev`'s usage of `verify_system_packages()` is *always* a `use_docker==False` invocation. Empty implementations should also be provided for macOS and Windows.

It may also be worth considering refactoring whether the right place to do this is as tool integration (briefcase?) that wraps all the "what Linux are we on", "what packages are installed", etc. Some of this logic is currently at in LinuxMixin, but might be easier to re-use as a separate tool.

freakboy3742:
[...]

PR #1143, opened 2023-03-20
Added code for supporting arch system package

proneon267:
## Testing instructions:

If you're on an Arch Linux system, or any derivative of an Arch system (e.g., Manjaro), you can test this branch with:
```
mkdir bee-test
cd bee-test
python3 -m venv venv
source venv/bin/activate
pip install git+https://github.com/proneon267/briefcase.git@patch-1
briefcase new --template https://github.com/beeware/briefcase-template --template-branch arch
```
Accept all the default values from the template, then run:
```
cd helloworld
briefcase run
```
You should get a running Toga GTK app. Then, run:
```
briefcase package
```
This should generate a `dist/helloworld-0.0.1-x86_64.pkg.tar.zst` file, which you can install with `pacman -U dist/helloworld-0.0.1-x86_64.pkg.tar.zst`

If you have any problems building, running or installing this package, or the package that is generated doesn't match your expectations as an Arch* user, let us know in the comments below, including details of the specific distro that you are using for testing.

---

For creating arch packages, we need to build a `PKGBUILD` file which will be used by `makepkg` utility to create the `.pkg.tar.zst` file. 
A sample `PKGBUILD` file:
```
# Maintainer: Your Name <youremail@domain.com>
pkgname=NAME
pkgver=VERSION
pkgrel=1
epoch=
pkgdesc=""
arch=()
url=""
license=('GPL')
groups=()
depends=()
makedepends=()
checkdepends=()
optdepends=()
provides=()
conflicts=()
replaces=()
backup=()
options=()
install=
changelog=
source=("$pkgname-$pkgver.tar.gz"
        "$pkgname-$pkgver.patch")
noextract=()
md5sums=()
validpgpkeys=()

prepare() {
	cd "$pkgname-$pkgver"
	patch -p1 -i "$srcdir/$pkgname-$pkgver.patch"
}

build() {
	cd "$pkgname-$pkgver"
	./configure --prefix=/usr
	make
}

check() {
	cd "$pkgname-$pkgver"
	make -k check
}

package() {
	cd "$pkgname-$pkgver"
	make DESTDIR="$pkgdir/" install
}
```

There are many parameters but we do not need to use them all. As you can see among all these functions, only `package()` is necessary, others are optional.
`makepkg` doesn't necessarily need to compile the stub binary. Since, `briefcase build` does the stub binary compilation, we can skip this step.
A simplified and tested `PKGBUILD` file:
```
pkgname=helloworld
pkgver=0.0.1
pkgrel=1
pkgdesc="This is a package"
arch=('x86_64')
url="https://github.com"
license=('GPL')
depends=('python' 'gobject-introspection' 'gobject-introspection-runtime' 'glibc' 'gtk3')
source=("$pkgname-$pkgver.tar.gz")
md5sums=('SKIP')
changelog=CHANGELOG
package() {
  cp -r "$srcdir/$pkgname-$pkgver"/usr/ "$pkgdir"/usr/
}
```

I will add more details later on. 
This PR relates to an issue: #1064 
## PR Checklist:


proneon267 (commented):
corrected a typo

freakboy3742 (changes requested):
I've updated your PR to fix some indentation and formatting issues, and tried it out; even with those changes, I get the following error:
```
[togatest] Building .pkg.tar.zst package...
Building source archive... done
Generating pkgbuild layout... done

Entering Docker context...
Docker| ------------------------------------------------------------------
Docker| ==> ERROR: togatest is not available for the 'x86_64' architecture.
Docker| ==> ERROR: changelog file (/Users/rkm/beeware/briefcase/local/togatest/CHANGELOG) does not exist or is not a regular file.
Docker| ------------------------------------------------------------------
Leaving Docker context.

Building Arch package...

Error while building .pkg.tar.zst package for togatest.
```

I'm also slightly dubious whether this will work in general. You're shipping the `/usr/bin/<myapp>` binary as part of the distribution artefact - won't that cause problems if the person installing the binary doesn't have exactly the same versions of the package? I thought Arch packages needed to be completely compiled from source - which means the distribution artefact needs to *not* include the binary, include the `bootstrap` folder, and perform a make as part of the installation process.

proneon267:
Hello
The first error seems to be due to togatest, which I guess needs to have test for this. But, i haven't written any tests. 

The second error must be due, not being able to find the `CHANGELOG` file, which I will fix.

You are right the stub binary won't work if the libraries are not compatible. ( sorry, force of habit of distributing only for a single architecture)

I'll make the changes and update the pull request soon. Thanks.

freakboy3742:
> Hello The first error seems to be due to togatest, which I guess needs to have test for this. But, i haven't written any tests.

Tests *will* be needed, but that's got nothing to do with this error. Togatest is the name of my test application. It's nothing more than the "Hello World" generated by Briefcase, but with a customised name (because I need to test against PySide, PPB, Pygame,...)
 
> The second error must be due, not being able to find the `CHANGELOG` file, which I will fix.

For the record - the referenced filename exists, and is a regular file. It's the package builder that is failing.
 
> You are right the stub binary won't work if the libraries are not compatible. ( sorry, force of habit of distributing only for a single architecture)

To be clear: Its not just "single architecture" that is the issue here - it's "incompatible version of libraries". Distributions like Ubuntu are able to provide guarantees on the specific version (and thus ABI) that various libraries will provide - locking a specific set of ABIs is the major reason for having tagged releases. AFAIK, Arch doesn't do this - every user is compiling everything from scratch, so there's no way to provide ABI guarantees.

proneon267 (commented):
Modified to tackle the errors.

src/briefcase/platforms/linux/system.py
```diff
@@ -1038,21 +1038,25 @@ def _package_pkg(self, app: AppConfig, **kwargs):
...
+                # Copy the CHANGELOG file to the bundle_path so it can be included in the source archive
+                self.tools.shutil.copy(changelog_source, self.bundle_path(app) / "CHANGELOG")
+                
                 self.tools.shutil.make_archive(
                     pkgbuild_path / f"{app.app_name}-{app.version}",
                     format="gztar",
-                    root_dir=self.bundle_path(app),
-                    base_dir=f"{app.app_name}-{app.version}",
+                    root_dir=self.build_path(app),
+                    base_dir=self.bundle_path(app),
```
freakboy3742, line 1059: This archive will include a pre-build binary unless you either exclude `usr/bin/{app.app_name}` as a path, or disable the call to `make` during the Build command.

src/briefcase/platforms/linux/system.py
```diff
@@ -1038,21 +1038,25 @@ def _package_pkg(self, app: AppConfig, **kwargs):
 with details about the release.
 """
             )
-
+            
         # Generate the pkgbuild layout
-        pkgbuild_path = self.bundle_path(app) / "pkgbuild"
+        pkgbuild_path = self.build_path(app) / "pkgbuild"
```
freakboy3742, line 1043: This isn't correct - it definitely needs to be `bundle_path`, or you'll generate code in a directory outside the Arch bundle.

src/briefcase/platforms/linux/system.py
```diff
@@ -1038,21 +1038,25 @@ def _package_pkg(self, app: AppConfig, **kwargs):
...
         with self.input.wait_bar("Generating pkgbuild layout..."):
             if pkgbuild_path.exists():
                 self.tools.shutil.rmtree(pkgbuild_path)
             (pkgbuild_path).mkdir(parents=True)
 
             # Build the source archive
             with self.input.wait_bar("Building source archive..."):
+                
+                # Copy the CHANGELOG file to the bundle_path so it can be included in the source archive
+                self.tools.shutil.copy(changelog_source, self.bundle_path(app) / "CHANGELOG")
```
freakboy3742, line 1053: Why does it need to be copied? Couldn't you add a single file directly to the archive?

src/briefcase/platforms/linux/system.py
proneon267: yes, we need to generate the code outside the `bundle_path` as the whole bundle_path will be archived for source tarball.

src/briefcase/platforms/linux/system.py
proneon267: Hello
I mean, doesn't `briefcase build` just run `make` and `make install`. So, we could just `briefcase create` and then `briefcase package`. If this is not the intended behavior, then I will modify it.

src/briefcase/platforms/linux/system.py
proneon267: Since we are archiving the whole `bundle_path` directory, I thought first copying and then archiving would be better.

proneon267:
About how the package system works.
`makepkg` reads `PKGBUILD` file and compiles the source(if you specify it to), and then installs(copies,) to the "$pkgdir". Then it packages them into a `.pkg.tar.zst` file. 
Then the package can be installed with `pacman`.  `pacman` copies the contents of the "$pkgdir" on to the root folder of the target system. That's it. It doesn't compile from source and is just a package manager. 
I think you are confusing arch package system with gentoo package system, where everything needs to be compiled from source.

Moreover, probbyou are conf

We do not need to run `make` and `make install` in the `PKGBUILD` file. As, the same is accomplished by running `briefcase build`. Hence, we can skip the `build()` function in `PKGBUILD` file.

src/briefcase/platforms/linux/system.py
freakboy3742: I can tell you for free - It's not going to be easy to remove the build step conditionally based on the packaging format, because there's a whole lot of logic that is shared across all backends that determines how to infer that a build is needed. It's going to be a lot easier to skip the specific call to make in the arch packaging case.

src/briefcase/platforms/linux/system.py
freakboy3742: Copying is a last resort; by copying, you're creating a duplicate file, which could lead to drift between the source and the packaged artefact. Packaging the source file directly is vastly preferred, and I can't see any reason it couldn't be done here.

src/briefcase/platforms/linux/system.py
freakboy3742: This setup is going to lead to a briefcase file layout that looks like:

* build
   * myapp
     * arch
       * pkgbuild
         * ...
       * rolling
         * bootstrap
           * ... 
         * myapp-1.2.3
           * usr
             * ...

That's the wrong location for the `pkgbuild` folder. It should be inside the `rolling` folder. This is, at the very least, needed to support Docker builds, because Docker mounts `build/myapp/arch/rolling` as `/app`; so the `pkgbuild` folder won't be visible inside the Docker container. More generally, all the artefacts of the `arch/rolling` build should be in the `arch/rolling` folder.

freakboy3742:
>

freakboy3742:
(Apologies on the close - hit the wrong button)

> Hello About how the package system works. `makepkg` reads `PKGBUILD` file and compiles the source(if you specify it to), and then installs(copies,) to the "$pkgdir". Then it packages them into a `.pkg.tar.zst` file. Then the package can be installed with `pacman`. `pacman` copies the contents of the "$pkgdir" on to the root folder of the target system. That's it. It doesn't compile from source and is just a package manager. I think you are confusing arch package system with gentoo package system, where everything needs to be compiled from source.

Sure - in that regard, you're describing a process that is functionally identical to both Debian and Redhat packages.

However, Debian and Redhat have a significant difference with Arch: they are able to guarantee binary compatibility. If I have Debian 11, you know *exactly* what version of (say) GTK I have, because that's part of the "GTK" package definition for the Debian 11 release. Arch *doesn't* provide that guarantee, because it doesn't tag releases in that way.

So - if *you* build a package for your Toga app, and you give it to *me*... how can I guarantee that the package will work? You know nothing about my system, other than the fact that I'm running Arch. If I'm running on an arch system where GTK was built 2 years ago, and you build your package on a "bleeding edge" Arch install... how do you know that your app will run on my machine?

> Moreover, probably you are confusing with the AUR system.

I have no idea *what* I'm confusing with what, because I'm not an Arch user. :-) 

Functionally, I want to be able to guarantee that user X can build a package, give it to user Y, and that package will Just Work. I was under the impression that the only way that would happen on Arch is if the compiled components were compiled on the *installing* system; but if that's not the case, then I'll gladly stand corrected.

src/briefcase/platforms/linux/system.py
```diff
@@ -994,6 +1018,115 @@ def _package_rpm(self, app: AppConfig, **kwargs):
...
+                    "\n".join(
+                        [
+                            f"# Maintainer: {app.author} <{app.author_email}>",
+                            f"pkgname={app.app_name}",
+                            f"pkgver={app.version}",
+                            f"pkgrel={getattr(app, 'revision', 1)}",
+                            f'pkgdesc="{app.description}"',
+                            f"arch=('any')",  # We should use {self.linux_arch}, but doing this as the architecture error 
+                                              # was probably caused due to this parameter.
+                                              # Moreover arch wiki recommends 'any' for architecture independent packages.
```
freakboy3742, line 1088: This doesn't smell right to me. 

If you have `numpy` (or any other binary dependency) in your app, Briefcase will be installing an architecture-specific binary wheel. Python is able to guarantee that the wheel will work because it adheres to the `manylinux` specification, which constraints the libraries that the wheel can link against; but it means that the artefacts in the wheel are platform specific. If you build your package on x86-64, you won't be able to deploy it on arm64.

src/briefcase/platforms/linux/system.py
proneon267: Sure, skipping the make call should solve the problem.

src/briefcase/platforms/linux/system.py
proneon267: Thanks! I was confused about the directory structure. I will modify it.

src/briefcase/platforms/linux/system.py
proneon267: Sure. I will change it.

proneon267 (commented):
Modified for addressing issues

freakboy3742:
> For dependencies compatibility `PKGBUILD` file provides the `depends=()` argument...

Yes - you've just given a description of how dependencies work in *every* environment. What you're missing are the practical application of these dependency specification for Arch as a distribution.

On every other Linux distro, we don't need to do any version pinning, because the version of GTK that is provided on Ubuntu 22.04 (or any other distro) is implicitly pinned. All Ubuntu 22.04 machines have binary compatible versions of GTK - that's why they tag releases. If I produce a binary package for Ubuntu 22.04 that depends on GTK, I know that package will still work in 2 years time.

On Arch, a description like "latest" isn't helpful when you're building an app for distribution. If I build a package with a dependency on the "latest" GTK, and 2 weeks pass, the "latest" isn't the same version any more, so there's no guarantee that someone who downloads a 2 week old package with a dependency on the "latest" GTK will be able to install it.

So, if an Arch package contains a pre-compiled binary, we need to guarantee the specific versions of *every* dependency - which means hard-pins on *every* library version (or, at best, *very* narrow ranges). This leads to a bad user experience as well - if I download a 2 year old package, I will discover that I have to downgrade my version of GTK just so I can support the binary requirements of this old package - something that may be mutually incompatible with other packages that I'm running.

This is why I'm suggesting that shipping a pre-compiled binary in an Arch package isn't a viable option. At the very least, we would need to have specific (or very narrow) version pins on every dependency, which is a maintenance overhead for app developers - and for Briefcase as well - as we need to take on the burden of ensuring that the version pins in the templates we provide are up to date with what Arch is currently publishing.

proneon267 (commented):
Modified to use the current working solution for the changelog.

proneon267:
So, the `CHANGELOG` file needs to be in the same directory where the `PKGBUILD` file is located. This is only possible if the `CHANGELOG` is either copied or a hardlink for the `CHANGELOG` file is created in the directory where the `PKGBUILD` is located.

This is because the `CHANGELOG`'s file path `self.base_path / "CHANGELOG"` won't be visible inside the docker and we would get the previous error. 

Also, we cannot use`"$srcdir/CHANGELOG"` to reference the file as the options in the `PKGBUILD` are evaluated first and the `"$srcdir"` variable will not be set until the source tarball is extracted.

Since, you are not too keen to copy the `CHANGELOG` file, we could create a hardlink(points to the same data on the disk) of the `CHANGELOG` file on the directory where `PKGBUILD` file is present.

Until you decide, I am keeping the copying solution, as it is the only working solution.

src/briefcase/platforms/linux/system.py
proneon267: The architecture error is usually caused if the host machine and target package architectures are different. That means, the previous error was caused due to mismatch of host architecture string. I have modified it to use `self.tools.host_arch`. Now, it should work fine.

freakboy3742 (changes requested):
Ok - I've done another review pass, correcting a couple of errors, performing a few cleanups, and updating tests so they pass.

(Out of interest - are you actually running this branch to verify that it works? There's a couple of bugs - like the output file name and the spelling of `vender`/`vendor` that suggest that you haven't actually executed this code.)

I can't claim I fully understand the logic of including a binary on a "rolling" release - but then, I don't really understand the appeal of using Arch in the first place, so I'll defer to someone who actually uses the Arch and their expression of Arch user expectations.

One of the other changes I've made is to remove the hard-coded gobject-introspection (et al) system requirements; those have been moved to the base briefcase template.

All that is missing now is test coverage for the new functionality.

freakboy3742:
Regarding the CHANGELOG copy: I now see the reason why a copy is required. I thought the CHANGELOG needed to be in the source tarball; I now see it needs to be in the pkgbuild directory, which is then passed into the `makepkg` command. In this case, copying is the right approach.

proneon267:
Hello
Thanks for helping. I haven't actually executed the code because the code seemed intuitive enough :) I will write the test coverage and submit for review.

freakboy3742:
While I admire your persistence at trying to build a PR without ever checking out the code - you know this would go a *lot* faster if you checked out the code and ran the tests locally, right?

proneon267:
Hello
will you take a look and let me know what I have missed, as coverage report says coverage is about 99.9% and the combine & check coverage fails.

proneon267:
The failed coverage seems to be due to the following partial from `src/briefcase/platforms/linux/system.py:`:
```
[773]elif app.packaging_format == "pkg":                      773 ↛ exit
[774]    self._verify_pkg_tools()
```
And the error is:
```
line 773 never returned from function _verify_pkg_tools(), because the condition on line 773 was never false.
```

This may be stemming from the fact that in `test_package.py` I did:
```
[162]:
def test_package_pkg_app(package_command, first_app):
      """An Arch app can be packaged"""
      # Set the packaging format
      first_app.packaging_format = "pkg"
  
      # Set the host architecture for test purposes.
      package_command.tools.host_arch = "x86_64"

```
Which I did because in `test_package.py`, we have:
```
def package_command(first_app, tmp_path):
    command = LinuxSystemPackageCommand(
        logger=Log(),
        console=Console(),
        base_path=tmp_path / "base_path",
        data_path=tmp_path / "briefcase",
    )
    command.tools.home_path = tmp_path / "home"

    # Set the host architecture for test purposes.
    command.tools.host_arch = "wonky"
    ```
    which is a problem for arch as "wonky" is not a valid architecture.

**Comment by proneon267** [comment]
I am not sure how should I proceed further. So, take a look at it and let me know.

**Comment by freakboy3742** [comment]
> The failed coverage seems to be due to the following partial from `src/briefcase/platforms/linux/system.py:`:

The problem is a bit of an edge case. Taken in isolation, the code:
```
if format == 'rpm':
   # do rpm stuff
elif format == 'deb': 
   # do deb stuff
elif format == 'pkg':
   # do arch stuff
```

has an implied branch where format is none of the above. In *practice*, this never gets executed, because format can't take any other value - but coverage has no way to know that with certainty.

There are two possible fixes; the first is to add a dummy else, and annotate the line with `# pragma: nocover` to let coverage know the missing line is OK; the second is to re-write the code so that there's no ambiguity. I've done that, along with a couple of other cleanups (including collapsing the couple of dozen commits into a single commit).
 
> which is a problem for arch as "wonky" is not a valid architecture.

Yes - but this is by design. The test code is written deliberately to *not* use a real architecture, to expose where a developer may have inadvertently hard-coded an architecture choice (usually `x86_64`) because it works on their machine. The tests all use "wonky" architecture to ensure that this doesn't happen.

**Comment by proneon267** [comment]
Thank you so much for helping. I couldn't have done it without you :)

**Comment by proneon267** [comment]
Hello
I tried to test this according to the test instructions at the top of the PR. Everything went smoothly for the following:
```
mkdir bee-test
cd bee-test
python3 -m venv venv
source venv/bin/activate
pip install git+https://github.com/proneon267/briefcase.git@patch-1
briefcase new --template https://github.com/beeware/briefcase-template --template-branch arch
cd helloworld
```

When I tried doing: `briefcase run`, I get the following error message:
```
    ~/bee-test  cd helloworld                       ✔  5s   bee-test  
    ~/b/helloworld  briefcase run                          ✔  bee-test  

[helloworld] Finalizing application configuration...
Targeting manjaro:rolling (Vendor base arch)
Determining glibc version... done
Targeting glibc 2.37
Targeting Python3.10

Using app template: https://github.com/beeware/briefcase-linux-system-template.git, branch v0.3.13

[helloworld] Installing support package...
No support package required.

** WARNING: Can't verify system packages                               **

    Briefcase doesn't know how to verify the installation of system
    packages on your Linux distribution. If you have any problems
    building this app, ensure that the packages listed in the app's
    `system_requires` setting have been installed.

[helloworld] Installing application code...
Installing src/helloworld... done

[helloworld] Installing requirements...
Collecting toga-gtk~=0.3.0
  Using cached toga_gtk-0.3.0-py3-none-any.whl (50 kB)
Collecting toga-core==0.3.0
  Using cached toga_core-0.3.0-py3-none-any.whl (499 kB)
Collecting pygobject>=3.14.0
  Downloading PyGObject-3.44.1.tar.gz (720 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 720.5/720.5 kB 772.3 kB/s eta 0:00:00
  Installing build dependencies: finished with status 'error'
  error: subprocess-exited-with-error
  
  × pip subprocess to install build dependencies did not run successfully.
  │ exit code: 1
  ╰─> [41 lines of output]
      Collecting setuptools
        Using cached setuptools-67.6.0-py3-none-any.whl (1.1 MB)
      Collecting wheel
        Using cached wheel-0.40.0-py3-none-any.whl (64 kB)
      Collecting pycairo
        Downloading pycairo-1.23.0.tar.gz (344 kB)
           ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 344.6/344.6 kB 997.5 kB/s eta 0:00:00
      Building wheels for collected packages: pycairo
        Building wheel for pycairo (pyproject.toml): started
        Building wheel for pycairo (pyproject.toml): finished with status 'error'
        error: subprocess-exited-with-error
      
        × Building wheel for pycairo (pyproject.toml) did not run successfully.
        │ exit code: 1
        ╰─> [12 lines of output]
            running bdist_wheel
            running build
            running build_py
            creating build
            creating build/lib.linux-x86_64-cpython-310
            creating build/lib.linux-x86_64-cpython-310/cairo
            copying cairo/__init__.py -> build/lib.linux-x86_64-cpython-310/cairo
            copying cairo/__init__.pyi -> build/lib.linux-x86_64-cpython-310/cairo
            copying cairo/py.typed -> build/lib.linux-x86_64-cpython-310/cairo
            running build_ext
            'pkg-config' not found.
            Command ['pkg-config', '--print-errors', '--exists', 'cairo >= 1.15.10']
            [end of output]
      
        note: This error originates from a subprocess, and is likely not a problem with pip.
        ERROR: Failed building wheel for pycairo
      Failed to build pycairo
      ERROR: Could not build wheels for pycairo, which is required to install pyproject.toml-based projects
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× pip subprocess to install build dependencies did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
Installing app requirements...

Unable to install requirements. This may be because one of your
requirements is invalid, or because pip was unable to connect
to the PyPI server.

Log saved to /home/proneon267/bee-test/helloworld/logs/briefcase.2023_03_25-06_15_41.run.log
```
I have attached the generated log file.
[briefcase.2023_03_25-06_15_41.run.log](https://github.com/beeware/briefcase/files/11068624/briefcase.2023_03_25-06_15_41.run.log)

**Comment by proneon267** [comment]
So, the previous error was caused by pip not being able to install `pycairo`, which in turn was not installing because it couldn't find the `pkg-config` tool, provided by the arch package `pkgconf`, which was not installed on the manjaro system.

Now, after properly installing `pkgconf` and `pycairo`, when I run: `briefcase run`, I get the following error:
```
    ~/bee-test  cd helloworld                              ✔  bee-test  
    ~/b/helloworld  briefcase run                          ✔  bee-test  

[helloworld] Finalizing application configuration...
Targeting manjaro:rolling (Vendor base arch)
Determining glibc version... done
Targeting glibc 2.37
Targeting Python3.10

Using app template: https://github.com/beeware/briefcase-linux-system-template.git, branch v0.3.13
Using existing template (sha 98e41db3fe8691019c8e2561acc3514cee62a659, updated Fri Mar 10 08:39:10 2023)

[helloworld] Installing support package...
No support package required.

** WARNING: Can't verify system packages                               **

    Briefcase doesn't know how to verify the installation of system
    packages on your Linux distribution. If you have any problems
    building this app, ensure that the packages listed in the app's
    `system_requires` setting have been installed.

[helloworld] Installing application code...
Installing src/helloworld... done

[helloworld] Installing requirements...
Collecting toga-gtk~=0.3.0
  Using cached toga_gtk-0.3.0-py3-none-any.whl (50 kB)
Collecting pygobject>=3.14.0
  Using cached PyGObject-3.44.1.tar.gz (720 kB)
Collecting gbulb>=0.5.3
  Using cached gbulb-0.6.4-py3-none-any.whl (17 kB)
Collecting toga-core==0.3.0
  Using cached toga_core-0.3.0-py3-none-any.whl (499 kB)
Collecting pycairo>=1.17.0
  Using cached pycairo-1.23.0-cp310-cp310-linux_x86_64.whl
Collecting travertino>=0.1.3
  Downloading travertino-0.2.0-py3-none-any.whl (16 kB)
Building wheels for collected packages: pygobject
  Building wheel for pygobject (pyproject.toml): started
  Building wheel for pygobject (pyproject.toml): finished with status 'done'
  Created wheel for pygobject: filename=PyGObject-3.44.1-cp310-cp310-linux_x86_64.whl size=854193 sha256=995bc2df7aff9fecc260aebe07465ca5a7ca6a8ffd3c984b0df36b7b7f062c96
  Stored in directory: /home/proneon267/.cache/pip/wheels/33/99/2c/061d0c6934509a9cc08470980d68fbfc972bd22a45c3f306d6
Successfully built pygobject
Installing collected packages: travertino, pycairo, toga-core, pygobject, gbulb, toga-gtk
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
caffeine-ng 4.0.2 requires click<8.0,>=7.1, but you have click 8.1.3 which is incompatible.
Successfully installed gbulb-0.6.4 pycairo-1.23.0 pygobject-3.44.1 toga-core-0.3.0 toga-gtk-0.3.0 travertino-0.2.0
Installing app requirements... done

[helloworld] Installing application resources...
Unable to find src/helloworld/resources/helloworld-16.png for 16px application icon; using default
Unable to find src/helloworld/resources/helloworld-32.png for 32px application icon; using default
Unable to find src/helloworld/resources/helloworld-64.png for 64px application icon; using default
Unable to find src/helloworld/resources/helloworld-128.png for 128px application icon; using default
Unable to find src/helloworld/resources/helloworld-256.png for 256px application icon; using default
Unable to find src/helloworld/resources/helloworld-512.png for 512px application icon; using default

[helloworld] Removing unneeded app content...
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/pygtkcompat/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/gi/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/gi/overrides/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/gi/repository/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/cairo/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/travertino/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga/hardware/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga/constants/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga/style/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga/sources/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga/resources/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga/widgets/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga_gtk/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga_gtk/hardware/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga_gtk/widgets/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga_gtk/widgets/internal/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga_gtk/widgets/internal/buttons/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga_gtk/widgets/internal/rows/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/toga_gtk/libs/__pycache__
Removing directory helloworld-0.0.1/usr/lib/helloworld/app_packages/gbulb/__pycache__
Removing unneeded app bundle content... done

[helloworld] Created build/helloworld/manjaro/rolling

[helloworld] Building application...
Build bootstrap binary...
Building bootstrap binary...

Log saved to /home/proneon267/bee-test/helloworld/logs/briefcase.2023_03_25-06_32_24.run.log

  File "/home/proneon267/.local/bin/briefcase", line 8, in <module>
    sys.exit(main())
  File "/home/proneon267/.local/lib/python3.10/site-packages/briefcase/__main__.py", line 25, in main
    command(**options)
  File "/home/proneon267/.local/lib/python3.10/site-packages/briefcase/commands/run.py", line 287, in __call__
    state = self.build_command(
  File "/home/proneon267/.local/lib/python3.10/site-packages/briefcase/commands/build.py", line 110, in __call__
    state = self._build_app(
  File "/home/proneon267/.local/lib/python3.10/site-packages/briefcase/commands/build.py", line 70, in _build_app
    state = self.build_app(app, test_mode=test_mode, **full_options(state, options))
  File "/home/proneon267/.local/lib/python3.10/site-packages/briefcase/platforms/linux/system.py", line 551, in build_app
    self.tools[app].app_context.run(
  File "/home/proneon267/.local/lib/python3.10/site-packages/briefcase/integrations/subprocess.py", line 133, in inner
    return sub_method(sub, args, **kwargs)
  File "/home/proneon267/.local/lib/python3.10/site-packages/briefcase/integrations/subprocess.py", line 381, in run
    return self._run_and_stream_output(args, **kwargs)
  File "/home/proneon267/.local/lib/python3.10/site-packages/briefcase/integrations/subprocess.py", line 439, in _run_and_stream_output
    with self.Popen(args, **kwargs) as process:
  File "/home/proneon267/.local/lib/python3.10/site-packages/briefcase/integrations/subprocess.py", line 547, in Popen
    return self._subprocess.Popen(
  File "/usr/lib/python3.10/subprocess.py", line 971, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.10/subprocess.py", line 1847, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'make'
```
I have attached the log file.
[briefcase.2023_03_25-06_32_24.run.log](https://github.com/beeware/briefcase/files/11068659/briefcase.2023_03_25-06_32_24.run.log)

**Comment by proneon267** [comment]
Turns out the error was due to `base-devel` package being not installed on the Manjaro Box. After installing the package and running `briefcase run`, it successfully runs a Toga GTK app.

But, when I do `briefcase package`, I get the following error:
```
    ~/b/helloworld  briefcase package               ✔  5s   bee-test  

[helloworld] Finalizing application configuration...
Targeting manjaro:rolling (Vendor base arch)
Determining glibc version... done
Targeting glibc 2.37
Targeting Python3.10

** WARNING: Can't verify system packages                               **

    Briefcase doesn't know how to verify the installation of system
    packages on your Linux distribution. If you have any problems
    building this app, ensure that the packages listed in the app's
    `system_requires` setting have been installed.

Briefcase doesn't currently know how to build system packages in PKG format.

Log saved to /home/proneon267/bee-test/helloworld/logs/briefcase.2023_03_25-06_42_03.package.log

```

I have attached the log file.
[briefcase.2023_03_25-06_42_03.package.log](https://github.com/beeware/briefcase/files/11068668/briefcase.2023_03_25-06_42_03.package.log)

**Comment by MarkusH** [comment]
Packaging and installation worked fine. However, I'm getting an error "No module named helloworld" with `briefcase run` or when running `helloworld` after installation:

```
$ helloworld                                                                        ││Install path: /usr                                                                              ││Pre-initializing Python runtime...                                                              ││PYTHONPATH:                                                                                     ││- /usr/lib/python3.10                                                                           ││- /usr/lib/python3.10/lib-dynload                                                               ││- /usr/lib/helloworld/app                                                                       ││- /usr/lib/helloworld/app_packages                                                              ││Configure argc/argv...                                                                          ││Initializing Python runtime...                                                                  ││Running app module: helloworld                                                                  ││---------------------------------------------------------------------------                     ││Application quit abnormally (Exit code -1)!                                                     ││/tmp/bee-test/venv/bin/python3: No module named helloworld
```

**Comment by freakboy3742** [comment]
@MarkusH Thanks for that. Not sure where that error is coming from - did the package install Python code under /usr/lib/helloworld/app?

**Comment by proneon267** [comment]
Turns out my previous error was because I was running the new branch in a venv instance but the main branch briefcase was installed in the system environment.

Anyways, I intalled the generated package with `sudo pacman -U ./helloworld-0.0.1-1-x86_64.pkg.tar.zst` and it installed successfully. I tried to run from the terminal like:
```
   ~/b/helloworld/dist  helloworld                                     ✔ 
Install path: /usr
Pre-initializing Python runtime...
PYTHONPATH:
- /usr/lib/python3.10
- /usr/lib/python3.10/lib-dynload
- /usr/lib/helloworld/app
- /usr/lib/helloworld/app_packages
Configure argc/argv...
Initializing Python runtime...
Running app module: helloworld

(helloworld:7157): Pango-CRITICAL **: 07:05:02.178: pango_font_description_set_size: assertion 'size >= 0' failed
/usr/lib/helloworld/app_packages/toga_gtk/app.py:43: DeprecationWarning: Gtk.Window.set_wmclass is deprecated
  self.native.set_wmclass(app.interface.name, app.interface.name)
```
and it works as expected.

Hello @MarkusH
Could you confirm, whether you are running the test on `arch` or `manjaro`. Also, I have found that manually installing the following packages help as they are not automatically being installed:
```
base-devel
pkgconf
cairo
gtk3
gobject-introspection-runtime
```
Please check if any of them are missing or out of date and try rebuilding and repackaging, the template. Also, could you attach the log file( as I caught all of the errors until now through the log file).

**Comment by proneon267** [comment]
For a system without installation of `base-devel` package, I get the following error:
```
    ~/b/h/d/helloworld  briefcase run                          ✔  dist  

[helloworld] Finalizing application configuration...
Targeting manjaro:rolling (Vendor base arch)
Determining glibc version... done
Targeting glibc 2.37
Targeting Python3.10

Using app template: https://github.com/beeware/briefcase-linux-system-template.git, branch v0.3.14
Template branch v0.3.14 not found; falling back to development template
Using existing template (sha 84602eaf4df034c0011a322ead43417d8424fa5a, updated Fri Mar 10 21:02:30 2023)

[helloworld] Installing support package...
No support package required.

Unable to build helloworld due to missing system dependencies. Run:

    sudo ['pacman'] install base-devel

to install the missing dependencies, and re-run Briefcase.

Log saved to /home/proneon267/bee-test/helloworld/dist/helloworld/logs/briefcase.2023_03_25-07_28_11.run.log

```

which is a wrong instruction message for installing the `base-devel` package. Need to correct it.

**Comment by MarkusH** [comment]
> @MarkusH Thanks for that. Not sure where that error is coming from - did the package install Python code under /usr/lib/helloworld/app?

Yes, that's where the majority of the code is. I can take a closer look next week, if needed.

**Comment by proneon267** [comment]
Added a `system_installer_install_flag` so that appropriate install flags for each package manager could be shown on the instruction. Also added related tests.

**Comment by freakboy3742** [comment]
> > @MarkusH Thanks for that. Not sure where that error is coming from - did the package install Python code under /usr/lib/helloworld/app?
> 
> Yes, that's where the majority of the code is. I can take a closer look next week, if needed.

Well that was a nice theory... some other thoughts:

1. Does it behave differently if you invoke helloworld *outside* your virtual environment? That *shouldn't* make any difference, but...
2. Does the helloworld app appear in GNOME app menus etc? Does it work any better if invoked from there?

**Comment by proneon267** [comment]
Hello
After installing the generated package, the helloworld app appears in GNOME app menu. The following is a screenshot on a new Manjaro Box after installing the generated package:

Here is the details shown on Manajro's pamanc app:
image
image
image

**Comment by proneon267** [comment]
Now it shows, the packager name correctly:
```
    ~/b/helloworld  pacman -Qip ./dist/helloworld-0.0.1-1-x86_64.pkg.tar.zst
Name            : helloworld
Version         : 0.0.1-1
Description     : My first application
Architecture    : x86_64
URL             : https://example.com/helloworld
Licenses        : BSD license
Groups          : None
Provides        : None
Depends On      : glibc>=2.37  python3  gtk3  gobject-introspection-runtime
Optional Deps   : None
Conflicts With  : None
Replaces        : None
Compressed Size : 1232.21 KiB
Installed Size  : 3055.50 KiB
Packager        : Jane Developer <jane@example.com>
Build Date      : Sun 26 Mar 2023 01:05:38 AM
Install Script  : No
Validated By    : None
Signatures      : None
```
But the `Validated By ` and Signatures` field remain `None` as the option `validpgpkeys=('')` in `PKGBUILD` file is not specified. This is because there is no mechanism to generate or specify a PGP key for the project in briefcase. This can become a major drawback as some systems are configured to not allow packages with unverified signatures to be installed. Moreover, some systems show a negative warning message before installing the package which will degrade user experience.

**Comment by freakboy3742** [comment]
Package signing is definitely something we're interested in adding; we have it already for macOS, and there's a PR in flight adding it for Windows. 

I'd consider it a separate feature to this PR, though; we need to ensure that whatever approach to signing we take, it is as broadly applicable as possible.

**Comment by proneon267** [comment]
Yes you are right, package signing should be a separate PR. Researching on the subject, linux package signing is a bit of mess. I mean the signing part is easy but the part where the user verifies and installs the app is a mess.

In Debian and derivatives, although package signing and verifying tools exists but by default Debian and derivatives verify the signature of the repository instead of the package. So, even if we would sign the generated package, it would not be verified, as this verifying tool and feature is disabled by default.

And in Arch and derivatives system, signed packages cannot be installed on the system, until the public key is imported into pacman's keyring. But it will let users install packages without any signatures without raising any warnings. 

In RHEL, I am not sure but it seems like there would be some sort of such nuances. 

All in all, signing is easy but verifying and installing the package on users' system is difficult.

**Comment by MarkusH** [comment]
Okay, here's the problem. If I run the commands as outlined in the PR description, this is the package content:

```
❯ pacman -Ql helloworld             
helloworld /usr/
helloworld /usr/bin/
helloworld /usr/bin/helloworld
helloworld /usr/lib/
helloworld /usr/lib/helloworld/
helloworld /usr/lib/helloworld/app/
helloworld /usr/lib/helloworld/app/README
helloworld /usr/lib/helloworld/app_packages/
helloworld /usr/lib/helloworld/app_packages/README
helloworld /usr/share/
helloworld /usr/share/applications/
helloworld /usr/share/applications/com.example.helloworld.desktop
helloworld /usr/share/doc/
helloworld /usr/share/doc/helloworld/
helloworld /usr/share/doc/helloworld/changelog.gz
helloworld /usr/share/doc/helloworld/copyright
helloworld /usr/share/icons/
helloworld /usr/share/icons/hicolor/
helloworld /usr/share/icons/hicolor/128x128/
helloworld /usr/share/icons/hicolor/128x128/apps/
helloworld /usr/share/icons/hicolor/128x128/apps/com.example.helloworld.png
helloworld /usr/share/icons/hicolor/16x16/
helloworld /usr/share/icons/hicolor/16x16/apps/
helloworld /usr/share/icons/hicolor/16x16/apps/com.example.helloworld.png
helloworld /usr/share/icons/hicolor/256x256/
helloworld /usr/share/icons/hicolor/256x256/apps/
helloworld /usr/share/icons/hicolor/256x256/apps/com.example.helloworld.png
helloworld /usr/share/icons/hicolor/32x32/
helloworld /usr/share/icons/hicolor/32x32/apps/
helloworld /usr/share/icons/hicolor/32x32/apps/com.example.helloworld.png
helloworld /usr/share/icons/hicolor/512x512/
helloworld /usr/share/icons/hicolor/512x512/apps/
helloworld /usr/share/icons/hicolor/512x512/apps/com.example.helloworld.png
helloworld /usr/share/icons/hicolor/64x64/
helloworld /usr/share/icons/hicolor/64x64/apps/
helloworld /usr/share/icons/hicolor/64x64/apps/com.example.helloworld.png
helloworld /usr/share/man/
helloworld /usr/share/man/man1/
helloworld /usr/share/man/man1/helloworld.1.gz
```

As you can see, all the Python files are missing.

**Comment by proneon267** [comment]
Hello
Could you also check if the files are also missing from the generated folder `/bee-test/helloworld/build/helloworld/manjaro/rolling/pkgbuild/src/helloworld-0.0.1/usr/lib/helloworld/app/helloworld/`

**Comment by MarkusH** [comment]
I've just ran `pip install` and `briefcase new` from the first code block again and now I have a working app. Both, `briefcase run` and `briefcase package` yield the expected results, such that `pacman -U` worked as desired as well:

```shell
$ pacman -Qi helloworld 
Name            : helloworld
Version         : 0.0.1-1
Description     : My first application
Architecture    : x86_64
URL             : https://example.com/helloworld
Licenses        : BSD license
Groups          : None
Provides        : None
Depends On      : glibc>=2.37  python3  gtk3  gobject-introspection-runtime
Optional Deps   : None
Required By     : None
Optional For    : None
Conflicts With  : None
Replaces        : None
Installed Size  : 2.98 MiB
Packager        : Jane Developer <jane@example.com>
Build Date      : Sun 26 Mar 2023 04:34:35 PM CEST
Install Date    : Sun 26 Mar 2023 04:35:00 PM CEST
Install Reason  : Explicitly installed
Install Script  : No
Validated By    : None
```
image
image

proneon267:
Hello
what packages did you install with pip?

MarkusH:
> Hello what packages did you install with pip?

I reran these commands:

```shell
~$ cd /tmp/bee-test
/tmp/bee-test$ source venv/bin/activate
(venv) /tmp/bee-test$ rm -rf helloworld
(venv) /tmp/bee-test$ pip install git+https://github.com/proneon267/briefcase.git@patch-1
(venv) /tmp/bee-test$ briefcase new --template https://github.com/beeware/briefcase-template --template-branch arch
(venv) /tmp/bee-test$ cd helloworld
(venv) /tmp/bee-test/helloworld$ briefcase run
(venv) /tmp/bee-test/helloworld$ briefcase package
(venv) /tmp/bee-test/helloworld$ sudo pacman -U dist/helloworld-0.0.1-1-x86_64.pkg.tar.zst
```

proneon267:
I am able to reproduce the error. On a fresh new installation of Manjaro, I got the same `No module named helloworld` error. Everything went smoothly until I ran `briefcase run`. 

In between the previous attempt at running the app, did you install or change something that would correct this error?

proneon267:
Here is the directory structure, of `helloworld` project when the error inside the `venv` occurs:
```
./helloworld
├── build
│   └── helloworld
│       └── manjaro
│           └── rolling
│               ├── bootstrap
│               │   ├── helloworld
│               │   ├── main.c
│               │   ├── Makefile
│               │   └── pyversion.h
│               ├── briefcase.toml
│               ├── Dockerfile
│               ├── helloworld-0.0.1
│               │   └── usr
│               │       ├── bin
│               │       │   └── helloworld
│               │       ├── lib
│               │       │   └── helloworld
│               │       │       ├── app
│               │       │       │   └── README
│               │       │       └── app_packages
│               │       │           └── README
│               │       └── share
│               │           ├── applications
│               │           │   └── com.example.helloworld.desktop
│               │           ├── doc
│               │           │   └── helloworld
│               │           │       ├── changelog.gz
│               │           │       └── copyright
│               │           ├── icons
│               │           │   └── hicolor
│               │           │       ├── 128x128
│               │           │       │   └── apps
│               │           │       │       └── com.example.helloworld.png
│               │           │       ├── 16x16
│               │           │       │   └── apps
│               │           │       │       └── com.example.helloworld.png
│               │           │       ├── 256x256
│               │           │       │   └── apps
│               │           │       │       └── com.example.helloworld.png
│               │           │       ├── 32x32
│               │           │       │   └── apps
│               │           │       │       └── com.example.helloworld.png
│               │           │       ├── 512x512
│               │           │       │   └── apps
│               │           │       │       └── com.example.helloworld.png
│               │           │       └── 64x64
│               │           │           └── apps
│               │           │               └── com.example.helloworld.png
│               │           └── man
│               │               └── man1
│               │                   └── helloworld.1.gz
│               └── helloworld.1
├── CHANGELOG
├── LICENSE
├── logs
│   ├── briefcase.2023_03_26-15_17_01.run.log
│   └── briefcase.2023_03_26-15_18_29.run.log
├── pyproject.toml
├── README.rst
├── src
│   └── helloworld
│       ├── app.py
│       ├── __init__.py
│       ├── __main__.py
│       └── resources
│           ├── helloworld.icns
│           ├── helloworld.ico
│           ├── helloworld.png
│           └── __init__.py
└── tests
    ├── helloworld.py
    ├── __init__.py
    └── test_app.py

38 directories, 36 files
```

proneon267:
Confirmed this error is caused when the requirements are not installed on a fresh system. For example, on a new system, the following error would be raised:
```
[helloworld] Installing support package...
No support package required.

Unable to build helloworld due to missing system dependencies. Run:

    sudo ['pacman'] ['-Syu'] base-devel gobject-introspection

to install the missing dependencies, and re-run Briefcase.

Log saved to /home/manjaro/io/helloworld/logs/briefcase.2023_03_26-15_43_19.run.log

```

Now install the missing dependencies with `sudo pacman -Syu base-devel gobject-introspection`. Re-run `briefcase run`. It won't raise any errors during the build process, but will show:
```

[helloworld] Finalizing application configuration...
Targeting manjaro:rolling (Vendor base arch)
Determining glibc version... done
Targeting glibc 2.37
Targeting Python3.10

[helloworld] Building application...
Build bootstrap binary...
make: Entering directory '/home/manjaro/io/helloworld/build/helloworld/manjaro/rolling/bootstrap'
python3.10 -c "import sys; print(f'#define PY_TAG \"{sys.version_info.major}.{sys.version_info.minor}\"')" > pyversion.h
gcc -o helloworld main.c pyversion.h -I/usr/include/python3.10 -I/usr/include/python3.10  -Wno-unused-result -Wsign-compare -march=x86-64 -mtune=generic -O3 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection -g -ffile-prefix-map=/build/python/src=/usr/src/debug/python -flto=auto -ffat-lto-objects -DNDEBUG -g -fwrapv -O3 -Wall  -L/usr/lib -lpython3.10 -lcrypt -ldl  -lm -lm  -fPIE
mkdir -p ../helloworld-0.0.1/usr/bin
cp helloworld ../helloworld-0.0.1/usr/bin
make: Leaving directory '/home/manjaro/io/helloworld/build/helloworld/manjaro/rolling/bootstrap'
Building bootstrap binary... done
Installing license... done
Installing changelog... done
Installing man page... done
Update file permissions...
Updating file permissions... done
Stripping binary... done

[helloworld] Built build/helloworld/manjaro/rolling/helloworld-0.0.1/usr/bin/helloworld

[helloworld] Starting app...
Install path: /home/manjaro/io/helloworld/build/helloworld/manjaro/rolling/helloworld-0.0.1/usr
Pre-initializing Python runtime...
PYTHONPATH:
- /usr/lib/python3.10
- /usr/lib/python3.10/lib-dynload
- /home/manjaro/io/helloworld/build/helloworld/manjaro/rolling/helloworld-0.0.1/usr/lib/helloworld/app
- /home/manjaro/io/helloworld/build/helloworld/manjaro/rolling/helloworld-0.0.1/usr/lib/helloworld/app_packages
Configure argc/argv...
Initializing Python runtime...
Running app module: helloworld
Application quit abnormally (Exit code -1)!
/usr/bin/python3: No module named helloworld

Problem running app helloworld.

Log saved to /home/manjaro/io/helloworld/logs/briefcase.2023_03_26-15_45_47.run.log
```
Again running `briefcase run` will show the same error. However, if you exit this instance of `venv` and again source this instance of `venv`, but this time `briefcase run` will run successfully. 

Looks like something is getting cached, so `briefcase build` doesn't re-run again properly. I have attached the log files of both before and after installing the missing dependencies.
[[Before Installing Dependencies]briefcase.2023_03_26-15_43_19.run.log](https://github.com/beeware/briefcase/files/11072245/Before.Installing.Dependencies.briefcase.2023_03_26-15_43_19.run.log)
[[After installing Dependencies]briefcase.2023_03_26-15_45_47.run.log](https://github.com/beeware/briefcase/files/11072246/After.installing.Dependencies.briefcase.2023_03_26-15_45_47.run.log)

rmartin16 (changes requested):
This is great; glad we're able to round this out with arch packaging.

Toga worked fine for me including with webview and numpy.

PySide6 was a different story....because ofc it was 🙃 
- `makepkg` is eager to strip the binaries in the package which is likely to cause problems for libraries from PyPI wheels
  - Adding `options=('!strip')` to `PKGBUILD` mitigates this
- I needed `qt6-base` at runtime for the pyside6 helloworld app
  - Interestingly, `makepkg` complained unless `qt6-base` was also installed at build time...

Other Notes
- Adding `libcanberra` as a dependency avoids the typical canberra module error
- I hit an error for missing `gcc` for `briefcase package`; installing `base-devel` fixed this. We should add this to the BeeWare tutorial as well like we did with `build-essential`.
  - I presume I got the error from `pip` about `gcc` missing because of what's outlined in #1137.

src/briefcase/platforms/linux/system.py
```diff
@@ -994,6 +1023,104 @@ def _package_rpm(self, app: AppConfig, **kwargs):
...
+        # The description *must* exist.
+        # pkgdesc has 80 char limit.
+        if app.description is None:
+            raise BriefcaseCommandError(
+                "App configuration does not define `description`. "
+                "Arch projects require a description."
+            )
+        # The changelog should exist.
+        changelog_source = self.base_path / "CHANGELOG"
+        if not changelog_source.exists():
```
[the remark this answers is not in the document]
proneon267: is `if not ((changelog_source.is_file()) and (changelog_source.exists())) good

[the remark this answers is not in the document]
rmartin16: It works but it's overkill; `exists()` simply verifies that `stat` returns something for the path while `is_file()` confirms that `stat` believes the path is a regular file (or symlink). So, `is_file()` effectively implies `exists()` making the `exists()` call redundant.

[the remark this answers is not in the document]
proneon267: ok should I also use `is_file()` in the tests also?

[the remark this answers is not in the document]
rmartin16: ehh, it's less important, per se, in the tests since we already know it's a regular file....because the test just made the file right before the test ran. The use of `is_file()` here is just to protect against unexpected user behavior where the changelog is a directory or something weird.

src/briefcase/platforms/linux/system.py
```diff
@@ -442,7 +464,7 @@ def verify_system_packages(self, app: AppConfig):
                 f"""\
 Unable to build {app.app_name} due to missing system dependencies. Run:
 
-    sudo {system_installer} install {' '.join(missing)}
+    sudo {system_installer} {system_installer_install_flag} {' '.join(missing)}
 
 to install the missing dependencies, and re-run Briefcase.
```
rmartin16, line 469: Not caused by this PR...but these values are `list`s, so the errors prints them as though they are optional.

```
Unable to build helloworld due to missing system dependencies. Run:

    sudo ['pacman'] ['-Syu'] gtk3

to install the missing dependencies, and re-run Briefcase.
```

Additionally, debian builds aren't checking for basic packages like this....so, they just bomb out later if GTK isn't installed. This may be better addressed in a separate PR, though.

rmartin16: on second look....I notice this PR _is_ setting `system_installer` and `system_installer_install_flag` as lists while debian and rhel were using strings. This should be reconciled one way or the other.

rmartin16:
@proneon267: Related to the empty build directories and differences between `venv` and system python....

I believe this is stemming from the fact the app template is rolled out in the `build` directory during the `create` command but then the `build` command fails on the missing dependencies. Once you install the missing dependencies and re-run `briefcase run`, it skips the steps to install the app's source and `requires` packages because it "looks" as though they've already been installed. If you run `briefcase run -ur`, it will perform those steps and avoid the situation. Once #1137 is resolved, I think situations like this will be prevented.

proneon267:
> If you run briefcase run -ur, it will perform those steps and avoid the situation. Once https://github.com/beeware/briefcase/issues/1137 is resolved, I think situations like this will be prevented.

So, what do you suggest we should do with this issue for now?

rmartin16:
> > If you run briefcase run -ur, it will perform those steps and avoid the situation. Once #1137 is resolved, I think situations like this will be prevented.
> 
> So, what do you suggest we should do with this issue for now?

Well, unless you want to fix #1137 (which I don't think would be appropriate for this PR), we probably shouldn't do anything. Once Briefcase is updated to confirm all of the needed system requirements are installed _before_ making changes to the filesystem, this issue shouldn't happen anymore then.

src/briefcase/platforms/linux/system.py
proneon267: Hello
It may look like it is due to these values being lists, but it is intentional. This is because, not all users would prefer to use `pacman`. 

In systems like Manjaro, there is another package manager in addition to `pacman`, called `pamac`. `pamac` has different syntax than `pacman`. 

Ex: For installing `gtk3`, an user can do `sudo pamac install gtk3` as well as `sudo pacman -Syu gtk3`. 

So, both the `system_installer` and `system_installer_install_flag` are infact optional and can be substituted by the user as per their choice.

rmartin16: That makes sense.

However, this implementation in `_system_requirement_tools` leads to two different Types of outputs: sometimes `Tuple[List[str], List[str], str, List[str]]` and sometimes `Tuple[List[str], List[str], List[str], List[str]]` (and meanwhile they're all `Optional[]`).

I'm just saying the method should really only have a single Type for its output to help prevent subtle bugs in not understanding there can be varied output Types.

Either way, though, the command that's printed for users should be runnable via copy/paste (i.e. printed without the brackets).

proneon267: You are right. It doesn't make sense that `system_installer` and `system_installer_install_flag` should be 2 different entities. So, I have updated the code  to use `system_installer_with_install_flag`. Now only one type of output will be produced: `Tuple[List[str], List[str], List[str]]`. Also, I have made the output copy/paste friendly.

proneon267:
Created `system_installer_with_install_flag` and corrected a condition which was checking for `system_installer` as `None`, instead of `system_verify` as `None`. Also, made the output copy/paste friendly.

src/briefcase/platforms/linux/system.py
```diff
@@ -994,6 +1017,104 @@ def _package_rpm(self, app: AppConfig, **kwargs):
...
+                            f"depends=({system_runtime_requires})",
+                            "changelog=CHANGELOG",
+                            'source=("$pkgname-$pkgver.tar.gz")',
+                            "md5sums=('SKIP')",
+                            "package() {",
+                            '    cp -r "$srcdir/$pkgname-$pkgver/usr/" "$pkgdir"/usr/',
+                            "}",
+                        ]
+                    )
+                )
```
rmartin16, line 1095: (creating discrete comment from review)

`makepkg` is eager to strip the binaries in the package which is likely to cause problems for libraries from PyPI wheels; this was also a problem for RPM packaging. Adding `options=('!strip')` to `PKGBUILD` will disable all stripping.
```suggestion
            with (pkgbuild_path / "PKGBUILD").open("w", encoding="utf-8") as f:
                f.write(
                    "\n".join(
                        [
                            f"# Maintainer: {app.author} <{app.author_email}>",
                            f'export PACKAGER="{app.author} <{app.author_email}>"',
                            f"pkgname={app.app_name}",
                            f"pkgver={app.version}",
                            f"pkgrel={getattr(app, 'revision', 1)}",
                            f'pkgdesc="{app.description}"',
                            f"arch=('{self.tools.host_arch}')",
                            f'url="{app.url}"',
                            f"license=('{app.license}')",
                            f"depends=({system_runtime_requires})",
                            "changelog=CHANGELOG",
                            'source=("$pkgname-$pkgver.tar.gz")',
                            "md5sums=('SKIP')",
                            "options=('!strip')",
                            "package() {",
                            '    cp -r "$srcdir/$pkgname-$pkgver/usr/" "$pkgdir"/usr/',
                            "}",
                        ]
                    )
                )
```

freakboy3742:
@MarkusH Awesome - thanks for that testing. Looks like you might have been bitten by #1137 on your first pass. That's a failure mode that will be particularly common when building system packages, so it's something we're definitely going to need to look at, but it's not an inherent problem with this Arch packaging PR, so we probably won't consider it a blocker on this PR.

On a more general note: the package works, but is it... "right"? i.e., does the way we're packaging the app conform to your expectations as an Arch user? In particular, the fact that we're including a pre-compiled binary on a system that doesn't have a hard concept of dependency pinning (in the sense of Debian/Ubuntu/Fedora that provide binary ABI guarantees for "Bullseye"/"22.04"/"37" et al) seems odd to me, but I also don't have a good read on Arch user expectations. 

@rmartin16 Unfortunately, I can't get a working Arch/Manjaro system in my test config (Parallels doesn't have full integration for either AFAICT); I can *build* packages in Docker, but I can't do a non-Docker build or test the results. It looks like you're able to do runtime tests - are you happy with where this patch is at?

rmartin16:
> @rmartin16 Unfortunately, I can't get a working Arch/Manjaro system in my test config (Parallels doesn't have full integration for either AFAICT); I can _build_ packages in Docker, but I can't do a non-Docker build or test the results. It looks like you're able to do runtime tests - are you happy with where this patch is at?

Yeah - I think so; the only thing that seems off is toga apparently isn't respecting dark mode in Gnome 43 in arch. On my primary machine with Gnome 42, it does respect dark mode. Everything else is working though 🦾

rmartin16 (approved):

freakboy3742 (approved):
@proneon267 Thanks for the contribution!

proneon267:
@freakboy3742 
Hello
If you are having problems installing Manjaro/Arch on a virtual machine then check for the following:
* Make sure the virtual hard disk is about 64 GB. I know this sounds arbitrary but the installation will fail with some rsync error code, even though it doesn't use anywhere near that much of space.
* Also, the installation may at first succeed but it won't be bootable. In that case boot with the live iso, chroot and reinstall and configure grub.
* Also there is a high chance that the installation may succeed without any problems on an EFI virtual machine. If you do install it in such a virtual machine, make sure that the efi partition is populated with the grub efi image. 
* The virtual machine is a 64-bit.
* If Parallels isn't working then maybe try virtualbox, it has a developer preview for M1/M2 mac, and i have heard that it works really well and is more configurable. 

Also, I am planning to add support for generating a package which could be used for distributing through AUR. I mean it won't be like a traditional package per se, just a tarball containing the `PKGBUILD` , the source tarball(uncompiled sources) and related things. So, the user can host the source tarball somewhere and add its address to the generated `PKGBUILD`. The developer would then distribute the `PKGBUILD` through the AUR. This would then download the source tarball, compile it and then install it. But, that's for a separate PR though.

freakboy3742:
@proneon267 Thanks for the tips - I am able to get Manjaro installed; the issue is with the Parallels integration with the native system (such as mounting the native system drive as a network drive) which is what allows me to efficiently test multiple platforms concurrently. Moving to virtualbox won't fix those problem, as virtualbox doesn't provide those features (to the best of my knowledge).

As for the AUR package - it sounds like what you're describing is what is called a "source distribution" on Debian/Redhat platforms. That definitely sounds like an option worth adding; I've opened #1152 to track the feature.

If you decide to work on this feature, can I please ask that you do your development work locally, rather than through the Github editing interface. The approach you took on this PR results in lots of individual commits without meaningful commit messages, and causes a lot of CI resources to be wasted.

proneon267:
Sure, I will do the development work locally this time. Thank you for helping. I couldn't have done it without you.

PR #1145, opened 2023-03-23
Update all `subprocess` calls to use `errors="backslashreplace"`

### LINKED ISSUE #1141: Truncated Output or Killed Child Process When Output Streamer Suppresses `UnicodeDecodeError` Exceptions from `popen_process.stdout.readline()`
### Describe the bug

Practically all processes are opened in `text` mode such that `subprocess` automatically decodes all piped output. While these processes are running, the output streamer continuously reads `stdout` with `readline()` so it can be logged. Additionally, this is wrapped in a `try/except` for `ValueError` which is suppressed because the process should be dead; this can manifest when a user spams CTRL+C, for instance, causing the process to close `stdout` on its end.

Given all this, the call to `readline()` on `stdout` can throw a `UnicodeDecodeError` for invalid data given the codec for decoding; since `UnicodeDecodeError` subclasses `ValueError`, this exception is suppressed.

Once this occurs, the output streamer waits 3 seconds for the processes to terminate...after which it forcibly kills it. This can result in truncated output (if the process happens to terminate within 3 seconds) or the process can be forcibly killed for seemingly no reason.

### Steps to reproduce

I have been encountering this when using a [forked copy](https://github.com/freakboy3742/linuxdeploy-plugin-gtk/blob/lib64-fix/linuxdeploy-plugin-gtk.sh) of the [linuxdeploy GTK plugin](https://github.com/linuxdeploy/linuxdeploy-plugin-gtk/blob/master/linuxdeploy-plugin-gtk.sh) when the `DEBUG` environment variable is set while testing #1132.

I haven't been able to assess whether this is something in my environment...or an issue with the plugin.

### Expected behavior

The "bad" data should be accommodated such that normal operation of Briefcase isn't affected. So, some setting of `errors` should probably be used with `subprocess`.

### Screenshots

_No response_

### Environment

- Python version: 3.10
- Software versions:
  - Briefcase: `0.3.13.dev180+g18f59eef.d20230319`

### Logs

This log includes the exception since I updated the output streamer to re-raise.
[briefcase.2023_03_19-18_52_42.build.log](https://github.com/beeware/briefcase/files/11012547/briefcase.2023_03_19-18_52_42.build.log)

### Additional context

_No response_

rmartin16:
## Changes
- Update calls to `subprocess` to include `errors="backslashreplace"` to avoid `UnicodeDecodeError` exceptions
- Limit the ability for the output streamer to ignore `ValueError` exceptions to _only_ those arising from the process unexpectedly closing `stdout`.

## Example of effect of `errors="backslashreplace"`
```
Docker| [gtk/stdout] ‘girepository-1.0/Atk-1.0.typelib’ -> ‘/app/Hello World.AppDir//usr/lib/girepository-1.0/Atk-1.0.typelib’                          subprocess.py:641
Docker| [gtk/stdout] \xe2\x80[gtk/stderr] ++ realpath --relative-to=/usr/lib64 /usr/lib64/girepository-1.0                                              subprocess.py:641
Docker| [gtk/stderr] + cp girepository-1.0 --archive --parents '--target-directory=/app/Hello World.AppDir//usr/lib' --verbose                          subprocess.py:641
Docker| \x98girepository-1.0/SoupGNOME-2.4.typelib’ -> ‘/app/Hello World.AppDir//usr/lib/girepository-1.0/SoupGNOME-2.4.typelib’                        subprocess.py:641
Docker| [gtk/stdout] ‘girepository-1.0/GdkPixbuf-2.0.typelib’ -> ‘/app/Hello World.AppDir//usr/lib/girepository-1.0/GdkPixbuf-2.0.typelib’              subprocess.py:641

```

## Issues
- Fixes #1141 

## PR Checklist:

### DIFF
--- a/changes/1141.bugfix.rst
+++ b/changes/1141.bugfix.rst
@@ -1,0 +1,1 @@
+When commands produce output that cannot be decoded to Unicode, Briefcase now writes the bytes as hex instead of truncating output or canceling the command altogether.
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -221,25 +221,36 @@ class Subprocess(Tool):
                 "encoding", os.device_encoding(sys.__stdout__.fileno()) or "UTF-8"
             )
 
+        # When text mode is enabled, subprocess defaults to "strict"
+        # handling for errors arising from decoding output to Unicode.
+        # To avoid Unicode exceptions from misbehaving commands, set
+        # `errors` so output that cannot be decoded for the specified
+        # encoding are replaced with hex of the raw bytes.
+        if any(kwargs.get(kw) for kw in ["text", "encoding", "universal_encoding"]):
+            kwargs.setdefault("errors", "backslashreplace")
+
         # For Windows, convert start_new_session=True to creation flags
         if self.tools.host_os == "Windows":
             try:
                 if kwargs.pop("start_new_session") is True:
                     if "creationflags" in kwargs:
                         raise AssertionError(
-                            "Subprocess called with creationflags set and start_new_session=True.\n"
-                            "This will result in CREATE_NEW_PROCESS_GROUP and CREATE_NO_WINDOW being "
-                            "merged in to the creationflags.\n\n"
-                            "Ensure this is desired configuration or don't set start_new_session=True."
+                            "Subprocess called with creationflags set and "
+                            "start_new_session=True.\nThis will result in "
+                            "CREATE_NEW_PROCESS_GROUP and CREATE_NO_WINDOW "
+                            "being merged in to the creationflags.\n\nEnsure "
+                            "this is desired configuration or don't set "
+                            "start_new_session=True."
                         )
-                    # CREATE_NEW_PROCESS_GROUP: Makes the new process the root process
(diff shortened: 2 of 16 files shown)


freakboy3742 (approved):
All looks good to me. One minor tweak to the formatting, mostly in service of making Black easier to deal with, but otherwise good to go.

I have a minor reservation about the `sub_kw` optimisation - it's a local test configuration we need to remember to use, and I don't *necessarily* consider verbosity a flaw in tests if the verbosity is explicit. However, I don't feel strongly enough to revert this myself (or ask you to revert it), and I can see some value in having a common definition so it's obvious and explicit when we're not using default values. 

As for testing - I can't think of any obvious additional tests either. As you note, any testing of subprocess output really falls into the "testing that subprocess does what it says it does", which doesn't really improve quality.

rmartin16:
Tbh, the consolidation was more about avoiding having to update 100 tests again in the future. I think this was the third time I've done it because something in `Subprocess` was changed.

freakboy3742:
I figured as much :-) You will insist on working on gnarly bits of code, though... so you have no-one to blame by yourself :-P

PR #1147, opened 2023-03-24
Mitigate likelihood of ill-timed context switches in streamer tests

rmartin16:
- In at least one [known case](https://github.com/beeware/briefcase/actions/runs/4507530299/jobs/7935404412), the thread created in the "stuck streamer" [test](https://github.com/beeware/briefcase/blob/788dd35b4be07e18ee559eceafa34edc6830b267/tests/integrations/subprocess/test_Subprocess__stream_output.py#L100) did not exit as expected and injected output in to the `stdout` of another test causing it to fail.
- This change sets the thread's exit flag sooner and slightly increases the amount of time the thread waits before checking the flag.

## PR Checklist:

### DIFF
--- a/changes/1147.misc.rst
+++ b/changes/1147.misc.rst
@@ -1,0 +1,1 @@
+The "stuck streamer" test was updated to mitigate the possibility of a race condition failing it or other tests.
--- a/tests/integrations/subprocess/test_Subprocess__stream_output.py
+++ b/tests/integrations/subprocess/test_Subprocess__stream_output.py
@@ -113,7 +113,7 @@ def test_stuck_streamer(mock_sub, streaming_process, monkeypatch, capsys):
 
     def monkeypatched_blocked_streamer(*a, **kw):
         """Simulate a streamer that blocks longer than it will be waited on."""
-        time.sleep(1)
+        time.sleep(1.5)
         if monkeypatched_streamer_should_exit.is_set():
             return
         print("This should not be printed while waiting on the streamer to exit")
@@ -129,14 +129,14 @@ def test_stuck_streamer(mock_sub, streaming_process, monkeypatch, capsys):
     with pytest.raises(KeyboardInterrupt):
         mock_sub.stream_output("testing", streaming_process, stop_func=send_ctrl_c)
 
+    monkeypatched_streamer_should_exit.set()
+
     # fmt: off
     assert capsys.readouterr().out == (
         "Stopping...\n"
         "Log stream hasn't terminated; log output may be corrupted.\n"
     )
     # fmt: on
-
-    monkeypatched_streamer_should_exit.set()
 
 
 def test_stdout_closes_unexpectedly(mock_sub, streaming_process, monkeypatch, capsys):
--- a/tests/integrations/subprocess/test_Subprocess__stream_output.py
+++ b/tests/integrations/subprocess/test_Subprocess__stream_output.py
@@ -107,16 +107,19 @@ def test_stuck_streamer(mock_sub, streaming_process, monkeypatch, capsys):
     mock_time = mock.MagicMock(side_effect=range(1000, 1005))
     monkeypatch.setattr(time, "time", mock_time)
 
-    # Flag for the mock streamer to exit and prevent it
-    # potentially printing in the middle of a later test.
+    # Flag that Briefcase has finished simulating its waiting on the output
+    # streamer to exit normally; so, it should now exit.
     monkeypatched_streamer_should_exit = Event()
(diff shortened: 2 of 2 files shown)


freakboy3742 (commented):
I can see how these changes might reduce the likelihood of the problem; however, I'm not sure it's a permanent fix. 

In the current implementation, there will be 0.25 + 3 * 0.1 == 0.55s of delay between the handling of the fake CTRL-C. That's *way* less than the 1s that it's currently waiting; if we're seeing an issue with 1s, there's no particular reason to believe we won't see problems with 1.5s either - and the longer hard sleep makes the test run slower to boot (ok - not by much, but it all adds up).

The move of the Event setting call is slightly more impactful; it at least guarantees that at the point we read stdout, we're on the "safe exit" path. If there's an analogous slowdown, we'll at least see that failure in *this* test, rather than the effect leaking into another test because the daemon thread has persisted in the background and is dirtying the output.

However, that leads me to wonder if that doesn't point to the real fix here. The current test is a "non-existence" test - but a non-existence test that *can* outlive the test case itself. I wonder if a better test might be an explicit check.

1. Replace the time.sleep() call in the blocked streamer with an Event.wait(timeout=1)
2. Generate that event in the same way that streamer_should_exit is currently used
3. If the timeout is reached, set a "timeout_reached" event that indicates the streamer wasn't exited.
4. Modify the side effect of subprocess.cleanup() to capture the popen instance of the underlying thread 
5. At the end of the test, wait for/assert that the popen has actually exited, and that the "timeout_reached" event wasn't set.

That way, we're certain that the daemon thread has exited (so the test is isolated from a process state point of view); and the test has been written so that it's not dependent on the side effects of stdout processing and the non-existence of a signal in stdout as a test pass.

Does that make any sense?

rmartin16:
> In the current implementation, there will be 0.25 + 3 * 0.1 == 0.55s of delay between the handling of the fake CTRL-C. That's _way_ less than the 1s that it's currently waiting; if we're seeing an issue with 1s, there's no particular reason to believe we won't see problems with 1.5s either - and the longer hard sleep makes the test run slower to boot (ok - not by much, but it all adds up).

These changes shouldn't slow down the test; none of the code is actually waiting on the thread because we're short circuiting the 3 seconds it would normal wait around for it. The current design of this test just assumes the thread will quietly disappear sometime after this test confirms that the misbehaving streamer was eventually ignored and Briefcase moved on. (Although, in a failure case for this test, the full wait time of this `sleep` should impact the test's runtime.)

> However, that leads me to wonder if that doesn't point to the real fix here. The current test is a "non-existence" test - but a non-existence test that _can_ outlive the test case itself. I wonder if a better test might be an explicit check.
> 
> 1. Replace the time.sleep() call in the blocked streamer with an Event.wait(timeout=1)
> 2. Generate that event in the same way that streamer_should_exit is currently used
> 3. If the timeout is reached, set a "timeout_reached" event that indicates the streamer wasn't exited.
> 4. Modify the side effect of subprocess.cleanup() to capture the popen instance of the underlying thread
> 5. At the end of the test, wait for/assert that the popen has actually exited, and that the "timeout_reached" event wasn't set.
> 
> That way, we're certain that the daemon thread has exited (so the test is isolated from a process state point of view); and the test has been written so that it's not dependent on the side effects of stdout processing and the non-existence of a signal in stdout as a test pass.
> 
> Does that make any sense?

I don't think I follow the part about a `Popen` instance....but I do follow using another `Event` to tell the fake streamer to go ahead and exit because we successfully tested whether a "stuck streamer" can hold up all of Briefcase. I'll make these changes; thanks.

freakboy3742:
> I don't think I follow the part about a `Popen` instance....but I do follow using another `Event` to tell the fake streamer to go ahead and exit because we successfully tested whether a "stuck streamer" can hold up all of Briefcase. I'll make these changes; thanks.

Ignore me. What I was driving at is that `stream_output` is internally creating a daemon thread, but the existence of that thread is never surfaced externally. I thought it was being passed as an argument to `cleanup()`; but on a second read, I see it's the same popen we passed in originally, not the thread instance.

rmartin16:
Oook, I think this is in line with what you were saying. Let me know what you think.

(note: i added the `monkeypatched_streamer_exited` flag after realizing if i only had a `wait(timeout=1)` on the `monkeypatched_streamer_was_improperly_awaited` flag....it would _always_ wait that full second)

(note2: now to see if any of these timeouts get hit randomly during CI....:/....)

freakboy3742 (approved):
Yup - this is pretty much what I had in mind; AFAICT, it should be more robust, and avoids any stdout leakage issues.

PR #1153, opened 2023-03-27
Verify the Docker buildx plugin is installed

rmartin16:
Briefcase implicitly relies on the `buildx` plugin being installed. This is because Briefcase depends on Docker using the BuildKit backend and `buildx` is the necessary frontend component to enforce using BuildKit.

A growing collection of users are reporting this error; it currently manifests like this:
```
[beethings] Building Docker container image...

Entering Docker context...
Docker| ------------------------------------------------------------------
Docker| DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
Docker|             Install the buildx component to build images with BuildKit:
Docker|             https://docs.docker.com/go/buildx/
Docker| 
Docker| unknown flag: --progress
Docker| See 'docker build --help'.
Docker| ------------------------------------------------------------------
Leaving Docker context.
```

Fortunately, more recent versions of Docker on linux come bundled with the `buildx` plugin. Docker Desktop and rootless Docker (which are more prevalent on other platforms although popularity is likely to increase for linux) have bundled the `buildx` plugin for quite a bit longer.

This also sets up transitioning the image build step to use `buildx` directly via `docker buildx build ...`.

## PR Checklist:

### DIFF
--- a/changes/1153.bugfix.rst
+++ b/changes/1153.bugfix.rst
@@ -1,0 +1,1 @@
+If the Docker ``buildx`` plugin is not installed, users are now directed by Briefcase to install it instead of Docker failing to build the image.
--- a/src/briefcase/integrations/docker.py
+++ b/src/briefcase/integrations/docker.py
@@ -19,7 +19,6 @@ version {docker_version}. Visit:
     {install_url}
 
 to download and install an updated version of Docker.
-{extra_content}
 """
     UNKNOWN_DOCKER_VERSION_WARNING = """
 *************************************************************************
@@ -70,7 +69,7 @@ Visit:
     {install_url}
 
 to download and install Docker manually.
-{extra_content}
+
 If you have installed Docker recently and are still getting this error, you may
 need to restart your terminal session.
 """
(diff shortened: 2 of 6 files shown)


rmartin16:
On Ubuntu 20.04, I installed version 19.03.15 of `docker-ce`. Without the `buildx` plugin installed (and without this PR), Briefcase is able to successfully build a deb or AppImage in Docker. With version 20.10.13 of `docker-ce`, the `buildx` plugin is included. Finally, with version 23.0.1, the `buildx` plugin is recommended for install (therefore included by default) as the `docker-buildx-plugin` package.

Nonetheless, with the `docker-buildx-plugin` packaged installed with any of these versions, the image can be built.

Docker 19 went EOL back in Jan 2021; if you want me to bump minimum version support to Docker 20, let me know.

freakboy3742 (approved):
Looks fairly straightforward, and should be a good improvement. The only downside is that I don't have an easy way to personally verify that `docker buildx version` raises an error if `buildx` it isn't installed - but the that would make sense as the behavior, so I'll take your word for it.

I've made one small tweak to a related error message - `--no-docker` is no longer a flag available everywhere, so we shouldn't recommend its use.

As for increasing the default Docker version; Yes, Docker 19 is EOL, but for our purposes, that doesn't functionally matter. If the user has an old version of Docker, it won't impact on Briefcase's ability to run, so there's no particular reason to force the update. If that ever changes (e.g., we need a specific flag or option that was introduced in a particular version), that's the trigger for a version update.

rmartin16:
> Looks fairly straightforward, and should be a good improvement. The only downside is that I don't have an easy way to personally verify that `docker buildx version` raises an error if `buildx` it isn't installed - but the that would make sense as the behavior, so I'll take your word for it.

FWIW, I'm depending on `docker` to raise its `docker: <command> is not a docker command` message. So, the error from `docker buildx version` should be the same as the error from `docker weird_command version`.

PR #1154, opened 2023-03-28
Removed the usage of `sys.platform` in favour of `platform.system()`

proneon267:
Modified to Removed the usage of `sys.platform`
in favour of `platform.system()`, which has more
consistent behaviour.

## PR Checklist:


proneon267:
A majority of the file changes are due to `isort` and `black` and not due to the actual changes.

mhsmith:
> Modified to Removed the usage of `sys.platform` in favour of `platform.system()`, which has more consistent behaviour.

`sys.platform` is clearly documented and easy enough to use. What's the benefit of this change?

> A majority of the file changes are due to `isort` and `black` and not due to the actual changes.

Then you must have used `pre-commit` or `black` incorrectly, because the previous code was already black-compatible. Please fix this, and reduce the PR to just the relevant changes.

proneon267:
Hello
I am planning to add support for freebsd and `sys.platform` will change depending on the version of freebsd. For example, on FreeBSD 12 it will be `freebsd12` and on FreeBSD 13 it will be `freebsd13`. But the base system is still FreeBSD, which is consistently produced by `platform.system()`. I need that consistent result for adding FreeBSD support.

Also, if the previous code was black-compatible, shouldn't `isort` and `black` would have just skipped formatting the files, instead of reformatting them?

rmartin16:
@proneon267, it looks like you probably aren't running `pre-commit` in your local environment. I recommend taking a look at the [contributing code](https://briefcase.readthedocs.io/en/latest/how-to/contribute-code.html) doc; it lays out how to ensure `pre-commit` checks pass locally before you can even commit the code and push it up to GitHub.

proneon267:
Yes there is support for testing FreeBSD with CI. https://github.com/freebsd/freebsd-ci

freakboy3742:
> Yes there is support for testing FreeBSD with CI. https://github.com/freebsd/freebsd-ci

That repository describes a Jenkins configuration. We don't have a Jenkins configuration - we use Github Actions.

proneon267:
In that case https://github.com/vmactions/freebsd-vm

proneon267:
The FreeBSD support will be a separate PR independent of this.

freakboy3742 (changes requested):
Ok - so... sure - *technically* possible. That workflow launches a virtualbox container and does QEMU emulation to get a working FreeBSD install... which... is not a great way to be running tests. 

Even if this wasn't the case, I'm not especially enthusiastic about the idea of adding a FreeBSD backend. 

FreeBSD isn't a platform I (or anyone else in the core team, AFAIK) knows much about; and in the entire time I've been running the project, I'm not aware of a single FreeBSD user who has asked for support. If we merged a FreeBSD backend, we'd have to take on the maintenance burden of keeping that backend working. We gain a whole lot of extra headaches... and very little tangible gain.

If you are especially enthused about the idea of Briefcase for FreeBSD, my suggestion would be to build it as an external plugin. Briefcase is specifically designed to be pluggable - although it has a bunch of backends and formats internally, those  are all built as plugins to a core. It should be possible (in theory) for a third-party to build a completely independent packaging format or platform, and plug that into briefcase using setup.cfg entry points. For example, you could define a `briefcasex_freebsd` package (briefcasex being a prefix to identify that it isn't an officially supported plugin), and then add:
```
[options.entry_points]
briefcase.platforms =
    freebsd = briefcasex_freebsd.freebsd
briefcase.formats.freebsd =
    ports = briefcasex_freebsd.freebsd.ports
```

to the project's setup.cfg file.

If it turns out that this *isn't* possible - i.e., that the plugin interface isn't rich enough to support freebsd, then I'd consider *that* a bug that is in scope, and any patch needed to address that bug would be worth consideration. It's possible some of the changes you've listed in *this* PR fall into that category. 

However, that's not what this PR is doing; and the base motivation of being "more consistent" doesn't hold up for any of the cases being tested here. On that basis, I'm going to close this PR. If you want to tackle this again from the point of view of making a Briefcase FreeBSD plugin, then any tests would need to exercise those extension points.

freakboy3742:
You appear to be very concerned about a problem we're not having :-)

We're currently running CI on linux, and it's working very consistently. The "linuxX" problem isn't one we're seeing.

There's a mild argument to be made in the name of consistency; but as PEP8 begins: "a foolish consistency is the hobgoblin of little minds :-). 

`sys.platform` is just as reliable as `platform.system()` for all our use cases; so unless there's a class of bugs that will be avoided, or features that will be unlocked by this change, I'm inclined to avoid the code churn.

proneon267:
Sure. Thanks for explaining.

mhsmith:
> wouldn't we have problems down the line, since the output of `sys.platform` on linux is dependent on the kernel version, like the output may be `linux515` , `linux6`, etc.

As the documentation explains, `sys.platform` has no version number on Linux since Python 3.3.

PR #1156, opened 2023-03-30
Flake8 tweaks

rmartin16:
## Changes
- Remove unnecessary `noqa` markers and clarify unqualified markers
- Simplify `flake8` configuration
  - Remove `max-complexity` instead of disabling complexity checks by excluding `C901`
  - Instead of replicating [default](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-ignore) checks for `exclude`, only list the additional checks to exclude
- Update CI to cancel the current run if a PR receives new commits

## Notes
I identified the unnecessary instances of `noqa` with `yesqa`. It has a `pre-commit` hook but since it requires running `flake8` itself, I didn't think it was necessary to have it run for everyone's commits.

```yaml
  - repo: https://github.com/asottile/yesqa
    rev: v1.4.0
    hooks:
      - id: yesqa
```

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,11 @@ defaults:
   run:
     shell: bash  # https://github.com/beeware/briefcase/pull/912
 
+# Cancel active CI runs for a PR before starting another run
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   pre-commit:
     name: Pre-commit checks


freakboy3742 (approved):
These all look like reasonable updates. As far as the flake8 config is concerned, leaning on  defaults will always be preferable IMHO; but I agree that running yesqa on every commit is probably overkill.

PR #1167, opened 2023-04-07
Correct some markup errors in docs.

freakboy3742:
Doing a local build of docs, I noticed some non-critical warnings and typographical inconsistencies. 

## PR Checklist:

- [x] All new features have been tested
- [x] All new features have been documented
- [x] I have read the **CONTRIBUTING.md** file
- [x] I will abide by the code of conduct

PR #1168, opened 2023-04-07
Docs Verification in CI and Cleanup

rmartin16:
## Changes
- Docs cleanup...many of which are necessary for docs verification
- Implement two new tox commands
  - `docs-lint`: verifies all hyperlinks in docs are valid and performs spellchecking
  - `docs-all`: builds the entire docs with error sensitivity cranked up
- Run `tox -e docs-lint` on RTD prior to building the docs

## Notes
- This builds off the changes in #1167.

## PR Checklist:


rmartin16:
I've introduced several (potentially) opinionated changes to `tox.ini` for docs; so, I welcome feedback to align with entrenched or otherwise established opinions.

Further, I guess I'm seeing this as a step towards introducing these CI checks as a workflow that will be implemented in `beeware/.github` so other repos can use it. Additionally, that would seem like a good opportunity to unify how each repo is building docs (such as where this "build" directory should be).

.github/workflows/ci.yml
```diff
@@ -32,6 +32,29 @@ jobs:
...
+        python-version: 3.x
+
+    - name: Install tox
+      run: python -m pip install .[dev]
+
+    - name: Check links
+      run: tox -e docs-links
+
+    - name: Check build
+      run: tox -e docs-all
```
freakboy3742, line 56: The RTD task is already doing a docs build; Rather than duplicating the build step, we can add a [`sphinx.fail_on_warning`](https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx-fail-on-warning) configuration to `.readthedocs.yaml`. 

RTD doesn't expose a "check links" option, or any other way to configure the sphinx arguments AFAICT, so we still need the "docs linting" pass in GitHub Actions.

docs/background/releases.rst
```diff
@@ -452,7 +452,7 @@ Features
   device model name and unique ID e.g. a Pixel 3a shows up as ``Pixel 3a
   (adbDeviceId)``. (#433)
 * Android apps are now packaged in Android App Bundle format. This allows the
-  Play Store to dynmically build the smallest APK appropriate to a device
+  Play Store to dynamically build the smallest APK appropriate to a device
```
freakboy3742, line 455: Side note - I'd *love* to get a spell checker in as part of the linting pass...

tox.ini
```diff
@@ -24,11 +24,35 @@ deps =
...
+# -E: force rebuild of environment
+# -a: read/parse all files
+# -n: nitpick mode
+sphinx_verbose_args = -v -W --keep-going -E -a -n
+
+[testenv:docs]
+change_dir = {[docs]change_dir}
+extras = {[docs]extras}
+commands =
+    sphinx-build -W -b html . _build/html
```
freakboy3742, line 43: While `sphinx-build` is installed as an entry point, I prefer the `python -m ` entry point because of the [issues associated with execution on Windows](https://snarky.ca/why-you-should-use-python-m-pip/)

tox.ini
```diff
@@ -24,11 +24,35 @@ deps =
...
+
+[testenv:docs-links]
```
freakboy3742, line 45: `docs-links`  is *currently* just a link check, but I can imagine other checks (most notably, a spellchecker), so `docs-lint` is probably a better name

tox.ini
freakboy3742: Also - would it be worth turn on `linkcheck` here? It doesn't seem overly expensive, and it would seem beneficial to verify that all links are valid as part of a "default" docs build.

freakboy3742 (changes requested):
A couple of comments/suggestions, but generally looks like a bunch of changes in the right direction.

.github/workflows/ci.yml
rmartin16: That's true; although, we won't get "nitpick mode" via the `-n` flag in RTD....not exactly the end of the world.

tox.ini
rmartin16: > While `sphinx-build` is installed as an entry point, I prefer the `python -m ` entry point because of the [issues associated with execution on Windows](https://snarky.ca/why-you-should-use-python-m-pip/)

I agree; although, `python -m sphinx-build` doesn't work. Based on my testing, `python -m sphinx` seems to default to the same behavior as a naked `sphinx-build` call....but I didn't dig in to confirm.

> Also - would it be worth turn on `linkcheck` here? It doesn't seem overly expensive, and it would seem beneficial to verify that all links are valid as part of a "default" docs build.

I guess it depends on how you want to think about these commands.

The `linkcheck` doesn't actually build the docs; it just checks the links. So, if we add the link checking here, it'll need to run before or after the existing call to actually build the docs. My thought was this might be annoying if you just want to quickly see your updates in HTML.

docs/background/releases.rst
rmartin16: I agree; however, when I ran (PyCharm's) spellcheck on the docs, at least 90% of the results were false-positives...

freakboy3742: [`sphinxcontrib.spelling`](https://sphinxcontrib-spelling.readthedocs.io/en/latest/) allows for custom dictionaries...

tox.ini
freakboy3742: Ah - that nitpicking technicality detail of "it doesn't work" :-)

As for a having a "fast build" option - that definitely makes sense. I was basing my suggestion on the fact that `linkcheck` was returning fairly quickly for me - but that appears to be because the `flathub.org` link was causing a fast fail. Once I fixed that, I can see a link check takes a moment (which shouldn't be surprising, given what it needs to do), and it doesn't seem to restrict the link check to files that have changed. 

On that basis, I agree that `docs` makes sense as a "fast path", with `docs-lint` being a full clean environment lint check.

On a related note, we probably want to add a note about these new options in the docs contribution guide, so that contributors know how to reproduce the failures they're likely to see in CI.

.github/workflows/ci.yml
rmartin16: All docs-related CI is now happening on RTD.

docs/background/releases.rst
rmartin16: Spellchecking is now live.

tox.ini
rmartin16: Updated.

tox.ini
rmartin16: Updated docs.

freakboy3742 (approved):
This is awesome - nice work. 

I'd be totally on board with these changes (or their analogs) being ported to Toga and Rubicon as well.

PR #1178, opened 2023-04-13
Further cleanup of docs spelling

freakboy3742:
The SSL error on macOS was caused because the PyPI name filter was turned on. This made every single one of the 440k names of PyPI packages a valid word in spelling. So... that needs to be turned off.

That removes the need for the SSL instructions in the contribution guide. 

It also reveals a bunch of additional spelling problems. The only ones that *aren't* legitimate are the ones tied to a package name; for those, I've adopted the policy of:
* If it's being used in a "code" context, use the name as a literal
* If it's being used in a "text" context, ensure preferred project naming and capitalisation is used (hence: pytest, Pygame, PyScript) 

I've also updated all the links that were raising redirects. Ideally, redirects could be raised as an error, but that doesn't appear to be an option with Sphinx.

## PR Checklist:


docs/background/releases.rst
```diff
@@ -242,43 +178,24 @@ Misc
...
-  Docker progress messages has also been improved. (#774)
-* Users can now define a ``BRIEFCASE_HOME`` environment variable. This allows
-  you to specify the location of the Briefcase tool cache, allowing the user to
-  avoid issues with spaces in paths or disk space limitations. (#789)
-* Android emulator output is now printed to the console if it fails to start
-  properly. (#799)
-* ``briefcase android run`` now shows logs from only the current process, and
-  includes all log tags except some particularly noisy and useless ones. It also
-  no longer clears the Logcat buffer. (#814)
+* Linux apps can now be packaged in Flatpak format. (#359) * SDKs, tools, and other downloads needed to support app builds are now stored in an OS-native user cache directory instead of ``~/.briefcase``. (#374)
```
mhsmith, line 181:
```suggestion
* Linux apps can now be packaged in Flatpak format. (#359) 
* SDKs, tools, and other downloads needed to support app builds are now stored in an OS-native user cache directory instead of ``~/.briefcase``. (#374)
```

mhsmith (approved):

PR #1181, opened 2023-04-13
Allow to run tox on single test files

MarkusPiotrowski:
Running `tox` always runs through the whole test suite, which can take a while, even when doing `tox -e py`. This small change in the `tox.ini` allows to pass single files, which I found quite helpful during the dev process.

## PR Checklist:

### DIFF
--- a/changes/1181.misc.rst
+++ b/changes/1181.misc.rst
@@ -1,0 +1,1 @@
+The possibility to run ``tox`` on single test files was added.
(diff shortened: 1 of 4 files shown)


freakboy3742 (approved):
Thanks for the contribution. FWIW, I tend to run `pytest` directly in my local environment for day-to-development, and leave tox for "full tests" in CI (or locally when I need to test multiple Python versions); but this is a simple modification, so we might as well include it.

PR #1183, opened 2023-04-13
Allow to pack a Windows app as zip file

### LINKED ISSUE #457: Create Windows executable instead of MSI
Hi, I have two questions: 
Is it possible to create a self-contained Windows executable file instead of an installable version? Basically I'm looking for something similar to how the packaging is done already for linux.
If it isn't at the moment, will it ever be possible? Thanks.

MarkusPiotrowski:
As discussed in issue #457 and elsewhere on the discord channel, a Windows app packed as zip file could be a nice alternative for shipping and will sometimes be preferable to an MSI installer.

I also got the impression that @freakboy3742 wouldn't block such a PR per se, so I'll give it a try ;-)

Fixes #457 

## PR Checklist:

### DIFF
--- a/changes/457.feature.rst
+++ b/changes/457.feature.rst
@@ -1,0 +1,1 @@
+On Windows, it is now possible to build a zip folder for distribution by using ``package -p zip``.
--- a/changes/457.feature.rst
+++ b/changes/457.feature.rst
@@ -1,1 +1,1 @@
-On Windows, it is now possible to build a zip folder for distribution by using ``package -p zip``.
+Support for packaging Windows apps as simple ZIP files was added.
--- a/docs/reference/platforms/windows/app.rst
+++ b/docs/reference/platforms/windows/app.rst
@@ -14,6 +14,9 @@ enabled. To ensure .NET Framework 3.5 is enabled:
 3. Select "Turn Windows features On or Off"
 4. Ensure that ".NET framework 3.5 (includes .NET 2.0 and 3.0)" is selected.
 
+Instead of building an MSI installer, you can also pack the folders for
+distribution to a zip file using the ``briefcase package -p zip`` command.
+
 Icon format
 ===========
 
--- a/docs/reference/platforms/windows/app.rst
+++ b/docs/reference/platforms/windows/app.rst
@@ -3,7 +3,11 @@ Windows App
 ===========
 
 A Windows app is a stub binary, allow with a collection of folders that contain
-the Python code for the app and the Python runtime libraries.
+the Python code for the app and the Python runtime libraries. Briefcase supports
+two packaging formats for a Windows app:
+
+1. As an MSI installer
+2. As a ZIP file containing all files needed to run the app
 
 Briefcase uses the `WiX Toolset <https://wixtoolset.org/>`__ to build an MSI
 installer for a Windows App. WiX, in turn, requires that .NET Framework 3.5 is
@@ -13,9 +17,6 @@ enabled. To ensure .NET Framework 3.5 is enabled:
 2. Traverse to Programs -> Programs and Features
 3. Select "Turn Windows features On or Off"
 4. Ensure that ".NET framework 3.5 (includes .NET 2.0 and 3.0)" is selected.
-
-Instead of building an MSI installer, you can also pack the folders for
-distribution to a zip file using the ``briefcase package -p zip`` command.
 
 Icon format
 ===========
--- a/docs/reference/platforms/windows/index.rst
+++ b/docs/reference/platforms/windows/index.rst
@@ -7,7 +7,7 @@ The default output format for Windows is an :doc:`app <./app>`.
 Briefcase also supports creating a :doc:`Visual Studio project <./visualstudio>`
 which in turn can be used to build an app.
 
-Both output formats support packaging as a Microsoft software installer (MSI) or
+Both output formats support packaging as a Microsoft Software Installer (MSI) or
 as a ZIP file.
 
 .. toctree::
--- a/docs/reference/platforms/windows/index.rst
+++ b/docs/reference/platforms/windows/index.rst
@@ -7,7 +7,8 @@ The default output format for Windows is an :doc:`app <./app>`.
 Briefcase also supports creating a :doc:`Visual Studio project <./visualstudio>`
 which in turn can be used to build an app.
 
-Both output formats support packaging as an MSI.
+Both output formats support packaging as a Microsoft software installer (MSI) or
+as a ZIP file.
 
 .. toctree::
    :maxdepth: 1
--- a/src/briefcase/platforms/windows/__init__.py
+++ b/src/briefcase/platforms/windows/__init__.py
@@ -4,6 +4,7 @@ import subprocess
 import uuid
 from pathlib import Path
 from typing import List
+from zipfile import ZIP_DEFLATED, ZipFile
 
 from briefcase.commands import CreateCommand, PackageCommand, RunCommand
 from briefcase.config import AppConfig, parsed_version
@@ -278,6 +279,8 @@ class WindowsPackageCommand(PackageCommand):
 
         if app.packaging_format == "zip":
             self._package_zip(app)
+            # The zip can't be signed, so:
+            return
         else:
             self._package_msi(app)
 
@@ -376,16 +379,18 @@ class WindowsPackageCommand(PackageCommand):
 
         self.logger.info("Building zip file...", prefix=app.app_name)
         with self.input.wait_bar("Packing..."):
-            zip_root_path = self.bundle_path(app) / f"{app.formal_name}-{app.version}"
-            self.tools.shutil.copytree(
-                self.bundle_path(app) / self.packaging_root,
-                zip_root_path,
-                dirs_exist_ok=True,
-            )
-            self.tools.shutil.make_archive(
-                self.distribution_path(app).with_suffix(""),
-                "zip",
-                self.bundle_path(app),
-                f"{app.formal_name}-{app.version}",
-            )
-            self.tools.shutil.rmtree(zip_root_path)
+            source = self.bundle_path(app) / self.packaging_root  # /src
+            try:
+                os.mkdir(self.dist_path)  # /dist
+            except FileExistsError:
+                pass
+
+            with ZipFile(self.distribution_path(app), "w", ZIP_DEFLATED) as archive:
+                for path, _, files in os.walk(source):
+                    for file in files:
+                        src_path = os.path.join(path, file)
+                        rel_zip_path = os.path.normpath(
+                            f"{app.formal_name}-{app.version}//"
+                            f"{src_path.replace(str(source), '')}"
+                        )
+                        archive.write(src_path, rel_zip_path)
--- a/src/briefcase/platforms/windows/__init__.py
+++ b/src/briefcase/platforms/windows/__init__.py
@@ -118,7 +118,7 @@ class WindowsRunCommand(RunCommand):
 class WindowsPackageCommand(PackageCommand):
     @property
     def packaging_formats(self):
-        return ["msi"]
+        return ["msi", "zip"]
 
     @property
     def default_packaging_format(self):
@@ -259,6 +259,16 @@ class WindowsPackageCommand(PackageCommand):
         :param timestamp_url: Timestamp authority server to use in code signing.
         :param timestamp_digest: Hashing algorithm to request from the timestamp server.
         """
+
+        # Just pack the 'binary' folders in a zip file:
+        if app.packaging_format == "zip":
+            self.logger.info("Building zip file...", prefix=app.app_name)
+            self.tools.shutil.make_archive(
+                self.distribution_path(app).with_suffix(""),
+                "zip",
+                self.bundle_path(app) / self.packaging_root,
+            )
+            return
 
         if sign_app and not identity:
             sign_app = False
--- a/src/briefcase/platforms/windows/__init__.py
+++ b/src/briefcase/platforms/windows/__init__.py
@@ -2,7 +2,7 @@ import os
 import re
 import subprocess
 import uuid
-from pathlib import Path
+from pathlib import Path, PurePath
 from typing import List
 from zipfile import ZIP_DEFLATED, ZipFile
 
@@ -380,17 +380,10 @@ class WindowsPackageCommand(PackageCommand):
         self.logger.info("Building zip file...", prefix=app.app_name)
         with self.input.wait_bar("Packing..."):
             source = self.bundle_path(app) / self.packaging_root  # /src
-            try:
-                os.mkdir(self.dist_path)  # /dist
-            except FileExistsError:
-                pass
+            zip_root = f"{app.formal_name}-{app.version}"
 
             with ZipFile(self.distribution_path(app), "w", ZIP_DEFLATED) as archive:
-                for path, _, files in os.walk(source):
-                    for file in files:
-                        src_path = os.path.join(path, file)
-                        rel_zip_path = os.path.normpath(
-                            f"{app.formal_name}-{app.version}//"
-                            f"{src_path.replace(str(source), '')}"
-                        )
-                        archive.write(src_path, rel_zip_path)
+                for file_path in source.glob("**/*"):
+                    archive.write(
+                        file_path, zip_root / PurePath(file_path).relative_to(source)
+                    )
--- a/src/briefcase/platforms/windows/__init__.py
+++ b/src/briefcase/platforms/windows/__init__.py
@@ -23,7 +23,8 @@ class WindowsMixin:
         return self.bundle_path(app) / self.packaging_root / f"{app.formal_name}.exe"
 
     def distribution_path(self, app):
-        return self.dist_path / f"{app.formal_name}-{app.version}.msi"
+        suffix = "zip" if app.packaging_format == "zip" else "msi"
+        return self.dist_path / f"{app.formal_name}-{app.version}.{suffix}"
 
 
 class WindowsCreateCommand(CreateCommand):
@@ -260,16 +261,6 @@ class WindowsPackageCommand(PackageCommand):
         :param timestamp_digest: Hashing algorithm to request from the timestamp server.
         """
 
-        # Just pack the 'binary' folders in a zip file:
-        if app.packaging_format == "zip":
-            self.logger.info("Building zip file...", prefix=app.app_name)
-            self.tools.shutil.make_archive(
-                self.distribution_path(app).with_suffix(""),
-                "zip",
-                self.bundle_path(app) / self.packaging_root,
-            )
-            return
-
         if sign_app and not identity:
             sign_app = False
 
@@ -284,6 +275,22 @@ class WindowsPackageCommand(PackageCommand):
                 timestamp_digest=timestamp_digest,
             )
             self.sign_file(app=app, filepath=self.binary_path(app), **sign_options)
+
+        if app.packaging_format == "zip":
+            self._package_zip(app)
+        else:
+            self._package_msi(app)
+
+        if sign_app:
+            self.logger.info("Signing package...", prefix=app.app_name)
+            self.sign_file(
+                app=app,
+                filepath=self.distribution_path(app),
+                **sign_options,
+            )
+
+    def _package_msi(self, app):
+        """Build the msi installer."""
 
         self.logger.info("Building MSI...", prefix=app.app_name)
         try:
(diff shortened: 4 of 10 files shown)


changes/457.feature.rst
```diff
@@ -0,0 +1 @@
+On Windows, it is now possible to build a zip folder for distribution by using ``package -p zip``.
```
freakboy3742, line 1:
```suggestion
Support for packaging Windows apps as simple Zip files was added.
```

docs/reference/platforms/windows/app.rst
```diff
@@ -14,6 +14,9 @@ enabled. To ensure .NET Framework 3.5 is enabled:
 3. Select "Turn Windows features On or Off"
 4. Ensure that ".NET framework 3.5 (includes .NET 2.0 and 3.0)" is selected.
 
+Instead of building an MSI installer, you can also pack the folders for
```
freakboy3742, line 17: This should be incorporated into the earlier paragraph introducing WiX; something like:

```
Briefcase supports 2 packaging formats for a Windows app:

1. As an MSI installer
2. As a Zipfile containing all files needed to run the app
```
then lead into the description of how WiX is used to build MSIs.

A similar comment/update is needed on the VisualStudio format as well.

freakboy3742 (changes requested):
Thanks for the contribution! The broad strokes of this look good; I've left some comments inline about some of the specifics.

MarkusPiotrowski:
Dear @freakboy3742 ,

Thank you very much for your detailed review and I'm sorry for confronting you with a PR with so much issues. One of the problems I had was that I had a running implementation on briefcase 0.3.12 (where app signing and distribution folders didn't exist) but then I was pushed back with each new release of briefcase since they had some significant changes for Windows packaging. Most of the time for this PR I spend trying to understand the testing environment (and obviously I did not well), for two days I believed that my implementation is not longer working because the tests failed until I realized that the implementation was OK and the issues were 'only' the tests... :-)
I hope I can address all the issues you mentioned, I'll try it.

tests/platforms/windows/app/test_package.py
```diff
@@ -30,9 +31,27 @@ def package_command(tmp_path):
     return command
 
 
+@pytest.fixture
+def package_zip_command(tmp_path):
```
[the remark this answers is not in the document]
MarkusPiotrowski: I guess you're right. Regarding creating a new fixture that extends an existing one I failed. Fixtures cannot be called directly and I didn't find the way to do so.

src/briefcase/platforms/windows/__init__.py
```diff
@@ -260,6 +260,16 @@ def package_app(
         :param timestamp_digest: Hashing algorithm to request from the timestamp server.
         """
 
+        # Just pack the 'binary' folders in a zip file:
+        if app.packaging_format == "zip":
+            self.logger.info("Building zip file...", prefix=app.app_name)
+            self.tools.shutil.make_archive(
+                self.distribution_path(app).with_suffix(""),
```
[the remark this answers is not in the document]
MarkusPiotrowski: I was thinking about a root folder, wasn't sure if this is wanted. The path issue origins from the fact that the distribution path as defined earlier in the code does already contain the `.msi` suffix. 
After writing the shutil command from scratch I finally realized that zip packaging is also used in the Web backend, so I took it from there.

tests/platforms/windows/app/test_package.py
```diff
@@ -207,6 +226,19 @@ def test_package_msi(package_command, first_app_config, tmp_path):
...
+    """A Windows app can be packaged as a zip file."""
+
+    first_app_config.packaging_format = "zip"
+    package_zip_command.package_app(first_app_config)
+
+    # The zip file exists and contains the content of the src folder
+    archive_file = tmp_path / "base_path" / "dist" / "First App-0.0.1.zip"
+    assert archive_file.exists()
+    with ZipFile(archive_file) as archive:
+        assert archive.namelist() == ["First App.exe"]
```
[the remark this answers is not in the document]
MarkusPiotrowski: It does not only give you coverage, it will also show you that creating a zip folder with the correct name does work.
My first guess was that within the testing environment I find some real hello world project that is used for building, running, packaging and which I can use. When I realized that e.g. the msi packager tests does not actually create an MSI installer, I wondered how important testing the zip content is considered crucial here. Of course, I can put the correct folder structure in the zip and add some more mock files, still, this wouldn't show that the packed app will run.

tests/platforms/windows/app/test_package.py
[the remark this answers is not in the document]
MarkusPiotrowski: Got it.

tests/platforms/conftest.py
```diff
@@ -18,6 +18,7 @@ def first_app_config():
         sources=["src/first_app"],
         requires=["foo==1.2.3", "bar>=4.5"],
         test_requires=["pytest"],
+        packaging_format="",
```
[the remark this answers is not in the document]
MarkusPiotrowski: Made a local `conftest.py` file in the platform/windows file.

src/briefcase/platforms/windows/__init__.py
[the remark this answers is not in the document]
freakboy3742: The web backend also does zip packaging, but the use case there explicitly *doesn't* need the root folder. You unpack a web project into the folder where you want it to be served. 

As for the issue with `distribution_path` already containing `msi` - that's my point - it shouldn't contain `msi` if you're packaging a zip file. At the very least, it's returning the wrong extension. `distribution_path` should be adapting to the selected packaging format.

freakboy3742:
> Thank you very much for your detailed review and I'm sorry for confronting you with a PR with so much issues. 

Not a problem at all - and although there's a lot of comments, I don't consider the issues to be *that* bad - it's lots of little things that are, in many cases, wrapped up in details that you probably don't know about unless you've been soaking in the code for years like I have. This is a really good first-time project contribution - the fact that you've submitted a PR that fully passes CI on your first attempt (even if there's room for improving those tests) puts you at the top of the class in terms of contributions :-)

src/briefcase/platforms/windows/__init__.py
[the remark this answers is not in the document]
MarkusPiotrowski: For the moment, I see two options to make a root folder:
- creating a folder in the build folder (e.g. on the same level as src) and copy the content of the src folder into it (this I have implemented now). In this case, should we delete the root folder afterward or just keep it?
- temporarily rename the build/.../scr folder to `MyApp-1.2.3`. This seems a little bit 'hackish' to me.
What would you recommend? Unfortunately, neither `ZipFile` nor `shutil.make_archive` have a 'put in a root folder before zipping' option...

src/briefcase/platforms/windows/__init__.py
```diff
@@ -260,6 +260,16 @@ def package_app(
         :param timestamp_digest: Hashing algorithm to request from the timestamp server.
         """
 
+        # Just pack the 'binary' folders in a zip file:
```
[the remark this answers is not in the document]
MarkusPiotrowski: Regarding app signing: Should the final zip file also be signed (if signing is requested), like the msi file?

docs/reference/platforms/windows/index.rst
```diff
@@ -7,7 +7,8 @@ The default output format for Windows is an :doc:`app <./app>`.
 Briefcase also supports creating a :doc:`Visual Studio project <./visualstudio>`
 which in turn can be used to build an app.
 
-Both output formats support packaging as an MSI.
+Both output formats support packaging as a Microsoft software installer (MSI) or
```
freakboy3742, line 10:
```suggestion
Both output formats support packaging as a Microsoft Software Installer (MSI) or
```

src/briefcase/platforms/windows/__init__.py
```diff
@@ -285,6 +276,22 @@ def package_app(
...
+        else:
+            self._package_msi(app)
+
+        if sign_app:
+            self.logger.info("Signing package...", prefix=app.app_name)
+            self.sign_file(
+                app=app,
+                filepath=self.distribution_path(app),
+                **sign_options,
+            )
```
freakboy3742, line 290: This will attempt to sign the .zip file, which won't work because .zip files don't allow for signing.

src/briefcase/platforms/windows/__init__.py
```diff
@@ -364,10 +371,21 @@ def package_app(
...
+        """Package the app as simple zip file."""
+
+        self.logger.info("Building zip file...", prefix=app.app_name)
+        with self.input.wait_bar("Packing..."):
+            zip_root_path = self.bundle_path(app) / f"{app.formal_name}-{app.version}"
+            self.tools.shutil.copytree(
+                self.bundle_path(app) / self.packaging_root,
+                zip_root_path,
+                dirs_exist_ok=True,
+            )
```
freakboy3742, line 384: This works, but it (a) results in a copy of the entire project, which might not be small, and (b) could be prone to stale files on the second build.

At the very least, the copied folder should be a one-use temporary folder, or deleted after use; however a better approach may be to *not* use `shutil`, and use zip file directly - that gives you much better control over the structure of the final zip file. For example, see [this StackOverflow](https://stackoverflow.com/a/32640292)  answer.

freakboy3742 (changes requested):
Getting really close - a couple more comments inline.

src/briefcase/platforms/windows/__init__.py
MarkusPiotrowski: Yes, I asked this in the conversation above, but then I didn't wait for your answer. Sorry. Google search results have let me believe that this is possible. Seems that it is not possible with Windows' SDK SignTool.

src/briefcase/platforms/windows/__init__.py
MarkusPiotrowski: I have considered some options. The 'problem' here is that we want to put the files in a root folder _that does not already exist_ before zipping. This is different to the situation described in the StackOverflow question/answer (that I also found). My impression was that neither `shutil.make_archive` nor `ZipFile` allows the creation/building of a root folder 'on the fly' while packing, so I was thinking about either to temporarily rename the `src` folder to `MyApp-1.2.3` (what I found 'hackish') or to make a plain copy and delete it afterwards (what I'm doing now).

OK, after reading this [StackOverflow answer](https://stackoverflow.com/questions/458436/adding-folders-to-a-zip-file-using-python), I see that it is possible.

MarkusPiotrowski:
Possibly, there was an unrelated error during the Python 3.9 test on macOs (CI / Unit test).
Can someone please restart the tests?

rmartin16:
hmm...`test_stuck_streamer` has been problematic before...I thought we had fixed it, though; seems another look is necessary. Thanks for the ping.

src/briefcase/platforms/windows/__init__.py
```diff
@@ -354,10 +374,23 @@ def package_app(
...
-            )
+    def _package_zip(self, app):
+        """Package the app as simple zip file."""
+
+        self.logger.info("Building zip file...", prefix=app.app_name)
+        with self.input.wait_bar("Packing..."):
+            source = self.bundle_path(app) / self.packaging_root  # /src
+            try:
+                os.mkdir(self.dist_path)  # /dist
+            except FileExistsError:
```
MarkusPiotrowski, line 385: OK, this `except` is missing coverage.

@freakboy3742 @rmartin16: I saw that testing creates several temporary folders, for each test one (?). So when I want to test a condition that a folder was already created (like here), I need to test this within the same `test_...()` function, otherwise I end up in a fresh test folder, is this correct?

rmartin16: > I saw that testing creates several temporary folders, for each test one (?).

Yes; Pytest provides the `tmp_path` fixture for each test such that it's a unique empty directory that's ultimately cleaned up by Pytest.

> So when I want to test a condition that a folder was already created (like here), I need to test this within the same `test_...()` function, otherwise I end up in a fresh test folder, is this correct?
> 
> Or I create the `dist` folder in the fixture from the beginning, so no need to run the `_package_zip()` two times.

I'd probably need to see a specific test example...but moreover, you shouldn't need to create the `dist_path` directory here; it is [already created](https://github.com/beeware/briefcase/blob/5c83de4a0967eef40b6145e42025fbe37d2ba540/src/briefcase/commands/package.py#L93) by the base `PackageCommand`.

src/briefcase/platforms/windows/__init__.py
```diff
@@ -354,10 +374,23 @@ def package_app(
...
+                pass
+
+            with ZipFile(self.distribution_path(app), "w", ZIP_DEFLATED) as archive:
+                for path, _, files in os.walk(source):
```
freakboy3742, line 389: We generally prefer the `pathlib` form of these commands to the `os` form - so `source.glob("*/**")` in this case. Similar updates are needed for the other `os` calls in this block.

src/briefcase/platforms/windows/__init__.py
freakboy3742: The mkdir call definitely isn't needed. It will be [created by the base package command](https://github.com/beeware/briefcase/blob/main/src/briefcase/commands/package.py#L93). That doesn't guarantee it exists for your *test* - so you may need to create the dist folder as part a fixture or other test setup.

tests/platforms/windows/app/test_package.py
```diff
@@ -27,12 +28,50 @@ def package_command(tmp_path):
         version="81.2.1.0",
         arch="groovy",
     )
+    command.tmp_path = tmp_path
```
freakboy3742, line 31: This isn't needed - `tmp_path` is guaranteed to be consistent within a given test. You don't need to annotated it onto the command.

tests/platforms/windows/app/test_package.py
```diff
@@ -27,12 +28,50 @@ def package_command(tmp_path):
...
+    )
+    package_command.files = (
+        "First App.exe",
+        "python.exe",
+        "python3.dll",
+        "vcruntime140.dll",
+        "app/first-app/app.py",
+        "app/first-app/resources/__init__.py",
+        "app_packages/clr.py",
+        "app_packages/toga_winforms/command.py",
```
freakboy3742, line 63: Although this evidently works on Windows, it would be preferable to use Pathlib objects and make the path separator unambiguous.

tests/platforms/windows/app/test_package.py
```diff
@@ -27,12 +28,50 @@ def package_command(tmp_path):
...
+    )
+    for path in paths:
+        (src_path / path).mkdir(parents=True)
+    for file in package_command.files:
+        open(f"{src_path}/{file}", "x").close()
```
freakboy3742, line 68: We've got a `create_file()` utility method in `tests/utils` for this purpose - it creates a file with known content, and ensures the directory exists.

freakboy3742 (changes requested):
A few more cleanups - but we're getting really close!

tests/platforms/windows/app/test_package.py
MarkusPiotrowski: I'd prefer to let it stand as it is here; this list is also used to check the content of the ZIP file, and the namelist of the ZIP file consists of strings in this format.
However, the entries will be converted to Pathlib objects when creating the file with `create_file()`, as you suggested.

tests/platforms/windows/app/test_package.py
MarkusPiotrowski: OK, got it. These pytest fixtures a still enigmatic to me.

src/briefcase/platforms/windows/__init__.py
MarkusPiotrowski: OK, did so.

tests/platforms/windows/app/test_package.py
MarkusPiotrowski: Used it now.

tests/platforms/windows/app/test_package.py
freakboy3742: But that's not what this part of the test is for. It's creating fixtures. It's *essential* that they are created with the correct, platform appropriate path separators. The test fixture should *not* also be providing the test validation data - the test should be explicitly specifying what it expects, not relying on the input from a fixture.

freakboy3742: This is especially important when what is being validated isn't in the same format as what is being put into the system in the first place.

MarkusPiotrowski: OK, changed it.

src/briefcase/platforms/windows/__init__.py
freakboy3742: This looks like the only outstanding item.

freakboy3742 (changes requested):
This is almost ready to land - the only outstanding issue I can see is the "package signing" issue. Indent that block so it only applies to MSI packaging, and we should be good.

It would also be desirable to have a test of that case, confirming that when a ZIP package is signed, only the app binary is signed. It doesn't need to actually do the signing; mocking ` `sign_app` and confirming the calls made should be sufficient.

MarkusPiotrowski:
@freakboy3742 Just to be sure that I got it right: At the moment, when a zip has been packaged, it will not go on for signing, because I put a `return` in the `if` clause (so principally I could also leave out the `else`). Indenting the last signing block makes it visual clearer that it only applies to the msi package, but then I can also remove the `return`, can't I?
I look what I can do regarding an `test_zip_file_with_signing`.

freakboy3742:
> @freakboy3742 Just to be sure that I got it right: At the moment, when a zip has been packaged, it will not go on for signing, because I put a `return` in the `if` clause (so principally I could also leave out the `else`). Indenting the last signing block makes it visual clearer that it only applies to the msi package, but then I can also remove the `return`, can't I?

My apologies - I completely missed the return in my scan of the code. 

As a general rule, we prefer to avoid using `return` as a flow control statement - "methods always enter at the top, and exit at the bottom" is a good expectation to set. There are some exceptions to this - usually when you need to shortcut logic due to an error or "no action required" case - but it's far too easy to accidentally miss an inline `return` statement (as I did in this case), so it's better to avoid the problem entirely. So - while the end result will be identical, the "indent" approach you've described would be preferred.

rmartin16:
If I may, a small nitpick while we're here: the `description` for `WindowsVisualStudioPackageCommand` and `WindowsAppPackageCommand` both explicitly reference `MSI` as the package format. There's some variation in how other platforms word this, but updating these to just `Package a Visual Studio project.` and `Package a Windows app.`, respectively, probably makes sense.

MarkusPiotrowski:
OK,  CI / Unit tests on macOS (for now: 3.8 and 3.9) throw a number of DeprecationWarnings.

freakboy3742:
> OK, CI / Unit tests on macOS (for now: 3.8 and 3.9) throw a number of DeprecationWarnings. This is not the well known `test_stuck_streamer` issue : `DeprecationWarning: pkg_resources is deprecated as an API`

Oh wow - you're having the most remarkable luck hitting CI problems that aren't your fault at all :-)

This one is because there's been a recent update to setuptools that is causing dmgbuild, the tool that we use to build macOS DMG installers, to raise a warning that pytest is considering a failure. I know how to fix it, but I'll need to push out a new dmgbuild release and then update briefcase to use that new version.

MarkusPiotrowski:
It' `test_stuck_streamer`-issue time again 🤪

rmartin16:
I've opened #1197 to hopefully resolve the issues with `test_stuck_streamer`. #1196 should resolve the new `pip` issues once it's merged.

freakboy3742 (approved):
Awesome - Thanks for the contribution! The issue with the failing macOS test is unrelated to this patch, and CI is passing, so I'm happy to merge this, and deal with the intermittent test failure independently.

MarkusPiotrowski:
Great! Thank you for your guidance with this PR.
Now I saw by chance, because of the recent discussion on discord about the briefcase templates, that there is an open ticket regarding leftover artifacts after the building process: https://github.com/beeware/briefcase-windows-app-template/issues/6#issue-1563430471
From that I learned that `python.exe` is an unwanted leftover, which is a pity, because that is one of the filenames that I put in the test module. Thus, I'm thinking about looking at this issue. My question: Except the two Python `.exe`s, is there other stuff that could/should be removed? If so, where could I learn about this?

freakboy3742:
I'm not overly concerned that the specific files in the test case include `python.exe` - the list of files generated by the template is always going to change; the test case doesn't have to be a *perfect* match for reality - just a representative example of "files that might be packaged". 

As for whether there's other stuff that can be removed - I honestly don't know. I haven't done a full audit; the exes are the obvious ones, but there might be some additional files that aren't needed. In terms of how to learn about it... I'm not sure there's much by way of documentation. The best approach will be to try and work out what role each file in the archive actually serves, and build up a list of candidates for elimination.

Issue #1182, opened 2023-04-13
RTD is Showing Docs for `v0.3.13` as `stable` After Release of `v0.3.14`


rmartin16:
### Describe the bug

The RTD build for `v0.3.14` [failed](https://readthedocs.org/projects/briefcase/builds/20086650/) on a misspelling for Pygame; although, this has since been [resolved](https://github.com/beeware/briefcase/commit/d2824a97c448b5a2a7c2716e76e57cd87f15c7db).

Looking back at the `docs` branch on `rmartin16/briefcase`, the commits that added Pygame support are indeed missing. I suppose this is an uncommon example of where the "Require branches to be up to date before merging" setting would have been quite useful.

### Steps to reproduce

See https://briefcase.readthedocs.io/en/stable

### Expected behavior

Show docs for `v0.3.14`.

I'm not sure the recommended way to resolve this; the only apparent way that occurs to me is moving the tag for `v0.3.14`...but that doesn't seem great for plenty of other reasons...

### Environment

n/a

freakboy3742:
Well... that bites. 

As an immediate workaround, I've updated RTD to show "latest". I'll leave this ticket open as a TODO reminder to revert that setting for the 0.3.15 release.

The root cause here seems to be that the RTD build failure isn't exposed as a blocking issue when doing a release. On a PR, an RTD build failure is visualised as a blocking item; however, that doesn't surface on release tags.

I can see a couple of possible systemic fixes:

1) As @rmartin16 suggests, add the "require branches to be up to date" setting. This works, but would be cumbersome when we have a lot of PRs to merge (as we do on Monday mornings when Dependabot runs, or during sprint conditions). Given that our test suite takes ~20 minutes to run, it would be desirable to minimise this sort of rebuild churn.
2) Enable the merge queue. That should minimize the number of builds that are done, but I'm not sure it's *that* much better.
3) Add a manual step to the release documentation to check the status of the RTD build before cutting the final release. This would prevent the problem, but would have also caused a headache in this case, as the 0.3.14 build would have been successfully pushed to TestPyPI, preventing a second 0.3.14 build from being created.
4) Add a docs build step to the release CI. This technically duplicates what RTD is doing, but it would surface the failure explicitly in Github's processes, and avoids the (3) problem because a successful docs lint & build would be a pre-requisite of pushing to TestPyPI.

Of these, (4) seems like the most likely to avoid this problem again in future, with a minimum of rebuild churn on end-users. Thoughts?

[...]

PR #1186, opened 2023-04-15
Verify docs were built on RTD before creating a GitHub Release

rmartin16:
## Changes
- Introduces a dependency on RTD successfully building docs for the new `v*` tag for creating a GitHub Release.

## Implementation
- When a new `v*` tag is pushed to the repo, GitHub initiates this repo's `release` workflow (and in turn the `publish` workflow as a consequence of a Release being "published").
- At the same time, presumably through a webhook between GitHub and RTD, the docs for the tag are built on RTD.
- So, this simply inserts a loop in the `release` workflow that confirms RTD successfully built the docs for the new version and allows the release process to progress.
  - This checks the `built` parameter of the payload returned from RTD's [version detail endpoint](https://docs.readthedocs.io/en/stable/api/v3.html#version-detail) which is defined as "the version has at least one successful build".
  - Since this is the first time RTD has seen the version, a value of `true` should imply the build that was just triggered completed successfully.
- As an example, consider the `v0.3.14` release from April 11 @ 0311 UTC:
  - GitHub `release` workflow [run](https://github.com/beeware/briefcase/actions/runs/4674210807)
  - RTD [build](https://readthedocs.org/projects/briefcase/builds/20086650/)

## Notes
- Depends on beeware/.github#30.
- Ref: #1182

## PR Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,11 +7,20 @@ on:
 
 jobs:
   ci:
+    name: CI
     uses: ./.github/workflows/ci.yml
+
+  docs:
+    name: Verify Docs Build
+    uses: beeware/.github/.github/workflows/docs-build-verify.yml@main
+    secrets: inherit
+    with:
+      project-name: "briefcase"
+      project-version: ${{ github.ref_name }}
 
   release:
     name: Create Release
-    needs: ci
+    needs: [ ci, docs ]
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


freakboy3742 (approved):
We can't really test this without cutting a release; but by inspection, it looks good. If it isn't, I guess I'll frantically fix it when we cut v0.3.15 :-)

PR #1188, opened 2023-04-16
Stop unintentionally creating a folder during tests on Windows

### LINKED ISSUE #1187: Running test suite leaves a top level folder "a-path-that-will-cause-an-OSError"
### Describe the bug

Running the test suite on Windows creates a top level folder named "a-path-that-will-cause-an-OSerror", which is not removed after the tests have finished.
E.g.:
D:/
|_ some other folder
|_ briefcase (the dev folder)
|_ a-path-that-will-cause-an-OSError
|_ more folders

### Steps to reproduce

1. Set up a briefcase development environment on Wndows as described here: https://briefcase.readthedocs.io/en/latest/how-to/contribute-code.html
2. Run `tox -e py`

### Expected behavior

All changes outside the dev folder and the /user/.../AppData/ folders should be reverted

### Screenshots

_No response_

### Environment

- Operating System: Windows 10 Home 21H2
- Python version: 3.8.8
- Software versions:
  - Briefcase: 0.3.14.dev151+g27996ba4

### Logs

```

```

### Additional context

_No response_

rmartin16:
## Changes
- Testing failure from writing the log file was actually depending on the call to `mkdir()` to fail.
  - This was assuming creating a root directory would fail...and it does on non-Windows systems.
  - On Windows, the call to create the root directory succeeds....but writing the logfile then fails.
```
Failed to save log to \a-path-that-will-cause-an-OSError...\logs\briefcase.2023_04_16-11_10_46.dev.log: [WinError 3] The system cannot find the path specified: '\\a-path-that-will-cause-an-OSError...\\logs'
```
- Instead of depending on actual filesystem interaction to test failure while writing the log file, this updates the test to mock the filesystem error when the file write occurs.
- This also introduces a test case for failing to create the directory....although, it's admittedly a little obtuse and probably fragile.

## Reference
- Fixes #1187

## PR Checklist:

### DIFF
--- a/tests/console/test_Log.py
+++ b/tests/console/test_Log.py
@@ -1,13 +1,13 @@
 import datetime
 from io import TextIOBase
-from pathlib import Path
 from unittest.mock import MagicMock, PropertyMock
 
 import pytest
 from rich.traceback import Trace
 
 import briefcase
-from briefcase.console import Log
+from briefcase.commands.dev import DevCommand
+from briefcase.console import Console, Log
 from briefcase.exceptions import BriefcaseError
 
 TRACEBACK_HEADER = "Traceback (most recent call last)"
@@ -27,6 +27,16 @@ def now(monkeypatch):
     datetime_mock.now.return_value = now
     monkeypatch.setattr(briefcase.console, "datetime", datetime_mock)
     return now
+
+
+@pytest.fixture
+def command(tmp_path) -> DevCommand:
+    """Provides a mocked DevCommand."""
+    command = MagicMock(spec_set=DevCommand(Log(), Console()))
+    command.base_path = tmp_path
+    command.command = "dev"
+    command.tools.os.environ = {}
+    return command
 
 
 def test_capture_stacktrace():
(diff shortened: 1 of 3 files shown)


tests/console/test_Log.py
```diff
@@ -1,28 +1,44 @@
...
+from briefcase.commands.dev import DevCommand
+from briefcase.console import Console, Log
 from briefcase.exceptions import BriefcaseError
 
 TRACEBACK_HEADER = "Traceback (most recent call last)"
 EXTRA_HEADER = "Extra information:"
 
 
-@pytest.fixture
+@pytest.fixture(autouse=True)
```
freakboy3742, line 17: I'm not a big fan of `autouse` - outside of "whole of test suite" configuration setup, I find explicit is better than implicit. For the sake of including a `now` fixture in a handful of tests, which makes it clear that "this test has a specific time interpretation".

freakboy3742 (changes requested):
Looks good; one suggested cleanup/reversion around the `autouse` on the `now` fixture.

tests/console/test_Log.py
rmartin16: That's fair; also removed it from Windows SDK tests.

freakboy3742 (approved):
Looks good!

PR #1208, opened 2023-04-24
Document use_full_install_path on Windows

### LINKED ISSUE #1199: Make the author name an optional part of the MSI install path

When an MSI installer runs, it installs the app in an `AppData/<author name>/<app name>` path. This follows the preferred scheme for Microsoft packaging, but it isn't *required*. 

It's not uncommon to see apps packaged in an `AppData/<app name>` path; and, if a solo developer is building the app, it may be preferable to use this scheme, rather than putting the author's name into the path.

A new option  `use_full_install_path` (open to suggestions for a better name) should exist for MSI installers. This will default to `True`; but if a value of `False` is specified, the author name will be excluded from the install path.

1. Do nothing; require the fully qualified install path
2. Instead of a boolean option, let the user configure everything after the AppData path. e.g., have a `install_path` setting that defaults to `{app.author_name}/{app.formal_name}`, but can be explicitly set to anything else the user may want.

I've suggested the boolean option rather than the fully-customised option because the need isn't for *full* customisation - the install path should always be the app's formal name; it's only a question of whether the author name is included as well.

### Additional context

Requested by @davidfokkema on Discord.

davidfokkema:
There is now a new option `use_full_install_path` which defaults to `True` to install the application in `Programs/<author name>/<app name>`. If you include the option in the `pyproject.toml` with a value of `False`, it installs the application into `Programs/<app name>`.

Fixes #1199.
See also: beeware/briefcase-windows-app-template#11, beeware/briefcase-windows-VisualStudio-template#15

## PR Checklist:

### DIFF
--- a/changes/1199.feature.rst
+++ b/changes/1199.feature.rst
@@ -1,0 +1,1 @@
+The company/author name in the installation path for Windows MSI installers is now optional.
--- a/docs/reference/platforms/windows/app.rst
+++ b/docs/reference/platforms/windows/app.rst
@@ -62,6 +62,16 @@ If ``true`` the installer will attempt to install the app as a per-machine app,
 available to all users. If ``false``, the installer will install as a per-user
 app. If undefined the installer will ask the user for their preference.
 
+``use_full_install_path``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Controls whether the app will be installed using a path which includes both the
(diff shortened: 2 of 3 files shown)


freakboy3742 (approved):
I've added a change note - but otherwise, this looks great. Thanks for the contribution!

(I've also merged the template PRs)

davidfokkema:
Great, thanks!!

Issue #476, opened 2020-08-18
Read requirements from setup.cfg


Czaki:
**Is your feature request related to a problem? Please describe.**

If program is provided as python package it contains description of all dependencies if files like setup.cfg or pyproject.toml (ex for poetry). It will be nice to not double this entry in files

**Describe the solution you'd like**
To avoid double providing information it will be nice to read requirements from `setup.cfg` if present and append it to `requires` entry.  

**Describe alternatives you've considered**
Manual manage or customs scripts which will overwrite `pyproject.toml` 

**Additional context**

freakboy3742:
I'm slightly hesitant on this request, because `setup.cfg` and `pyproject.toml` are for different purposes, and should be expressing dependencies in a different way.

`setup.cfg` is for packaging a *library*. A library *should not* be hard pinning package versions - it should be specifying ranges that are valid, so that pip can resolve dependencies.

`pyproject.toml` (in the context of Briefcase, anyway) is for packaging *apps*. An app *should* have hard pins for *all* it's dependencies, because the specific versions used to deploy a particular version of an app is part of the definition of the app.

Having Briefcase to read requirements from `setup.cfg` is essentially reading the wrong thing - it's reading a library configuration into an app.

However, I *could* be convinced to expand Briefcase to use Poetry lock formats (or, better still, a PyPA-standardized format for locked requirements). Poetry can (apparently) handle different lock lists for different platforms; we'll need to incorporate that into Briefcase's usage.

I'll modify the ticket title to reflect that usage; there's a lot more design work needed here, though.

Czaki:
> However, I _could_ be convinced to expand Briefcase to use Poetry lock formats (or, better still, a PyPA-standardized format for locked requirements). Poetry can (apparently) handle different lock lists for different platforms; we'll need to incorporate that into Briefcase's usage.

In such approach I also suggest support for requirements file. This file has support in `pip-tools`  (`pip-compile` command) and tools like dependabot, requires.io etc which simplify managing and updating dependecies.

freakboy3742:
We can certainly try; the problem is that requirements files *don't* have enough flexibility (by themselves). A project can only have a single `requirements.txt` file, and there's no provision for "optional" requirements. Briefcase needs to be able to specify different package lists for different platforms.

We could use different requirements files for different platforms - but AFAIK that then breaks the automated tooling around requirements files, as those tools assume a single `requirements.txt` file.

Czaki:
[...]

Issue #807, opened 2022-08-02
On Windows, `briefcase create` immediately followed by `briefcase run` skips the `build` step


mhsmith:
**Describe the bug**
```
(bw-dev) smith@xanadu:~/git/beeware/hello$ briefcase create

Application 'helloworld' already exists; overwrite [y/N]? y

[helloworld] Removing old application bundle...

[helloworld] Generating application template...
Using app template: https://github.com/beeware/briefcase-windows-app-template.git, branch 3.8
Using existing template (sha e5e330faf36c2490c74a5459c30951b516b41fa1, updated Mon Aug  1 15:58:39 2022)
Assigning helloworld an application GUID of 93bbf2ae-12d0-580c-aeb0-965dc4c04b69

[helloworld] Installing support package...
Using support package https://briefcase-support.org/python?platform=windows&version=3.8&arch=amd64
... using most recent revision
python-3.8.3-embed-amd64.zip already downloaded
Unpacking support package... done

[helloworld] Installing dependencies...
Collecting toga-core>=0.3.0.dev35
  Using cached toga_core-0.3.0.dev35-py3-none-any.whl (491 kB)
Collecting toga-winforms>=0.3.0.dev35
  Using cached toga_winforms-0.3.0.dev35-py3-none-any.whl (326 kB)
Collecting travertino>=0.1.3
  Using cached travertino-0.1.3-py3-none-any.whl (15 kB)
Collecting pythonnet>=3.0.0a1
  Using cached pythonnet-3.0.0rc4-py3-none-any.whl (250 kB)
Collecting clr-loader>=0.1.7
  Using cached clr_loader-0.1.7-py3-none-any.whl (45 kB)
Collecting cffi>=1.13
  Using cached cffi-1.15.1-cp38-cp38-win_amd64.whl (178 kB)
Collecting pycparser
  Using cached pycparser-2.21-py2.py3-none-any.whl (118 kB)
Installing collected packages: travertino, pycparser, toga-core, cffi, clr-loader, pythonnet, toga-winforms
Successfully installed cffi-1.15.1 clr-loader-0.1.7 pycparser-2.21 pythonnet-3.0.0rc4 toga-core-0.3.0.dev35 toga-winforms-0.3.0.dev35 travertino-0.1.3
Installing app dependencies... done

[helloworld] Installing application code...
Installing src/helloworld... done

[helloworld] Installing application resources...
Installing src/helloworld/resources/helloworld.ico as application icon... done

[...]

PR #1209, opened 2023-04-24
Override requirements (at a given level) from a requirements file.

moshez:
In that case, instead of merging, we override

Fixes: #476

## PR Checklist:

### DIFF
--- a/tests/commands/create/test_install_app_requirements.py
+++ b/tests/commands/create/test_install_app_requirements.py
@@ -193,6 +193,56 @@ def test_app_packages_relock_requires(
     assert myapp.test_requires is None
 
 
+def test_app_packages_prelocked_requires(
+    create_command,
+    myapp,
+    app_packages_path,
+    app_packages_path_index,
+):
+    """If an app has a valid list of requirements, pip is invoked."""
+    myapp.requires = ["first", "second==1.2.3", "third>=3.2.1"]
+    myapp.requires_lock = "requirements-test-lock.txt"
+    create_command.tools[myapp].app_context.run.return_value.stdout = (
+        "first==1.1.1\n" "second==1.2.3\n" "third==4.1.1\n"
+    )
+    (create_command.base_path / myapp.requires_lock).write_text(
+        "first==1.1.1\n" "second==1.2.3\n" "third==4.0.0\n"
+    )
+
+    create_command.install_app_requirements(myapp, test_mode=False, relock=False)
+
+    # A request was made to install requirements
+    create_command.tools[myapp].app_context.run.assert_called_with(
+        [
+            sys.executable,
+            "-u",
+            "-m",
+            "pip",
+            "install",
+            "--upgrade",
+            "--no-user",
+            f"--target={app_packages_path}",
+            "first==1.1.1",
+            "second==1.2.3",
+            "third==4.0.0",
+        ],
+        check=True,
+        env={
+            "PYTHONPATH": str(
+                create_command.bundle_path(myapp)
+                / "path"
+                / "to"
+                / "support"
+                / "platform-site"
+            )
+        },
+    )
+
+    # Original app definitions haven't changed
+    assert myapp.requires == ["first", "second==1.2.3", "third>=3.2.1"]
+    assert myapp.test_requires is None
+
+
 def test_app_packages_valid_requires_no_support_package(
     create_command,
     myapp,


tests/commands/update/conftest.py
```diff
@@ -46,7 +46,7 @@ def verify_app_tools(self, app):
 
     # Override all the body methods of a UpdateCommand
     # with versions that we can use to track actions performed.
-    def install_app_requirements(self, app, test_mode):
+    def install_app_requirements(self, app, test_mode, relock=False):
         self.actions.append(("requirements", app.app_name, test_mode))
```
freakboy3742, line 50: This should also be tracking the provided argument in the list of actions.

tests/commands/create/conftest.py
```diff
@@ -118,7 +118,7 @@ def generate_app_template(self, app):
     def install_app_support_package(self, app):
         self.actions.append(("support", app.app_name))
 
-    def install_app_requirements(self, app, test_mode):
+    def install_app_requirements(self, app, test_mode, relock=False):
         self.actions.append(("requirements", app.app_name, test_mode))
```
freakboy3742, line 122: This should also be tracking the provided argument in the list of actions.

src/briefcase/commands/create.py
```diff
@@ -504,9 +546,7 @@ def install_app_requirements(self, app: BaseConfig, test_mode: bool):
         :param app: The config object for the app
         :param test_mode: Should the test requirements be installed?
```
freakboy3742, line 547: There's a missing docstring here.

src/briefcase/commands/create.py
```diff
@@ -487,7 +488,48 @@ def _install_app_requirements(
...
+            lock_file.write_text(res.stdout)
+
+        requires = [
+            line.split("#", 1)[0].strip() for line in lock_file.read_text().splitlines()
+        ]
+        requires = [potential for potential in requires if potential != ""]
+        return requires
+
+    def install_app_requirements(
+        self, app: BaseConfig, test_mode: bool, relock: bool = False
```
freakboy3742, line 531: `relock` shouldn't have a default value here; we're in a position to be explicit whenever it's used.

freakboy3742 (changes requested):
So - there's a start of a locking implementation here, but there's still a lot of work needed.

I've flagged a bunch of stuff inline that stood out on a first pass - the most notable of which is that the code doesn't run as provided. The invocation of piptools straight up crashes in my testing... and I can't see how it has ever worked.

There's three "bigger picture" issues that need some work as well.

1. It seems like there's a bit of a missed opportunity for a simplification. This PR is using pistols to produce a requirements.txt file, which is then being post-processed back into a list of requirements. If we're going to the trouble of building a requirements file for the purposes of locking, it seems foolish to *not* to just use it as an input to `pip`.
2. It misses the overlap with the existing `install_app_requirements()` invocation that is already requirements file based. Some platforms (notably, Android, plus the AppImage and System Linux backends) don't use pip to install requirements directly - they write a requirements file based on the requires list provided.
3. Continuing that theme - there is some non-trivial backend-specific handling of requirements in the Linux, Android and web backends. It's not immediately clear to me that the changes made in this PR will be compatible with that backend handling (and, because the PR doesn't run as provided, I can't test those theories).
4. Lastly - the general workflow wasn't obvious to me on first pass. I would have expected a workflow where `create` generated a lock file when requirements were installed (possibly requiring a `--lock` argument); the user could then choose to adopt that lock file by adding it to their `pyproject.toml`. I'm not absolutely wedded to that as a UX - but requiring the user to modify the `pyproject.toml` to reference a `requires_lock` file that doesn't yet exist in order to generate a lock file with that name was definitely contrary to what I would have expected.

src/briefcase/commands/base.py
```diff
@@ -673,6 +673,12 @@ def _add_update_options(
             help=f"Update requirements for the app{context_label}",
         )
 
+        parser.add_argument(
+            "--relock",
+            action="store_true",
+            help=f"Recalculate locked requirements for the app{context_label}",
+        )
```
mhsmith, line 680: Why "relock" rather than simply "lock"? "Relock" implies that it was locked before, which isn't necessarily the case.

freakboy3742: In the way it's defined in the PR at present, it *is* a relock, because the initial create call always passes `relock=True`, automatically creating the lock file if the setting exists. This is part of my issue with this PR, as you need to specify a lock file that doesn't exist to get the initial invocation to work as you'd expect. So, "re" part of the relocking is present, but it won't always be clear how you get to the relock in the first place.

If the setting is also exposed to the `create` command, then it would definitely make sense for the option to be `--lock`. It might even make sense regardless, describing "locking" as an imperative action that is part of a build.

docs/reference/configuration.rst
```diff
@@ -302,6 +302,21 @@ application level, *and* platform level, the final set of requirements will be
...
+The name of a lock file.
+If the file exists,
+its contents will be used instead of the
+value of the
+``requires``
+field.
+
+Use
+``briefcase update -r --relock``
+to recompute the lock file.
```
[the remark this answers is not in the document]
mhsmith: How about generalizing this to `requires_file`, so it can be used for any requirements file, not just lock files? See #1270 for an example of where this would be useful.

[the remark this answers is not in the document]
freakboy3742: Agreed that it would be nice to support requirements files in general, as an alternative to pyproject.toml `requires` definitions. However, I think this is a supplement to, rather than a replacement for `requires_lock`.

The choice to introduce `requires_lock` was to allow a project to define loose requirements based on core functionality, but then lock a specific solution to the full package set required at runtime. This makes the maintenance task as an author simpler - just define the high level packages you need, and then periodically generate a new lock of the full solution so that builds are reproducible.

Before we introduced `requires_lock`, one ideas I had was to allow `requires` to be either a list or a string; if it's a list, it's a list of packages; if it's a string, it's a path to a file. I can't see any reason we couldn't still do this - `requires` accepts both list and str forms, and `requires_lock` defines a locking file. This would allow for `requires` to continue to be an additive setting (so you can define global, macOS and Windows requirements files) but `requires_lock` is a complete solution file.

An alternative to having 2 type alternatives for `requires` would be to add `requires_file`; however, I'm not 100% about the ergonomics of having 1 setting with 2 possible types, vs having 2 settings and then needing to document and implement error handling (or at least a resolution order) around projects that define both.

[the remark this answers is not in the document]
mhsmith: > I'm not 100% about the ergonomics of having 1 setting with 2 possible types, vs having 2 settings and then needing to document and implement error handling (or at least a resolution order) around projects that define both.

I think it's better to have two separate settings. Not only would that be more explicit and less error-prone, it could also be useful to define both requirements files and explicit requirements, especially when adding a BeeWare app to an established project. Such projects might also find it useful if the `requires_file` setting accepted a list of multiple files.

But I agree it's better for this PR to only deal with the locking issue: I misunderstood the title and thought it had a larger scope.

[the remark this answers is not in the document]
freakboy3742: To clarify - are you suggestion that we support both settings, and the final list of installed packages is the union of `requires` *and* `requires_file`, merged across all levels (app, platform, format etc)? I was originally thinking it would require picking one to have primacy (or, alternatively, raise an error if both are defined); supporting both is an interesting option I hadn't considered.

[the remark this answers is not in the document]
mhsmith: Yes, and I'm basing this on the experience I had developing the Android app for Electron Cash. That project has a core Python library with its own requirements file, and three apps using the library (PyQT, iOS and Android). So the app needs to use the library's requirements file, and then add some requirements of its own, either directly or via a [second requirements file](https://github.com/Electron-Cash/Electron-Cash/blob/4.2.14/android/app/build.gradle#L46). I think this would be a fairly common scenario.

mhsmith:
> I would have expected a workflow where `create` generated a lock file when requirements were installed (possibly requiring a `--lock` argument); the user could then choose to adopt that lock file by adding it to their `pyproject.toml`.

That seems a bit awkward. Instead of requiring the user to pick a filename, why not just auto-generate the filename based on the platform and format names? Then pyproject.toml doesn't need to be edited; instead, we have the following scenarios:
* Every command which can install requirements has a `--lock` option which generates or re-generates the lock file.
* If the lock file exists, it's used automatically in any subsequent commands.
* Maybe we also have an `--unlock` option to remove the lock file. In that case, the the user would never need to touch the file directly, and its name, location and format would become internal implementation details which we'd be free to change in the future. 

I don't have a clear idea of the file's name or location, but maybe we could store it under a `briefcase` or `.briefcase` subdirectory of the project. This could be used to store other things which are internal to Briefcase, but should still be checked into version control, unlike `build`.

freakboy3742:
> > I would have expected a workflow where `create` generated a lock file when requirements were installed (possibly requiring a `--lock` argument); the user could then choose to adopt that lock file by adding it to their `pyproject.toml`.
> 
> That seems a bit awkward. Instead of requiring the user to pick a filename, why not just auto-generate the filename based on the platform and format names? Then pyproject.toml doesn't need to be edited; instead, we have the following scenarios:

So, one reason is that people can be... unreasonably opinionated about the names they use for their lock files :-) 
 
> * Every command which can install requirements has a `--lock` option which generates or re-generates the lock file.
> * If the lock file exists, it's used automatically in any subsequent commands.
> * Maybe we also have an `--unlock` option to remove the lock file. In that case, the the user would never need to touch the file directly, and its name, location and format would become internal implementation details which we'd be free to change in the future.

I do like this as a workflow. My hesitation is around the slightly opaque user experience - "I've added foobar==1.2.3 to my requirements, but it's not getting installed" becomes even harder to diagnose. This is always going to be the case when lock files are involved, but when the lock file is completely implied, there's no evidence in the config file that directs the user's attention to the fact that locking is being performed.
 
> I don't have a clear idea of the file's name or location, but maybe we could store it under a `briefcase` or `.briefcase` subdirectory of the project. This could be used to store other things which are internal to Briefcase, but should still be checked into version control, unlike `build`.

This is where Opinions (tm) get involved :-) The only conventions I've seen around requirements are either (1) in the project root, or (2) in a `requirements` directory - both are workable here.

mhsmith:
> My hesitation is around the slightly opaque user experience - "I've added foobar==1.2.3 to my requirements, but it's not getting installed" becomes even harder to diagnose.

I think changing the requirements list should always invalidate an existing lock file. We could store the original (merged but unlocked) requirements file next to the lock file, and compare it with the current configuration.

This also touches on the automatic update idea discussed at https://github.com/beeware/briefcase/issues/807#issuecomment-1203354768.

> The only conventions I've seen around requirements are either (1) in the project root, or (2) in a `requirements` directory - both are workable here.

If we do decide that the lock file is an internal implementation detail, I think it should be stored in a directory whose name makes that clear.

freakboy3742:
> > My hesitation is around the slightly opaque user experience - "I've added foobar==1.2.3 to my requirements, but it's not getting installed" becomes even harder to diagnose.
> 
> I think changing the requirements list should always invalidate an existing lock file. We could store the original (merged but unlocked) requirements file next to the lock file, and compare it with the current configuration.
> 
> This also touches on the automatic update idea discussed at [#807 (comment)](https://github.com/beeware/briefcase/issues/807#issuecomment-1203354768).

Ah - the overlap with #807 changes things significantly. I was considering the lock file as something you could (concievably, although you're not likely to) write by hand, in which case manual specification and customisation options make more sense. Having the lock autogenerated on any change to requirements is another matter entirely.
 
> > The only conventions I've seen around requirements are either (1) in the project root, or (2) in a `requirements` directory - both are workable here.
> 
> If we do decide that the lock file is an internal implementation detail, I think it should be stored in a directory whose name makes that clear.

Agreed; although the lock file *is* something that would be checked into version control, so we need to be careful not to make it *too* hidden; and we need to make sure that it any "#807 state" information that feeds the lock file invalidation process is also storable in version control.

freakboy3742:
I'm going to close this - it's been almost 2 years since the last update, and the PR has drifted significantly from the current codebase.

PR #1210, opened 2023-04-24
provide user a more informative error when a git clone fails

rmblau:
This change provides the user with a more informative error message and a solution when the git clone fails.

When a user runs `briefcase create` and the clone fails, a ValueError is raised. To a newcomer the error is non obvious what the root cause is. This change provides an explicit error and a solution to resolve the error. 

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -809,10 +809,10 @@ Did you run Briefcase in a project directory that contains {filename.name!r}?"""
                 # Template cache path exists, but isn't a git repository
                 # Just use the template directly, rather than attempting an update.
                 cached_template = template
-            except ValueError:
+            except ValueError as e:
                 raise BriefcaseCommandError(
                     f"Git repository in a weird state, delete {cached_template} and try briefcase create again"
-                )
+                ) from e
         else:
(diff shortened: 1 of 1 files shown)


freakboy3742 (approved):
Looks good - thanks!

src/briefcase/commands/base.py
```diff
@@ -809,6 +809,10 @@ def update_cookiecutter_cache(self, template: str, branch="master"):
                 # Template cache path exists, but isn't a git repository
                 # Just use the template directly, rather than attempting an update.
                 cached_template = template
+            except ValueError:
+                raise BriefcaseCommandError(
+                    f"Git repository in a weird state, delete {cached_template} and try briefcase create again"
+                )
```
rmartin16, line 815: `from e` pls....wonder if there's a linter for this
```suggestion
            except ValueError as e:
                raise BriefcaseCommandError(
                    f"Git repository in a weird state, delete {cached_template} and try briefcase create again"
                ) from e
```

freakboy3742: Agreed that an automated check for this would be nice - not sure what that would look like in terms of implementation, though...

PR #1214, opened 2023-04-24
Fix up macOS signing options for `beeware package`

### LINKED ISSUE #865: Remove the `--no-sign` option and warn the user about `--adhoc-sign` on macOS applications
The `package` command provides `--no-sign` and `--adhoc-sign` options. 

This is only used on macOS. Previously, macOS allowed completely unsigned apps and DMGs. However, as of macOS 12 (Monterey), apps *must* be signed to be able to run at all; and adhoc-signed apps won't be usable on any computer other than the one that created the app. Adhoc signing is useful for local testing, but completely unsigned apps are no longer possible on macOS.

The `--no-sign` option should be removed. AdHoc signing should become the default signing option for all packaging on all platforms. This should be interpreted as "perform the minimal signing necessary to allow this app to run" - on Windows and Linux, this means "no signing".

Using the `--adhoc-sign` option on macOS should prominently warn the user that the app will work, but will not be re-distributable. 

We could retain the `--no-sign` option and make it raise an error; however, that's not a great user experience. As macOS is the only platform that performs signing at present, it makes sense to repurpose "adhoc" signing as "do the minimum legal signing".

basepi:
For `beeware package` on macOS:
- Deprecate `--no-sign`, as all macOS packages must be signed in modern macOS
- Change identity selection to always prompt unless `--adhoc-sign` or `--identity` are defined.
  - This is a change for users with only a single identity configured. Previously if no identity args were used, we would just use that identity automatically. Now we always prompt, and add the ad-hoc identity to that 
- Add the `ad-hoc` identity to the identity selection prompt
- Fix up tests for the above ^

Closes #865 

## PR Checklist:

Thank you to @mhsmith for the testing help!


rmartin16:
A thought came to mind: #865 was written before Windows code signing was implemented. A lot of console programs support `--no-X` style flags to override any other flags being used. This could remain a use-case here for signing outside of macOS.

src/briefcase/commands/package.py
```diff
@@ -6,6 +8,16 @@
...
+class DeprecatedAction(argparse.Action):
+    def __call__(option_strings, *args, **kwargs):
+        """Use warnings.warn to tell the user that this argument is deprecated and will
+        have no effect."""
+        warnings.warn(
+            f"The {option_strings} argument is no longer used, and has no effect.",
+            DeprecationWarning,
+        )
+
+
```
basepi, line 20: I could easily move this somewhere central if we think it would be useful elsewhere. Alternatively, I'm open to other suggestions, including raising an error as opposed to just warning.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -241,21 +241,20 @@ def select_identity(self, identity=None):
...
+            'Select code signing identity to use. "ad-hoc" will result '
+            "in a package that can run locally, but cannot be "
+            "redistributed. If your identity is not listed, please see "
+            "https://briefcase.readthedocs.io/en/stable/how-to/code-signing/macOS.html\n"
+            "To skip this prompt in the future, use `--adhoc-sign` or `--identity=<identity>`:"
+        )
+        self.input.prompt()
+        identity = select_option(identities, input=self.input)
+        identity_name = identities[identity]
+        self.logger.info(
```
freakboy3742, line 256: This will be misleading if the selected option is the adhoc identity.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -241,21 +241,20 @@ def select_identity(self, identity=None):
...
-            identity, identity_name = list(identities.items())[0]
-        else:
-            self.input.prompt()
-            self.input.prompt("Select code signing identity to use:")
-            self.input.prompt()
-            identity = select_option(identities, input=self.input)
-            identity_name = identities[identity]
-            self.logger.info(
-                f"""
+        identities["-"] = "ad-hoc"
```
freakboy3742, line 244: This label should be human readable, and should incorporate the "testing only" advice, rather than putting that in the prompt.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -241,21 +241,20 @@ def select_identity(self, identity=None):
...
+        self.input.prompt()
+        self.input.prompt(
+            'Select code signing identity to use. "ad-hoc" will result '
+            "in a package that can run locally, but cannot be "
+            "redistributed. If your identity is not listed, please see "
+            "https://briefcase.readthedocs.io/en/stable/how-to/code-signing/macOS.html\n"
+            "To skip this prompt in the future, use `--adhoc-sign` or `--identity=<identity>`:"
```
freakboy3742, line 251: The "use -i in future" advice isn't needed; it's provided after an identity is selected.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -577,48 +574,46 @@ def package_app(
...
-                if notarize_app is None:
-                    notarize_app = True
 
-                identity, identity_name = self.select_identity(identity=identity)
+            identity = "-"
+            self.logger.info(
+                (
+                    "Signing app with ad-hoc identity... Note that this app "
+                    "will run, but cannot be re-distributed."
+                ),
```
freakboy3742, line 598: This should be 2 messages - an info saying that the ad-hoc identiy being used and a warning that the app won't be redistributable.

tests/commands/package/test_call.py
```diff
@@ -232,14 +190,14 @@ def test_identity_arg_package_one_app(package_command, first_app, tmp_path):
...
+    """If there is one app,and an --adhoc argument, package signs the app using ad-hoc
     option."""
     # Add a single app
     package_command.apps = {
         "first": first_app,
     }
 
-    # Configure an adhoc signing option
+    # Configure an ad-hoc signing option
     options = package_command.parse_options(["--adhoc"])
```
freakboy3742, line 201: I'm not sure how this passed previously... the argument should be `--adhoc-sign`

tests/platforms/macOS/app/test_package.py
```diff
@@ -550,8 +459,8 @@ def test_dmg_with_installer_icon(package_command, first_app_with_binaries, tmp_p
     # Package the app without signing or notarization
     package_command.package_app(
         first_app_with_binaries,
-        sign_app=False,
         notarize_app=False,
+        adhoc_sign=True,
```
freakboy3742, line 463: To confirm - this test has gone from "no signing" to "adhoc signing" - which is fine, given the discussion we've had; but we need to make sure we've disabled/mocked the actual signing code, or else this is going to massively slow down the tests. Can you confirm these have been mocked?

freakboy3742:
> A thought came to mind: #865 was written before Windows code signing was implemented. A lot of console programs support `--no-X` style flags to override any other relevant flags being used. So, `--no-sign` could remain a use-case for signing outside of macOS.

I guess the question is whether we interpret "adhoc signing" as "minimum viable signing", which is "no signing" on Windows and Linux, or whether we try and find an "adhoc" interpretation that works on Windows and Linux.

src/briefcase/commands/package.py
freakboy3742: Honestly - my inclination is to just remove it completely. APIs need to be very careful about backwards compatibility, but this is a command that will (mostly) be executed by humans - they're a lot more adaptable.

freakboy3742 (changes requested):
This is a great first pass; There's a couple of complex interactions that come of out removing adhoc-sign; plus some mostly cosmetic cleanups around the prompt.

src/briefcase/commands/package.py
basepi: Good news. Because I actually did this very wrong anyway.... 😅

rmartin16:
> > A thought came to mind: #865 was written before Windows code signing was implemented. A lot of console programs support `--no-X` style flags to override any other relevant flags being used. So, `--no-sign` could remain a use-case for signing outside of macOS.
> 
> I guess the question is whether we interpret "adhoc signing" as "minimum viable signing", which is "no signing" on Windows and Linux, or whether we try and find an "adhoc" interpretation that works on Windows and Linux.

Ahhh...yeah, I see what you mean. In that case, I would _not_ let `--adhoc-sign` mean "do not sign even if `--identity` is specified" for other platforms; users will need to ensure `--identity` isn't passed if they don't want signing to happen. And adhoc signing would not make any sense on Windows....or on Linux most likely. Making this a macOS setting only probably makes sense (as well as for other relevant platforms).

Additionally, this gets rid of `sign_app` for the `_package_app()` call (which is good) but it is still being respected for windows code signing. This may also be a good time to clarify the help text for `--identity` since it's specific to macOS right now.

basepi:
> This may also be a good time to clarify the help text for --identity since it's specific to macOS right now.

Done.

rmartin16:
> > This may also be a good time to clarify the help text for --identity since it's specific to macOS right now.
> 
> Done.

Oh, sorry; I should have been more clear. The `--identity` setting is also used by the Windows package command to codesign Windows apps. I meant that the helptext currently says "40-digit hex checksum, or the full name of the identity" and the "full name of the identity" is only supported on macOS; Windows requires the 40 char cert thumbprint.

Moreover, though, I think it may be best to remove all the signing options from the base package command and implement them individually for each platform....although, that's really only a half-baked idea right now.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -577,48 +583,67 @@ def package_app(
...
-                    "Signing app with adhoc identity...", prefix=app.app_name
+                    "Signing app with ad-hoc identity...",
+                    prefix=app.app_name,
+                )
+                self.logger.warning(
+                    (
+                        "Because you are signing with the ad-hoc identity, this "
+                        "app will run, but cannot be re-distributed."
+                    ),
+                    prefix=app.app_name,
```
freakboy3742, line 631: The duplication of the adhoc warnings here suggests there's an opportunity to clean up the logic - could we restrict the initial logic to `if adhoc: identity = '-'; else identity= select_identity()`, and then do all the validation assuming that we *have* a signing identity?

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -429,7 +436,7 @@ def team_id_from_identity(self, identity_name):
             return re.match(r".*\(([\dA-Z]*)\)", identity_name)[1]
         except TypeError:
             raise BriefcaseCommandError(
-                "Couldn't extract Team ID from signing identity {identity!r}"
+                f"Couldn't extract Team ID from signing identity {identity_name!r}"
```
freakboy3742, line 439: Good catch!

freakboy3742 (changes requested):
This is really close; The logic around the identity selection is the only major issue with what is here.

I need to do a pass over the windows backend as well, just to confirm if there are any side effects even though the code hasn't been changed.

basepi:
Anything I can do to get this over the finish line?

freakboy3742:
@basepi Apologies for the delay - I've been digging myself out of the pile of PRs resulting from the sprints. This one was literally first on my list for this morning :-) I should have an update for you in an hour or two.

rmartin16:
@freakboy3742, I think it's safe to drop the `sign_app` parameter for Windows packaging if you wanted to include getting rid of that.

freakboy3742:
> @freakboy3742, I think it's safe to drop the `sign_app` parameter for Windows packaging if you wanted to include getting rid of that.

Yup - you caught me in the middle of making that exact change :-)

freakboy3742 (approved):
I've made a couple of cosmetic tweaks to documentation, and made the corresponding changes to the Windows signing backend - but otherwise, this is great! Thanks for the PR!

PR #1216, opened 2023-04-24
Refs #1201: Parallelisation of calls to app signing for macOS

### LINKED ISSUE #1201: Parallelise calls to app signing

When signing a macOS app, briefcase needs to sign each individual .so file. This is currently being done sequentially, which can take a while.

Briefcase should parallelise the calls to sign individual files in the bundle, with the degree of parallelisation reflecting the number of CPUs available on the machine.

Do nothing. Single threading works; it's just slower.

### Additional context

[ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) should be a fairly approachable way to implement this.

nondescryptid:
# Changes 

- Modified ```sign_app()`` (src > briefcase > platforms > macOS > __init.py__) to use ThreadPoolExecutor (as suggested in #1201) such that signing occurs in parallel. The [default value for the max_workers param](https://docs.python.org/3/library/concurrent.futures.html) is currently used (min(32, oc.cpu_counts() + 4).  
- Modified test in test_sign_app to allow files to not be signed in a specific sequence. My assumption here is that signing things in a specific order cannot be expected with parallelisation, but note that I am very inexperienced in this regard. 

This PR speeds up signing of files. I tried building the BeeWare helloworld tutorial app sequentially, and then with the parallelised implementation. The sequential signing took about 10 seconds, while the parallelised one took around 2.4s according to my stopwatch. 
Screenshot 2023-04-24 at 1 05 35 PM

## PR Checklist:

## Discussion needed: What parts of apps require signing to occur in a specific sequence, and if so, should parallelisation occur? 
There are comments in the sign_app() function that state: 
```
# Signs code objects in reversed lexicographic order to ensure nesting order is respected
# (objects must be signed from the inside out)
```
I am not sure about how strict this requirement is. If it is a strict requirement and lots of things might break because of not signing files in order, writing files sequentially may be the preferred implementation.
I tested my parallelised implementation with the BeeWare helloworld tutorial, which only uses Toga. I have not tested this on larger, more complex apps. 

Fixes #1201


src/briefcase/platforms/macOS/__init__.py
```diff
@@ -360,17 +361,26 @@ def sign_app(self, app, identity):
...
+        # Run signing through a ThreadPoolExecutor so that they run in parallel
+        # for every path in sign_targets, group by the lambda which returns the parent of the path. 
+        # Groups which have the same parent will be returned together in the iterator "names". Names is later cast to a list
+        for _, names in itertools.groupby(sign_targets, lambda name: name.parent):
+            with concurrent.futures.ThreadPoolExecutor() as executor:
+                names = list(names)
+                futures = []
+                progress_bar = self.input.progress_bar()
+                task_id = progress_bar.add_task("Signing App", total=len(names))
+                with progress_bar:
```
freakboy3742: This creates a new progress bar for every group/folder; there should be a single progress bar that is reused.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -360,17 +361,26 @@ def sign_app(self, app, identity):
...
+                    for path in sorted(names, reverse=True):
```
freakboy3742: This sorts the files in folders, but it doesn't guarantee the folders themselves are in reverse order.

freakboy3742 (approved):
Nice work! I've cleaned up a couple of code style issues that were causing a problem with CI, added a test for a failure case that wasn't being caught, and made a small tweak to how progress bars and group sorting are applied - but this is really solid! Thanks for the contribution!

Issue #1207, opened 2023-04-23
Crash on Linux for `run` Command with `--target` Because Docker Tool is not Verified Before Use


rmartin16:
### Describe the bug

The `briefcase run linux system --target <target>` command crashes. Releases `0.3.13` and `0.3.14` are affected.

```
❯ briefcase -V
0.3.14

❯ briefcase run --target ubuntu

[helloworld] Finalizing application configuration...
Checking Docker target image ubuntu...

Log saved to /home/russell/tmp/beeware/helloworld/logs/briefcase.2023_04_23-18_38_02.run.log

Traceback (most recent call last):
  File "/home/russell/tmp/beeware/venv-del/bin/briefcase", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/russell/tmp/beeware/venv-del/lib/python3.11/site-packages/briefcase/__main__.py", line 25, in main
    command(**options)
  File "/home/russell/tmp/beeware/venv-del/lib/python3.11/site-packages/briefcase/commands/run.py", line 273, in __call__
    self.finalize(app)
  File "/home/russell/tmp/beeware/venv-del/lib/python3.11/site-packages/briefcase/commands/base.py", line 529, in finalize
    self.finalize_app_config(app)
  File "/home/russell/tmp/beeware/venv-del/lib/python3.11/site-packages/briefcase/platforms/linux/system.py", line 188, in finalize_app_config
    self.tools.docker.prepare(app.target_image)
    ^^^^^^^^^^^^^^^^^
AttributeError: 'ToolCache' object has no attribute 'docker'
```

### Steps to reproduce

`briefcase new --no-input && cd helloworld && briefcase run --target ubuntu`

### Expected behavior

Briefcase should verify the `Docker` tool before it is used in `LinuxSystemPassiveMixin.finalize_app_config()`.

### Environment

- Operating System: ubuntu 23.04
- Python version: 3.11
- Software versions:
  - Briefcase: `0.3.13`, `0.3.14`, `0.3.15.dev50+g2d19e158`

### Logs

<details>

<summary>Briefcase Log File</summary>

```
Date/Time:       2023-04-23 18:53:26 
Command line:    /home/user/github/beeware/briefcase/venv-3.11-briefcase/bin/briefcase run --target ubuntu

OS Release:      Linux 6.2.0-20-generic
OS Version:      #20-Ubuntu SMP PREEMPT_DYNAMIC Thu Apr  6 07:48:48 UTC 2023
[...]

PR #1222, opened 2023-04-24
refactor linux mixins to remove target option for run command

cguardia:
Make ``use_docker`` False by default in ``LinuxSystemPassiveMixin`` and move docker processing and related ``target`` option to ``LinuxSystemMostlyPassiveMixin``.

This removes the ``target`` option from the run command, which avoids a confusing error when trying to use that option. It was decided  that the option does not make sense for the run command.

refs #1207 

## PR Checklist:

- [ x] All new features have been tested
- [x ] I will abide by the code of conduct

### DIFF
--- a/changes/1207.bugfix.rst
+++ b/changes/1207.bugfix.rst
@@ -1,0 +1,1 @@
+Refactored Linux system mixins to remove ``target`` option and related processing from ``run`` command.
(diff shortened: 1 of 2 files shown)


changes/1207.bugfix.rst
```diff
@@ -0,0 +1 @@
+Refactored Linux system mixins to remove ``target`` option and related processing from ``run`` command.
```
freakboy3742, line 1:
```suggestion
The option to run Linux system packages through Docker was removed.
```

freakboy3742 (changes requested):
This looks great - thanks! I've got one small tweak for the change note, but otherwise this looks really solid.

Issue #1115, opened 2023-03-01
Provide command line option to override app settings


freakboy3742:
### What is the problem or limitation you are having?

PR #1107 initially failed CI because the verify-app tests require the use of a specific template branch on each repository.

### Describe the solution you'd like

It should be possible to override any app setting at the command line. For example:

```
briefcase create --app-setting template_branch=foobar
```
or
```
briefcase create -S template_branch=foobar
```

would be the equivalent of editing `pyproject.toml` and adding a template_branch setting for the platform being built. 

It should be possible to provide multiple setting overrides:

```
briefcase create -S template=https://example.com/template -S template_branch=foobar
```

Argparse `action="append"` allows for this.

### Describe alternatives you've considered

It is possible to override the template branch by changing the app configuration file. However, it is undesirable to modify a configuration file in a CI context, because the modification will become unnecessary as soon as CI passes.

### Additional context

While the motivating use case for this feature is PRs that require a custom template branch, another use of this feature would be modifying the version at build time. By passing in `--app-setting version=1.2.3`, it would be possible to use CI to control the app version.

PR #1106 adds a `finalize()` method to the base command; the end of this method would seem a reasonable place to apply any overrides. This then guarantees that command line arguments are the final applied values for any app setting.

mhsmith:
I think `--config` would be a more readable spelling of this option, and it's also consistent with the wording used in the [documentation](https://briefcase.readthedocs.io/en/stable/reference/configuration.html). Possibly shortened to `-c` as in the [`git` command](https://git-scm.com/docs/git), though I haven't checked whether that conflicts with any existing option.

Ewelina179:
Hi @freakboy3742 Can I work on this task?

freakboy3742:
@Ewelina179 Absolutely! We don't have a formal process for accepting/assigning tasks; if it's open, it's available to be worked on. Drop a comment on the ticket to indicate that you're working on it (we're about to start the PyCon US sprints, so we need to do a *little* coordination to make sure we don't waste anyone's time); and start hacking!

Ewelina179:
Thanks ;) Great to know. So I'm starting to work on that task.

PR #1224, opened 2023-04-24
Adds briefcase project version to briefcase cli

rmblau:
This change adds a feature to the briefcase cli that allows the user to get their project version with the command `briefcase project version`. The output of `project version` is, for example,  `project.version=0.0.1`

This PR implements the feature requested in issue #360

## PR Checklist:

### DIFF
--- a/changes/360.feature.rst
+++ b/changes/360.feature.rst
@@ -1,0 +1,1 @@
+add `briefcase project version` command to briefcase
--- a/docs/reference/commands/index.rst
+++ b/docs/reference/commands/index.rst
@@ -12,5 +12,6 @@ Command reference
    build
    run
    package
+   project
    publish
    upgrade
--- a/docs/reference/commands/project.rst
+++ b/docs/reference/commands/project.rst
@@ -22,4 +22,4 @@ The following options can be provided at the command line.
 ``version``
 ---------------------------------------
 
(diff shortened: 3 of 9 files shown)


freakboy3742:
Thanks for the PR; however, in discussing one minor detail about the implementation (should the output include the `project.version=` prefix), I think I've been convinced that this isn't a good fit for Briefcase in the first place.

Firstly - although this has been exposed as a new command, the Briefcase configuration file actually allows for version to be overridden at the level of individual platforms (so you could have a 1.2.3.post1 point release on a particular platform.

Secondly - the original use case for this was (as I understand it) CI, and being able to determine the version that was built by Briefcase. On reflection, extracting the version from the Briefcase configuration is the wrong way to do this - #1115 would allow you to specify the version from the command line; and the full implication of tickets like #1235 is that we should allow for dynamic version specifiers.

I'm also having difficulty thinking of any other attribute that we might want to add to a `project` command; so - we're going to end up with a top level command that isn't that useful, and doesn't have a meaningful expansion path.

Apologies for not catching this earlier - I should have done a stronger audit of the idea before the sprints.

rmblau:
Thank you for your feedback @freakboy3742! I shall continue to look for ways to contribute to the project :)

Issue #1104, opened 2023-02-22
Web / Add Support for "Pyodide Packages"


rmartin16:
### What is the problem or limitation you are having?

Pyodide [maintains](https://github.com/pyodide/pyodide/tree/main/packages) a [collection](https://pyodide.org/en/stable/usage/packages-in-pyodide.html#packages-in-pyodide) of popular Python packages that are compatible with the pyscript runtime.

It is currently difficult to use these packages within the Briefcase web workflows.

### Describe the solution you'd like

Briefcase static web format should support adding Pyodide packages as requirements to the app.

### Describe alternatives you've considered

Manually add the needed package to `pyscript.toml`. This needs to happen each time the app is rebuilt.
```
packages = [
    "/static/wheels/helloworld-0.0.1-py3-none-any.whl",
    "/static/wheels/toga_core-0.3.0-py3-none-any.whl",
    "/static/wheels/toga_web-0.3.0-py3-none-any.whl",
    "/static/wheels/travertino-0.1.3-py3-none-any.whl",
    "numpy",
]
```

See the package loaded in the browser console:
```
[pyscript/main] Packages to install:  (5) ['/static/wheels/helloworld-0.0.1-py3-none-any.whl', '/static/wheels/toga_core-0.3.0-py3-none-any.whl', '/static/wheels/toga_web-0.3.0-py3-none-any.whl', '/static/wheels/travertino-0.1.3-py3-none-any.whl', 'numpy']0: "/static/wheels/helloworld-0.0.1-py3-none-any.whl"1: "/static/wheels/toga_core-0.3.0-py3-none-any.whl"2: "/static/wheels/toga_web-0.3.0-py3-none-any.whl"3: "/static/wheels/travertino-0.1.3-py3-none-any.whl"4: "numpy"]
pyodide.asm.js:10 Loading numpy
pyodide.asm.js:10 Loaded numpy
```

It does seem eventually that library maintainers will be able to provide packages on PyPI that will be compatible with browser runtimes.

In theory, it should also probably be possible to download Pyodide's numpy [wheel](https://cdn.jsdelivr.net/pyodide/v0.21.3/full/numpy-1.22.4-cp310-cp310-emscripten_3_1_14_wasm32.whl) but `pip` gets upset about the platform for the downloaded wheel.
```
>>> Running Command:
[...]

PR #1225, opened 2023-04-24
allow micropip to install pyodide wheels

reynoldsnlp:
Currently, briefcase tries to load compiled wheels from the host platform, which leads to errors when loading in the browser. This PR parses the whl filename and any wheels that are not pure python or wasm32 are replaced with simply the package name, so that micropip can install it. We do not attempt to bundle the whl file compiled for the host platform.

In the future, when Pyodide starts hosting an index, we should download the pyodide wheel and bundle that wheel.

## PR Checklist:

### DIFF
--- a/changes/1225.bugfix.rst
+++ b/changes/1225.bugfix.rst
@@ -1,0 +1,1 @@
+Web platform: Allow micropip to install binary dependencies.
--- a/changes/1225.bugfix.rst
+++ b/changes/1225.bugfix.rst
@@ -1,1 +1,1 @@
-Web platform: Allow micropip to install binary dependencies.
+Web platform: Allow micropip to attempt to install binary dependencies if wasm32 binary wheel is not bundled.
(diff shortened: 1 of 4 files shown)


rmartin16:
I think this may be related:
- https://github.com/beeware/briefcase/issues/1104

reynoldsnlp:
@rmartin16 Yes, #1104 and #1227 should both be linked to this PR so they close upon merge.

freakboy3742 (changes requested):
I can see what you're doing here, and it's clever... a little bit too clever :-) We're downloading a platform wheel, then using the existence of a platform wheel to determine whether the wheel reference should be ignored. This will work, but it's going to result in a lot of unused files in the static folder; and potentially inconsistencies in package versions (since the version request in the requires list will be ignored in the resolution to pyscript).

I think I'd prefer to see this addressed at the original installation phase, rather than the "use" phase. In the simplest approach, the list of packages published by Pyodide is known, so we can filter what is pip installed (and that what wheels exist on disk); alternatively, we can potentially self-host copies of the pyodide wheels on a private index; better still would be to convince Pyodide to publish a simple index of their packages.

PR #1228, opened 2023-04-24
add coverage report to tox.ini

cguardia:
Added a coverage report at the end of the test output

it's helpful to have coverage information when there's a coverage check in the CI.

## PR Checklist:

### DIFF
--- a/changes/1228.misc.rst
+++ b/changes/1228.misc.rst
@@ -1,1 +1,1 @@
-Added coverage report to tox.ini and updated docs to mention it.
+Added coverage command to tox.ini and updated docs to mention it.
--- a/changes/1228.misc.rst
+++ b/changes/1228.misc.rst
@@ -1,0 +1,1 @@
+Added coverage report to tox.ini and updated docs to mention it.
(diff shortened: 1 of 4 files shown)


freakboy3742 (dismissed):
Providing a tox-based mechanism for reporting coverage is definitely desirable; however, we can't do it like this, because it will break CI. If a CI run invokes `combine`, then the reports produced by CI will be "pre-combined", which will prevent us from evaluating coverage on a project-wide basis.

mhsmith (changes requested):
There are a couple of problems with making it a separate testenv section:
* It loses the ability to run coverage on a specific Python version
* It duplicates all the rest of the content in the section: e.g. you've already duplicated the `extras` and the `coverage run` command, and for completeness it should probably have the  `passenv` and `download` lines as well.

Instead, how about using the same technique as the `testenv:docs` section further down in the file? That creates environments `docs`, `docs-lint`, and `docs-all`, and runs different commands depending on whether `lint` or `all` were used. We could do something similar with the optional coverage commands.

mhsmith:
Separately, you'll see at the bottom of the PR that the docs build failed because of a spelling error. In this case, the word "codebase" is a real word, so you can fix the failure by adding it to docs/spelling_wordlist.

mhsmith (approved):

PR #1231, opened 2023-04-25
Issue 1202 - Make wizard work with existing projects

### LINKED ISSUE #1202: Provide an wizard for existing apps

`briefcase new` works well for bootstrapping a new project from scratch. However, if you have an existing project, the wizard can't really be used - your only option is to write (or update) your `pyproject.toml` config from scratch

Briefcase should have an option (`briefcase new --sources src/myapp`) that can be used to set up an existing project with Briefcase. This mode would ask the usual wizard questions, use the location specified by `--sources`, and write a first draft of a `pyproject.toml` - updating the existing `pyproject.toml` if it already exists.

Treat this problem as a documentation issue.

### Additional context

The easiest approach will likely be to use the cookiecutter new project template to generate a full dummy project config into a temporary directory, then merge the `pyproject.toml` in the dummy with the one that the project already has, and then clean up the temporary directory. 

When the user specifies a sources of `src/myapp`, there should be some error detection to ensure that:
1. The specified path exists
2. It contains a `__main__.py`

Once that has been verified, there's no need to ask the user for the appname - it should be the tail of the path given to `--sources` (i.e., `--sources src/myapp` would be an app name of `myapp`). The user *should* still be asked for formal name, bundle etc.

It may also be necessary to provide a `--test-sources` option so that the user can optionally specify the location of their test code. 
1. Specifying `--test-sources` but not `--sources` should be an error; 
2. if `--sources` is specified, but `--test-sources` isn't, the template should be written as if there are *no* test sources; 
3. If `--sources` *and* `--test-sources` is specified, the testing entry script generated by the dummy template should be copied into the test sources folder.

When in `--sources` mode, the wizard should also audit the project to highlight any extra files that may be needed. For example, the default template generates a LICENSE and CHANGELOG file - if these aren't present in the existing project, the user should be warned that they should add them (or copy over the versions from the dummy project with a warning that dummy content has been added).

yngvem:
Fixes #1202 (co-authored with @MarieRoald)

To make the wizard work with existing projects, we have added a `--target` option to `briefcase new`. The help text of this option states

```raw
  --target TARGET       The root path of the briefcase project. If the path
                        exists, then the wizard assumes that it contains an
                        existing project and will attempt to set up briefcase
                        with it (by merging the pyproject.toml files).
                        Otherwise, a new Briefcase project will be created at
                        the specified path.
```

If ``--target`` is specified, the wizard will run in a temporary directory. Then, dependent on whether the target directory exists, it will either merge the ``pyproject.toml`` files or copy all files generated by the wizard.

When the wizard is run and ``--target`` points to an existing directory, the user is additionally asked for the source directory and test source directory. The source directory must contain a ``__main__.py`` file, and the test source directory must exist (but may be empty). Furthermore, if the target directory is missing either a LICENSE file or a CONTRIBUTING file, then the wizard will show a warning message.

One important point is that for this change to work, we also need to change the template (see [this PR](https://github.com/beeware/briefcase-template/pull/67) for that).

## PR Checklist:


src/briefcase/commands/new.py
```diff
@@ -516,16 +645,89 @@ def verify_tools(self):
...
+                        f"License file not found in {target}. You should either create a new license file in {target}"
+                        " or rename an already existing license file to LICENSE."
+                    )
+
+                changelog_file = target / "CHANGELOG"
+                if not changelog_file.is_file():
+                    self.logger.warning(
+                        f"Changelog file not found in {target}. You should either create a new changelog file in"
+                        f" {target} or rename an already existing changelog file to CHANGELOG."
+                    )
```
freakboy3742, line 731: There's one more resource that needs to be copied - the test entry script. This will be `tests/appname.py` in the generated template; that entry script should be copied into the `tests` folder.

src/briefcase/commands/new.py
```diff
@@ -327,6 +410,43 @@ def build_app_context(self):
...
+
+For example, if you have an existing project, myapp, and the path supplied to
+--target points to the root directory of this application (e.g. myapp) and the
+tests are located in ``myapp/tests``, then you should set test source directory
+``tests``.
+""",
+                variable="test source directory",
+                default="tests",
+                validator=self.create_test_source_dir_validator(target),
+            )
```
freakboy3742, line 445: There's an edge case here of a project that doesn't have tests. This question will require the existence of an actual tests folder.

freakboy3742 (changes requested):
This is definitely on the right track. I've flagged a couple of edge cases inline, and some areas where the testing needs some minor changes, but this isn't far from being mergable.

src/briefcase/commands/new.py
```diff
@@ -516,16 +645,89 @@ def verify_tools(self):
...
+                    )
+
+                changelog_file = target / "CHANGELOG"
+                if not changelog_file.is_file():
+                    self.logger.warning(
+                        f"Changelog file not found in {target}. You should either create a new changelog file in"
+                        f" {target} or rename an already existing changelog file to CHANGELOG."
+                    )
+            else:
+                shutil.copytree(project_dir, target)
```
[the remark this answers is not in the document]
MarieRoald: Agreed! We have attempted to refactor it out into sub methods now. While we didn't change the tests, the test coverage is still 100%. Is this OK, or should we also add tests explicitly for these new utility methods before merging?

tests/commands/new/conftest.py
```diff
@@ -40,7 +41,27 @@ def verify_app_tools(self, app):
...
+placeholder = "abc"
+"""
+            )
+
+        with open(out_dir / "CHANGELOG", "w") as file:
+            file.write("CHANGELOG")
+        with open(out_dir / "LICENSE", "w") as file:
+            file.write("LICENSE")
+
+        return out_dir
```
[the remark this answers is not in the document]
MarieRoald: We tried making a `create_new_app_output` fixture that creates these files. Is that the correct approach here?

src/briefcase/commands/new.py
MarieRoald: We're a bit unsure about what to do here. When looking closer at this, we found something we are unsure if is a problem. If we understand this correctly, we can specify a folder for the tests in pyproject.toml. But in [`tests/appname.py`](https://github.com/beeware/briefcase-template/blob/main/%7B%7B%20cookiecutter.app_name%20%7D%7D/tests/%7B%7B%20cookiecutter.module_name%20%7D%7D.py), it seems to be hardcoded (on [line 18](https://github.com/beeware/briefcase-template/blob/6e8a0d2777244d13b34a6572a5be45bd71a9a229/%7B%7B%20cookiecutter.app_name%20%7D%7D/tests/%7B%7B%20cookiecutter.module_name%20%7D%7D.py#L18)). I guess we could change the cookiecutter to use the provided test source directory (which we added to [our branch](https://github.com/yngvem/briefcase-template/blob/c2f67cb9ced448f6e6caf9d99ee9d3a7bf49a12e/%7B%7B%20cookiecutter.app_name%20%7D%7D/pyproject.toml#L21)) in [`tests/appname.py`](https://github.com/beeware/briefcase-template/blob/main/%7B%7B%20cookiecutter.app_name%20%7D%7D/tests/%7B%7B%20cookiecutter.module_name%20%7D%7D.py) as well, but we're unsure if hardcoding this path is the correct option. Could we somehow read this from the pyproject file or simply the name of the directory the tests lie in, or could this cause some issues?

src/briefcase/commands/new.py
MarieRoald: Our thought here was that it's better to always ask the user for a test directory that exists since this will prevent a misspelling (e.g. writing test instead of tests) from breaking the briefcase setup. If a project has no tests, can the user create an empty tests directory? What do you think? Should the wizard create an empty test directory if it doesn't exist and show a warning instead?

freakboy3742: Validation is useful for catching the mis-spelling of a tests folder that *exists*, but it doesn't help with the "there aren't any tests to find" case. 

In both cases, we also need to do something about the test bootstrapping script. Briefcase requires that the test script has a termination sentinel, which existing test suites won't have.

I feel like we need to either (a) have an additional "Does you have a test suite?" question, or (b) allow "" as a valid answer to the question for the "I don't have tests" case. If there's no test suite, then we can copy in the version that has been templated; if they *do* have a test suite, we can copy just the bootstrap `tests/appname.py` script.

src/briefcase/commands/new.py
```diff
@@ -327,6 +411,43 @@ def build_app_context(self):
         # The module name can be completely derived from the app name.
         module_name = self.make_module_name(app_name)
 
+        default_source_dir = f"src/{module_name}"
```
freakboy3742, line 414: This workflow feels a little backwards. If a user has specified a `--target`, and that folder contains a `pyproject.toml` or `setup.cfg`, then we should be able to extract a candidate formal name from that metadata, and use that to generate a candidate appname. 

At the very least, there should be some acknowledgement in the prompt text that we're converting an existing app. The current process start with "Let's build a new Briefcase app", which doesn't suggest anything about migrating and existing app. In my initial testing, I wasn't sure if I'd activated the right branch, because the workflow was identical until I was asked for a source directory.

src/briefcase/commands/new.py
```diff
@@ -516,16 +646,154 @@ def verify_tools(self):
...
+            self.migrate_necessary_files(project_dir, target)
+
+    def create_and_move_new_app(
+        self,
+        template: Optional[str] = None,
+        template_branch: Optional[str] = None,
+        target: Path = None,
+        **options,
+    ):
+        """Create a new briefcase app in a given directory.
```
freakboy3742, line 753: I'm not sure I follow why creating in a temp folder and copying is needed for the "new app" case. There doesn't appear to be any re-use of the "generate into temp folder" logic (i.e., always generate into a template folder, but varying what is copied out) - there's two independent uses of the `tmp_base_path` context manager.

If the issue is creating into a specific folder - wouldn't that be better served by making the base path configureable when the template is rolled out?

freakboy3742 (changes requested):
I've raised a couple of high-level workflow issues, but what is here looks really solid. It's probably close enough that I could carry it the rest of the way without much effort; but if you'd rather keep working on it yourself, let me know and I'll continue to provide reviews as you go.

As an aside - I didn't realise that you two were the people behind the computer aided embroidery poster presentation! That was really nice stuff!

yngvem:
We'd be happy to work on it, but our schedule is unfortunately quite busy until the first week of June (ish). If you want to get this merged before mid June, then it's probably best if you continue yourself, otherwise we'll get back to it in June.

freakboy3742:
> We'd be happy to work on it, but our schedule is unfortunately quite busy until the first week of June (ish). If you want to get this merged before mid June, then it's probably best if you continue yourself, otherwise we'll get back to it in June.

Totally understood - I don't think this is *super* urgent, so if you're interested in continuing, we can wait until you're available again in ~June.

src/briefcase/commands/new.py
MarieRoald: That makes sense! We have now updated the code so it supports the second option: "" for "I don't have tests". In that case, the whole test source directory is copied. If they have tests, only `tests/appname.py` is copied and otherwise, the full test-dir is copied.

src/briefcase/commands/new.py
MarieRoald: Good point! So far we have changed the text descriptions to reflect that we are setting up an existing app is set up for Briefcase whenever the `--target` directory is provided and it exists. We could also try to extract the metadata if they have a pyproject.toml file that follows [PEP621](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata). What do you think?

src/briefcase/commands/new.py
MarieRoald: The reason we're creating a temp folder is that currently, `--target` is used to specify the exact directory where Briefcase should be set up, and the cookiecutter template (as far as we understand) creates a new directory with the name `app_name`.

So `__call__` essentially has three options right now:

 * No `--target`: no change to behaviour
 * `--target` exists: Briefcase assumes that `--target` points to a python project and attempts to set it up for briefcase
 * `--target` doesn't exist, but the parent of `--target` does exist: Briefcase runs the wizard and creates a new project with root directory `--target` (i.e. not `--target/app_name`)

We could remove the last option altogether and only accept existing targets, but check if there's a `setup.py`/`setup.cfg`/`pyproject.toml` file in the target. Then, if there is no setup-file, Briefcase could create a new directory `--target/app_name` where the project is initialised. In that case, we wouldn't need the temporary directory when creating new Briefcase apps.

What do you think?

src/briefcase/commands/new.py
freakboy3742: I'll likely tweak the specific text before landing this, but this definitely sounds like the right direction. In particular, we should be "hinting" as much as possible from as many sources as possible - if PEP621 metadata exists, we should be using it.

src/briefcase/commands/new.py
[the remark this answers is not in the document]
freakboy3742: If coverage is still 100%, there's no need to add tests. If there's a new edge case introduced by potential arguments, it might be worth adding some additional explicit tests; but otherwise, I'm comfortable with the logic being tested implicitly by usage higher up.

src/briefcase/commands/new.py
freakboy3742: I don't think this is quite right - or, at least, it's potentially skipping over a decision that we should make explicitly.

To my eyes, there's 4 possible cases at runtime:
1. No `--target`
2. `--target` points a non-existent directory
3. `--target` points at an existing directory that does *not* contain `setup.py/setup.cfg/pyproject.toml`
4. `--target` points at an existing directory that contains `setup.py/setup.cfg/pyproject.toml`

The first and last cases are the easy ones. Case 1 is existing behavior; the wizard asks for the new project name, and creates a directory with that name. Case 4 is the "migrate existing" case; the wizard reads whatever it can from the metadata in the target folder, and generates pyproject.toml accordingly.

The question is what to do with the remaining 2 cases. I can see two possibilities, depending on whether we consider `--target` to be a "migration" option or a "destination" option.

If it's a "migration" option, case 2 and 3 are errors. If you specify a target that doesn't contain project metadata, it's not an existing project, so we should error out and let the user know that the specified target isn't an existing project.

If it's a "destination" option, then we're specifying where we want the final product to end up. In that case, it doesn't matter whether the directory exists or not; what we're specifying is where we want the final files to end up. If there's metadata there, that's just another source of wizard hints (which is case 4); but if there isn't metadata, the last name part of the target folder becomes the app name, and the final `pyproject.toml` is generated in whatever target was specified. If the folder *didn't* exist, then we can copy in all the template code; if it did, we just create the pyproject.toml.

I guess there's also middle ground: Make case 1 and case 2 the same, except that in case 1, you have to ask the user for the app name; and make case 3 and 4 the same, except that case 3 needs to take the app name from the directory, rather than app metadata. 

Based on your explanation, I *think* you're leaning towards the "migration" definition - and I think I agree. The "destination" interpretation is going to be a more complicated to implement, and will be prone to randomly creating directories full of code because someone mistyped a directory name - and at the end of the day, it's a lot of complexity for a use case that I'm not sure really exists. You're either starting a new project, or you're converting an existing one, and for all the complications of cases 2 and 3, the alternate implementation is `mv oldname newname`. :-)
 
My only hesitation with this comes from the naming of `--target`. The name `target` makes it sound like it's a "destination" argument - which it definitely isn't under the "migration" interpretation. I'm tempted to suggest that we should replace ``--target`` with a simple boolean ``--convert`` option, and *only* support converting the current working directory. I'm not sure there's much of a use case for "migrate that project in that directory over there"; "make my current project Briefcase compatible" is a lot more likely. 

The other option, of course, would be to have a different command entirely - possibly sharing a lot of common implementation, but `briefcase convert` converts a pyproject.toml in your current folder, and `briefcase new` creates an entirely new project directory.

Thoughts?

src/briefcase/commands/new.py
freakboy3742: The way Briefcase runs tests is to run `python -m tests.appname` (or its equivalent) - so the `appname.py` is part of the briefcase contract; and it contains content that won't exist in any existing test harness (it needs to emit a termination sentinel). This is required because Briefcase isn't a literal Python interpreter, so you can't just "run `pytest` over a folder" - we need to set up the test suite inside the packaged app.

`pyproject.toml` specifies the additional source folders you want to copy into your app when it's running in test mode; the simplest approach is to put all your tests in a `tests` folder (and most projects follow this convention anyway, so it's not much of a burden). However, just having those tests isn't enough - the `tests.appname` module entry point needs to exist.

tests/commands/new/conftest.py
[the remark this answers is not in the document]
freakboy3742: Yes - this looks to be on the right track.

src/briefcase/commands/new.py
yngvem: That's a good point! We agree that the boolean `--convert` option might be easier to use and more maintainable than the `--target` option, and the other option of having a `briefcase convert` command sounds even better. 

Having a `briefcase convert` command would probably be more discoverable for new users -- it'd be immediately visible by just writing `briefcase` in the terminal. This option would also maybe be easier to maintain since we would more clearly separate the logic for converting an app for `briefcase` and setting up a new app for briefcase.

MarieRoald:
We made an attempt at adding a new convert-command to briefcase (as discussed [here](https://github.com/beeware/briefcase/pull/1231#discussion_r1232944917)). Currently, it will read PEP621-formatted metadata to try and help the wizard, but with options to overrule the inferred values whenever needed.

We plan to rebase on top of main soon, add tests, and revert the changes to `briefcase new`, but we wanted to check a couple of things with you first:

1. Do you think this new layout makes sense? You can run it yourself by running `briefcase convert --template=gh:yngvem/briefcase-template --template-branch=explicit-source-dir-field`
2. We still need a way to copy dependencies automatically - does Briefcase support standard PEP508-formatted dependency strings (e.g. `package_name > 1.2.1; os_name != 'nt'`)? In that case, we could copy the dependency list from the `[project]` field in the pyproject.toml file into `[tool.briefcase.app.{{app_name}}]`.

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,866 @@
...
+        """Verify that the tools needed to run this command exist.
+
+        Raises MissingToolException if a required system tool is missing.
+        """
+        super().verify_tools()
+        git.verify_git_is_installed(tools=self.tools)
+
+    @wraps(NewCommand.add_options)
+    def add_options(self, parser):
+        return NewCommand.add_options(self, parser)
```
freakboy3742, line 86: This pattern feels like it's trying to implement subclassing without actually doing subclassing. AFAICT, if ConvertCommand is a subclass of NewCommand (or vice versa), most of this implementation would no longer be required.

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,866 @@
...
+            intro += " Based on the PEP-621 formatted pyproject.toml file, we believe it could be"
+            intro += f"\n'{default}'"
+        else:
+            default = "My first application"
+
+        description = self.input_text(
+            intro=intro,
+            variable="description",
+            default=default,
+        )
```
freakboy3742, line 347: We don't need *both* a PEP621 description and a `tool.briefcase.description`. If they've got PEP621 metadata, we can use it as-is and omit that key from the `tool.briefcase` section.

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,866 @@
...
+        if test_source_dir:
+            test_entry_script = project_dir / test_source_dir / f"{project_dir.name}.py"
+            copy2(
+                test_entry_script,
+                self.base_path / test_source_dir / f"{project_dir.name}.py",
+            )
+        else:
+            copytree(project_dir / "tests", self.base_path / "tests")
+
+    def setup_existing_app_for_briefcase(
```
freakboy3742, line 723: Following the pattern of other commands, this entry point should be `convert_app()`. The fact that it calls a command called `new_app` that unpacks the app from a template suggests that Convert should extend New.

freakboy3742 (changes requested):
> We made an attempt at adding a new convert-command to briefcase (as discussed [here](https://github.com/beeware/briefcase/pull/1231#discussion_r1232944917)). Currently, it will read PEP621-formatted metadata to try and help the wizard, but with options to overrule the inferred values whenever needed.
> 
> We plan to rebase on top of main soon, add tests, and revert the changes to `briefcase new`, but we wanted to check a couple of things with you first:
> 
> 1. Do you think this new layout makes sense? You can run it yourself by running `briefcase convert --template=gh:yngvem/briefcase-template --template-branch=explicit-source-dir-field`

I've left some comments in the code about the broad code structure; I think there's some potential for cleanup by making better use of subclassing. 

In terms of the workflow - I think the general approach is headed in the right direction; however, there's a couple of questions that will be redundant. I've flagged them inline as well.

> 2. We still need a way to copy dependencies automatically - does Briefcase support standard PEP508-formatted dependency strings (e.g. `package_name > 1.2.1; os_name != 'nt'`)? In that case, we could copy the dependency list from the `[project]` field in the pyproject.toml file into `[tool.briefcase.app.{{app_name}}]`.

Not at present. It supports the version specifier part, but not the "qualifier" part (including and after the semicolon). Supporting PEP508 environment specifiers is arguably something we should support; however, we need to be careful how those specifiers and the platform-specific sections of the `[tool.briefcase]` configuration interact.

src/briefcase/commands/convert.py
yngvem: We were considering that. However, we weren't sure if it made sence to say that a `ConvertCommand` is a special case of a `NewCommand` (or vice versa). One possible solution is to abstract the shared behaviour into a `SetupProjectCommand` that both `ConvertCommand` and `NewCommand` inherits from. What do you think of that solution?

src/briefcase/commands/convert.py
yngvem: Hmmm, a solution to that might be changing the template so if the description field is empty, then it's skipped?

src/briefcase/commands/convert.py
freakboy3742: Based on some of the subsequent logic, my initial impression is that "Convert" is a subclass of "New" - after all, Convert will be invoking the same "ask questions and roll out a template" logic, but with an additional "merge" step, and some variations on which questions are asked.

However, a common base class also seems like it could be a viable option. It's difficult to say for certain which would be better without having worked through all the details myself.

src/briefcase/commands/convert.py
freakboy3742: I guess it depends what impact that has on the "merge" part of the process. Preference will always be for a simpler template; but if having a slightly more complicated template makes the merge process significantly easier, then that seems like a reasonable price to pay. However, my initial impression is that it should be possible to put a dummy value in place, and then exclude the briefcase-level description if there's a PEP621 description during the process of merging the metadata.

mhsmith:
> Not at present. It supports the version specifier part, but not the "qualifier" part (including and after the semicolon). Supporting PEP508 environment specifiers is arguably something we should support; however, we need to be careful how those specifiers and the platform-specific sections of the `[tool.briefcase]` configuration interact.

We might never have directly tested the environment markers, but it looks to me like they work fine. Are you aware of any problems with them?

freakboy3742:
> > Not at present. It supports the version specifier part, but not the "qualifier" part (including and after the semicolon). Supporting PEP508 environment specifiers is arguably something we should support; however, we need to be careful how those specifiers and the platform-specific sections of the `[tool.briefcase]` configuration interact.
> 
> We might never have directly tested the environment markers, but it looks to me like they work fine. Are you aware of any problems with them?

Not from experience. I was assuming the syntax wasn't legal as a command line option to pip - but if they *are* legal, then I guess that simplifies things.

MarieRoald:
We finally managed to pick this up again. Quite a few changes since last time (commit 0a93c1f8f8c3b35697ef007b48e90914e657906d).

 * We rebased ontop of main
 * We updated the pyproject merging so fields that fully overlap with what's in the pyproject.toml are removed (we based this on the `briefcase.config.merge_pep621_config` function)
 * We inherit from NewCommand, which simplifies the code substantially
 * We renamed `ConvertCommand.new_app` to `ConvertCommand.convert_app`
 * We reset all changes to `NewCommand` so it matches what's in main exactly.

There are still some things to do, mainly making tests, but we wanted to run the code by you to see what you think of these changes before we spend much time migrating the unit tests to this new code layout.

We also need to reconsider how we handle the source directory and test source directory. We could either update the cookiecutter template and the `NewCommand.build_app_context` to set it without asking (like we had in 0a93c1f8f8c3b35697ef007b48e90914e657906d), or we can modify the output pyproject.toml file when we merge them. Do you have a preference there?

src/briefcase/commands/convert.py
MarieRoald: Agreed, we have now changed the name of the entry point to `convert_app`

src/briefcase/commands/convert.py
MarieRoald: I guess this is irrelevant now that Briefcase also reads PEP621 formatted pyproject.toml files? We updated the merging so overlapping fields are removed if they are equal.

src/briefcase/commands/convert.py
MarieRoald: We opted for subclassing, and it seemed to work nicely.

freakboy3742 (commented):
Hi, and welcome back! 

I've flagged a few things inline, but I think this is generally heading in the right direction. Most of the things I've flagged are, cosmetic changes to match patterns elsewhere in the Briefcase codebase, opportunities for refactoring, or opportunities to streamline the overall conversion workflow.

Regarding the handling of `sources`/`test_sources` - my main requirement is that any inline comments and formatting in the template pyproject.toml should be preserved. That would seem to preclude editing inline, as libraries like tomllib or tomli_w aren't format/comment preserving, and I'm not sure if the weight of carrying another TOML library just to preserve formatting in one very specific use case is worth it. 

The other thing I wanted to flag - there's a couple of places where our lives might be made simpler by improving Briefcase's PEP621 compliance, rather than trying to map PEP621 to Briefcase. For example, as PEP621 *has* a format for LICENSE handling, we should be trying to improve Briefcase's usage of that format, rather than wrangling a PEP621-compliant project into Briefcase's preferred usage. I don't want to derail this PR into a PEP621-compliance rabbit hole (that's what #1235 is for 😄) - but if we're faced with an opportunity to move closer to that compliance, we should take that opportunity.

src/briefcase/commands/convert.py
yngvem: Agreed, skipping unambiguous fields is a good idea. We've tried to remove the prompts for the unambiguous fields now. Do you think we should show a message along the lines of

```raw
We need a name that can serve as a machine-readable Python package name for your application.
This name must be PEP508-compliant - that means the name may only contain letters, numbers, hyphens
and underscores; it can't contain spaces or punctuation, and it can't start with a hyphen or underscore.

Based on your PEP621 formatted pyproject.toml file, we assume it is {app_name}.
```

(and similar for the description)? Or is it fine to just skip to the next field?

Also, the way we delete overlapping fields is with `tomli_w`, but as you mentioned here: https://github.com/beeware/briefcase/pull/1231#pullrequestreview-1830961522, this approach doesn't preserve formatting. Alternatively, since all the fields we remove are single-line fields, we could delete the lines manually by checking e.g. `line.startswith("description=")` etc, but that might be a bit less fault tolerant? What do you think?

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,835 @@
...
+        self,
+        template: Optional[str] = None,
+        template_branch: Optional[str] = None,
+        **options,
+    ):
+        # Confirm host compatibility, and that all required tools are available.
+        # There are no apps, so finalize() will be a no op on app configurations.
+
+        self.finalize()
+        return self.setup_existing_app_for_briefcase(
```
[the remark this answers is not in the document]
yngvem: We made an attempt at structuring the code like this so it should hopefully match the rest of Briefcase better now.

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,835 @@
...
+        self.input.prompt()
+
+        context = self.build_app_context()
+
+        self.logger.info()
+        self.logger.info(
+            f"Generating required files to set up '{context['formal_name']}' with Briefcase"
+        )
+
+        # If a branch wasn't supplied through the --template-branch argument,
```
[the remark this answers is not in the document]
yngvem: Good catch, we updated `NewCommand` slightly to facilitate more code sharing here.

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,835 @@
...
+
+            if not (self.base_path / test_source_dir).is_dir():
+                raise ValueError("The source directory must exist and be a directory.")
+            if (
+                test_entry := self.base_path / test_source_dir / f"{app_name}.py"
+            ).exists():
+                raise ValueError(
+                    f"{test_entry} is reserved for the test entry script,"
+                    " but it already exists."
+                )
```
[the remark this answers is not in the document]
yngvem: We have now added a more informative error message. We could always make it succeed with a warning, but I usually find that a bit confusing when a command succeeds, but the output doesn't work. Do you think the new error message makes it clear enough?

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,835 @@
...
+        """
+        app_name = self.input_app_name()
+        formal_name = self.input_formal_name(app_name)
+        # The class name can be completely derived from the formal name.
+        class_name = make_class_name(formal_name)
+        # The module name can be completely derived from the app name.
+        source_dir = self.input_source_dir(app_name)
+        module_name = Path(source_dir).name
+        test_source_dir = self.input_test_source_dir(app_name)
+        project_name = self.input_text(
```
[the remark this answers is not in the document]
yngvem: We updated `NewCommand` to avoid reproducing the text.

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,835 @@
...
+        """
+        self.merge_or_copy_pyproject(project_dir / "pyproject.toml")
+
+        # Copy license file if not already there
+        license_file = self.pep621_data.get("license", {}).get("file")
+        if license_file is not None and Path(license_file).name != "LICENSE":
+            self.logger.warning(
+                f"License file found in {self.base_path}, but its name is not LICENSE. Renaming the file to LICENSE."
+            )
+            (self.base_path / license_file).rename(self.base_path / "LICENSE")
```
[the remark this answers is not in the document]
yngvem: I agree, is this something we should sort out in this PR or should it be a separate issue?

yngvem:
We have tried to answer most of your questions and added questions to the ones where we were a bit unsure of how to proceed. One more question though, we noticed the `project_overrides` in the `new` command. Should we add that to the `convert` command as well, or is that unnecessary?

src/briefcase/commands/convert.py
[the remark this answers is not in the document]
freakboy3742: Agreed that a warning isn't really appropriate, because there's no sensible "recovery" that can be performed. The new error message is much clearer.

freakboy3742:
> We have tried to answer most of your questions and added questions to the ones where we were a bit unsure of how to proceed. One more question though, we noticed the `project_overrides` in the `new` command. Should we add that to the `convert` command as well, or is that unnecessary?

For consistency, it should probably be included.

src/briefcase/commands/convert.py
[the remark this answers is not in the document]
freakboy3742: I think it's fine to be treated as a separate issue. There's already #1235 tracking this general class of improvements; l'm happy to leave it at that.

src/briefcase/commands/convert.py
freakboy3742: > Agreed, skipping unambiguous fields is a good idea. We've tried to remove the prompts for the unambiguous fields now. Do you think we should show a message along the lines of ... (and similar for the description)? Or is it fine to just skip to the next field?

I think the approach should mirror what happens with project_overrides - we don't need to echo the explanatory text, but we *should* indicate the value that was selected.

> Also, the way we delete overlapping fields is with `tomli_w`, but as you mentioned here: [#1231 (review)](https://github.com/beeware/briefcase/pull/1231#pullrequestreview-1830961522), this approach doesn't preserve formatting. Alternatively, since all the fields we remove are single-line fields, we could delete the lines manually by checking e.g. `line.startswith("description=")` etc, but that might be a bit less fault tolerant? What do you think?

I agree that a pattern match approach will be less fault tolerant, so that's not the right way to go.

There is another option - ignore the problem. Having a duplicated key isn't actually an issue - Briefcase will always use the "most specific" entry; it's mildly undesirable from a "don't repeat yourself" perspective, but given the alternatives, and the specific circumstances (migrating an existing project), maybe that's an acceptable price?

yngvem:
We have just pushed a version with tests where we tried to address all the comments so far (we're still missing docstrings for some tests, but I think the code is ready for a new review now).

You can run `briefcase convert --template http://github.com/yngvem/briefcase-template --template-branch convert-command -Q author="Some Author" -Q unused=override` to test it in action.

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,810 @@
...
+            )
+
+            if url == "Other" and not override_value:
+                url = self.input_text(
+                    intro="\nWrite the url.",
+                    variable="application URL",
+                    default=self.make_project_url("com.example", app_name),
+                    validator=self.validate_url,
+                )
+        elif options and not override_value:
```
freakboy3742, line 322: I don't think there's a need to differentiate between the N=1 and N>1 case - the N=1 case will fall back to a selection with 2 options, one of which is "other", and one of which is the default.

freakboy3742 (changes requested):
This is definitely converging on a final form. A lot of comments inline, but most of them are either wording changes, or opportunities for simplification and consolidation with other features. It definitely feels like we're getting close!

I haven't done a full review of the tests; partially because they'll need some changes after the response to this review, but also because they're currently showing some coverage gaps.

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,810 @@
...
+
+    def validate_source_dir(self, source_dir: str) -> bool:
+        """Determine if the source_dir is valid.
+
+        :param source_dir: The candidate source directory
+        :returns: True. If there are any validation problems, raises ValueError with a
+            diagnostic message.
+        """
+        if not (self.base_path / source_dir).is_dir():
+            raise ValueError("The source directory must exist and be a directory.")
```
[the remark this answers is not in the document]
MarieRoald: We have updated the error message to mention the `__main__.py` file.

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,810 @@
...
+
+    @cached_property
+    def pep621_data(self):
+        """The pyproject["project"] dictionary.
+
+        Empty dict if [project] is not in the pyproject.toml file.
+        """
+        return self.pyproject.get("project", {})
+
+    def create_test_source_dir_validator(self, app_name: str) -> Callable[[str], bool]:
```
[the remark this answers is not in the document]
MarieRoald: Good spot!

src/briefcase/commands/new.py
```diff
@@ -361,6 +363,93 @@ def input_text(self, intro, variable, default, validator=None, override_value=No
...
+            default=default,
+            options=list(zip(options, options)),
+        )
+
+    def show_welcome_prompt(self) -> None:
+        """Show a welcome prompt that describes what this command will do."""
+        self.input.prompt()
+        self.input.prompt("Let's build a new Briefcase app!")
+
+    def get_version_and_template_info(self, template, template_branch):
```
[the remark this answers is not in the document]
MarieRoald: We have tried to address this now

src/briefcase/commands/new.py
```diff
@@ -619,33 +682,43 @@ def _gui_bootstrap_choices(self, bootstraps):
 
         return bootstrap_choices
 
-    def new_app(
+    def safe_generate_template(
```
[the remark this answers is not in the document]
MarieRoald: Good idea! We have tried to address that now

MarieRoald:
We have tried to address all your comments, and while testing our implementation, we also noticed that the module name must be based on the app name, so there is a lot less flexibility in the source directory than we first thought. We updated the `get_source_dir_hint`-method to raise an error if there are no valid source directories based on the provided app name. Does that sound like a good solution?

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,761 @@
...
+
+            if "tool" in pyproject and "briefcase" in pyproject["tool"]:
+                raise BriefcaseCommandError(
+                    f"[tool.briefcase] already in {self.base_path / 'pyproject.toml'}."
+                    " Cannot initialise briefcase for an application that already is packaged with briefcase."
+                )
+
+    def validate_not_empty_project(self) -> None:
+        """Cannot setup new app for a project containing only a child directory with log
+        files or no child directories."""
```
freakboy3742, line 719: I'm not sure that this validation step is required. This validation will pass if there are *any* subdirectories that don't contain logs - so a `venv` or `tmp` folder would be enough to make it pass.

In addition, one of the first things validated in `build_app_context` is the source folder, which will also validate that a directory exists, but will be a more rigorous check, as it validates that a *specific* directory exists. 

The only case I can see that this *will* catch is a truly "bare" project that only contains `pyproject.toml` and `myapp.py` (and maybe a `README` or other stray files) - but that project will get flagged as soon as you try to find the source folder.

On balance, it seems to me that while this validation step isn't technically *incorrect*, it isn't actually adding any useful validation in practice.

Is there some other use case that is a source of concern that this validating?

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,761 @@
...
+        )
+
+        project_dir = tmp_path / context["app_name"]
+        self.migrate_necessary_files(
+            project_dir, context["test_source_dir"], context["module_name"]
+        )
+
+    def validate_pyproject_file(self) -> None:
+        """Cannot setup new app if it already has briefcase settings in pyproject."""
+        if (self.base_path / "pyproject.toml").exists():
```
freakboy3742, line 707: Is there any value in letting this validation pass if there is *no* `pyproject.toml` file? Most of the benefit provided by this wizard is provided by parsing PEP621 configuration data; if you don't have that... why not just make the existence of *a* pyproject.toml file a requirement for running this command? 

If nothing else, it means the project has been at least *superficially* modernised - the *existence* of a `pyproject.toml` file has an impact on how pip operates, so we even if it isn't fully PEP621 compliant, it won't be relying on some archaic legacy setup.py behavior that will be implicitly altered by adding a Briefcase configuration.

tests/commands/convert/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+        super().validate_not_empty_project()
+        self.actions.append(("validate-not-empty-project",))
+
+    def convert_app(self, **kwargs):
+        self.actions.append(("new", kwargs))
+        return full_options({"new_state": "done"}, kwargs)
+
+
+@pytest.fixture
+def tmp_path_generator(tmp_path):
```
freakboy3742, line 58: I'm not sure I follow why this fixture is needed. `tmp_path` *should* be a unique path for each test case, AFAIK

freakboy3742 (changes requested):
A lot of comments here - but we're definitely getting close now! The issues I've picked up are mostly nitpick style things, or minor edge cases and cleanups. 

Regarding `get_source_dir_hint()` - you're correct that the module name is very tightly constrained; I think the approach you've taken is a good compromise (allowing for the edge case that I've highlighted inline). It's better to fail with a clear error than to try and be smart and generate a project that won't work. I can imagine a world where we dig into configurations for common packaging libraries like setuptools to introspect source directory configurations and entry points... but honestly, if a user has gone so far off base that their source directory isn't PEP621 name in the project root or `src`, then I don't hold much hope of their app working out of the box with Briefcase anyway :-)

The last missing piece is documentation. As we're adding a new command, we need a section in the docs (`docs/reference/commands/convert.rst`) to describe how to drive that command. Don't get too hung up on the specific language - a good first draft that covers all the major details is all we really need. Given how similar `convert` is to `new`, you can probably get 80% of the way by duplicating the `new.rst` in that folder and updating as appropriate.

freakboy3742:
Also - don't worry about the ReadTheDocs test failure; that's a known issue that was resolved in #1695. If you merge main, you'll get a fix for that problem.

freakboy3742:
@MarieRoald @yngvem Let me know when you're ready for a review; I can see this is passing CI, but I'm not sure if you've made all the changes you're expecting in this pass.

src/briefcase/commands/convert.py
```diff
@@ -0,0 +1,761 @@
...
+        :returns: The description text for the source dir prompt.
+        """
+        if (self.base_path / f"src/{module_name}").is_dir():
+            default = self.base_path / f"src/{module_name}"
+        elif (self.base_path / module_name).is_dir():
+            default = self.base_path / module_name
+        elif possible_src_dirs := list(self.base_path.glob(f"**/{module_name}/")):
+            default = min(
+                possible_src_dirs, key=lambda p: (len(p.parents), len(str(p)))
+            )
```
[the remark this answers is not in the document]
yngvem: Good idea!

src/briefcase/commands/convert.py
yngvem: We added this because there was this one time when we tried running the wizard in an empty directory twice. The first time, we got an error in the validation step, while we did not the second time because of the log-directory, which we thought were a bit confusing.

tests/commands/convert/conftest.py
yngvem: Some tests (e.g. for merge_or_copy_pyproject) needs two different temporary directories, so we made this generator instead of using `tempfile.TemporaryDirectory`

yngvem:
> A lot of comments here - but we're definitely getting close now! The issues I've picked up are mostly nitpick style things, or minor edge cases and cleanups.
> 
> Regarding `get_source_dir_hint()` - you're correct that the module name is very tightly constrained; I think the approach you've taken is a good compromise (allowing for the edge case that I've highlighted inline). It's better to fail with a clear error than to try and be smart and generate a project that won't work. I can imagine a world where we dig into configurations for common packaging libraries like setuptools to introspect source directory configurations and entry points... but honestly, if a user has gone so far off base that their source directory isn't PEP621 name in the project root or `src`, then I don't hold much hope of their app working out of the box with Briefcase anyway :-)
> 
> The last missing piece is documentation. As we're adding a new command, we need a section in the docs (`docs/reference/commands/convert.rst`) to describe how to drive that command. Don't get too hung up on the specific language - a good first draft that covers all the major details is all we really need. Given how similar `convert` is to `new`, you can probably get 80% of the way by duplicating the `new.rst` in that folder and updating as appropriate.

We have added a draft for the documentation now so it should be ready for a review

src/briefcase/commands/base.py
```diff
@@ -149,6 +151,12 @@ class BaseCommand(ABC):
     # introduced for a platform such that older versions of a template are incompatible
     platform_target_version: str | None = None
 
+    # We specify the default app template as a class attribute so that it can be
+    # overwritten by properties of subclasses. This is necessary since if we had it as
+    # an instance attribute, it would cause an AttributeError since the property would
+    # not have a setter.
+    app_template_url = None
```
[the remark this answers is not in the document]
freakboy3742: I can see that `create` is removing the template details from the context, but I don't think there's a particularly *strong* reason why that is the case. It's more an accident of history - we originally did a little bit of cleanup of the template context; we reverted that for `new` so that we could output the specific template version in the template, but we never made the corresponding change to `create` (and, for some reason, we used "template_source", rather than "template" for the template variable). It can't think of any reason it would hurt to pass in extra template variables - if anything, it would be beneficial, as we could include the template version in the platform templates - so for the sake of simplifying `generate_template`, we might as well include them.

As for `fallback_template` - you're correct there's no existing usage of `a=x or y`; that's because the idiom we use is `a = x if x else y`.

freakboy3742 (commented):
I've made a couple of minor tweaks, and I want to take a final look at it in the morning with some fresh eyes and a couple more practical tests, but I think you've done it! Congratulations on bringing this whale of a PR to completion!

freakboy3742 (approved):
A new morning with strong coffee and fresh eyes revealed a couple more edge cases, and a couple more cleanups for output messages. I've flagged the problems I found inline for posterity, but otherwise, this looks good to go! 

Thanks again for your persistence with this PR - it's a huge piece of work! I hope I'll see you both at PyCon US this year so I can congratulate you in person.

MarieRoald:
What a great message to start the day with! We're coming to PyCon US this year too, so we'll probably see you there 😄

Issue #1155, opened 2023-03-29
Underscore isn't a valid character in bundle names on iOS


freakboy3742:
### Describe the bug

An app name of `hello_world` is accepted as legal; however, if you build an iOS app with this name, you will get an error.

### Steps to reproduce

1. Generate an app with an app name of `hello_world`
2. `briefcase run iOS`
3. During build, you get the warning:
```
warning: invalid character in Bundle Identifier. This string must be a uniform type identifier (UTI) that contains only alphanumeric (A-Z,a-z,0-9), hyphen (-), and period (.) characters. (in target 'Hello World' from project 'Hello World')
```

The app will start successfully; however it's reasonable to assume that the app would be rejected from the App Store.

### Expected behavior

The generated app shouldn't raise a warning - the bundle identifier should be normalised to a valid form.

### Environment

- Operating System: iOS (any)
- Python version: Any
- Software versions:
  - Briefcase: 0.3.13

### Additional context

The problem occurs because `_` isn't a valid character in a domain name. We currently accept both `_` and `-` in an app name (as both are accepted by PEP508), and normalise `-` to `_` for the module name (which makes a valid python module). What we require is the reverse normalization - converting `_` to `-` to make a valid "bundle_name" which is passed to the template context when generating a project. It may be worth adding a full `bundle_identifier` (== `f'{bundle}.{bundle_name}'`) for convenience.

The iOS backend also needs to use the full bundle identifier to launch the app.

This has been observed as a problem on iOS, but it likely affects any other platform using bundles - macOS (app and Xcode), Android, and Linux (System, AppImage and Flatpak). On some of these platforms, the use of the bundle name is cosmetic, but we should use the same bundle naming scheme for consistency.

PR #1234, opened 2023-04-25
Added `bundle_identifier` for consistent bundle/app IDs across platforms

kevin-brown:
This introduces two new dynamic properties to `AppConfig`:
* `bundle_name` which is a normalized version of `app_name` that replaces underscores with dashes
* `bundle_identifier` which is the combination of `bundle` and `bundle_name` to give a fully qualified identifier

The `bundle_identifier` is shown in the following places:
* `repr(AppConfig())` in addition to the `app_name` (example: `<com.example.first-app (first_app) v1.2.3 AppConfig>`)
* `briefcase create` as the `App-ID` line

The `bundle_identifier` has been propagated to the following platforms:
* iOS for the package name
* Windows for the domain used for apps
* Linux for the binary paths and docker image names

It has not been propagated to Android as that platform requires underscores instead of dashes in their application IDs.

This fixes https://github.com/beeware/briefcase/issues/1155.

## PR Checklist:

### DIFF
--- a/src/briefcase/config.py
+++ b/src/briefcase/config.py
@@ -431,6 +431,23 @@ class AppConfig(BaseConfig):
         return self.app_name.replace("-", "_")
 
     @property
+    def bundle_name(self):
+        """The bundle name for the app.
+
+        This is derived from the app name, but:
+        * all `_` have been replaced with `-`.
+        """
+        return self.app_name.replace("_", "-")
+
+    @property
+    def bundle_identifier(self):
+        """The bundle identifier for the app.
+
+        This is derived from the bundle and the bundle name, joined by a `.`.
+        """
+        return f"{self.bundle}.{self.bundle_name}"
+
+    @property
     def class_name(self):
         """The class name for the app.
 
--- a/src/briefcase/config.py
+++ b/src/briefcase/config.py
@@ -419,7 +419,7 @@ class AppConfig(BaseConfig):
             )
 
     def __repr__(self):
-        return f"<{self.bundle}.{self.app_name} v{self.version} AppConfig>"
+        return f"<{self.bundle_identifier} v{self.version} AppConfig>"
 
     @property
     def module_name(self):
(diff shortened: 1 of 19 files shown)


src/briefcase/config.py
```diff
@@ -419,7 +419,7 @@ def __init__(
             )
 
     def __repr__(self):
-        return f"<{self.bundle}.{self.app_name} v{self.version} AppConfig>"
+        return f"<{self.bundle_identifier} ({self.app_name}) v{self.version} AppConfig>"
```
freakboy3742, line 422: Is there a particular reason you've added the app name here? The bundle identifier should be unique enough for diagnostic purposes, which is what the `repr` is being used for.

tests/platforms/iOS/xcode/test_run.py
```diff
@@ -72,7 +72,7 @@ def test_run_multiple_devices_input_disabled(run_command, first_app_config):
         run_command.run_app(first_app_config, test_mode=False, passthrough=[])
 
 
-def test_run_app_simulator_booted(run_command, first_app_config, tmp_path):
+def test_run_app_simulator_booted(run_command, underscore_app_config, tmp_path):
```
freakboy3742, line 75: My assumption is that this has been changed to `underscore_app_config` to piggyback a test of underscore handling onto an existing test. That makes some sense - however, it's not immediately obvious that the test is serving a double purpose here. If a future iteration of the code didn't ned to test booting the simulator (or needed to test it a different way), it would be easy to lose this testing case. 

Even though there will be some duplicated code, I'd suggest making an explicit "run_app_with_underscore" test.

freakboy3742 (changes requested):
Looks good - a couple of fairly minor tweaks inline. 

There's also a need for updates to the templates for apps - the iOS, for example [generates an app bundle ID](https://github.com/beeware/briefcase-iOS-Xcode-template/blob/main/%7B%7B%20cookiecutter.format%20%7D%7D/%7B%7B%20cookiecutter.formal_name%20%7D%7D.xcodeproj/project.pbxproj#L367), which will need to match what is generated here.

src/briefcase/config.py
kevin-brown: Got it, I wasn't sure if this was for diagnostic purpose or for repl cases. The only reason to include it would be to reduce confusion caused by the automatic conversion of the bundle identifier by making it clear that yes, the app name is still used elsewhere.

I'll make the change shortly to remove the `app_name` from the repr.

freakboy3742 (approved):
Awesome - thanks for the fix!

PR #1240, opened 2023-04-25
Added option to update support to update command

### LINKED ISSUE #856: Add ability to refresh support package
The `briefcase update` command updates the user's code in a project, with options to also update dependencies (`-d`), and resources (`-r`). However, there is no ability to update the support package. This need doesn't arise often, but for completeness, it would be desirable to have this option.

A new option `-s`/`--update-support` that updates the support package in an existing app.

The current workaround is to rebuild the app if you need to update the support package. This works, but is a sledgehammer approach.

**Additional context**

The general behavior will be very close to the behavior of the `-d` update dependencies option, so that implementation can be used as a point of comparison.

The code installing the support package will need to delete the existing support folder prior to unpacking the support package. This will need to check if the project *has* a support package - if the template doesn't define a `support_path`, then `update -s` will be a no-op.

melaniearbor:
Adds a flag to `briefcase update` to allow explicit updating of the application support package.

Fixes #856

## PR Checklist:


src/briefcase/commands/update.py
```diff
@@ -53,6 +55,10 @@ def update_app(
             self.logger.info("Updating application resources...", prefix=app.app_name)
             self.install_app_resources(app=app)
 
+        if update_support:
+            self.logger.info("Updating application support...", prefix=app.app_name)
+            self.install_app_support(app=app, update=True)
```
freakboy3742, line 60: Should this be `install_app_support_package`

src/briefcase/commands/create.py
```diff
@@ -302,6 +302,8 @@ def install_app_support_package(self, app: BaseConfig):
         except KeyError:
             self.logger.info("No support package required.")
         else:
+            if update:
```
freakboy3742, line 305: Does this need an `and support_path.exists() and support_path.is_dir()` qualifier? It's not a likely edge case, but it's conceptually possible.

freakboy3742 (approved):
Looks good - a couple of minor issues flagged inline, plus docs and the coverage issue, and we should be good to go!

src/briefcase/commands/create.py
```diff
@@ -292,16 +292,19 @@ def _unpack_support_package(self, support_file_path, support_path):
...
         :param app: The config object for the app
+        :param upsate: The force update the support package
         """
         try:
             support_path = self.support_path(app)
         except KeyError:
             self.logger.info("No support package required.")
         else:
+            if update and support_path.exists() and support_path.is_dir():
+                support_path.rmdir()
```
freakboy3742: `Path.rmdir()` requires that the directory is empty. You can use `shutil.rmtree(path)` instead; that will purge the directory before removing it. 

It's also a good idea to add some logging output so it's clear (to the user, and to the logs) that an action has been performed.

src/briefcase/commands/create.py
```diff
@@ -292,16 +292,19 @@ def _unpack_support_package(self, support_file_path, support_path):
         except (shutil.ReadError, EOFError) as e:
             raise InvalidSupportPackage(support_file_path) from e
 
-    def install_app_support_package(self, app: BaseConfig):
+    def install_app_support_package(self, app: BaseConfig, update: bool = False):
```
freakboy3742: I can see why you've added this boolean flag; however, in practice, I don't think it's needed. If we simplify the logic so that we *always* delete the support package if it always exists, that logic won't ever trigger for a clean build, and will still clean up on an update.

tests/commands/create/test_install_app_support_package.py
```diff
@@ -43,7 +43,50 @@ def test_install_app_support_package(
...
+        )
+    )
+
+    # Mock shutil so we can confirm that unpack is called,
+    # but we still want the side effect of calling it
+    create_command.tools.shutil = mock.MagicMock(spec_set=shutil)
+    create_command.tools.shutil.unpack_archive.side_effect = shutil.unpack_archive
+
+    # Install the support package
+    create_command.install_app_support_package(myapp, update=True)
```
freakboy3742: This test gives us coverage, but doesn't verify that the old support package was successfully deleted. Creating some dummy "old" support files, and then verifying that they've been deleted would have caught the `rmdir()` issue I flagged.

freakboy3742 (approved):
I've made a couple of minor docs updates, a small simplification around the `update` argument, and expanded the test of support cleanup (details inline) - but otherwise, this is really good. Thanks for the contribution!

freakboy3742 (changes requested):
Bother. CI has done it's job, and revealed a major problem: Windows apps Linux AppImages don't have a standalone folder for their support packages. On those targets, the support files are co-mingled with the app itself, so when the support folder is deleted, it deletes the stub app and other necessary components.

I'll add some notes on the original feature request ticket to discuss options.

freakboy3742 (approved):
I've pushed an update that does best-effort cleanup on Windows and AppImage, and warns the user. With that, I think we can call this done!

melaniearbor:
Thanks for the collab and cleanup @freakboy3742! So excited to have helped BeeWare ❤️ 🐝

PR #1243, opened 2023-04-26
Ensure that the run command populates the target_image.

freakboy3742:
Follow up to #1239; #1222 removed the `--target` option from the Run command. However, when run infers a build, the base command needs to have some options describing that docker is not being used.

This has manifested as CI failures in Toga (e.g. [this failed run](https://github.com/beeware/toga/actions/runs/4803945790/jobs/8549026939))

## PR Checklist:

### DIFF
--- a/tests/platforms/linux/system/test_run.py
+++ b/tests/platforms/linux/system/test_run.py
@@ -1,5 +1,7 @@
 import os
 import subprocess
+import sys
+from pathlib import Path
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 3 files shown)


tests/platforms/linux/system/test_run.py
```diff
@@ -44,6 +52,87 @@ def test_unsupported_host_os(run_command, host_os):
...
+                / "first-app-0.0.1"
+                / "usr"
+                / "bin"
+                / "first-app"
+            )
+        ],
+        cwd=tmp_path / "home",
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        bufsize=1,
```
rmartin16, line 124: You can use `**sub_stream_kw` here if you want :)

rmartin16 (approved):
I ran through the linux system workflows with and without Docker on my host linux machine, Fedora 37, and Ubuntu 22.04. They all worked as expected.

PR #1244, opened 2023-04-28
Add support for accurate local coverage reports

rmartin16:
## Changes
- Explicitly test `os.altsep` for local requirements
- Ensure `os.X_OK` is unmocked for Android SDK download test
- Include test case for linux system not updating file perms
- Explicitly test `project_path()`

## PR Checklist:


rmartin16:
@freakboy3742,
Concerning the final piece of coverage for macOS...would you like me to do anything about this? (personally, I only really wanted coverage to work on Linux...but macOS mostly came along for the ride.)

I can easily add `pragma: no cover` or go a little bigger with something like [coverage-conditional-plugin](https://github.com/wemake-services/coverage-conditional-plugin).

freakboy3742 (changes requested):
Conceptually, I'm definitely in favor of the idea of being able to return a "clean" coverage report on a single platform without needing to do a full CI run. However, I'm not sure I'm completely on board with this as an approach. 

If I'm reading this right, the approach you've taken is to essentially find all the non-covered platform branches, and backfill platform mocks to make "non-Windows look like Windows", etc. There's two reasons this makes me nervous.

Firstly, mocks are inevitable for a project like Briefcase (because we're closely bound to tools that have side effects), but we should be trying to minimize the use of mocks where possible (or, at least, be careful about what we mock). Every mock is a potential source of error, because you need to verify that the mocked behavior accurately mirrors "real" behavior. In this case, we're adding tests that mock the behavior of every non-local test condition (so, macOS on Windows, and Windows on macOS, etc). It strikes me as far too easy to introduce a mock that incorrectly shadows "real" behavior, but gives us a test pass with 100% coverage.

Secondly, this approach to testing doesn't have a failure condition. Let's say I add a new block of code to handle macOS code signing. That code will only run on macOS, so I write a platform-gated test. That passes my local macOS testing with 100% coverage; and I push to CI, and it passes there as well, because the combined CI report will merge a macOS and Windows report. Unless we turn on a requirement of 100% code coverage on every *individual* platform, we can't procedurally prevent pushing a change that will break CI on a different platform; and if we *do* that, there's no benefit to having the combined CI report.

FWIW, I took a look at conditional-coverage late last year as a potential solution, and came to the conclusion that it wasn't a good fit, for much the same reason. conditional-coverage means each individual platform returns 100% coverage; but as a result, there's no way to do a combined report that verifies "This project as a whole has 100% coverage over all platforms". Essentially, it's possible to introduce an "if False" condition to a line/branch that won't be covered on any platform, but also won't trigger as code that isn't tested. 

If you can come up with a way to make the general idea work, I'm definitely open to the general idea - I'm just wary of doing something purely for the sake of "number go up".

rmartin16:
> Conceptually, I'm definitely in favor of the idea of being able to return a "clean" coverage report on a single platform without needing to do a full CI run. However, I'm not sure I'm completely on board with this as an approach.
> 
> If I'm reading this right, the approach you've taken is to essentially find all the non-covered platform branches, and backfill platform mocks to make "non-Windows look like Windows", etc. There's two reasons this makes me nervous.

I would not characterize my approach like this; although, it is true I simply found all the uncovered code....and covered it.

Here's a summary of my changes:
- Data directory name for macOS is different
  - New test that mocks the host OS for that branch
  - Existing tests run the same
- Use of `os.altsep` for confirming if a requirement is local
  - New micro test for that specific function that mocks `os.altsep`
- `PYTHONMALLOC` is set for Windows
  - New micro test for that specific function that mocks host OS
- `open` commands for macOS and Linux
  - Mocks host OS to confirm `subprocess` command
- `open` command for Windows
  - New test that mocks `os.startfile()` since it only exists for Windows
  - Wanted to ensure the test finds `os.startfile()` itself on Windows
- Android SDK binary is executable
  - Mocked `os.access()` only for Windows
- Add explicit tests for `command.project_path()`
- Update Linux System file perms test for false negative missed branch
- Re-implement `verify_tools()` tests for platforms so they actually test something now

So, for the most part, the tests were already nearly cross-platform compatible (at least between Linux and macOS). I do not think I compromised the test suite with these changes. Most of the new mocks were to make the `verify_tools()` tests much more useful.

> Firstly, mocks are inevitable for a project like Briefcase (because we're closely bound to tools that have side effects), but we should be trying to minimize the use of mocks where possible (or, at least, be careful about what we mock). Every mock is a potential source of error, because you need to verify that the mocked behavior accurately mirrors "real" behavior. In this case, we're adding tests that mock the behavior of every non-local test condition (so, macOS on Windows, and Windows on macOS, etc). It strikes me as far too easy to introduce a mock that incorrectly shadows "real" behavior, but gives us a test pass with 100% coverage.

I agree; although, I don't think this PR's use of mocking goes any further than our current use. In my mind, the tests should be designed to fully and accurately test the target platform. Ensuring single platform coverage should not require compromising that goal by mocking beyond everything we already mock.

Now, if we wanted to get 100% coverage on Windows, I think this concern would be much more heightened....because _a lot_ more would need to be mocked out to test certain things on Windows.

> Secondly, this approach to testing doesn't have a failure condition. Let's say I add a new block of code to handle macOS code signing. That code will only run on macOS, so I write a platform-gated test. That passes my local macOS testing with 100% coverage; and I push to CI, and it passes there as well, because the combined CI report will merge a macOS and Windows report. Unless we turn on a requirement of 100% code coverage on every _individual_ platform, we can't procedurally prevent pushing a change that will break CI on a different platform; and if we _do_ that, there's no benefit to having the combined CI report.

This was actually something I had in my original design to accommodate. However, I removed it because it seemed like overkill given the small number of affected tests.

In fact, this was already a problem; hence the change to the linux system file perms test.

Nonetheless, it's pretty straightforward to add a custom marker like `pytest.mark.single_platform_coverage` that could be used to disable certain tests. So, we could disable the single platform tests in CI to avoid this type of problem.

> If you can come up with a way to make the general idea work, I'm definitely open to the general idea - I'm just wary of doing something purely for the sake of "number go up".

My motivation here was getting coverage before having to push to GitHub. Commonly, I make a bunch of changes, try to ensure coverage, but can only really find out in CI. Maybe if `coverage.py` could provide a diff of before and after...

At the end of the day, I think these concerns can be managed by ensuring the current quality of the tests is maintained.

With that, do you think this successfully manages your abstract concerns above? And do you have any concrete concerns with the changes? I'm definitely willing to implement the `pytest` marker if that's the barrier. OTOH, I understand if you think this exposes the project to too much nebulous risk from false coverage.

freakboy3742:
> > Conceptually, I'm definitely in favor of the idea of being able to return a "clean" coverage report on a single platform without needing to do a full CI run. However, I'm not sure I'm completely on board with this as an approach.
> > If I'm reading this right, the approach you've taken is to essentially find all the non-covered platform branches, and backfill platform mocks to make "non-Windows look like Windows", etc. There's two reasons this makes me nervous.
> 
> I would not characterize my approach like this; although, it is true I simply found all the uncovered code....and covered it.
...
> So, for the most part, the tests were already nearly cross-platform compatible (at least between Linux and macOS). I do not think I compromised the test suite with these changes. Most of the new mocks were to make the `verify_tools()` tests much more useful.

On a closer examination, it's not as "platform-mocky" as I first thought. 

> Now, if we wanted to get 100% coverage on Windows, I think this concern would be much more heightened....because a lot more would need to be mocked out to test certain things on Windows.

Yeah - and I guess this is where my remaining hesitation comes from. We're adding more tests (ok - not a *lot* of them, but some) - and in practice, it's not *functionally* increasing our coverage - it's increasing the reported coverage in a very specific subset of use cases. In some regards, we're *duplicating* tests - something that would be tested by virtue of running under windows is now being tested under macOS and Linux as well, except that we now need a separate test. 

The "can't enforce in CI" comment from before is also still true - this only reports 100% coverage if you merge *all* Python versions on a given platform. The individual python + platform version tests in CI report lots of CI misses, especially on macOS and Windows. 

> > If you can come up with a way to make the general idea work, I'm definitely open to the general idea - I'm just wary of doing something purely for the sake of "number go up".
> 
> My motivation here was getting coverage before having to push to GitHub. Commonly, I make a bunch of changes, try to ensure coverage, but can only really find out in CI. 

Completely agreed that this is a desirable outcome. FWIW, my approach has been to generate a clean report on main, and then compare with a report on the branch to check there aren't any new lines. Not an ideal approach, but it works (...ish). 

> With that, do you think this successfully manages your abstract concerns above? And do you have any concrete concerns with the changes? I'm definitely willing to implement the `pytest` marker if that's the barrier. OTOH, I understand if you think this exposes the project to too much nebulous risk from false coverage.

I don't have any concrete concerns about the tests, beyond the "test duplication" issue I mentioned before. 

On reflection, I guess it's not so much "false coverage" that is the issue - it's that we go to the effort (and duplication) of fixing things now, but we don't have any real way to ensure things "stay fixed". At least as currently constructed, there's no step we can turn on a 100% coverage requirement - we'd need to add a "platform specific" coverage report (and then only enforce 100% on Linux - maybe macOS if we fix the one missing path there). At that point, we've made the tests more complex, but that complexity hasn't really made our lives easier. Yes, there's a delay in getting merged coverage reports out of CI, but it's not *that* much of a delay, and once you have that report, you've got a shortlist of the lines you've missed in testing - which will appear in a local single python version test run.

rmartin16:
So, it doesn't sound like I'll be able to convince you the trade-offs go the other way :) but that's ok; the gains are certainly not objectively better than the losses...I'm still waffling myself.

Nonetheless, instead of 100% coverage for Linux and macOS, I reoriented the PR just to incorporate the less controversial changes that I think make the tests stronger regardless. Although, I'm least confident about that for the `os.altsep` test. Additionally, I moved the updates to testing `verify_tools()` in to the `Tool` base class PR.

I've found adding this to `[tool.coverage,report]` effectively addresses the goal:
```
exclude_also = [
    '^ {20}app_name = "org.beeware.briefcase"$',
    '^ {20}subprocess.run\($',
    '^ {16}if platform.system\(\) == "Windows":$',
    '^ {8}if self.platform == "windows":$',
    '^ {8}if self.tools.host_os == "Windows":$',
    '^ {8}elif self.tools.host_os == "Darwin":$',
    '^ {12}if self.tools.host_os != "Windows":$',
]
```

> The "can't enforce in CI" comment from before is also still true 

fwiw, this wouldn't be too hard to deal with. If we prepend the platform name to the `.coverage` files before uploading them, then we'd just need to add an additional step in calculating total coverage for calculating platform coverage as well.

freakboy3742 (approved):
Can't argue that this subset of tests is a lot less controversial. I've added two more tweaks for good measure - a `#pragma` exclusion for the `dmgbuild` import failure (since that import failure is incidental, and caught by other platform validation), and a cleanup of a platform message.

As for adding an `exclude_also` definition: Yes, that gets coverage to 100% - but it also masks coverage failures. We don't care about the `if windows` branches on macOS - but we *do* care about them on Windows. What we really need is a *conditional* `exclude_also` - a set of definitions that only applies on each platform. Not sure how we'd manufacture that, though.

rmartin16:
> As for adding an `exclude_also` definition: Yes, that gets coverage to 100% - but it also masks coverage failures.

True true; it's just something I can keep in `git stash` and pop out prior to local testing.

> What we really need is a _conditional_ `exclude_also` - a set of definitions that only applies on each platform. Not sure how we'd manufacture that, though.

I think [coverage-conditional-plugin](https://github.com/wemake-services/coverage-conditional-plugin) would support this...since a cursory review suggest it can operate on arbitrary logic.

freakboy3742:
> I think [coverage-conditional-plugin](https://github.com/wemake-services/coverage-conditional-plugin) would support this...since a cursory review suggest it can operate on arbitrary logic.

Based on my investigations last year, conditional-coverage ends up doing the same problem masking. If you have a conditional coverage setup that ignores a Windows branch on macOS, then when that coverage report is merged, any failure to hit the Windows branch when on Windows is masked.

However, I'm happy to be proven wrong on that one - and if I *am* wrong, then I'd be very happy to see it integrated. On paper, it seems like exactly the capability we need - we just need to work out how to drive it.

Maybe we just need to take a different approach to coverage reporting, and produce merged *per platform* coverage reports, each of which is expected to be clean. So - we introduce conditional coverage, add branch markers so that windows branches aren't included on macOS, but we generate an independent macOS coverage report and Windows coverage report. That makes the coverage report a *little* harder to interpret (since there will be 3 coverage reports), but it shouldn't be *that* hard to explain.

Devolving this even further - we could add in conditional flags for python versions as well, so *every* pytest run has a coverage report that should hit 100%.

rmartin16:
I believe the coverage exclude rule apply at report build time....not at the time coverage evaluation is actually happening. Assuming that's true, my thought was simply to disable any of these platform rules in CI. I am investigate further.

freakboy3742:
> I believe the coverage exclude rules apply at report build time....not at the time coverage evaluation is actually happening. Assuming that's true, my thought was simply to disable any of these platform rules in CI. I can investigate further.

Definitely worth some further investigation - as I've said before, the *goal* of returning 100% coverage locally is definitely worthwhile, if we can make it work.

I've merged this PR on the basis that it makes a bunch of good incremental improvements; a move to conditional coverage will be (mostly) orthogonal to the changes in this PR.

PR #1252, opened 2023-04-30
`pytest` and `pre-commit` cannot run concurrently in `tox`

rmartin16:
- Checks like `isort` can create temporary files in the `src` or `tests` and this can cause `pytest` to fail to load properly.

## PR Checklist:


freakboy3742 (approved):
Makes sense to me; haven't seen the problem in practice, but it's easy to see how it could eventuate - and it doesn't hurt for tox to mirror what's happening in CI.

rmartin16:
FWIW...this was only happening in my virtual machines where disk IO is extra slow. I think that was exacerbating the problem.

PR #1258, opened 2023-04-30
Ensure `tox` uses oldest supported Python for coverage

rmartin16:
https://github.com/nedbat/coveragepy/issues/1572#issuecomment-1522546425

## PR Checklist:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,6 +23,8 @@ commands =
 
 [testenv:coverage{,-html}{,-fail}]
 depends = py{,38,39,310,311,312}
+# coverage should run on oldest supported Python
+base_python = py38,py39,py310,py311
 parallel_show_output = True
 extras = {[testenv]extras}
 # 2023-04-22 see pkgenv ↑
--- a/changes/1258.misc.rst
+++ b/changes/1258.misc.rst
@@ -1,0 +1,1 @@
+The ``tox`` coverage environment will now use the oldest available version of Python.


tox.ini
```diff
@@ -23,6 +23,8 @@ commands =
 
 [testenv:coverage{,-html}{,-fail}]
 depends = py{,38,39,310,311,312}
+# coverage should run on oldest supported Python
+base_python = py38,py39,py310,py311
```
freakboy3742, line 27: Clarifying the interpretation here - The behavior in CI definitely seems to be using the py38 environment; but does this definition mean it will prefer the py38 environment, but if you don't have one, it will use py39, and so on? The tox documentation isn't clear about this (to me, anyway), and it's difficult to test.

freakboy3742 (commented):
Looks good - but one question about the exactly what is being added/changed here.

tox.ini
rmartin16: Yeah....everything about `tox` just feels a little cryptic. Nonetheless, `tox` supports a whole "discovery" algorithm for finding versions of Python; depending on `PATH` is mostly working for me, though.

This setting means use the first available version of Python in this list.

The easiest way I found to test it is to make sure Python 3.8 or Python 3.9 is in `PATH` and start `tox` from a virtual env using a different Python version. You should see `tox` create an environment using 3.8...sometimes adding `-v` to `tox` makes things clearer....or even more confusing...

Finally, CI is was using Python 3.8 even before this because that's what's installed for that CI job and the default choice for `which python`.

freakboy3742 (approved):
Ok - that's enough of a clarification for me; it's working in my (limited) testing, and in CI, so I'm happy to run with that.

PR #1260, opened 2023-05-03
Add conditional coverage flags for accurate platform coverage

rmartin16:
## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,17 +29,17 @@ jobs:
     uses: beeware/.github/.github/workflows/towncrier-run.yml@main
 
   package:
-    name: Python Package
+    name: Python package
     uses: beeware/.github/.github/workflows/python-package-create.yml@main
 
   unit-tests:
     name: Unit tests
-    needs: [pre-commit, towncrier, package]
+    needs: [ pre-commit, towncrier, package ]
     runs-on: ${{ matrix.platform }}-latest
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
-        platform: [ "macos", "ubuntu", "windows" ]
+        platform: [ "macOS", "Ubuntu", "Windows" ]
         python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12-dev" ]
         include:
         - experimental: false
@@ -47,7 +47,7 @@ jobs:
         - python-version: "3.12-dev"
           experimental: true
         # Run tests against the latest Windows Store Python
-        - platform: "windows"
+        - platform: "Windows"
           python-version: "winstore"
           experimental: false
     steps:
@@ -57,7 +57,7 @@ jobs:
         fetch-depth: 0
 
     - name: Set up Python
-      if: matrix.python-version != 'winstore'
+      if: startswith(matrix.python-version, '3')
       uses: actions/setup-python@v4.6.0
       with:
         python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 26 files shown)


rmartin16:
100% coverage on all platforms/versions. Still need to ensure everything is coherent and probably some docs updates.

@freakboy3742, if you have a second, curious if you have thoughts on this failure? maybe just the fact this is a 3.12 alpha?

https://github.com/beeware/briefcase/actions/runs/4889292359/jobs/8727757471

freakboy3742:
> @freakboy3742, if you have a second, curious if you have thoughts on this failure? maybe just the fact this is a 3.12 alpha?

That... makes no sense to me at all. Unless I'm misreading things, the line it's referencing [doesn't even seem like a candidate for execution](https://github.com/beeware/briefcase/blob/80a1cda7543212c7c45d39ff55ebbb16f8cbac06/src/briefcase/platforms/macOS/__init__.py#L572)

.github/workflows/ci.yml
```diff
@@ -57,52 +57,129 @@ jobs:
...
+      if: success() || failure()
       uses: actions/upload-artifact@v3.1.2
       with:
         name: coverage-data
         path: ".coverage.*"
         if-no-files-found: ignore
 
-    - name: Report platform coverage
-      run: tox -e coverage
+  coverage-platform:
```
freakboy3742, line 96: If what we're generating is a "platform and python version" specific coverage report, and it will pass with 100% on every platform/python combination... is there any reason to *not* tag this to the end of the testing pass for a platform and python, rather than doing it as a separate step with it's own matrix fanout?

rmartin16: In addition to wanting to provide a full analysis of coverage before failing, the only way to assess coverage across Python versions for a particular platform is to coalesce the coverage files at the end.

But this must be done using the OS for that platform. So, we end up with at least the 3 "Platform coverage" jobs alongside "Project coverage" after the tests.

At that point, if those 3 additional coverage jobs are necessary, the benefit of performing specific Python version coverage there becomes more palatable to 1) allow their failure to fail all of CI but not fail testing and 2) provide better consolidation of the coverage analysis in CI.

P.S.
> But [platform-specific coverage] must be done using the OS for that platform

This is necessary in order for the new conditional coverage rules to work. They operate on `sys.platform` to know when to require/ignore certain code blocks. We could consider spoofing this, though, to potentially allow coverage analysis of a platform while not on that platform. I can at least experiment with that.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -167,12 +165,12 @@ def run_app(
             raise BriefcaseCommandError(f"Unable to start app {app.app_name}.")
         finally:
             # Ensure the App also terminates when exiting
-            if app_pid:
+            if app_pid:  # pragma: no-cover-if-is-py310
                 with suppress(ProcessLookupError):
                     self.tools.os.kill(app_pid, SIGTERM)
```
rmartin16, line 170: I...don't know what's happening on 3.10 here...

tests/commands/dev/test_get_environment.py
```diff
@@ -1,12 +1,29 @@
...
+from unittest.mock import PropertyMock
 
 import pytest
 
 PYTHONPATH = "PYTHONPATH"
 PYTHONMALLOC = "PYTHONMALLOC"
 
 
+@pytest.mark.parametrize("platform", ["windows", "darwin", "linux", "wonky"])
+def test_pythonmalloc_only_set_for_windows(
```
rmartin16, line 12: This is to get branch coverage on any individual platform.

tests/integrations/android_sdk/AndroidSDK/test_properties.py
```diff
@@ -13,6 +13,7 @@
         ("Windows", "win"),
         ("darwin", "mac"),
         ("Darwin", "mac"),
+        ("Wonky", "wonky"),
```
rmartin16, line 16: more branch coverage for an individual platform.

.github/workflows/ci.yml
rmartin16: This turned out to be a fruitful experiment; I was able to incorporate python version-specific and platform-specific coverage reporting in-line with the existing job matrices.

Although, this has two larger status-quo changes:
- Unit tests no longer fast fail
  - This is to prevent coverage reporting from failing a job and thus the whole matrix.
  - I think ideally developers will have access to the full coverage information.
- The Verify App Build tests will not run now until unit tests and python version-specific coverage is passing.
  - Since failures for unit tests and python-specific coverage are represented by a single flag, either failure prevents app builds.
  - I haven't found an existing flag to avoid this....of course, we could always create one, I guess.

- #1262

rmartin16:
Closing in favor of #1262

PR #1262, opened 2023-05-05
Conditional Coverage Reporting with Platform Spoofing

rmartin16:
## Changes
- Implements [conditional-coverage-plugin](https://github.com/wemake-services/coverage-conditional-plugin) to allow for platform-aware coverage reporting.
- This PR is an alternative to #1260.
  - Python version-specific coverage reporting is performed in the `unit-tests` job
  - Platform-specific coverage is spoofed and performed in the `coverage` job
- Insofar as the spoofing is concerned, it allows `tox` to produce reports from `coverage` files created on other platforms....it does not allow you to create `coverage` files for other platforms.
- New `tox` commands
  - Labels:
    - `tox p -m test`: only runs tests for all available platforms
    - `tox p -m test-fast`: only runs tests with multi-process for all available platforms
    - `tox p -m test-platform`: runs tests for all available platforms and produces coverage report
    - `tox p -m ci`: intended to most closely mimic GitHub CI
  - A naked `tox` (or `tox p`) helps cover the most likely default case where host env only has a single Python in `PATH`
  - To run tests and coverage for a specific version of Python, use `tox -e py310,coverage310`
  - My go to has become `tox p -qmci`

## Notes
- I didn't expect to like this PR better because I was concerned it would be difficult to allow a Python version-specific coverage report to fail while still capturing the rest of coverage analysis.
- To allow capturing all coverage analysis, CI will no longer fast fail if `pytest` fails....but I don't think that's too bad because the tests run pretty quickly and all jobs should finish or also error quickly.
- Additionally, the Build App jobs will not run until at least Python version-specific coverage passes.
  - It may make sense to just update the Build Apps job to depend on the `coverage` job instead of `unit-tests`.
- Example run with missing Python version-specific coverage: 
  - https://github.com/beeware/briefcase/actions/runs/4897506370
- Example run with Python versions-specific coverage but missing platform-specific coverage:
   - https://github.com/beeware/briefcase/actions/runs/4898497859/jobs/8747658142

## PR Checklist:


src/briefcase/platforms/macOS/__init__.py
```diff
@@ -167,12 +165,12 @@ def run_app(
             raise BriefcaseCommandError(f"Unable to start app {app.app_name}.")
         finally:
             # Ensure the App also terminates when exiting
-            if app_pid:
+            if app_pid:  # pragma: no-cover-if-is-py310
                 with suppress(ProcessLookupError):
                     self.tools.os.kill(app_pid, SIGTERM)
```
rmartin16, line 171: I don't know why this is reported uncovered only on 3.10...

rmartin16:
@freakboy3742, as long as you don't think I went _too_ overboard with the `tox` commands, I'll' update the documentation to include (at least some of) them as well as explain that if multiple Python versions are available via `PATH`, then `tox` will use them....and some of the esoteric nature of `coverage` needing to run with the oldest version of Python used to produce the files.

tox.ini
```diff
@@ -21,26 +25,51 @@ use_develop = fast: True
...
+    # disable coverage exclusions for host platform to test entire project
+    project: COVERAGE_EXCLUDE_PLATFORM=disable
+commands_pre =
+    python --version
+    python -c 'if 1: \
+        import os; \
+        print("COVERAGE_FILE", os.environ.get("COVERAGE_FILE", "<not set>")); \
+        print("COVERAGE_PLATFORM", os.environ.get("COVERAGE_PLATFORM", "<not set>")); \
+        print("COVERAGE_EXCLUDE_PYTHON_VERSION", os.environ.get("COVERAGE_EXCLUDE_PYTHON_VERSION", "<not set>")); \
+        print("COVERAGE_EXCLUDE_PLATFORM", os.environ.get("COVERAGE_EXCLUDE_PLATFORM", "<not set>"))'
```
freakboy3742, line 68: I get that these may be useful for debugging purposes, but do they serve a purpose in the long term?

tests/commands/dev/test_get_environment.py
```diff
@@ -1,12 +1,29 @@
...
+    platform,
+    monkeypatch,
+):
+    """PYTHONMALLOC env var is only set on Windows."""
+    monkeypatch.setattr(
+        type(dev_command), "platform", PropertyMock(return_value=platform)
+    )
+    env = dev_command.get_environment(first_app, test_mode=False)
+    expected = "default" if platform == "windows" else "missing"
+    assert env.get(PYTHONMALLOC, "missing") == expected
```
freakboy3742, line 24: I'm not 100% sold on the approach here. There's effectively 2 options:

1. Mark the branch as no-cover, and only test it on Windows
2. monkeypatch platform and test the branch on every platform.

My inclination is that it's better to do the former, as our mock of a platform is more likely to be in error than coverage's identification of the platform.

tests/platforms/iOS/xcode/test_mixin.py
```diff
@@ -37,3 +40,39 @@ def test_distribution_path(create_command, first_app_config, tmp_path):
         match=r"WARNING: No distributable artefact has been generated",
     ):
         create_command.distribution_path(first_app_config)
+
+
+def test_verify(create_command, monkeypatch):
+    """If you're on macOS, you can verify tools."""
+    create_command.tools.host_os = "Darwin"
```
freakboy3742, line 47: What coverage is this adding? Is it purely to evaluate "verify macOS works" on Windows/Linux?

pyproject.toml
```diff
@@ -26,6 +27,37 @@ exclude_lines = [
     "if TYPE_CHECKING:",
 ]
 
+[tool.coverage.coverage_conditional_plugin.rules]
```
freakboy3742, line 30: How many of these are we actually using? Given that this is going to be a bit noisy, my inclination is only to include the one's we're actually using. There's a couple of very specific Python versions that are interesting - mostly for "does importlib have the API we need" reasons, rather than for Python version reasons.

freakboy3742 (changes requested):
I think the additional tox targets all make sense; no objections there. My bigger concerns are around the testing and CI configs:

1. The approach of mocking the platform for a test, rather than `#pragma` excluding a test. As I've indicated inline, I think it's more likely that we'll mess up a mock than conditional-coverage will mess up a branch coverage; on that basis, I'd rather see a platform run *less* tests, but get coverage in the aggregate.
2. I'm not sure I understand the coverage reporting approach. If each individual platform + python version that we test has 100% coverage, any "per platform", "per python version", or "overall" coverage report doesn't really serve a purpose, does it? 
 
There's only 2 benefits I can see. 

Firstly, there's a minor improvement in developer ergonomics - if I add a new branch that isn't python version or platform specific, and miss coverage, then I'll then get 10+ failing coverage reports in CI rather than 1. That's bad... but then, if I can run tox on 1 platform and get the same report.

Secondly, there's a possible edge case where we have an "if False" conditional coverage equivalent. In that case, a unified report *might* report the dead code. However, I'm not sure how we'd get this in practice, outside of an error in the conditional specifier.

However, both of these cases seem pretty niche, especially given the complications that the CI configuration gains in order to support various merged reports. 

Am I missing something here? What's the benefit of this approach over having a simple coverage report at the end of each test run, which we require to be a 100%?

tox.ini
rmartin16: Definitely of debatable usefulness now. Because there really isn't any way to introspect what `condition-coverage-plugin` is doing, these were extremely helpful in getting everything working. They may prove useful in the future if things go wrong but we can just add them back if the information would be useful then.

tests/commands/dev/test_get_environment.py
rmartin16: Unless I'm not following entirely, I did implement option 2. I switched to option 1, though. I had to use `# pragma: no cover` since no platform will take both branches.

tests/platforms/iOS/xcode/test_mixin.py
rmartin16: I wouldn't consider this one to be adding coverage per se; although, it does add coverage on Windows and Linux.

In my mind, this adds "proper" testing and coverage for iOS verification since it explicitly ensures the expected tools are verified. I think that all of the platforms with tools should have this test.

Furthermore, I've added all of these tests in #1093.

rmartin16:
> I think the additional tox targets all make sense; no objections there.

👍🏼 

> My bigger concerns are around the testing and CI configs:
> 
> 1. The approach of mocking the platform for a test, rather than `#pragma` excluding a test. As I've indicated inline, I think it's more likely that we'll mess up a mock than conditional-coverage will mess up a branch coverage; on that basis, I'd rather see a platform run _less_ tests, but get coverage in the aggregate.

I did try to add `# pragma` everywhere without coverage on a platform; I made an exception for `PYTHONMALLOC`  because it requires a naked `# pragma: no cover`.

> 2. I'm not sure I understand the coverage reporting approach. If each individual platform + python version that we test has 100% coverage, any "per platform", "per python version", or "overall" coverage report doesn't really serve a purpose, does it?
> 
> There's only 2 benefits I can see.
> 
> Firstly, there's a minor improvement in developer ergonomics - if I add a new branch that isn't python version or platform specific, and miss coverage, then I'll then get 10+ failing coverage reports in CI rather than 1. That's bad... but then, if I can run tox on 1 platform and get the same report.
> 
> Secondly, there's a possible edge case where we have an "if False" conditional coverage equivalent. In that case, a unified report _might_ report the dead code. However, I'm not sure how we'd get this in practice, outside of an error in the conditional specifier.
> 
> However, both of these cases seem pretty niche, especially given the complications that the CI configuration gains in order to support various merged reports.
> 
> Am I missing something here? What's the benefit of this approach over having a simple coverage report at the end of each test run, which we require to be a 100%?

I don't think you're missing anything necessarily. Basically, if we don't run platform or project coverage with all the conditional coverage rules disabled, we won't have any real verification that everything is covered. I'm not sure how likely or easily someone could create the situation where platform + python version coverage passes but project coverage would not., though; i did create mock ups where such conditions would arise but they were _highly_ contrived.

As for the complications to the CI config, I think some of those should have already existed; for instance, checking if a specific step failed instead of just `failure()`. Managing the coverage files was already happening; now, we just have to control the naming. And finally, each of the platform coverage reports need to be produced appropriately. In aggregate, though, there's certainly much more going on now.

I'd say the most unnecessary thing is uploading all of these HTML reports; I'd definitely be fine with just uploading the final project coverage report. After all, if you download the coverage zip file, one could create any one of these HTML reports themselves.

At the end of the day, I think the platform and project coverage reporting is a good thing to do to verify any conditional coverage rules usage. However, I don't think it's absolutely critical.

tests/platforms/iOS/xcode/test_mixin.py
freakboy3742: I definitely agree that the rest of the test is a much more comprehensive test of the core iOS verification. I guess my question is whether we gain anything by running this test on Windows/Linux. However, it's only verifying that the right workflow is executed (i.e., the right verify methods are invoked), so I guess it doesn't hurt.

tests/integrations/android_sdk/AndroidSDK/test_verify.py
```diff
@@ -210,11 +225,20 @@ def test_download_sdk(mock_tools, tmp_path):
...
+            mock_tools.os.access.assert_called_with(
+                cmdline_tools_base_path / "latest" / "bin" / "sdkmanager",
+                os.X_OK,
+            )
+        else:
+            # When the actual host is non-Windows, ensure the unpacked binary is executable
+            assert os.access(
+                cmdline_tools_base_path / "latest" / "bin" / "sdkmanager",
+                os.X_OK,
+            )
```
freakboy3742, line 241: This is the kind of test I was referring to in my comments about mocking vs platform coverage. We're explicitly testing the behaviour of Linux on Windows, and we need to insert a platform mock of `os.access` to make that work. I'm not sure that's actually gaining us anything.

freakboy3742:
> > My bigger concerns are around the testing and CI configs:
> > 
> > 1. The approach of mocking the platform for a test, rather than `#pragma` excluding a test. As I've indicated inline, I think it's more likely that we'll mess up a mock than conditional-coverage will mess up a branch coverage; on that basis, I'd rather see a platform run _less_ tests, but get coverage in the aggregate.
> 
> I did try to add `# pragma` everywhere without coverage on a platform; I made an exception for `PYTHONMALLOC` because it requires a naked `# pragma: no cover`.

On another pass through, I think the only one that stands out for me is Android SDK download; comment inline.

Agreed the `PYTHONMALLOC` test is annoying; however, it's a 1 line block of logic we can verify in other ways, so I don't think I'm too concerned.
 
> > Am I missing something here? What's the benefit of this approach over having a simple coverage report at the end of each test run, which we require to be a 100%?
> 
> I don't think you're missing anything necessarily. Basically, if we don't run platform or project coverage with all the conditional coverage rules disabled, we won't have any real verification that everything is covered. I'm not sure how likely or easily someone could create the situation where platform + python version coverage passes but project coverage would not., though; i did create mock ups where such conditions would arise but they were _highly_ contrived.

I agree that an individual platform+python coverage report is needed, and the overall project coverage report (with conditional coverage disabled) is needed; but I'm not sure I see what the individual Linux/macOS/Windows coverage reports give us. These are the biggest new sections of CI config, and I'm not sure what we gain from them. 

I'm wondering if it might also allow us to simplify the `COVERAGE_*` variables... but I'm not sure if we need to preserve those so that `tox p` works for a local "all python versions" test run. It's worth noting that run will only ever pass if the developer has *all* Python versions (including a Py3.12 pre-release) installed, which is going to be uncommon for contributors outside of the core team.

> I'd say the most unnecessary thing is uploading all of these HTML reports; I'd definitely be fine with just uploading the final project coverage report. After all, if you download the coverage zip file, one could create any one of these HTML reports themselves.

Agreed. If the HTML was being hosted somewhere, that might be different - but getting and viewing the HTML report involves enough messing around that I doubt it's actually benefitting anyone.

tests/integrations/android_sdk/AndroidSDK/test_verify.py
rmartin16: ah, yeah; I think this is another one that'll require `# pragma: no cover`. I'll verify to be sure.

rmartin16:
> > > Am I missing something here? What's the benefit of this approach over having a simple coverage report at the end of each test run, which we require to be a 100%?
> > 
> > 
> > I don't think you're missing anything necessarily. Basically, if we don't run platform or project coverage with all the conditional coverage rules disabled, we won't have any real verification that everything is covered. I'm not sure how likely or easily someone could create the situation where platform + python version coverage passes but project coverage would not., though; i did create mock ups where such conditions would arise but they were _highly_ contrived.
> 
> I agree that an individual platform+python coverage report is needed, and the overall project coverage report (with conditional coverage disabled) is needed; but I'm not sure I see what the individual Linux/macOS/Windows coverage reports give us. These are the biggest new sections of CI config, and I'm not sure what we gain from them.

When the platform+python version checks run, both platform and python version conditional coverage rules are enabled; however, when whole platform coverage is run, the platform conditional rules are enabled while the python version rules are disabled.

So, strictly speaking, omitting the platform coverage checks exposes the risk that coverage may be incomplete for one platform but complete for another; this is something that only running project coverage would mask.

I don't think this is a deal breaker per se to remove platform coverage checks; it was kinda hard to even contrive some example code to do it. And as long as we're not too fancy with using the conditional coverage pragma, the likelihood should approach zero.

> I'm wondering if it might also allow us to simplify the `COVERAGE_*` variables... but I'm not sure if we need to preserve those so that `tox p` works for a local "all python versions" test run.

If platform coverage checks are omitted in CI, it could obviate the need for using `COVERAGE_FILE`. However, the other `COVERAGE_` env vars to disable platform or python version checks would remain necessary.

> It's worth noting that run will only ever pass if the developer has _all_ Python versions (including a Py3.12 pre-release) installed, which is going to be uncommon for contributors outside of the core team.

Well, _technically_, you can get away without having a version of Python 3.12 installed since coverage is identical for 3.12 and 3.11. (However, if I submit a PR to fix the `pytest` warnings I'm currently seeing with 3.12, that'll change).
 
> > I'd say the most unnecessary thing is uploading all of these HTML reports; I'd definitely be fine with just uploading the final project coverage report. After all, if you download the coverage zip file, one could create any one of these HTML reports themselves.
> 
> Agreed. If the HTML was being hosted somewhere, that might be different - but getting and viewing the HTML report involves enough messing around that I doubt it's actually benefitting anyone.

Ok. I'll probably leave uploading the project coverage report; I've actually used that one before.

With this, I think I'll leave it to you for whether platform coverage checks should run in CI. Personally, I'm leaning towards keeping them in CI...especially since the complication of HTML reports will be removed....but this isn't a string conviction at all.

freakboy3742:
> Ok. I'll probably leave uploading the project coverage report; I've actually used that one before.

That works for me.
 
> With this, I think I'll leave it to you for whether platform coverage checks should run in CI. Personally, I'm leaning towards keeping them in CI...especially since the complication of HTML reports will be removed....but this isn't a string conviction at all.

Missing platform coverage is definitely an edge case, but the *does* edge exists; since the CI environment is already configured and just needs another (relatively quick) command to generate and verify coverage, we might as well keep the platform reports. We definitely don't need the per-platform HTML reports, though.

So - that leaves:
1. That last windows mocking test
2. Removing the HTML reports for all but the overall coverage run
3. Contributor docs explaining the new targets

And this should be good to land.

src/briefcase/commands/dev.py
```diff
@@ -147,8 +147,8 @@ def get_environment(self, app, test_mode: bool):
         # On Windows, we need to disable the debug allocator because it
         # conflicts with Python.net. See
         # https://github.com/pythonnet/pythonnet/issues/1977 for details.
-        if self.platform == "windows":
-            env["PYTHONMALLOC"] = "default"
+        if self.platform == "windows":  # pragma: no branch
```
rmartin16, line 150: turns out `pragma: no branch` is a real thing

src/briefcase/integrations/android_sdk.py
```diff
@@ -307,12 +307,14 @@ def install(self):
...
 
+            # fmt: off
             # Python zip unpacking ignores permission metadata.
             # On non-Windows, we manually fix permissions.
-            if self.tools.host_os != "Windows":
+            if self.tools.host_os != "Windows":  # pragma: no branch no-cover-if-is-windows
                 for binpath in (self.cmdline_tools_path / "bin").glob("*"):
                     if not self.tools.os.access(binpath, self.tools.os.X_OK):
                         binpath.chmod(0o755)
+            # fmt: on
```
rmartin16, line 318: Black has `# fmt: skip`....but it only works if its the only comment on the line...

https://github.com/psf/black/issues/3450

tox.ini
```diff
@@ -18,41 +23,57 @@ commands = pre-commit run --all-files --show-diff-on-failure --color=always
...
+    platform-macos: COVERAGE_PLATFORM = darwin
+    platform-windows: COVERAGE_PLATFORM = win32
+    # use the coverage files created in CI
+    ci-platform-linux: COVERAGE_FILE = .coverage.ubuntu
+    ci-platform-macos: COVERAGE_FILE = .coverage.macos
+    ci-platform-windows: COVERAGE_FILE = .coverage.windows
+    # disable conditional coverage exclusions for Python version to test entire platform
+    {platform,project}: COVERAGE_EXCLUDE_PYTHON_VERSION=disable
+    # disable conditional coverage exclusions for host platform to test entire project
+    project: COVERAGE_EXCLUDE_PLATFORM=disable
```
rmartin16, line 64: I keep forgetting that `tox`'s `factor`s are intended to control how the environment runs....so, I converted the `COVERAGE_*` env vars to `factor`s. I'm not sure if I've sold myself on the `ci` factor yet....maybe there's a better option...

tox.ini
```diff
@@ -18,41 +23,57 @@ commands = pre-commit run --all-files --show-diff-on-failure --color=always
...
 passenv = LOCALAPPDATA
+setenv = COVERAGE_FILE = {env:COVERAGE_FILE:.coverage}
 extras = dev
 # 2023-04-22 see pkgenv ↑
 download = {[pkgenv]download}
 commands =
-    !fast : python -m coverage run -m pytest {posargs:-vv}
-    fast : python -m pytest {posargs:-vv -n auto}
+    !fast : python -m coverage run -m pytest {posargs:-vv --color yes}
+    fast : python -m pytest {posargs:-vv --color yes -n auto}
```
rmartin16, line 34: funny story if you miss it in my commit...

Upon realizing that `pytest` wasn't running in color after adding `FORCE_COLOR`, it dawned on me that `tox` was masking the env var for `pytest`. However, when I passed `FORCE_COLOR=1` to `pytest`, a lot of the tests failed. This is because we're using `capsys` to capture the output....which becomes full of ANSI control chars.

This turns out to have been a problem before `FORCE_COLOR` was added to CI....but fyi to keep in mind.

tox.ini
```diff
@@ -18,41 +23,57 @@ commands = pre-commit run --all-files --show-diff-on-failure --color=always
 [testenv:py{,38,39,310,311,312}{,-fast}]
 depends: pre-commit
 use_develop = fast: True
-skip_sdist = fast: True
```
rmartin16, line 21: Originally, I thought that `skip_sdist` would help because Briefcase was being installed in editable mode....but `tox` actually builds an "editable package" and installs that. So, `skip_sdist` wasn't doing anything....but building and installing the "editable package" is much faster.

src/briefcase/integrations/android_sdk.py
freakboy3742: Clarifying - this is because the `# pragma` hits line length limits, so black tries to reformat, breaking coverage?

rmartin16: yeah....i really wasn't a fan...

```python
            # Python zip unpacking ignores permission metadata.
            # On non-Windows, we manually fix permissions.
            if (
                self.tools.host_os != "Windows"
            ):  # pragma: no branch no-cover-if-is-windows
                for binpath in (self.cmdline_tools_path / "bin").glob("*"):
                    if not self.tools.os.access(binpath, self.tools.os.X_OK):
                        binpath.chmod(0o755)

```

rmartin16: > breaking coverage?

To saw this part....let me check; I just really didn't like the re-format.

rmartin16: This is what's necessary to get coverage on linux:

```python
            # Python zip unpacking ignores permission metadata.
            # On non-Windows, we manually fix permissions.
            if (  # pragma: no branch no-cover-if-is-windows
                self.tools.host_os != "Windows"
            ):
                for binpath in (self.cmdline_tools_path / "bin").glob("*"):
                    if not self.tools.os.access(binpath, self.tools.os.X_OK):
                        binpath.chmod(0o755)
```

rmartin16:
I'm going to finish up the docs tomorrow and do one more review with fresh eyes of the interplay between CI and `tox`. Free feel to let me know your thoughts on some of the consolidation in the meantime.

Also, the cert for https://docs.appimage.org expired....so, docs builds are failing.

freakboy3742:
> I'm going to finish up the docs tomorrow and do one more review with fresh eyes of the interplay between CI and `tox`. Free feel to let me know your thoughts on some of the consolidation in the meantime.

I'll do a quick pass to see if anything stands out.
 
> Also, the cert for https://docs.appimage.org expired....so, docs builds are failing.

🙄

src/briefcase/integrations/android_sdk.py
freakboy3742: Ugh. All bad options, IMHO. I have a mild "we chose black, so we should live with it's sub-optimal choices" preference, but not enough to actually block a merge if you disagree.

freakboy3742 (commented):
Modulo documentation, this is all looking solid. Pretty sure you've earned your Eagle Scout Merit Badge for Tox wrangling :-)

src/briefcase/integrations/android_sdk.py
rmartin16: for posterity....getting coverage on all platforms is even worse....

```python
            # Python zip unpacking ignores permission metadata.
            # On non-Windows, we manually fix permissions.
            if (  # pragma: no branch
                self.tools.host_os != "Windows"
            ):  # pragma: no-cover-if-is-windows
                for binpath in (self.cmdline_tools_path / "bin").glob("*"):
                    if not self.tools.os.access(binpath, self.tools.os.X_OK):
                        binpath.chmod(0o755)
```

docs/how-to/contribute-code.rst
```diff
@@ -17,6 +17,9 @@ The recommended way of setting up your development environment for Briefcase is
 to use a `virtual environment <https://docs.python.org/3/library/venv.html>`__,
 and then install the development version of Briefcase and its dependencies:
 
+Clone Briefcase and create virtual environment
```
rmartin16, line 20: Along with adding documentation for conditional coverage, I also wanted to make these docs easier to skim. So, I added a lot more headers so you can easily scroll down the page and understand what's included...on top of the TOC on the right.

Additionally, I did play around with adding a tl;dr section....but something was telling me you probably wouldn't like it....

image

Nonetheless, the motivation for such a section stems from the seemingly not insignificant number of contributors that seemingly don't do most of what's on this page...

tox.ini
```diff
@@ -83,8 +104,8 @@ commands =
     all  : python -m sphinx {[docs]sphinx_args_extra} -b html . {[docs]build_dir}/html
 
 [testenv:package]
-package_env = none
```
rmartin16, line 86: This [workaround](https://github.com/beeware/briefcase/pull/995/files#r1042849461) doesn't seem necessary [anymore](https://github.com/tox-dev/tox/pull/2800).

rmartin16:
> Pretty sure you've earned your Eagle Scout Merit Badge for Tox wrangling :-)

ha, this was definitely a journey. I think I've finally made it to the end, though. Now I can finish the `Tool` base class and maybe get rootless docker working.

docs/how-to/contribute-code.rst
```diff
@@ -354,18 +458,101 @@ After running the test suite, generate a coverage report by running:
 
       (venv) C:\...>tox -e coverage
 
-You can run the test suite and coverage together by including the testing
-environment to run, e.g. ``py,coverage`` or ``py310,coverage``.
+To run the test suite along with this coverage reporting, run:
```
freakboy3742, line 461: This seems a little misleading - `tox p` also runs the test suite and give coverage. The difference is that it *doesn't* do towncrier, docs and pre-commit checks. Feels like there's an "only" or "just" missing here.

docs/how-to/contribute-code.rst
```diff
@@ -354,18 +458,101 @@ After running the test suite, generate a coverage report by running:
...
+  .. group-tab:: Windows
+
+    .. code-block:: doscon
+
+      (venv) C:\...>tox p -m test
+
+Coverage report for host platform
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If all supported versions of Python are available to tox, then coverage for the
```
freakboy3742, line 486: The distinction between `-e coverage` and `-e coverage-platform` isn't clear to me based on this description - if `coverage` fills in the gaps for missing platforms, what is `-platform` for? Also - there's no mention of Python version specific variants. Functionally - how do I get 100% coverage for `tox -e py`?

freakboy3742 (changes requested):
I've made a couple of cosmetic tweaks; but I've also got some follow up questions about the usage of `-coverage` options.

docs/how-to/contribute-code.rst
```diff
@@ -223,35 +235,70 @@ and re-commit the change.
...
     .. code-block:: doscon
 
-      (venv) C:\...>tox p
+      (venv) C:\...>tox
+
+The full test suite can take a while to run. You can speed it up considerably by
+running tox in parallel, by running ``tox p`` (or ``tox run-parallel``). When
+you run the test suite in parallel, you'll get less feedback on the progress of
+the test suite as it runs, but you'll still get a summary of any problems found
+at the end of the test run.
```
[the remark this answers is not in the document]
rmartin16: > parallel mode seems to have some interesting failure modes if the test suite doesn't pass - I've seen a bunch of weird "can't clean up temp folder" errors, but only when the test fails.

I saw this too....but only on my macOS VM; wasn't sure if it was virtualization or something else going on. Good to have two datapoints, though.

docs/how-to/contribute-code.rst
rmartin16: I'm super steeped in this configuration at this point; so, this feedback helps me understand how it's perceived by others. I'll need to take this in to consideration to convey what's happening with some updates.

To answer the question for the time being, though, running `tox -e coverage` leaves both "python version exclusions" as well as "platform exclusions" enabled; running `tox -e coverage-platform` disables "python version exclusions" while keeping "platform exclusions" enabled.

Basically, `tox -e coverage` asks if the current platform and Python version is covered; `tox -e coverage-platform` asks if the current platform is covered for all Python versions.

docs/how-to/contribute-code.rst
rmartin16: Oh...this was only to show how you can run tests _and_ get coverage reporting.....as opposed to just running `tox -e coverage` to get the report.

But, tbh, the use-case to run _just_ `tox -e coverage` is a little niche....i mean, it just creates a coverage report from existing coverage files. The most likely use-case for getting coverage is to run the tests over code you _just_ changed and get the coverage report.

So, it might be better to first introduce methods to run tests and get the coverage report....and in note at the end mention that `tox -e coverage` can re-produce a coverage report.

docs/how-to/contribute-code.rst
freakboy3742: > Basically, `tox -e coverage` asks if the current platform and Python version is covered; `tox -e coverage-platform` asks if the current platform is covered for all Python versions.

So... either I'm misunderstanding your terminology, or... no it doesn't? 

`tox -e py && tox -e coverage` runs a single python version on the current platform... and then fails coverage. `

`tox -e py38,py39,py310,py311 && tox -e coverage` passes coverage when run on a single platform.

What am I missing here?

rmartin16: The difference is which conditional coverage rules are in effect when the coverage report is produced.

> `tox -e py && tox -e coverage` runs a single python version on the current platform... and then fails coverage.

This runs with both "platform exclusions" and "Python version exclusions" enabled; so, if the host platform and python version are fully covered, then this succeeds. (However, this has a caveat; `tox -e coverage` will look for the oldest version of Python installed; so, technically, `tox -e py` could run for the version of Python in your virtual environment, but `tox -e coverage` may find an older version of Python that exists on the system. So, safest way to run `tox -e py,coverage` is with an explicit Python version, i.e. `tox -e py311,coverage311`)

> `tox -e py38,py39,py310,py311 && tox -e coverage` passes coverage when run on a single platform.

As above, this also runs with "platform exclusions" and "Python version exclusions" enabled. Producing a coverage report for all Python versions only really makes sense for `tox -e coverage-platform`.

The difference between `tox -e coverage` and `tox -e coverage-platform` is whether "python version exclusions" should be disabled.....as it is for the `-e coverage-platform` command.

To round this out, running `tox -e coverage-project` disables the "platform exclusions" as well as the "python version exclusions; this'll only work, though, if you have coverage files produced on Linux, macOS, and Windows.

Does this help? I'm not sure I can tell where we're missing each other...

freakboy3742: > Does this help? I'm not sure I can tell where we're missing each other...

Not really... maybe I'm misunderstanding the "double negative" aspect of what "platform exclusions enabled" means. I read it as the `#pragma` lines for no-cover-if-windows (et al) are in effect, so you shouldn't get a coverage miss if "exclusions are enabled".

To take this from another angle: what invocation of coverage would let me run a single test run for Python3.10 on macOS (i.e., `tox -e py310`), and get a 100% coverage report?

And, the inverse - how would I generate a *failing* report out of `tox -e py38,py39,py310,py311` on macOS (i.e., I have coverage for all Python versions, but I haven't run on Linux and Windows)?

rmartin16: hmm...I was concerned this would remain as confusing as it was to me when I started this...

First, your understanding of "coverage exclusions" is correct; when an exclusion like `no-cover-if-is-windows` is "enabled", then coverage for that code on Windows is ignored.

Second, though, there are two levels of exclusions:
- Python version exclusions
  - When these are enabled, it doesn't matter which version of Python is used for coverage reporting
  - (however, there is the issue of "phantom branches" with using a different version for reporting coverage versus creating the coverage files.)
- Platform exclusions
  - When these are enabled, it doesn't matter which platform is used for coverage reporting

So, from the user perspective trying to create coverage reports:
- Running `tox -e coverage`
  - This has both "python version" and "platform" exclusions enabled
  - So, it doesn't matter the platform or Python version being used to produce the coverage report
  - (however, again, the caveat that `coverage` should be run using the same version of Python used to produce the coverage files because of how that all works.)
- Running `tox -e coverage-platform`
  - This disables "python version" exclusions...so, for instance, the `no-cover-if-gte-py310` exclusion is disabled (i.e. ignored) and coverage is verified for those code blocks
  - However, "platform" exclusions, like `no-cover-if-is-windows`, remain in effect and that coverage is ignored
- Running `tox -e coverage-project`
  - This disables "python version" _and_ "platform" exclusions; so, all of the conditional coverage rules are ignored

So, specifically:

> what invocation of coverage would let me run a single test run for Python3.10 on macOS (i.e., `tox -e py310`), and get a 100% coverage report?

Running `tox -e coverage310` after `tox -e py310` will produce a 100% coverage because, by default, all coverage exclusions are enabled.

> how would I generate a failing report out of `tox -e py38,py39,py310,py311` on macOS (i.e., I have coverage for all Python versions, but I haven't run on Linux and Windows)?

Running `tox -e coverage-project` would disable _all_ exclusions and detect that linux-specific and windows-specific code was not covered.

What about this? And how much of this should be in the docs? because clearly I've internalized all this but it is far from obvious...

freakboy3742: > hmm...I was concerned this would remain as confusing as it was to me when I started this...
> 
> First, your understanding of "coverage exclusions" is correct; when an exclusion like `no-cover-if-is-windows` is "enabled", then coverage for that code on Windows is ignored
 
Ok - good to know I haven't missed something fundamental here :-)

> What about this? And how much of this should be in the docs? because clearly I've internalized all this but it is far from obvious...

Eureka - I've finally worked out where the gap is!

`tox -e py` runs the test suite on the current default Python3 (3.11 on my machine). However, `tox -e coverage` runs on the *oldest version of Python installed on your machine* (3.8 for me). 

I was seeing a bunch of coverage failures, and assuming they were all platform/version mismatches, but didn't actually verify that they were caused by Python version/platform branches. It turns out they were all 3.8/3.11 branching discrepancies. Once I worked that out, everything fell into place.

That means on my machine, `tox -e py && tox -e coverage311` and `tox -e py38 && tox -e coverage` both pass clean. `tox -e py311 && tox -e coverage311-platform` fails on the python version branches.

So - the issue is with explaining the differences in "default python version" that is used between `py` and `coverage` targets. The fact that `tox -e py && tox -e coverage` doesn't pass clean (unless you only have a single Python version installed) definitely isn't intuitively obvious (as my ... journey... has revealed :-).

I guess the alternative would be a "test and coverage for a single python version" target, but I'm not sure I see an easy way to add that without another set of complications to the tox config.

rmartin16: So, considering all this and your other comment above....I'm thinking we just remove talking about `tox -e coverage` and instead keep it simple for a beginner and just say "use `tox -e py310,coverage310` to produce a coverage report" and mention that "310" should be swapped out for the version of Python they are using.

> I guess the alternative would be a "test and coverage for a single python version" target, but I'm not sure I see an easy way to add that without another set of complications to the tox config.

It wouldn't be too difficult to add additional labels for each Python version; so `tox -m test39`, `tox -m test310`, etc. to run the tests and produce the coverage report. And then, we could just talk about these python version-specific labels in the coverage section.....as well as producing platform and project coverage with `tox -m test-platform` and `tox -m test-project`, respectively.

Ultimately, there's a lot of options in this version of the `tox.ini`....but we don't have to expose it all to anyone up front. An adventurous dev can poke around in `tox.ini` to discover this.

What do you think?

P.S. a side concern is people not noticing the difference between `tox -e` and `tox -m`....especially since tox will run `tox -e test` and kinda look like it's something...but really do nothing at all... I haven't found a way to make tox error for an "unknown/undefined" environment.

freakboy3742: > So, considering all this and your other comment above....I'm thinking we just remove talking about `tox -e coverage` and instead keep it simple for a beginner and just say "use `tox -e py310,coverage310` to produce a coverage report" and mention that "310" should be swapped out for the version of Python they are using.

Sounds like a good approach to me.

> > I guess the alternative would be a "test and coverage for a single python version" target, but I'm not sure I see an easy way to add that without another set of complications to the tox config.
> 
> It wouldn't be too difficult to add additional labels for each Python version; so `tox -m test39`, `tox -m test310`, etc. to run the tests and produce the coverage report. And then, we could just talk about these python version-specific labels in the coverage section.....as well as producing platform and project coverage with `tox -m test-platform` and `tox -m test-project`, respectively.

This also seems like a good idea; it provides a simple path in to the most likely "just run the tests and give me coverage" use case.

> Ultimately, there's a lot of options in this version of the `tox.ini`....but we don't have to expose it all to anyone up front. An adventurous dev can poke around in `tox.ini` to discover this.
> 
> What do you think?

Definitely agreed. We don't need to document everything in the contributor guide. We just need to document enough to explain the common developer use cases. To my mind, that means:

1. Run the tests quickly to prove they all pass
2. Run the tests on one platform and confirm I've got coverage
3. Run as many versions of the tests as I can on one platform, and give me coverage
4. Do as much of a CI-alike run as I can on one platform.

> P.S. a side concern is people not noticing the difference between `tox -e` and `tox -m`....especially since tox will run `tox -e test` and kinda look like it's something...but really do nothing at all... I haven't found a way to make tox error for an "unknown/undefined" environment.

Yeah - that's not great - but if the "public testing interface" is all `tox -m` calls, then it's a lot less likely that people will get caught out.

rmartin16:
Update coverage docs to say to use `tox -m test311` for Python version-specific coverage and `tox -m test-platform` for platform coverage.

I also added the `tl;dr` section at the top and a "make a git branch" section that at least aspires to prevent people from committing to `main`...

freakboy3742 (approved):
One last pass of minor cosmetic doc tweaks, and I think we done! Nice work!

PR #1263, opened 2023-05-06
Install specific requirement from `setup.cfg` for testing

rmartin16:
## Changes
- Introduces a straightforward script to install requirements as defined in the project's metadata
- By default, only non-extra requirements can be installed.
- To install an extra requirement, the name of the extra much be specified in `--extra`

## Notes
- The motivation for this is the several places we install the entire project when we just need one package
- I have spent a lot of time trying to understand the "best way" to install a subset of dependencies from a project
- PyPA basically [wants nothing](https://github.com/pypa/pip/issues/8049) to do with this...and for a lot of good reasons
  - So, expecting something like `pip install --only-dep=tox briefcase` to ever be a thing seems very unlikely
- `tox` has [expressed](https://github.com/tox-dev/tox/issues/2301) interest in supporting something like this
- The most useful tool I found to provide an API of sorts to extract the requirement was `pip-tools`
  - While its for freezing requirements for deployments, it technically serves this purpose
  - For instance: `pip-compile --resolver=backtracking --extra dev -o - 2>/dev/null | grep '^tox='`
  - However, it's quite slow in this use-case

@freakboy3742, this is mostly to gauge interest after our quick messages about this. Let me know what you think. At the end of the day, there isn't _that_ much time spent installing all of Briefcase...but when you just need `coverage` or `tox`, it seems like a little much. If this doesn't seem worth the squeeze, I'm fine moving on.

## PR Checklist:


.github/workflows/ci.yml
```diff
@@ -80,9 +80,7 @@ jobs:
         # See https://github.com/pypa/pip/issues/11982 for details
         python -m pip install --upgrade "pip!=23.1"
         python -m pip install --upgrade setuptools
-        # We don't actually want to install briefcase; we just
-        # want the dev extras so we have a known version of tox.
-        python -m pip install $(ls dist/briefcase-*.whl)[dev]
+        python tests/env_utils.py install-req tox
```
freakboy3742, line 83: This will likely be opaque enough for first-time readers (or, lets be honest, me in 2 months :-) that it's worth retaining a comment so it's clear what this is doing.

freakboy3742 (commented):
tl;dr - I like this as a general direction; my only comments are around bike-sheddy details.

Specifically - Do we anticipate an "env util" other than installing dev requirements? And - if we *do* end up with a need for another tool, would there be any reason that we need to cram it into this tool, rather than just adding a new one?

My suggestion - make this an `install_requirement.py` script at the top level of the project (with a bunch of docs at the top of the file describing exactly why it exists). There's no reason it needs to be hidden in the test folder; and a single purpose "install a requirement" script is easier to explain than a generic "environment utils" script.

If I were going to go down the path of "adding features that I'm not sure we have a use case for", I might tweak the command line interface so that invocation was `python -m install_extra dev tox` - i.e., install a specific requirement from the named section, possibly with `python -m install_extra dev` would install all the dev requirements without installing the package itself.

If we *really* want to go down the rabbit hole, it might be worth publishing this as a standalone package on PyPI. If we treat it the same way we treat pip and setuptools (i.e., a tool that is allowed to be unpinned because main is always stable and compatible). We can then add it as a simple requirement, then use it as `install_extra dev tox`. This same requirement exists on Toga, rubicon, etc... and I can imagine it being useful to others...

rmartin16:
I've implemented the basic recommendations for this to remain as a relatively simple script.

However, as for usage such as `python -m install_extra dev tox`, based on my understanding, I think this would inherently necessitate installing a package with such an entry point defined...but that's mostly what we're trying to avoid. Admittedly, I don't have a rock solid understanding of entry points, so please let me know if you think I'm coming to the wrong conclusion.

Speaking of conclusions...I'm not really super enthused with a random script in the root of the project (although, the location isn't so much my hesitation as much as it's just a script that's divorced from everything else).

So, I'm considering this still to be an interim step towards a more formalized tool that would ideally be packaged and put on PyPI. Although, my hesitation there is I'm not particularly a packaging expert and I feel like there's a lot of holes and missing functionality given all the permutations of Python packaging. I'll do some more research, though, and see how this shakes out.

freakboy3742:
> However, as for usage such as `python -m install_extra dev tox`, based on my understanding, I think this would inherently necessitate installing a package with such an entry point defined...but that's mostly what we're trying to avoid. 

Running `install_extra` would require an entry point; `python -m install_extra` is effectively just `python install_extra.py` but  allowing for `install_extra` to be a directory that contains a `__main__.py`. 

> Speaking of conclusions...I'm not really super enthused with a random script in the root of the project (although, the location isn't so much my hesitation as much as it's just a script that's divorced from everything else).

I can empathise; however, from my perspective, it's a key piece of project infrastructure that's relatively easy to explain, so I can live with it.

> So, I'm considering this still to be an interim step towards a more formalized tool that would ideally be packaged and put on PyPI. Although, my hesitation there is I'm not particularly a packaging expert and I feel like there's a lot of holes and missing functionality given all the permutations of Python packaging. I'll do some more research, though, and see how this shakes out.

Oh, thar *absolutely* be dragons in here for the general case :-) But we don't need to solve the general case ourselves.

install_requirement/__main__.py
freakboy3742, line 1: Just confirming this change was deliberate - putting this in a directory isn't required for a single file module - a single `hello.py` can be invoked as `python -m hello`.

rmartin16: it was deliberate insofar as I was thinking it was necessary :) i didn't realize `python -m` would work like that. I'll revert it back in that case.

.github/workflows/ci.yml
```diff
@@ -81,10 +81,9 @@ jobs:
     - name: Install dev Dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install --upgrade setuptools
-        # We don't actually want to install briefcase; we just
-        # want the dev extras so we have a known version of tox.
-        python -m pip install $(ls dist/briefcase-*.whl)[dev]
+        python -m pip install --upgrade setuptools build wheel
```
freakboy3742, line 84: I'm not 100% clear on why `build` and `wheel` need to be on this list. Is this purely defensive, or is there a specific need?

freakboy3742 (changes requested):
A couple of questions/clarifications, but this mostly makes sense.

.github/workflows/ci.yml
rmartin16: This script depends on `build` and `wheel`.

freakboy3742: /me facepalms. Of course - missed the imports at the top of the script.

install_requirement.py
```diff
@@ -0,0 +1,173 @@
...
+        requirement
+        for requirement in map(Requirement, project_metadata.get_all("Requires-Dist"))
+        if not requirement.marker or requirement.marker.evaluate({"extra": extra_name})
+    ]
+
+    matching_requirements = [
+        requirement
+        for requirement in project_requirements
+        if any(requirement.name.startswith(req) for req in requested_requirements)
+    ]
```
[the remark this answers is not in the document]
rmartin16: Right; that is the intended use-case since many testing dependencies tend to be named with a prefix that's an implied dependency. Or in the least, that's true for the `pytest` and `coverage` plugins we're using.

This type of functionality should probably be supported with a regex but I didn't want to add that complication.

Using a literal may be more expected and allows the behavior to be more explicit.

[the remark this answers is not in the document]
freakboy3742: Yeah - this is relying on a very specific naming convention - while this may be valid for a lot of packages, there's not guarantee that it will always be true. I can also imagine a possible situation where we *literally* just want pytest and no plugins, so explicit will always be better than implicit.

freakboy3742 (approved):
Nice work - makes for a nice cleanup!

PR #1267, opened 2023-05-08
Disable progress bar and wait bar if `stdout` is not attached to a tty

rmartin16:
## Changes
- Disables the Progress Bar and Wait Bar if `stdout` if not attached to a tty

## Notes
- It was discovered in beeware/.github#36 that when `FORCE_COLOR` is set, Rich not only enables ANSI color and styling, but also turns on "live" elements like the Progress Bar
- Allowing the Progress Bar to run in CI can create really ugly [logs](https://github.com/beeware/.github/actions/runs/4902657729/jobs/8754641453)
- I initially considered disabling these dynamic elements when the `--no-input` switch is used....but I decided against it because:
  - It's a weak proxy for "non-interactive session"
  - It can disable the dynamic elements when they are still desired and reasonable to show
  - When users encounter the problem of polluted CI logs, they would be required to add an un-intuitive switch to _all_ invocations of Briefcase
- Rich's [recommendation](https://rich.readthedocs.io/en/stable/console.html#interactive-mode) for this situation is to instantiate `Console` with `force_terminal=True` and `force_interactive=False`
  - However, this recommendation is qualified with "you might want to do this in some CI environments"....and this only seems to make sense if the application using Rich is only really used in CI by the application developer
  - In Briefcase's case, we would need to establish some process for users to implement in their CI in order to activate these overrides for `Console`
  - Alternatively, we could always connect `sys.stdout.isatty()` to these overrides....although, I think I'd rather just let Rich do what it does by default and just target our known issues
    - After all, Rich has all the battle-tested logic for when terminal and interactive things should happen
- Thus, just disabling the specific uses of Rich's dynamic elements in Briefcase when a tty isn't available seems like the best way to use default functionality while avoiding known issues

## PR Checklist:

### DIFF
--- a/changes/1267.bugfix.rst
+++ b/changes/1267.bugfix.rst
@@ -1,0 +1,1 @@
+Briefcase's Progress Bar and Wait Bar will no longer be rendered if the ``FORCE_COLOR`` environment variable is set.
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -348,23 +348,29 @@ class Log:
 
 class Console:
     def __init__(self, printer=Printer(), enabled=True):
-        self.print = printer
-        self.input = printer.console.input
         self.enabled = enabled
+        self.print = printer
+        # Use Rich's input() to read from user
+        self.input = printer.console.input
         self._wait_bar: Progress = None
-        # Signal that Rich is dynamically controlling the console output.
-        # Therefore, all output must be printed to the screen by Rich to
-        # prevent corruption of dynamic elements like the Wait Bar.
+        # Signal that Rich is dynamically controlling the console output. Therefore,
+        # all output must be printed to the screen by Rich to prevent corruption of
+        # dynamic elements like the Wait Bar.
         self.is_console_controlled = False
 
-    def prompt(self, *values, markup=False, **kwargs):
-        """Print to the screen for soliciting user interaction.
+    @property
+    def is_interactive(self):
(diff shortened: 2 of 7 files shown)


changes/1267.bugfix.rst
```diff
@@ -0,0 +1 @@
+Briefcase's Progress Bar and Wait Bar will no longer be rendered if the ``FORCE_COLOR`` environment variable is set.
```
freakboy3742, line 1:
```suggestion
The Progress Bar and Wait Bar will no longer be rendered if Briefcase runs on a non-TTY console (e.g., in CI environments, or when being piped through another process).
```

freakboy3742 (changes requested):
Implementation looks good; one suggestion about the change note that I wanted to confirm before merging.

changes/1267.bugfix.rst
rmartin16: I had to think to remember why I choose this wording...

So, Briefcase already doesn't render progress bars or the wait bar if `stdout` isn't attached to a TTY. AFAICT, it is only when `FORCE_COLOR` is set that Rich will start rendering these without a TTY....or `force_terminal=True` is explicitly specified when the `Console` is created.

Mentioning `FORCE_COLOR` in the RN does seem especially esoteric...but it's most accurate.

freakboy3742: Ah - got it. Makes sense.

freakboy3742 (approved):

PR #1271, opened 2023-05-12
Bump JDK `8u242-b08` to `17.0.7+7`

### LINKED ISSUE #827: Android SDK XML warnings when using a user-provided SDK
**Describe the bug**
I used the ANDROID_SDK_ROOT environment variable to make Briefcase use my existing Android SDK. This resulted in the following warnings when Briefcase runs `sdkmanager --list_installed` (duplicate lines removed):

```
Warning: This version only understands SDK XML versions up to 2 but an SDK XML file of version 3 was encountered. This can happen if you use versions of Android Studio and the command-line tools that were released at different times.
Warning: unexpected element (uri:"", local:"base-extension"). Expected elements are <{}codename>,<{}layoutlib>,<{}api-level>
Warning: unexpected element (uri:"", local:"base-extension"). Expected elements are <{}codename>,<{}api-level>
Warning: unexpected element (uri:"", local:"base-extension"). Expected elements are <{}vendor>,<{}codename>,<{}abi>,<{}api-level>,<{}tag>
Warning: Errors limit exceeded. To receive all errors set com.sun.xml.bind logger to FINEST level.
```

This is because I usually manage the SDK using Android Studio's graphical SDK Manager (current version: Chipmunk), not the command-line sdkmanager. The cmdline-tools "latest" directory in my SDK is version 5, because I've never had any reason to update it. Version 6 and 7 both work without warnings, so this doesn't affect Briefcase-installed SDKs, which use version 6.

The warnings appear to be harmless, but Briefcase could fix them by installing and using a specific version of cmdline-tools, as discussed at https://github.com/beeware/briefcase/pull/766#issuecomment-1160947331.

**Environment:**
 - Python version: 3.8
 - Software versions:
   - Briefcase: 7a95f4bc

### LINKED ISSUE #1065: Bump OpenJDK Version
### Describe the bug

The JDK version has been superseded; putting this out there to consider bumping it.

`src/briefcase/integrations/java.py`
```
        # As of April 10 2020, 8u242-b08 is the current AdoptOpenJDK
        # https://adoptopenjdk.net/releases.html
        self.release = "8u242"
        self.build = "b08"
```

Version `jdk8u362-b09` is available.
Additionally, version 11 is also LTS.

There is also this banner at the [website](https://adoptopenjdk.net/releases.html):
24th July 2021: [AdoptOpenJDK is moving to the Eclipse Foundation](https://blog.adoptopenjdk.net/2021/03/transition-to-eclipse-an-update/) and rebranding.
Our July 2021 and future releases will come from [Adoptium.net](https://www.adoptium.net/)

### Steps to reproduce

`briefcase run android` will require the jdk.

### Expected behavior

Not entirely sure....but it seems like we want to avoid the JDK getting too old.

### Screenshots

_No response_

### Environment

- Python version: 3.11
- Software versions:
  - Briefcase: `0.3.12.dev359+gcf18ae8c.d20230128`

### Logs

_No response_

### Additional context

_No response_

### LINKED ISSUE #1112: KeytoolException: Failed to read key AndroidDebugKey from store: Invalid keystore format
Reported by warchants on Discord, with the following log:
```
           > Task :app:packageDebug FAILED                                                                                                                         subprocess.py:638
                                                                                                                                                                   subprocess.py:638
           FAILURE: Build failed with an exception.                                                                                                                subprocess.py:638
                                                                                                                                                                   subprocess.py:638
           * What went wrong:                                                                                                                                      subprocess.py:638
           Execution failed for task ':app:packageDebug'.                                                                                                          subprocess.py:638
           > A failure occurred while executing com.android.build.gradle.tasks.PackageAndroidArtifact$IncrementalSplitterRunnable                                  subprocess.py:638
              > com.android.ide.common.signing.KeytoolException: Failed to read key AndroidDebugKey from store "C:\Users\Shahab\.android\debug.keystore": Invalid  subprocess.py:638
           keystore format                                                                                                                                                          
```

The first time a debug build is done on any given machine, the Android Gradle plugin creates the debug.keystore file with a randomly-generated key, and then uses it for all subsequent debug builds of all apps on that machine. Unfortunately it looks like there's been a backward-incompatible change to the file's format, so files created by new Java versions (e.g. the version that comes with new versions of Android Studio) cannot be read by the Java version that Briefcase currently uses.

Based on [this discussion](https://stackoverflow.com/questions/68829204), we could fix this by moving to Java 8u301 or newer. I've confirmed that it works OK with Corretto version 1.8.0_352.

Such a new version is not available at https://github.com/AdoptOpenJDK/openjdk8-binaries, which Briefcase currently uses. But its successor, https://adoptium.net/, currently offers version 8u362. See #1065.

rmartin16:
## Changes
- Bumps JDK from `8u242-b08` to `17.0.7+7`
  - Fixes #1065
  - Fixes #1112 
  - Fixes #827
- Dependent on bumping Gradle and the Android Gradle plugin in the template
  - beeware/briefcase-android-gradle-template#68
- Implements version compatibility checking for templates
  - Now, when breaking changes are introduced for a platform's template, the version of Briefcase they are being introduced in to can be used as an epoch version to later determine compatibility
  - The target Briefcase version epoch for the template is specified via the `target_version` key in the `briefcase` section of `briefcase.toml`
  - The currently supported Briefcase version epoch for a platform is specified via `platform_target_version` on the `Command` for the platform
  - If a Briefcase platform declares an epoch version, the rolled out template must declare support for the same epoch version.

## PR Checklist:


rmartin16:
Adding `--add-opens=java.base/java.io=ALL-UNNAMED` to `jvmargs` in `gradle-wrapper.properties` let the build finish for me.

src/briefcase/integrations/java.py
```diff
@@ -70,7 +79,7 @@ def verify(cls, tools: ToolCache, install=True):
         install_message = None
 
         if tools.host_arch == "arm64" and tools.host_os == "Darwin":
-            # Java 8 is not available for macOS on ARM64, so we will require Rosetta.
+            # Java is not available for macOS on ARM64, so we will require Rosetta.
```
freakboy3742, line 82: This isn't true any more; we can get a native ARM Java.

freakboy3742 (commented):
One thought - to make the migration path cleaner, do we want to unpack into a `tools/java17` directory? That should dramatically simplify any issues around disambiguating java8 and java17 installs, or trying to do in-place upgrades.

src/briefcase/integrations/java.py
rmartin16: That's what I was thinking....so, all the logic to handle rosetta can come out?

freakboy3742: I think so - this was the only thing we were using it for, AFAIR.

rmartin16:
> One thought - to make the migration path cleaner, do we want to unpack into a `tools/java17` directory? That should dramatically simplify any issues around disambiguating java8 and java17 installs, or trying to do in-place upgrades.

Great idea; this'll definitely simplify handling changing the java version in user installs.

As far how this upgrade should interact with Gradle, I'll wait to see if Malcolm (or anyone else) has feedback here; while I got things working, I have practically no prior knowledge for all these interactions.

mhsmith:
Thanks, I'll have a look at this on Monday.

mhsmith:
> Adding `--add-opens=java.base/java.io=ALL-UNNAMED` to `jvmargs` in `gradle-wrapper.properties` let the build finish for me.

The Briefcase Android template (and most existing user build directories) is currently on Android Gradle plugin 4.2. I haven't tested that with Java 17, and probably neither has Google because they were still on Java 11 at that time. But as long as this workaround allows it to pass the Toga testbed, it should be fine.

Upgrading to Java 17 would also have the following benefits:
* It would fix #1112, so I've added that to the description above.
* It would allow us to upgrade the Android Gradle plugin, which would remove the warnings mentioned in [this comment](https://github.com/beeware/briefcase/issues/827#issuecomment-1217860885) (though not the `sdkmanager` warnings in the original report). 

> > One thought - to make the migration path cleaner, do we want to unpack into a `tools/java17` directory? That should dramatically simplify any issues around disambiguating java8 and java17 installs, or trying to do in-place upgrades.
> 
> Great idea; this'll definitely simplify handling changing the java version in user installs.

So the idea is that Briefcase would still only support one version of Java at a time, but unpack into a `java17` directory in case that changes in the future?

In which case when upgrading from an older version of Briefcase, I guess the existing `java` directory would just be ignored, and left there forever?

freakboy3742:
> So the idea is that Briefcase would still only support one version of Java at a time, but unpack into a `java17` directory in case that changes in the future?

Correct.
 
> In which case when upgrading from an older version of Briefcase, I guess the existing `java` directory would just be ignored, and left there forever?

That's my thinking, yes. There's a *very* small edge case (mostly affecting us as developers) where we might need to run/test both java8 and java17 in different configurations, so there is *some* benefit to keeping the old java; at some point in the future. It's not a *huge* amount of disk space (200MB), and it's easily removed manually if there's an issue. I guess we could add a cleanup pass to `briefcase upgrade`, but honestly, I don't think it's worth the effort.

mhsmith:
> Adding `--add-opens=java.base/java.io=ALL-UNNAMED` to `jvmargs` in `gradle-wrapper.properties` let the build finish for me.

@freakboy3742 pointed out updating the template wouldn't have any effect on existing build directories, which would fail to build with Java 17. While build directories are theoretically ephemeral, we do often recommend people to edit them to work around missing features in Toga, so we should avoid forcing people to regenerate them.

Here are some possible solutions:
* See if there's another way to pass this argument, e.g. on the command line or an environment variable.
* Make Briefcase edit existing gradle-wrapper.properties files to add the argument
* Make Briefcase support both Java 8 and Java 17, and have the template specify which one it wants, e.g. in briefcase.toml. Existing build directories which don't specify a version would default to Java 8.
  * This would require the Rosetta code to be retained, unless we switch over to Amazon Corretto, which provides macOS aarch64 builds for both Java 8 and 17. Since all JDKs have roughly the same file layout, I wouldn't expect this to be difficult.

rmartin16:
Hmm....so, I guess we need to decide which trade-offs are most important....and what should be upgraded at this time.


### Gradle Updates
I pushed beeware/briefcase-android-gradle-template#8 to upgrade Gradle and its Android plugin to their latest versions. That obviates the need for the `jvmargs` update....but it requires a different tweak to specify the package namespace.

Are you interested in updating Gradle at this time?

### Breaking Changes
As for the breaking changes in the `build` directory....we'd be introducing a lot of complexity to avoid them. And it feels like users should have a way to apply their customizations on demand....for CI or just creating their App on an arbitrary machine. And supporting multiple versions of JDK....would this be necessary beyond avoiding breaking changes?

I think I'm leaning towards making the breaking changes and requiring users to rebuild the `build` directory....but I will defer to you two for this decision.

mhsmith:
> Are you interested in updating Gradle at this time?

Yes: I'll probably have to drop support for Android Gradle plugin 4.2 in the next version of Chaquopy anyway, so it's good that you're looking at this now.

> I think I'm leaning towards making the breaking changes and requiring users to rebuild the `build` directory.

I'm OK with that, as long as the requirement is communicated clearly. Most users won't read release notes, and as it stands, a user running `pip install --upgrade briefcase` and then `briefcase build android` would get an incomprehensible Java error message, which is a pretty bad user experience.

What do you think of [this idea](https://github.com/beeware/briefcase/issues/827#issuecomment-1219201337) for a  mechanism to declare when Briefcase has broken backward compatibility with existing build directories? That should come in useful on other platforms as well.

rmartin16:
> > I think I'm leaning towards making the breaking changes and requiring users to rebuild the `build` directory.
> 
> I'm OK with that, as long as the requirement is communicated clearly. Most users won't read release notes, and as this PR stands, a user with an existing Android build directory who runs `pip install --upgrade briefcase` and then `briefcase build android` would get an incomprehensible Java error message, which is a pretty bad user experience.
> 
> What do you think of [this idea](https://github.com/beeware/briefcase/issues/827#issuecomment-1219201337) for a mechanism to declare when Briefcase has broken backward compatibility with existing build directories? That should come in useful on other platforms as well.

Yeah, I agree - we should definitely avoid surfacing such errors from underlying tools.

I think [that idea](https://github.com/beeware/briefcase/issues/827#issuecomment-1219201337) to track the source version of Briefcase in the `build` directory makes sense. There's also been discussion around tracking other "metadata" about a Briefcase project (like in #807)....I wonder if these should use a shared tracking infrastructure...

I'll play around a bit with it. Additional ideas/thoughts certainly welcome....I always love a ballooning scope 🥲

mhsmith:
> I think that idea to track the source version of Briefcase in the build directory makes sense.

My initial thought was that each template could simply have a placeholder for the Briefcase version in its briefcase.toml file, so that wouldn't require much new code.

rmartin16:
> My initial thought was that each template could simply have a placeholder for the Briefcase version in its briefcase.toml file, so that wouldn't require much new code.

Right....my hesitation with this approach has been that it would require _all_ users to re-run the `create` command each time Briefcase is updated. I would rather only require this to occur if it is necessary....however, as I've been thinking about it, I've realized this would only be possible if the information about compatibility between Briefcase's default templates and Briefcase itself was maintained out-of-band; that is, that information could not be a part of the template itself....because a cloned template cannot know which future versions of Briefcase it will be compatible with.

Hypothetically, an independent source of this compatibility map could be maintained somewhere on the internet....but then the trade-off of maintaining this map and checking it for _every_ command a user runs....vs just detecting that the current Briefcase is different than the Briefcase that cloned the template and prompting the user to have the `create` command re-ran for them....starts to become more competitive.

So, ok; I'll move forward with this approach. To make the data more structured than a comment in `briefcase.toml`, I'm leaning towards making a `briefcase` section with a `version` key. I could also imagine maybe a `metadata` section and a `briefcase_version` key... And there will probably need to be some special logic to avoid re-running the `create` command for each different SCM version of Briefcase. Thoughts/ideas welcome.

mhsmith:
> this would only be possible if the information about compatibility between Briefcase's default templates and Briefcase itself was maintained out-of-band; that is, that information could not be a part of the template itself

Like I said in the [original comment](https://github.com/beeware/briefcase/issues/827#issuecomment-1219201337), my idea is that this information would be in Briefcase itself, with each platform module having a *minimum* template version it supports. So if the next Briefcase version is 0.3.15, then the Android module would have a minimum template version of 0.3.15, while the other modules probably wouldn't need any minimum at all yet. Each module would only raise its minimum template version when necessary, so that would avoid forcing the user to rerun `create` for every new version of Briefcase.

rmartin16:
> > this would only be possible if the information about compatibility between Briefcase's default templates and Briefcase itself was maintained out-of-band; that is, that information could not be a part of the template itself
> 
> Like I said in the [original comment](https://github.com/beeware/briefcase/issues/827#issuecomment-1219201337), my idea is that this information would be in Briefcase itself, with each platform module having a _minimum_ template version it supports.

Ahh...I didn't realize you meant literally embed the compatibility information within Briefcase. Nonetheless, I considered this and maybe eliminated it from my implementation too hastily.

My primary reason for not wanting to do this was because of arbitrary user-specified templates. On the one hand, we're (kinda) introducing a mechanism to control the compatibility between Briefcase and templates. On the other hand, hardcoding the "oldest supported version" in to Briefcase really limits the usability of the mechanism for arbitrary templates.

I will need to think about these interactions more (and the one below) to fully flesh out an understanding.

(This also made me realize that the version of a template can be specified in `pyproject.toml`.....so, even if you're using a supported version of Briefcase to run the `create` command, you would actually need to verify that the target branch for the template is suitable for the current Briefcase version. Otherwise, a subsequent check of the Briefcase version against `briefcase.toml` will succeed...while the template is actually too old for this Briefcase....)

rmartin16:
In light of the edge cases I've found so far with tracking the version of Briefcase that cloned the repo inside the rolled out template to determine compatibility between Briefcase and a template, I'm taking a step back to make sure I understand the problem.

### Problem
Report incompatibility between the running version of Briefcase and a template that's already been rolled out.....or even potentially a template that's being considered for rolling out.

### History
The discussion thus far has centered around capturing the version of Briefcase that actually rolls out the template. The primary issue with this strategy is it uses the version of Briefcase as a proxy for the version of the template....and this proxying is what introduces the edge cases.

### Assessment
So, now stepping back, what are we actually trying to detect is incompatible between the two? For this PR, anyway, it is that Briefcase and the template agree about the requirements for the underlying dependencies to operate on the rolled out template. That being the case, dependency management and compatibility is a pretty well-trodden area.

### Design
My new approach, therefore, is simply to require that Briefcase and the template being used make compatible assertions about underlying dependencies.
- Briefcase will assert lower-bound (and potentially upper-bound) version requirements (if any exist).
- Templates will assert the exact version of the dependency used in operation.
- If a template doesn't assert a version for the dependency or the version is out of range of what Briefcase is asserting, then Briefcase will error.

### Implementation
- Introduce a new step in verification that confirms template dependency compatibility 
  - Create new section in `briefcase.toml`
    - New section would probably be along the lines of `briefcase.requires`
    - Add requirements for `gradle==8.0` and `android_gradle_plugin==8.0.1`
  - Create new Briefcase Command class variable to declare template requirements
- TODO: need to flesh out the details

Thoughts, opinions, omissions, etc. welcome.

freakboy3742:
> Thoughts, opinions, omissions, etc. welcome.

It's true that we've been using the Briefcase version as a proxy for version compatibility; however, I'm not sure I follow why this introduces edge cases - or, at least, why the edge case is any more than being overly conservative. 

I agree that for many releases, a vN template will be entirely compatible with vN+1, and doing a hard check on briefcase version will result in projects that would otherwise be fine from raising an error.

However, we generally encourage people to treat the `build` folder as ephemeral. There are some exceptions to this (most notably, iOS and Android projects where users need to modify permissions), but I'd suggest those exceptions are all examples of missing features. Since the folder is intended to be ephemeral, I don't see the conservative approach as being especially problematic - a "delete and start again" error message is a mostly acceptable solution.

We *could* get into micromanaging specific tool versions (in this case Java, but I guess the same could be true of Xcode, Visual Studio, or anything else); but I'm not sure the complexity is worth it. Ultimately, a check of "generated requiring Java==8" won't give you any more fidelity than "generated with Briefcase 0.3.15". 

The only reason per-tool dependencies would make sense to me is if we were going to maintain support for older toolchains (e.g., keeping support for Java 8 when we add support for Java 17) - but honestly, I'm not convinced that's worth the complexity. We have limited resources; we already have enough problems maintaining Python and OS versions without borrowing more trouble :-)

rmartin16:
I'm not sure I understand your conclusions. Are you arguing for status quo? or for stamping the Briefcase version in to the rolled out template? or something else perhaps?

I think I will have an easier time enumerating the compromises if I understand your current position better.

freakboy3742:
> I'm not sure I understand your conclusions. Are you arguing for status quo? or for stamping the Briefcase version in to the rolled out template? or something else perhaps?
> 
> I think I will have an easier time enumerating the compromises if I understand your current position better.

I'm not arguing for status quo - I'm saying that *per-tool* version/dependency checks are overkill, and a Briefcase version in the generated app template should be sufficient.

We definitely need *some* sort of version check to catch the "this app was generated expecting Java 8 and Gradle 4.2" case - but I think that check can be at the granularity of "This app was generated with Briefcase 0.3.X". We can then introduce checks on a per-backend basis to confirm if there are any known incompatibilities between the generated app and the *current* Briefcase version.

rmartin16:
> We definitely need some sort of version check to catch the "this app was generated expecting Java 8 and Gradle 4.2" case - but I think that check can be at the granularity of "This app was generated with Briefcase 0.3.X".

If such a check is performed using the Briefcase version as a proxy, there are two primary compromises I'm aware of:
- This assumes the template being cloned is already compatible with the version of Briefcase being used.
  - The most obvious example where this would occur is if an older branch of the Briefcase template is pinned.
  - I don't think there's a good use-case for this type of pinning except if someone is maintaining a cloned repo to track their own augmentation to the released Briefcase template.
  - So, the outcome being that a not near-zero likelihood of users encountering the errors we're trying to avoid after a Briefcase upgrade
- This also doesn't allow third parties to leverage this Briefcase <-> template dependency mechanism
  - Since we'd be tracking this dependency at the granularity of a Briefcase version, even unaffected templates would need to be re-created.
  - But additionally, we'd be stopping short of allowing third party templates a way to better control this dependency themselves
  - (This compromise is noticeably weaker and could reasonably be considered irrelevant/out-of-scope)

> We can then introduce checks on a per-backend basis to confirm if there are any known incompatibilities between the generated app and the current Briefcase version.

I'm not sure how this would be implemented practically....unless we'd want Briefcase to start manually inspecting specific files within the rolled out template. I haven't been too keen on this so far because it really starts to interlock Briefcase and the released templates....as well as potentially unobvious mandates for third party templates.

Moreover, though, this really just sounds like the dependency checking I outlined...given it basically just says Briefcase requires these versions and the template provides these versions, are they compatible?

In conclusion, I also feel a bit like all this dependency checking is somewhat overkill....however, I also want to make sure we're on the same page about why stamping the version of Briefcase in to the rolled out template and using that for compatibility checks later has holes. This is why my original design was to simply require re-running the `create` command for _every_ Briefcase upgrade.

If this is still considered much ado about nothing, no worries. I can implement one of these proxies with the accepted trade-offs. However, I'll probably need more insight in to how specific backend dependency checks could be otherwise implemented.

freakboy3742:
> > We definitely need some sort of version check to catch the "this app was generated expecting Java 8 and Gradle 4.2" case - but I think that check can be at the granularity of "This app was generated with Briefcase 0.3.X".
> 
> If such a check is performed using the Briefcase version as a proxy, there are two primary compromises I'm aware of:
> 
> * This assumes the template being cloned is already compatible with the version of Briefcase being used.

It has to be, by definition. Whenever we use a template, we clone a version-specific branch of that template. If the version-specific branch isn't compatible with that version of Briefcase, then that's a bug.
   
I feel like we're missing each other in the conversation here, but I can't work out where. 

The problem we're trying to avoid is that Briefcase 0.3.15  (assuming this change is in the next release) will verify Java 17. An Android project generated using Briefcase 0.3.14 will contain a template that requires Java 8. Building that project using Java 17 will raise errors that will be mostly impenetrable to the end user.

What I'm proposing (and, AFAIK, what @mhsmith has proposed) is:

* adding a `briefcase_version = '0.3.15'` tag to the `briefcase.toml` for the Android template
* At some point in the build process, the Android backend will perform a check that `briefcase_version >= 0.3.15`. If `briefcase.toml` template *doesn't* provide this key, we can assume that the project was generated on 0.3.14 or earlier. If this condition isn't met, raise a clean "we can't build this" error.

When 0.3.16 is released, projects generated with 0.3.15 should still be compatible (assuming there isn't some other big change), so the `version >= 0.3.15` check will still pass.

In this context `briefcase_version` is really only serving as a proxy for "Java 17, Android SDK vWhatever, Xcode 13.0.0, ...". It's encoding "This template should work with the tools that were provided by Briefcase version X", or "what is the oldest version of Briefcase that can run this template?".

The only edge case I can see is the case of a user providing a custom template branch; if they're doing that, then they're also responsible for updating the template. If they've forked the 0.3.14 Android template, then it won't run on Briefcase 0.3.15, because it won't be compatible with Java 17 (or, at least, we can't guarantee that). If it is, then they can update their template, or add `briefcase_version = 0.3.15` to their existing project's `briefcase.toml`.

I'm *not* suggesting that we add this version check to every project, or every backend - only to backends where we know there is a compatibility problem. For now, the Android backend's version of Java is the only compatibility issue that I'm aware of. 

> * This also doesn't allow third parties to leverage this Briefcase <-> template dependency mechanism  
>   * Since we'd be tracking this dependency at the granularity of a Briefcase version, even unaffected templates would need to be re-created.
>   * But additionally, we'd be stopping short of allowing third party templates a way to better control this dependency themselves
>   * (This compromise is noticeably weaker and could reasonably be considered irrelevant/out-of-scope)

I guess we *could* break this down to a finer grained level; I guess it doesn't really matter if we define the minimum briefcase version and then imply the version of Java, or specify the exact version of Java that is a literal dependency. I'm not sure there's really a use case for this, but I guess there's some appeal to defining "actual" requirements rather than implied ones.

rmartin16:
Ahh...I think I see at least part of our disconnect, if not all. Your proposed implementation is different than what's been discussed; it would hardcode the minimum supported version of Briefcase inside the template instead of within Briefcase itself. Previously, the implementation being discussed was hardcoding the minimum version of Briefcase within the Gradle platform implementation.

Nonetheless, this seems a lot more workable.

I am wondering about it not applying this to all platforms....in the very least, I don't see why the arbitrary introduction the `briefcase_version` tag in to a `briefcase.toml` wouldn't be respected for an arbitrary platform. Although, we would need a way to conditionally enforce the presence of this tag for certain platforms, namely Gradle, right now.

I'll put an implementation together in the next few days :) thanks for sticking with me

mhsmith:
> In this context `briefcase_version` is really only serving as a proxy for "Java 17, Android SDK vWhatever, Xcode 13.0.0, ...". It's encoding "This template should work with the tools that were provided by Briefcase version X"

Yes, I think this is the clearest way to look at it, and that's why it has to be hard-coded into the template, rather than introduced via a template placeholder variable as I proposed above. And that also allows the same mechanism to be used by third-party templates, or other branches of the default template, which are use cases which didn't occur to me before.

> it would hardcode the minimum supported version of Briefcase inside the template instead of within Briefcase itself. Previously, the implementation being discussed was hardcoding the minimum version of Briefcase within the Gradle platform implementation.

We would still need to specify versions in both places, but the`briefcase_version` in the template can serve as both a minimum and a maximum:
* If `briefcase_version` is greater than the actual Briefcase version, then we should always refuse to build, because we cannot know what changes were made in that future version.
* If `briefcase_version` is less than the actual Briefcase version, then we can proceed with the build as long as `briefcase_version` is above a certain minimum. The minimum should be set separately for each backend, on the basis of tool upgrades, or any other factor that breaks backward compatibility with old templates.

In other words, we can do the build as long as:  
`min_template_version` (in gradle.py) <= `briefcase_version` (in briefcase.toml) <= Actual Briefcase version

rmartin16:
> We would still need to specify versions in both places, but the`briefcase_version` in the template can serve as both a minimum and a maximum:

To make sure I understand, to what end would specifying the minimum version also within Briefcase serve?

Is it intended to be a forcing mechanism to ensure a minimum Briefcase version is enforced in the case where one is not specified in `briefcase.toml`?

So, as applied here, if these breaking changes are a part of Briefcase v0.3.15, we would need to ensure all Gradle templates are actually advertising a minimum supported version to get the benefit for upgrading users...or users of third party templates.

Just imagining a future state well beyond this upgrade, I'm kinda leaning towards just making `briefcase_version` in `briefcase.toml` required. So, for this upgrade, everyone would need to re-run the `create` command but then this feature would be fully implemented without all these caveats...

rmartin16:
ohhh....I think I see now....or at least, I see this:

If a template asserts a minimum supported version of `0.3.15`, for instance, then a currently running version of Briefcase can verify it is that version or newer. But if we iterate this process again and introduce another round of breaking changes in `0.3.16`, then Briefcase would still perceive that originally rolled out template as supported...when it is, in fact, not.

If, however, Briefcase itself was tracking the versions for the platforms that contain breaking changes (as you're saying @mhsmith), then that would be enough information to detect the rolled out template is no longer compatible after a version of Briefcase is installed with the subsequent iteration of breaking changes.

And I believe this can also serve as the forcing mechanism for requiring released templates to assert a minimum supported version....while not requiring an arbitrary template to do so until a breaking change has been introduced from Briefcase (such as WiX 4 will do for Windows). That is, if Briefcase asserts a minimum version for the platform, the template must assert one as well.

rmartin16:
All right; this approach seems pretty solid IMO:
- Templates can declare a minimum supported version of Briefcase via `briefcase.minimum_version`
- Platforms in Briefcase can declare a minimum Briefcase version such that a template must declare minimum support for that version or later
- These minimums will be checked prior to builds for any platform

Please let me know if this should be tweaked before I finalize the implementation.

src/briefcase/commands/build.py
```diff
@@ -14,13 +17,46 @@ def add_options(self, parser):
...
 
         :param app: The application to build
         """
         # Default implementation; nothing to build.
 
+    def verify_template_compatability(self, app: AppConfig):
+        """Verify the template satisfies the Command's requirements."""
+        self.logger.info("Verifying template compatability...", prefix=app.app_name)
+
+        with self.input.wait_bar("Verifying Briefcase version support..."):
```
freakboy3742, line 31: Not sure this needs a wait bar - it should be near instantaneous, and in the "normal" case there's nothing actionable from the perspective of the user.

freakboy3742 (commented):
This all looks to be on the right track to me. 

My only comments are in bikeshed territory. In a generated template, "minimum version" is really the version of briefcase that was used to *generate* the template; on a platform, "minimum_version" is the oldest compatible Briefcase version. My inclination would be to use `briefcase.version` in the template, and `oldest_compatible_briefcase` on platform.

freakboy3742:
> > In a generated template, "minimum version" is really the version of briefcase that was used to generate the template
> 
> Are you imagining an additional process that would ensure this is true? For instance, would this setting be incremented each time a new version branch is created in the template repo?

Why is that needed? We already pass in `briefcase.__version__` to the template context... it can be completely templated.

src/briefcase/commands/build.py
rmartin16: It is true it's near instantaneous but I'm considering it more to be communicating the action that is taking place and the `done` shows it succeeded. This isn't too dissimilar from a lot of other uses of the Wait bar (for instance, verifying glibc)...although, in this scenario, I think this would be clearer is additional aspects of the template were being verified than just if Briefcase is compatible.

rmartin16:
> > > In a generated template, "minimum version" is really the version of briefcase that was used to generate the template
> > 
> > 
> > Are you imagining an additional process that would ensure this is true? For instance, would this setting be incremented each time a new version branch is created in the template repo?
> 
> Why is that needed? We already pass in `briefcase.__version__` to the template context... it can be completely templated.

Well, this brings us full circle really. We initially discussed stamping the Briefcase version that's actively rolling out the template in to the `briefcase.toml`...but this creates holes in the verification logic. Namely, it assumes a template being rolled out is compatible with the version of Briefcase rolling it out.....but this is only sure to be true in the default situation where Briefcase is pulling a released template using the version branch based on the version of Briefcase. In other situations, this guarantee doesn't exist....and is why the template must independently assert its Briefcase compatibility. I'm planning to outline all the different workflows to help reveal these subtleties.

freakboy3742:
> > Why is that needed? We already pass in `briefcase.__version__` to the template context... it can be completely templated.
> 
> Well, this brings us full circle really. We initially discussed stamping the Briefcase version that's actively rolling out the template in to the `briefcase.toml`...but this creates holes in the verification logic. Namely, it assumes a template being rolled out is compatible with the version of Briefcase rolling it out.....

I acknowledge that the edge case you describe exists - However, I'm not sure I see it as a major issue.

The case we're particularly concerned about is an old project rolled out using Briefcase 0.3.14, which I try to build it with Briefcase 0.3.15. For this case, the fact that it was rolled out with 0.3.14 and is being re-used is enough to infer that it *worked* with 0.3.14; so the fact that the version encoded in the rolled out template is "dynamic" doesn't matter. We're trying to catch "has anything changed in briefcase since this project was generated".

If I'm rolling out a fresh project a 0.3.15 template branch using Briefcase 0.3.15... it had better work :-) If it doesn't, that's a bug. If it's an official template, that's our problem; if it's a custom template, that's the template developers problem.

If I'm understanding you correctly, you're concerned about the case where I've got a custom template that works with 0.3.14, which will stop working when it's used for 0.3.15. I don't think it's unreasonable to require that custom template maintainers pay attention to changes in Briefcase, and adapt accordingly (provided we flag these changes so they are provided actionable advice). However, I agree that it might be nice to catch predictable errors caused by fundamental shifts in Briefcase (like a Java version update). I'm not sure we can catch this case without going back to the approach of encoding specific tool version requirements in the template. 

There might be a way to simplify this approach though. The reason we're using the Briefcase version is that it's a convenient proxy for "the tool versions that were current as of Briefcase version X". That's all valid, but the Briefcase version changes more often than the tool versions do. This means using the briefcase version isn't a good option in the template when an upper bound is needed, because the value needs to be updated with every Briefcase version.

However, we can introduce a different proxy - let's call it "epoch" - which is a slow-moving proxy for a range of Briefcase versions. Right now, we're in Android Epoch 1 (Java 8 etc). When the Java 17 change lands, we'll be in Epoch 2. Epoch becomes a slow moving version number; the template can encode the epoch, and it won't need to be updated until the next substantial change. Any existing project has an implied Epoch of 1; it effectively means "briefcase version < 0.3.15". 

This way, we don't have to encode (and update) an upper bound in the template; we just need an epoch that we bump when we make a substantial change in a platform's tooling versions.

Does that make more sense?

mhsmith:
If we make the definition "This template should work with Briefcase version X", then I think these questions are fairly easy to answer.

First, I think we should call it something like `briefcase.target_version` rather than "minimum", because the main motivation for this feature is actually to impose a *maximum* Briefcase version on the template, i.e. the last version of Briefcase which will accept it. The fact that it also gives a minimum limit is just a nice bonus.

The target version does have to be hard-coded in the template and not passed in to the template context. Otherwise, for example, if we do introduce another round of breaking changes in Briefcase 0.3.16, you could still tell it to use the template branch `0.3.15` when running `briefcase create`, and the generated project would say it came from Briefcase 0.3.16, thus bypassing the compatibility check.

However, we shouldn't increment the template's target version with every Briefcase version, because that would remove the option of using a older Briefcase version with a newer template version, which might sometimes be useful. We should only increment it if the template has changed in a way which actually requires the new Briefcase version. And when that happens, we would usually increment the "minimum template target version" in Briefcase at the same time.

This would effectively be the same as the epoch number idea, but without the overhead of keeping track of how the epoch numbers map to Briefcase versions.

That way, a given version of Briefcase can safely use both older and newer versions of the template, either in existing build directories or freshly-`create`d ones, as long as their target versions meet the [limits I set out above](https://github.com/beeware/briefcase/pull/1271#issuecomment-1557963948). And the same goes for third-party templates.

freakboy3742:
> If we make the definition "This template should work with Briefcase version X", then I think these questions are fairly easy to answer.

If I'm understanding you correctly, you're effectively advocating for an Epoch-style solution, but the Epoch number is also the literal Briefcase version where the epoch started? The use-in-practice would look something like:

1. Project rolls out today with 0.3.14 templates. No target-version is added to the generated briefcase.toml.
2. Briefcase updates to 0.3.15, with a breaking Java change. The Android template adds a hard coded target-version of 0.3.15.
3. User tries to build the 0.3.14 project; it fails because there's no target-version declared in the project. 
4. User deletes their build, and regenerates with Briefcase 0.3.15. Build succeeds.
5. Briefcase updates to 0.3.16, with no breaking changes. The Android template and backend check don't change.
6. User builds their 0.3.15 project with 0.3.16 - it succeeds, because the backend looks for and finds target-version = 0.3.15 in the template.
7. Briefcase updates to 0.3.17, with some other breaking Android change. The Android template updates it's target-version to 0.3.17, and the gradle backend updates it's check to 0.3.17
8. User builds their 0.3.15 project with 0.3.17; it breaks because the project declares it's in 0.3.15 format.

This sequence also works for a custom template, forked from the 0.3.14 template. It will break at step 3 for the same reason; if the template maintainer updates to be 0.3.15-compatible, they can add the compatibility tag, and the template will work again at step 4 and 6, and break at step 8.

An alternative approach for this user be at step 4 would be: instead of deleting and re-generating, they manually add `target-version=0.3.15` to their *existing* briefcase.toml, and manually resolve any build complications that occur. This may not be easy, but at that point the user gets to keep all the shiny pieces when it breaks.

Have I understood correctly? If so - this seems entirely workable to me, and AFAICT addresses @rmartin16's concerns.

rmartin16:
> Have I understood correctly? If so - this seems entirely workable to me, and AFAICT addresses @rmartin16's concerns.

This is my understanding and is also what's implemented right now.

mhsmith:
> If I'm understanding you correctly, you're effectively advocating for an Epoch-style solution, but the Epoch number is also the literal Briefcase version where the epoch started?

Yes, except that the version number comparisons would still use `>=` rather than `==`. For example, if Briefcase 0.3.17 has a minimum target-version of 0.3.15, it can still accept a template with a target version of 0.3.16 or 0.3.17, because it knows that no template-breaking changes happened in either of those versions.

mhsmith:
Such templates could exist because they require something that was introduced in Briefcase 0.3.16, even though it wasn't a breaking change in itself.

rmartin16:
The implementation for supporting template upgrades in complete.
- If a Briefcase platform enforces a minimum Briefcase version, the rolled out template must declare support for that version or newer.
- If a rolled out template specifies a minimum version, the running Briefcase version must be that version or newer.

I tried to accurately but succinctly describe these situations in the error messages.

I feel like I should document this somewhere....but most of this is only really relevant to a custom template maintainer. Otherwise, "normal" users just run `briefcase create` and move on with life.

Probably a "breaking change" changenote, though?

src/briefcase/commands/build.py
```diff
@@ -14,16 +17,66 @@ def add_options(self, parser):
...
+            if (platform_target and platform_target > template_target) or (
+                template_target and template_target > current_version
+            ):
+                minimum_version = ".".join(map(str, platform_target or current_version))
+                raise BriefcaseCommandError(
+                    f"""\
+The app template is not compatible with this version of Briefcase since it is
+targeting version {'.'.join(map(str, template_target))}.
+
+If you are using BeeWare's default template, then running the create command
```
rmartin16, line 65: this grammar needs to be tweaked...i'll do it later if we're good with the rest of this.
```suggestion
If you are using BeeWare's default template, then run the create command
```

src/briefcase/commands/build.py
```diff
@@ -14,16 +17,66 @@ def add_options(self, parser):
...
 
         :param app: The application to build
         """
         # Default implementation; nothing to build.
 
+    def verify_template(self, app: AppConfig):
+        """Verify the template satisfies the Command's requirements."""
+        self.logger.info("Checking template compatability...", prefix=app.app_name)
+
+        with self.input.wait_bar("Verifying Briefcase support..."):
```
freakboy3742, line 31: Not sure a waitbar is needed here. Verification should be near immediate.

src/briefcase/commands/build.py
```diff
@@ -14,16 +17,66 @@ def add_options(self, parser):
...
+                raise BriefcaseCommandError(
+                    f"""\
+The app template must declare a target version of Briefcase to confirm it is
+compatible with this version of Briefcase.
+
+The template's target Briefcase version must be {'.'.join(map(str, platform_target))} or later.
+
+Once the template is updated, run the create command:
+
+    $ briefcase create {self.platform} {self.output_format}
```
freakboy3742, line 51: Can't verify this renders 100% correctly, but suggested text:
```suggestion
Briefcase requires that the app template explicitly declare that it is compatible with 
Briefcase {'.'.join(map(str, platform_target))} or later. However, the generated app's 
`briefcase.toml` has no `target_version` declaration. 

If the app was generated with an earlier version of Briefcase using the default Briefcase
template, you can run:

     $ briefcase create {self.platform} {self.output_format}

to re-generate your app. 

If you are using a custom template, you'll need to update the template to correct any 
compatibility problems, and then add the compatibility declaration.
```

src/briefcase/commands/build.py
```diff
@@ -14,16 +17,66 @@ def add_options(self, parser):
...
+"""
+                )
+
+            current_version = Version(briefcase.__version__).release
+            if (platform_target and platform_target > template_target) or (
+                template_target and template_target > current_version
+            ):
```
freakboy3742, line 58: This should be an `elif platform_target != template_target`. There's no need for the actual Briefcase version to be involved, and it's not a > comparison - we're looking for a "Briefcase version as epoch" marker. For this change to Java, the template_target will be `0.3.15`, and it will stay that way until we make another backwards incompatible change to the Android template format.

src/briefcase/commands/build.py
```diff
@@ -14,16 +17,66 @@ def add_options(self, parser):
...
+If you are using BeeWare's default template, then running the create command
+to update the app using a compatible version of the app template.
+
+If a custom template is being used, it must be updated to be compatible with
+Briefcase version {minimum_version} before re-running the create command.
+
+To run the create command:
+
+    $ briefcase create {self.platform} {self.output_format}
+"""
```
freakboy3742, line 74:
```suggestion
The app templated used to generate this app is not compatible with this version of Briefcase.
Briefcase requires a template that is compatible with version {platform_target}; the template
used to generate this app is compatible with version {template_target}.

If the app was generated with an earlier version of Briefcase using the default Briefcase
template, you can run:

     $ briefcase create {self.platform} {self.output_format}
     
to re-generate your app. 

If you are using a custom template, you'll need to update the template to correct any 
compatibility problems, and then add the compatibility declaration.
```

src/briefcase/commands/build.py
```diff
@@ -14,16 +17,66 @@ def add_options(self, parser):
...
 
-    def build_app(self, app: BaseConfig, **options):
+    def build_app(self, app: AppConfig, **options):
         """Build an application.
 
         :param app: The application to build
         """
         # Default implementation; nothing to build.
 
+    def verify_template(self, app: AppConfig):
```
freakboy3742, line 27: This should probably be in the base command, and called as part of a more generic check (`finalize?`). If you've got an app that was generated for 0.3.14, and you try to run it without building, you're going to have the same class of problem - it's not exclusively a *build* problem.

freakboy3742 (changes requested):
A couple of comments inline - the only major issue is the version check, which I don't think is correct - the Briefcase version shouldn't factor into the check. Either there's an exact match with the platform's "epoch" version, or there's no match.

I think the specific location of the check might also need tweaking, for the reason flagged inline.

src/briefcase/commands/build.py
rmartin16: IMO, this wait bar here is mostly intended to provide better continuity with the rest of the output.

While I consider the Wait Bar to be mostly about communicating a task is ongoing, I think the consistency of `[task description]... done` helps users understand what exactly is taking place in a consistent format. There's several small steps like this we're already wrapping with the Wait Bar elsewhere.

I suppose the end state of the Wait bar output could be manually printed...but that feels a little awkward.

Still want to remove/change this?

Example output:
```
Using app template: /home/russell/github/beeware/briefcase-android-gradle-template, branch v0.3.15

[helloworld] Installing support package...
No support package required.

[helloworld] Installing application code...
Installing src/helloworld... done

[helloworld] Installing requirements...
Writing requirements file... done

[helloworld] Installing application resources...
Unable to find src/helloworld/resources/helloworld-round-48.png for 48px round application icon; using default
Unable to find src/helloworld/resources/helloworld-round-72.png for 72px round application icon; using default
Unable to find src/helloworld/resources/helloworld-round-96.png for 96px round application icon; using default
Unable to find src/helloworld/resources/helloworld-round-144.png for 144px round application icon; using default
Unable to find src/helloworld/resources/helloworld-round-192.png for 192px round application icon; using default
Unable to find src/helloworld/resources/helloworld-square-48.png for 48px square application icon; using default
Unable to find src/helloworld/resources/helloworld-square-72.png for 72px square application icon; using default
Unable to find src/helloworld/resources/helloworld-square-96.png for 96px square application icon; using default
Unable to find src/helloworld/resources/helloworld-square-144.png for 144px square application icon; using default
Unable to find src/helloworld/resources/helloworld-square-192.png for 192px square application icon; using default

[helloworld] Removing unneeded app content...
Removing directory app/src/main/python/helloworld/__pycache__
Removing unneeded app bundle content... done

[helloworld] Created build/helloworld/android/gradle

[helloworld] Checking template compatibility...
Verifying Briefcase support... done

[helloworld] Updating app metadata...
Setting main module... done

[helloworld] Building Android APK...
Starting a Gradle Daemon, 1 incompatible Daemon could not be reused, use --status for details
```

src/briefcase/commands/build.py
rmartin16: Yeah...I initially tried to put it there but ran in to issue. I'll try to move it up again.

src/briefcase/commands/build.py
freakboy3742: I see what you're saying; I guess for me, "is this a valid template" is a fast validation check that should almost always pass. It's not "doing" anything - it's just verifying that everything else will work. 99% of the time, it won't be communicating anything - and the 1% of the time that it does, there will be a big "it won't work" error message. 

Some of the other "short lived" tasks (e.g., updating app metadata) are actually *doing* something - they're part of the build process, writing/modifiying files, etc. I can't think of another task that is (a) short lived and (b) doesn't write/add/delete anything to the disk, with the exception of no-op tasks that are flagged has doing something, but then the platform backend doesn't need to actually do anything. 

The closest I can think of is some of the Docker checks - but those (a) *can* take time, and (b) can fail in some creative ways, so flagging that you're in the middle of a Docker check can be useful debugging.

rmartin16: Ok...so, do you want this to print anything then? I feel like just printing the section header `[helloworld] Checking template compatibility...` with nothing else is also awkward.

That makes sense if so.

freakboy3742: Oh sorry - I thought that part was implied. My intention was to say that there should be *no* message when verifying the template version, unless there's a problem. Apologies for the ambiguity on my part.

src/briefcase/commands/build.py
mhsmith: I already discussed this [here](https://github.com/beeware/briefcase/pull/1271#issuecomment-1563470051) and [here](https://github.com/beeware/briefcase/pull/1271#issuecomment-1557963948).

rmartin16: this has definitely been a sticking point...i guess i won't say too much more except these approaches are effectively the same...with slightly different implementations.

I like the idea of an epoch....but I think using the Briefcase version is confusing....because people will think you're literally talking about the current version of Briefcase.

I also like the idea of supporting a version inside the template that's independent of the platform version support declaration in Briefcase. This allows third party templates to specify an arbitrary level of support with Briefcase. If we limit this to Briefcase's internal platform version, then we're limiting the number of potential breaking changes third party can protect themselves from....which could be missed breaking changes or even just nuances among versions.

At this point, I'm starting to feel we may just need an executive decision....if only because I'm a bit exhausted going round and round. I'm prepared to implement the epoch versioning or even something else. (I'm still working on promoting the check to all Commands since that's pretty impactful to the tests.)

freakboy3742: In which case, I'm going to put on my BDFN hat and say "epoch-style numbering, using the Briefcase version as the epoch marker". 

Yes, there's a source of confusion in that a 0.3.16 template will have marker saying it is for 0.3.15. However, as this discussion has revealed, *any* discussion of compatibility and version numbers is going to be confusing; picking an identifier that has a clear interpretation (vs a new "epoch" number that will need it's own explanation) is about as good as we're going to get, IMHO. 

It's also not especially user-visible - it's a value in `briefcase.toml`, which isn't something that run-of-the-mill users will ever need to see. This is a "confusing detail" that will mostly affect Briefcase maintainers, not the outside world.

Using the 0.3.15 version is also implicit documentation of which release notes you need to look at for a migration guide.

I acknowledge that there *could* be a need for a third-party template provider to lock against a different Briefcase version. However, we don't have any particular evidence that this requirement exists - it's mostly speculation and trying to foretell the future. On that basis, I'm comfortable living with the limitation until such time as we can clearly identify a specific use-case. *If* such a requirement were to develop, I can imagine adding an additional template option (or command line option) to ignore template version checks, or impose a *different* version check; however, I think we should cross that bridge when (and if) we come to it.

rmartin16:
I re-implemented the template verification using a target version epoch.
- Verifying app template for _all_ command
  - Moved app template verification in to `base.py`
  - This verification couldn't be done in `finalize()` since it runs before pre-requisite command run....therefore, a `build` command would call finalize before it called a pre-requisite `create` command and that isn't going to work
- Reoriented the configuration from the word `version` to the word `epoch`
  - Given the inherent confusion around all this...it seemed like using the word epoch at least avoids the easy confusion with just treating this config as a normal Briefcase version
- Simplified the error reporting in to a single error message
  - The two different error messages based on whether the app template contains a `target_epoch` or not weren't very different to being with....so, I combined the error conditions
  - I removed references to Briefcase versions (and especially saying something like "Briefcase 0.3.15 or later") since the actual version of Briefcase isn't _really_ relevant....just that they aren't the same
  - And plus....unless you already understand the epoch thing going on....showing a version seems more likely to just be confusing
- Added some typing
  - I did _not_ go full bore on this like I did `integrations`....I just fixed the most problematic typing (like using `BaseConfig` since practically no attributes are defined for it) and added future annotations
  - That said, I can type everything if you want...

rmartin16:
hmm....when verifying the test coverage is actually meaningful, I realized none of the Grade platform tests were exercising the target epoch logic because I put the declaration in the Gradle mixin. It should be on the actual Commands.

So, question: should we add the epoch declaration to all Gradle commands? or just the Gradle Build command since I think that's the only one actually dependent on JDK 17?

mhsmith:
> * Reoriented the configuration from the word `version` to the word `epoch`
>   * Given the inherent confusion around all this...it seemed like using the word epoch at least avoids the easy confusion with just treating this config as a normal Briefcase version

I'd prefer to keep it as `target_version`, so we won't need to rename it if we ever give it the more complex interpretation discussed before. 

From the Briefcase user's point of view, removing the version number from the error message should already be enough to prevent confusion. 

And as for custom template authors, they should be using the official templates as a guide anyway, so that should lead them to pick the correct value for this setting. And I don't think renaming the setting would make them any less likely to change it.

rmartin16:
> > * Reoriented the configuration from the word `version` to the word `epoch`
> >   
> >   * Given the inherent confusion around all this...it seemed like using the word epoch at least avoids the easy confusion with just treating this config as a normal Briefcase version
> 
> I'd prefer to keep it as `target_version`, so we won't need to rename it if we ever give it the more complex interpretation discussed before.

That's a fair point; I'll rename it.

> in the least, a different reason must be preventing the tests from failing as I'm expecting...

I also found the source of my missing expectations for testing. Currently, this app template compatibility check only runs if the template has been rolled out in the file system. (This is only required for the `dev` command since all other commands will have a rolled out template (except `upgrade` but it doesn't do app verification anyway).)

During testing, we don't always mock the app in the file system...and I don't think we should enforce that it is. Instead, we just populate members on the command that effectively mock the app.

So, I'll just re-architect the gate for template compatibility checking on whether template configuration/metadata exists for an app regardless of the source.

freakboy3742:
> I also found the source of my missing expectations for testing. Currently, this app template compatibility check only runs if the template has been rolled out in the file system. (This is only required for the `dev` command since all other commands will have a rolled out template (except `upgrade` but it doesn't do app verification anyway).)

Is it required for dev? What's the failure mode that requires a check on dev? As you say, you can run dev without having an app template at all - and this (AIUI) is entirely an issue with generated app templates.

>   * That said, I can type everything if you want...

I think I'd prefer to keep to "1 improvement per PR" as a general philosophy. Add types to new code, or code that is directly impacted by this change if it helps, but "I'm in base.py so I'll update all the types" only serves to make the PR harder to review, because the reviewer has to separate typing improvements from the core functionality that has changed.

rmartin16:
> > I also found the source of my missing expectations for testing. Currently, this app template compatibility check only runs if the template has been rolled out in the file system. (This is only required for the `dev` command since all other commands will have a rolled out template (except `upgrade` but it doesn't do app verification anyway).)
> 
> Is it required for dev? What's the failure mode that requires a check on dev? As you say, you can run dev without having an app template at all - and this (AIUI) is entirely an issue with generated app templates.

To be clear, I was saying that supporting a way to skip this template version check is only required for the `dev` command because it can run without a rolled out template. I didn't want to skip the check on `command = "dev"` or similar....since it should be possible to just discover if a template is rolled out and check it. Although, now it's problematic because `dev` command doesn't have an `output_format` specified...

freakboy3742:
> To be clear, I was saying that supporting a way to skip this template version check is only required for the `dev` command because it can run without a rolled out template. 

Ack - that makes more sense :-)

rmartin16:
I added explicit tests for Gradle to verify a template's compatibility. This was a bit less straightforward than I expected because verifying the template happens after any prerequisite commands have run....but before you actually get in to the Gradle-specific code. So, I had to mock that specific steps already occurred and skip some steps entirely since mocking them can be a little arduous.

I also tweaked skipping the template compatibility check; it is now skipped if a new dedicated error is raised for a missing `briefcase.toml` or if a command raises `NotImplementedError` suggesting that something about inspecting a template is not applicable to this Command. I went round and round with this and may have gotten tunnel vision....so fresh eyes may see better alternatives that I'd be open to hearing.

mhsmith:
Looks reasonable to me, but I'll let @freakboy3742 do the final review because he's more familiar with the code.

src/briefcase/commands/base.py
```diff
@@ -324,104 +337,102 @@ def binary_path(self, app):
...
-        return self._path_index[app]
+            return self._briefcase_toml[app]
+        except KeyError:
+            try:
+                with (self.bundle_path(app) / "briefcase.toml").open("rb") as f:
+                    self._briefcase_toml[app] = tomllib.load(f)
+            except OSError as e:
+                raise MissingAppMetadata(self.bundle_path(app)) from e
+            else:
+                return self._briefcase_toml[app]
```
rmartin16, line 356: FYI...I realized something about this coding style while testing this.

When Rich prints the `MissingAppMetadata` exception to the log file, it doesn't actually roll back up stack of exceptions in to the true source of the exception (that is, whatever code is trying to read `briefcase.toml`). If, however, you raise the `MissingAppMetadata` from the `KeyError` exception, then Rich includes the full stack in the exception print out.

I'm not sure if I think that's a good way to mitigate this problem....but it's something to thing about for this coding style.

freakboy3742 (approved):
Looks good, and works well in my testing.

I've made a couple of minor tweaks prior to landing:
1. Minor tweak to the language when downloading the new JDK
2. Removed the custom config classes - we weren't using them, and based on where we are now, I'm having difficulty imaging how we *would* use them
3. Testing was broken on ARM64 hardware, due to a couple of tests that were hard coding the Linux host OS, but not the host architecture. This wasn't picked up because we don't (and can't) do any CI testing on ARM.

PR #1272, opened 2023-05-14
Move default Flatpak runtime to `pyproject.toml` and upgrade to Freedesktop 22.08

rmartin16:
## Changes
- Updates the default Flatpak runtime from Freedesktop 21.08 -> 22.08
- Additionally, this default runtime will now be set in `pyproject.toml` at project creation
  - https://github.com/beeware/briefcase-template/pull/68

## PR Checklist:


freakboy3742 (commented):
Is there a particular motivation for this change? I'm hesitant to make the change here because it's a system default; it's the value that will be used in the absence of an explicit value in the user's pyproject.toml. Changing the constant here means that we'll be changing the behavior of any app that had the runtime version undefined; it would be akin to changing the Ubuntu base version on the AppImage template. Any app that isn't Toga or PySide will be in this category.

rmartin16:
That makes sense; however, 21.08 goes EOL in August. Furthermore, each version is only supported for 2 years anyway; so, what should be the update strategy given this? Maybe this should be defaulted in `pyproject.toml`?

freakboy3742:
> That makes sense; however, 21.08 goes EOL in August. Furthermore, each version is only supported for 2 years anyway; so, what should be the update strategy given this? Maybe this should be defaulted in `pyproject.toml`?

We should definitely be including a default value in `pyproject.toml`. 

Given that the EOL lifespan is so short, I'm inclined to suggest that the default for an unspecified flakpak image/version specifier should be an error. If the default project template provides an appropriate default, we can update it regularly as new versions are published; and any existing projects will raise an error on upgrade - but the migration path should be obvious from the error message. 

It might also be worthwhile taking this opportunity to modify our towncrier configuration to allow for explicit backwards compatibility notes - that way we can make the known migration issues explicit.

mhsmith:
> It might also be worthwhile taking this opportunity to modify our towncrier configuration to allow for explicit backwards compatibility notes - that way we can make the known migration issues explicit.

There's already a "removal" change note type, which displays as "Deprecations and Removals" by default, but can be renamed [like this](https://github.com/chaquo/chaquopy/blob/14.0.2/pyproject.toml).

rmartin16:
The Flatpak runtime is defaulted in `briefcase-template` now and Briefcase will error if a runtime is not specified in `pyproject.toml`. Feel free to tweak the error messages; tried to succinctly tell the user what's wrong.

As for the removal note, let me know if you want to tweak anything about towncrier.

freakboy3742 (approved):
Looks good - I've made a minor tweak to the error message, and added a towncrier config change to introduce the "Backwards incompatible changes" section.

PR #1281, opened 2023-05-21
Auto-update `pre-commit` hooks

brutusthebee:
Updated `pre-commit` hooks to their latest versions and ran updated hooks against the repo.


freakboy3742:
Closing on the basis that Docformatter 1.7.1 has a few too many bugs in it.

PR #1286, opened 2023-05-22
Remove timestamp and PID from Android log format

mhsmith:
This information is rarely helpful, and the added line length makes the log difficult to follow even in a moderately wide terminal.

## PR Checklist:

### DIFF
--- a/changes/1286.bugfix.rst
+++ b/changes/1286.bugfix.rst
@@ -1,0 +1,1 @@
+Android logs no longer include timestamp and PID, making them easier to read on narrow screens.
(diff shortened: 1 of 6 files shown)


freakboy3742 (approved):
A possible follow up that occurred to me: using a `-v` switch to control the amount of logging information that is displayed. For most day-to-day Toga apps, you really only care about `I/python.stdout`. Controlling the log filter to only show those lines and filter the tag prefix would give a very clean day-to-day log output; but if you turn on `-v`, you get all the other log lines plus the log prefix (and -vv turns on *all* the logs, including the ones that we're currently supressing).

PR #1290, opened 2023-05-25
Fix up support for Python 3.12

rmartin16:
## Changes
- Enable new tarfile extraction protections in Python 3.12
  - https://docs.python.org/3.12/whatsnew/3.12.html#other-language-changes
  - This applies the new ["data" filter](https://docs.python.org/3.12/library/tarfile.html#tarfile.data_filter) to unpacking tar files
  - This prevents the unpacked files from using absolute filepaths or symlinks from referencing absolute filepaths or filepaths outside of the target directory
  - Additionally, the permissions are sanitized to basically ensure they are sane; e.g. world execute permissions are removed if the owner doesn't have execute permissions.

## Notes
- `DeprecationWarning` from pytest for `ast` constants
  - https://github.com/pytest-dev/pytest/issues/10977
- `DeprecationWarning` from dateutil for `datetime.utcfromtimestamp()` usage
  - https://github.com/dateutil/dateutil/issues/1284

## PR Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ no-cover-if-is-macos = "'darwin' == os_environ.get('COVERAGE_PLATFORM', sys_plat
 no-cover-if-not-macos = "'darwin' != os_environ.get('COVERAGE_PLATFORM', sys_platform) and os_environ.get('COVERAGE_EXCLUDE_PLATFORM') != 'disable'"
 no-cover-if-is-windows = "'win32' == os_environ.get('COVERAGE_PLATFORM', sys_platform) and os_environ.get('COVERAGE_EXCLUDE_PLATFORM') != 'disable'"
 no-cover-if-not-windows = "'win32' != os_environ.get('COVERAGE_PLATFORM', sys_platform) and os_environ.get('COVERAGE_EXCLUDE_PLATFORM') != 'disable'"
+no-cover-if-lt-py312 = "sys_version_info < (3, 12) and os_environ.get('COVERAGE_EXCLUDE_PYTHON_VERSION') != 'disable'"
 no-cover-if-is-py310 = "python_version == '3.10' and os_environ.get('COVERAGE_EXCLUDE_PYTHON_VERSION') != 'disable'"
 no-cover-if-lt-py310 = "sys_version_info < (3, 10) and os_environ.get('COVERAGE_EXCLUDE_PYTHON_VERSION') != 'disable'"
 no-cover-if-gte-py310 = "sys_version_info > (3, 10) and os_environ.get('COVERAGE_EXCLUDE_PYTHON_VERSION') != 'disable'"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ no-cover-if-is-macos = "'darwin' == os_environ.get('COVERAGE_PLATFORM', sys_plat
 no-cover-if-not-macos = "'darwin' != os_environ.get('COVERAGE_PLATFORM', sys_platform) and os_environ.get('COVERAGE_EXCLUDE_PLATFORM') != 'disable'"
 no-cover-if-is-windows = "'win32' == os_environ.get('COVERAGE_PLATFORM', sys_platform) and os_environ.get('COVERAGE_EXCLUDE_PLATFORM') != 'disable'"
 no-cover-if-not-windows = "'win32' != os_environ.get('COVERAGE_PLATFORM', sys_platform) and os_environ.get('COVERAGE_EXCLUDE_PLATFORM') != 'disable'"
+no-cover-if-is-py312 = "python_version == '3.12' and os_environ.get('COVERAGE_EXCLUDE_PYTHON_VERSION') != 'disable'"
 no-cover-if-lt-py312 = "sys_version_info < (3, 12) and os_environ.get('COVERAGE_EXCLUDE_PYTHON_VERSION') != 'disable'"
 no-cover-if-is-py310 = "python_version == '3.10' and os_environ.get('COVERAGE_EXCLUDE_PYTHON_VERSION') != 'disable'"
 no-cover-if-lt-py310 = "sys_version_info < (3, 10) and os_environ.get('COVERAGE_EXCLUDE_PYTHON_VERSION') != 'disable'"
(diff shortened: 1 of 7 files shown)


rmartin16:
interesting....`setuptools` is no longer [installed](https://discuss.python.org/t/python-3-12-beta-1-and-feature-freeze-is-here/26982/3?u=rmartin16) in to virtual envs by default...fyi

rmartin16:
Test coverage still fails for Python 3.12. My guess is `coverage` hasn't been updated to account for ast changes in Python 3.12.

pyproject.toml
```diff
@@ -51,6 +52,10 @@ multi_line_output = 3
 testpaths = ["tests"]
 filterwarnings = [
     "error",
+    # Work around https://github.com/pytest-dev/pytest/issues/10977 for Python 3.12
+    'ignore:(ast\.Str|ast\.NameConstant|ast\.Num|Attribute s) is deprecated and will be removed.*:DeprecationWarning:',
```
freakboy3742, line 56: I'd rather *not* do this sort of masking for a pre-release; if we get closer to 3.12 final and it hasn't been fixed, then sure - but our 3.12 support is mostly for "canary" testing at this point (to catch issues like the tarball filter argument). I don't think we need to introduce workarounds for problems that *should* be fixed before 3.12 final.

rmartin16: That's makes sense; I was kinda leery of it since there isn't a good way to track when it can come back out.

For at least the `pytest` error, I'm expecting them to fix it pretty quickly and have a new release available...not sure about the `dateutil` one.

I mostly just want GitHub to add proper support for `allow-failure` so CI doesn't look like it failed because testing failed on a non-critical platform...:(

rmartin16: Furthermore, though, these changes won't get full coverage until the pytest `DeprecationWarning` is fixed.

rmartin16:
Pytest support for 3.12 is blocked on the 3.12.0b2 release.
- https://github.com/pytest-dev/pytest/pull/10894

rmartin16:
- Waiting on [new release](https://github.com/dateutil/dateutil/releases) for `python-dateutil`
- The missing [coverage](https://github.com/beeware/briefcase/actions/runs/5231611790/jobs/9445865866?pr=1290) for Python 3.12 seems to be [covered](https://github.com/beeware/briefcase/blob/96479a0cab4814ea5ed359d2431628d145e86698/tests/platforms/macOS/app/test_package__notarize.py#L484)...

rmartin16:
Sooo....traction with PRs in `python-dateutil` have been lethargic and Irit hasn't moved on python/cpython#105658.

Given that and the looming 3.12 release, I'm inclined to push a commit with configuration to suppress these errors.

Let me know if you'd rather go a different direction. Alternatively, we could merge this now and just include such mitigations with the updates to CI for 3.12.

freakboy3742:
> Sooo....traction with PRs in `python-dateutil` have been lethargic and Irit hasn't moved on [python/cpython#105658](https://github.com/python/cpython/issues/105658).
> 
> Given that and the looming 3.12 release, I'm inclined to push a commit with configuration to suppress these errors.
> 
> Let me know if you'd rather go a different direction. Alternatively, we could merge this now and just include such mitigations with the updates to CI for 3.12.

The coverage issue doesn't bother me that much - I'd be happy merging something to ignore that one on the basis that the line *is* covered (and the overall coverage report supports that), with an open ticket to revert the ignore once the underlying issue is resolved.

The dateutil issue is more of a concern though. Is that actually an *error*, or just a very loud warning that is being raised as an error by pytest for some reason? It's not clear to me why a DeprecationWarning is presenting as an error. If it *is* a hard error, I don't think we can ignore it; but if it's an ignorable warning and cookiecutter still *works*, then I'd be happy with an explicit ignore (again with a ticket to revisit once datetutil is fixed).

rmartin16:
> The dateutil issue is more of a concern though. Is that actually an _error_, or just a very loud warning that is being raised as an error by pytest for some reason? It's not clear to me why a DeprecationWarning is presenting as an error. If it _is_ a hard error, I don't think we can ignore it; but if it's an ignorable warning and cookiecutter still _works_, then I'd be happy with an explicit ignore (again with a ticket to revisit once datetutil is fixed).

We updated the `pytest` configuration a while back to report all warnings as errors.

```toml
[tool.pytest.ini_options]
testpaths = ["tests"]
filterwarnings = [
    "error",
]
```

Nonetheless, the `dateutil` error is just about `datetime.utcfromtimestamp()` being deprecating; it won't be an actual problem in 3.12.

Additionally, the transitive dependency is through `arrow` which currently includes the dependency `python-dateutil >= 2.7.0`....so, we'll automatically get the update once released.

```
briefcase==0.3.16.dev133+g77cf5df2.d20230829
├── cookiecutter [required: >=2.2,<3.0, installed: 2.3.0]
│   ├── arrow [required: Any, installed: 1.2.3]
│   │   └── python-dateutil [required: >=2.7.0, installed: 2.8.2]
```

freakboy3742:
> > The dateutil issue is more of a concern though. Is that actually an _error_, or just a very loud warning that is being raised as an error by pytest for some reason? It's not clear to me why a DeprecationWarning is presenting as an error. If it _is_ a hard error, I don't think we can ignore it; but if it's an ignorable warning and cookiecutter still _works_, then I'd be happy with an explicit ignore (again with a ticket to revisit once datetutil is fixed).
> 
> We updated the `pytest` configuration a while back to report all warnings as errors.
> 
> ```toml
> [tool.pytest.ini_options]
> testpaths = ["tests"]
> filterwarnings = [
>     "error",
> ]
> ```
> 
> Nonetheless, the `dateutil` error is just about `datetime.utcfromtimestamp()` being deprecating; it won't be an actual problem in 3.12.
> 
> Additionally, the transitive dependency is through `arrow` which currently includes the dependency `python-dateutil >= 2.7.0`....so, we'll automatically get the update once released.
> 
> ```
> briefcase==0.3.16.dev133+g77cf5df2.d20230829
> ├── cookiecutter [required: >=2.2,<3.0, installed: 2.3.0]
> │   ├── arrow [required: Any, installed: 1.2.3]
> │   │   └── python-dateutil [required: >=2.7.0, installed: 2.8.2]
> ```

Ah - that makes sense. In which case, I'm happy with smothering the warnings for now, with some open housekeeping tickets to clean up the code once the problems have been resolved upstream.

rmartin16:
image

freakboy3742 (approved):
/me sheds at tear at the beauty... :-)

Issue #463, opened 2020-08-01
Add support for ANDROID_HOME environment variable


ProgrammAbel:
**Describe the bug**
After creating a project, building for Android and accepting the license, I had to delete it (because of #462). Then, when creating a new, clean project and building it for Android again, it did not prompt me to accept a license but instead gave me an error that the licenses weren't accepted.

**To Reproduce**
1. Create a project
2. Build for Android and accept the license
3. Delete the folder of this project
4. Create another project
5. Build for Android
6. See error

**Expected behavior**
For the application to build successfully.

**Screenshots**
![Screenshots2020 08 01-19 05 43 screenshot](https://user-images.githubusercontent.com/51785314/89107607-1e486f80-d42a-11ea-865e-143848775b3a.png)

**Environment:**
 - Operating System: Linux 5.7.10-arch1-1
 - Python version: Python 3.8.4
 - Software versions:
   - Briefcase: 0.3.3
   - Toga: 0.3.0

praveen-manohar:
Hi @ProgrammAbel i got the same error once, you have to accept the license on android sdkmanager(not on project run), once you done that, you're error will be clear now.

freakboy3742:
Thanks for the report. The licenses are accepted on a per-computer basis, so the interaction with multiple projects is coincidental. 

I assume one of two things happened:
* The licenses weren't fully accepted on the first execution.
* The second execution introduced *new* licenses that needed to be accepted.

The license check at present is very simplistic - it looks for the existence of `~.briefcase/tools/android_sdk/licenses/android-sdk-license`. If that file exists, it assumes the licenses have been accepted. However, there are a lot of other license files in that folder (7 on my install); so it's possible we might need to improve that check. 

If you're still seeing this error, I'd be interested in knowing what license files (if any) currently exist in that folder (and, if there *is* an `android-sdk-license` file, what it contains). Hopefully, we only need to expand the license check to look for another file (or files). Worse case, we need to modify the license check to actually invoke licenses every time (or add a command flag to explicitly perform a license check).

ProgrammAbel:
[...]

PR #1306, opened 2023-06-10
No warning if JDK referred by macOS is invalid

### LINKED ISSUE #1305: Confusing message about JAVA_HOME
### Describe the bug

I got this message on macOS:
```
** WARNING: JAVA_HOME does not point to a Java 17 JDK                  **

    Android requires a Java 17 JDK, but the location pointed to by the
    JAVA_HOME environment variable:

    /Library/Java/JavaVirtualMachines/amazon-corretto-8.jdk/Contents/Home

    isn't a Java 17 JDK (it appears to be Java 1.8.0_352).

    Briefcase will use its own JDK instance.
```

This didn't make any sense, because I didn't have any JAVA_HOME environment variable. It turns out it was actually getting the path by running /usr/libexec/java_home.

### Steps to reproduce

* Install a Java version other than 17, and make sure its location is returned by /usr/libexec/java_home
* Remove or rename Briefcase's existing Java 17 installation, if any
* Unset JAVA_HOME
* Run `briefcase build android`

### Expected behavior

The message should correctly indicate where the path came from.

### Screenshots

_No response_

### Environment

- Operating System: macOS
- Python version: 3.8
- Software versions:
  - Briefcase: 96479a0cab4814ea5ed359d2431628d145e86698

### Logs

_No response_

### Additional context

_No response_

rmartin16:
## Changes
- The user is no longer warned about any issues detected with the JDK referred by macOS via `/usr/libexec/java_home`
- Fixes #1305

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/java.py
+++ b/src/briefcase/integrations/java.py
@@ -121,7 +121,6 @@ class JDK(ManagedTool):
     Briefcase will use its own JDK instance.
 
 *************************************************************************
-
 """
 
             except OSError:
@@ -139,7 +138,6 @@ class JDK(ManagedTool):
     Briefcase will use its own JDK instance.
 
 *************************************************************************
-
 """
 
             except subprocess.CalledProcessError:
@@ -166,7 +164,6 @@ class JDK(ManagedTool):
     from the command prompt.
 
 *************************************************************************
-
 """
 
             except IndexError:
@@ -193,7 +190,6 @@ class JDK(ManagedTool):
     from the command prompt.
 
 *************************************************************************
-
 """
 
         # macOS has a helpful system utility to determine JAVA_HOME. Try it.
@@ -215,8 +211,11 @@ class JDK(ManagedTool):
                     pass  # do not alert user if macOS found an unqualified JDK
 
         if java is None:
-            # If we've reached this point, any user-provided JAVA_HOME is broken;
-            # use the Briefcase one.
+            # Inform the user if the user-specified JDK wasn't valid
+            if install_message:
+                tools.logger.warning(install_message)
+
+            # Use the Briefcase JDK install
             java_home = tools.base_path / cls.JDK_INSTALL_DIR_NAME
 
             # The macOS download has a weird layout (inherited from the official Oracle
@@ -228,10 +227,6 @@ class JDK(ManagedTool):
 
             if not java.exists():
                 if install:
-                    # We only display the warning messages on the pass where we actually
-                    # install the JDK.
-                    if install_message:
-                        tools.logger.warning(install_message)
                     tools.logger.info(
                         f"A Java {cls.JDK_MAJOR_VER} JDK was not found; downloading and installing...",
                         prefix=cls.name,
--- a/src/briefcase/integrations/java.py
+++ b/src/briefcase/integrations/java.py
@@ -52,8 +52,8 @@ class JDK(ManagedTool):
         )
 
     @classmethod
-    def jdk_version_from_path(cls, tools: ToolCache, java_path: str | Path) -> bool:
-        """Return JDK version from a path by running ``<java_path>/bin/javac``.
+    def version_from_path(cls, tools: ToolCache, java_path: str | Path) -> str:
+        """Return a JDK's version from a path by running ``<java_path>/bin/javac``.
 
         This will fail if the path contains a JRE instead of a JDK.
 
@@ -61,6 +61,10 @@ class JDK(ManagedTool):
          - OSError - the ``javac`` executable doesn't exist
          - CalledProcessError - ``javac`` returned a non-zero value
          - IndexError - unparsable version value from ``javac``
+
+        :param tools: ToolCache of available tools
+        :param java_path: File path to a candidate JDK install
+        :return: JDK release version; e.g. "17.0.7"
         """
         output = tools.subprocess.check_output(
             [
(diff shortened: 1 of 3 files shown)


src/briefcase/integrations/java.py
```diff
@@ -51,6 +51,26 @@ def OpenJDK_download_url(self):
             f"{self.JDK_RELEASE}_{self.JDK_BUILD}.{extension}"
         )
 
+    @classmethod
+    def jdk_version_from_path(cls, tools: ToolCache, java_path: str | Path) -> bool:
```
mhsmith, line 55:
```suggestion
    def jdk_version_from_path(cls, tools: ToolCache, java_path: str | Path) -> str:
```

src/briefcase/integrations/java.py
```diff
@@ -74,38 +94,14 @@ def verify_install(cls, tools: ToolCache, install: bool = True, **kwargs) -> JDK
...
-                output = tools.subprocess.check_output(
-                    [
-                        os.fsdecode(Path(java_home) / "bin" / "javac"),
-                        "-version",
-                    ],
-                )
-                # This should be a string of the form "javac 17.0.7\n"
-                version_str = output.strip("\n").split(" ")[1]
+                version_str = cls.jdk_version_from_path(tools, java_home)
                 if version_str == cls.JDK_RELEASE:
```
mhsmith, line 102: This isn't testing the major version, but the exact version number. It looks like this was already incorrect, so we might as well fix it in this PR.

Extracting the major version isn't as simple as just taking the first number, because version 8 and earlier had a leading "1", e.g. "javac 1.8.0_242".

src/briefcase/integrations/java.py
```diff
@@ -196,6 +192,23 @@ def verify_install(cls, tools: ToolCache, install: bool = True, **kwargs) -> JDK
...
+                # If /usr/libexec/java_home doesn't exist, OSError will be raised
+                # If no JRE/JDK is installed, /usr/libexec/java_home raises an error
+                java_home = tools.subprocess.check_output(
+                    ["/usr/libexec/java_home"],
+                ).strip("\n")
+            except (OSError, subprocess.CalledProcessError):
+                pass  # No java on this machine
+            else:
+                try:
+                    if cls.jdk_version_from_path(tools, java_home) == cls.JDK_RELEASE:
```
mhsmith, line 207: Same issue here.

mhsmith:
Good call on skipping the warning altogether in this case. Unlike the environment variable, merely having a copy of Java installed doesn't mean the user is asking us to use it.

src/briefcase/integrations/java.py
rmartin16: Just verifying major version definitely makes sense to me. Right now, we're requiring users have the exact version of Java that Briefcase is using.

However, do we need to support parsing the versioning of Version 8? Since we're looking for `17`, this should just ignore any Version 8 installs of Java. Or am I missing something perhaps?

rmartin16:
Another thing is we're not _always_ informing the user when Briefcase ignores their `JAVA_HOME` setting. As pointed out [before](https://github.com/beeware/.github/pull/40#issuecomment-1575450350), if Briefcase rejects the JDK install at `JAVA_HOME` but a valid JDK already exists in Briefcase's data directory, Briefcase silently uses its data directory JDK install.

If `JAVA_HOME` is set but Briefcase doesn't use it, I think we should always inform users.

The downside to this, i suppose, is users will now be informed every time they run Briefcase that `JAVA_HOME` isn't valid instead of just the first time like it does now.

src/briefcase/integrations/java.py
mhsmith: I was imagining an error message saying "you have Java version 1". But it's better to keep the error message the way it is, with the full version number, because that's more helpful for debugging. In which case, you're right: the "1" prefix doesn't matter.

mhsmith:
Given that `JAVA_HOME` should be mentioned in the documentation (https://github.com/beeware/briefcase/issues/463#issuecomment-1217891624), I agree it's better to give a warning every time we detect an environment variable but don't use it.

rmartin16:
> Given that `JAVA_HOME` should be mentioned in the documentation ([#463 (comment)](https://github.com/beeware/briefcase/issues/463#issuecomment-1217891624)), I agree it's better to give a warning every time we detect an environment variable but don't use it.

Good deal; I'll update the documentation when I fix `ANDROID_SDK_ROOT` next.

src/briefcase/integrations/java.py
```diff
@@ -51,6 +51,30 @@ def OpenJDK_download_url(self):
...
+        :param java_path: File path to a candidate JDK install
+        :return: JDK release version; e.g. "17.0.7"
+        """
+        output = tools.subprocess.check_output(
+            [
+                os.fsdecode(Path(java_path) / "bin" / "javac"),
+                "-version",
+            ],
+        )
+        # This should be a string of the form "javac 17.0.7\n"
```
freakboy3742, line 75: This comment is misleading (/stale) - the return value doesn't include `javac `.

freakboy3742 (commented):
One minor nitpick in a stale comment, but otherwise looks good to me. I'll leave the final approval to Malcolm, but I'm happy with this.

src/briefcase/integrations/java.py
rmartin16: Are you seeing this somewhere in particular? I'm getting `javac 17.0.7` as the output.

freakboy3742: Possibly confusing placement/wording then. I read "this" as applying to return value, not the input. The input definitely has `javac `; but the return value is just 17.0.7.

mhsmith (approved):

PR #1307, opened 2023-06-11
Add support to specify an Android SDK in `ANDROID_HOME` (fixes #463)

### LINKED ISSUE #463: Add support for ANDROID_HOME environment variable

rmartin16:
## Changes
- Fixes #463
- The file path in `ANDROID_HOME` now takes precedence over `ANDROID_SDK_ROOT`
- The eventual file path for the SDK used is passed both as `ANDROID_HOME` and `ANDROID_SDK_ROOT` when Android commands are run in the shell.
- Behavior summary:
  - No environment variables set
    - Briefcase's SDK is used
  - Only `ANDROID_HOME` is set
    - If it's an SDK, it is silently used
    - If it's not an SDK, a warning is issued it isn't an SDK and Briefcase's SDK is used
  - Only `ANDROID_SDK_ROOT` is set
    - If it's an SDK, a warning is issued that `ANDROID_SDK_ROOT` is deprecated but it is used
    - If it's not an SDK, a warning is issued it isn't an SDK and Briefcase's SDK is used
  - Both `ANDROID_HOME` and `ANDROID_SDK_ROOT` are set to the same value
    - Falls back to as if only `ANDROID_HOME` was set
  - Both `ANDROID_HOME` and `ANDROID_SDK_ROOT` are set to _different_ values
    - A warning is issued that `ANDROID_HOME` takes precedence over `ANDROID_SDK_ROOT`
    - If `ANDROID_HOME` is a valid SDK, it is used
    - If `ANDROID_HOME` is _not_ a valid SDK, a warning is issued it isn't an SDK and Briefcase's SDK is used

## Notes
- I broke with the [recommended implementation](https://github.com/beeware/briefcase/issues/463#issuecomment-1217891624) because I think it can imposition the user too much.
- For instance, erroring if `ANDROID_HOME` and `ANDROID_SDK_ROOT` are set to different file paths.
  - This isn't inherently problematic for Briefcase...at least based on my implementation.
  - However, erroring like this would _require_ users to change their environment just to run Briefcase.
- As part of this, I am also setting the SDK file path in both `ANDROID_HOME` and `ANDROID_SDK_ROOT` for subprocess calls
  - This is primarily because if the user has a bad setting for `ANDROID_SDK_ROOT` and Briefcase only sets `ANDROID_HOME`, the `ANDROID_SDK_ROOT` setting will leak in to subprocess calls.....therefore, the user would need to unset `ANDROID_SDK_ROOT` to be able to run Briefcase.
  - So, it seems reasonable to me for Briefcase to simply mask this scenario since we already have a known valid SDK.

## PR Checklist:

### DIFF
--- a/docs/reference/platforms/android.rst
+++ b/docs/reference/platforms/android.rst
@@ -4,6 +4,21 @@ Android
 
 When generating an Android project, Briefcase produces a Gradle project.
 
+Gradle requires an installation of the Android SDK and a Java JDK.
+
+An existing installation of the Android SDK will be used by Briefcase if its
+file path is specified in the ``ANDROID_HOME`` environment variable.
+The environment variable ``ANDROID_SDK_ROOT`` can also specify the file path to
+the SDK but it is considered deprecated by Android.
+
+An existing Java JDK will be used by Briefcase if its file path is specified in
+the ``JAVA_HOME`` environment variable. Briefcase requires the JDK to be
+version 17. On macOS, Briefcase will use the ``/usr/libexec/java_home`` tool to
+actively find an existing JDK installation for use if ``JAVA_HOME`` is not set.
+
+If the above methods fail to find an Android SDK or Java JDK, Briefcase will
+install an isolated copy in its data directory.
+
 Icon format
 ===========
 
--- a/docs/reference/platforms/android.rst
+++ b/docs/reference/platforms/android.rst
@@ -8,8 +8,8 @@ Gradle requires an install of the Android SDK and a Java 17 JDK.
 
 If you have an existing install of the Android SDK, it will be used by Briefcase
 if the ``ANDROID_HOME`` environment variable is set. If ``ANDROID_HOME`` is not
-present in the environment, Briefcase will honor the ``ANDROID_SDK_ROOT``
-environment variable (the use of variable has been deprecated by Android).
+present in the environment, Briefcase will honor the deprecated
+``ANDROID_SDK_ROOT`` environment variable.
 
 If you have an existing install of a Java 17 JDK, it will be used by Briefcase
 if the ``JAVA_HOME`` environment variable is set. On macOS, if ``JAVA_HOME`` is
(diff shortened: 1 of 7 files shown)


freakboy3742 (commented):
I've pushed some minor language tweaks, but otherwise this looks good.

In terms of the specific behavior you've implemented - those changes both makes sense to me. Setting both environment variables removes all doubt over the environment that is used at runtime; and warning the user about a setting inconsistency but recovering acts as a canary that something isn't right while not getting in the way of progress. 

I'll leave the final approval to Malcolm in case I've missed something in my analysis, but as far as I'm concerned, I'm happy.

docs/reference/platforms/android.rst
```diff
@@ -4,6 +4,21 @@ Android
 
 When generating an Android project, Briefcase produces a Gradle project.
 
+Gradle requires an install of the Android SDK and a Java 17 JDK.
+
+If you have an existing install of the Android SDK, it will be used by Briefcase
+if the ``ANDROID_HOME`` environment variable is set. If ``ANDROID_HOME`` is not
+present in the environment, Briefcase will honor the ``ANDROID_SDK_ROOT``
+environment variable (the use of variable has been deprecated by Android).
```
mhsmith, line 12:
```suggestion
present in the environment, Briefcase will honor the deprecated
``ANDROID_SDK_ROOT`` environment variable.
```

mhsmith (approved):

PR #1356, opened 2023-07-11
Use Date/Time on Android Device for Fallback Logging

### LINKED ISSUE #1146: Android crash logs are hidden if phone and computer times aren't in sync
### Describe the bug

If an Android app is well behaved, Briefcase will start the app, detect the PID of the running app, then display the logs for that PID.

If an app crashes on startup, there may not be enough time for Briefcase to detect the PID; in this case, Briefcase moves into a fallback mode, and prints the last 10 seconds of logs from *any* PID, with a filter that captures important Toga/Chaquopy log sections.

Briefcase establishes this "10 second" window by using the local system time. However, this timestamp is passed to `adb` as a log filter (`-t`), and it is the *phone's* clock that is used to evaluate the log filter.

So - if the computer and the phone disagree on the current time, it is possible for the app logs to appear empty, even though there are plenty of logs - just outside the "10 second" window.

### Steps to reproduce

1. Generate a "hello world" app
2. Put `raise Exception()` at the top of `app.py`
3. Deliberately change your computer's clock so that it is 1 minute faster than your phone
4. Run the app on Android.

It may be necessary to use a physical phone; an emulator will be using the host system clock.

### Expected behavior

Logs for the crash should be generated in Briefcase.

### Screenshots

_No response_

### Environment

- Operating System: Any
- Python version: Any
- Software versions:
  - Briefcase: 0.3.13+

### Logs

```

```

### Additional context

One fix is to use the *phone's* perception of time to establish the start of the logging window. We currently use `datetime.now()` to get the start time prior to launching the app; we could instead use `adb shell date`, and parse the result. This would ensure that the time filter passed to  `adb logcat` will be using a common reference time.

rmartin16:
## Changes
- Fixes #1146
- When the PID for the app on an Android device is not found, the fallback app streaming method will use a datetime from the device instead of the host machine as the start time to show logs.

## Notes
- I'm operating under the assumption that Android's `date` command has supported this formatting for as long as we've needed it to.
- I'm also currently aborting if obtaining the device datetime goes wrong since such a contingency seems really unlikely to be necessary when things are working otherwise.

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1517,9 +1517,15 @@ Activity class not found while starting app.
             raise BriefcaseCommandError("Error stopping the Android emulator.") from e
 
     def datetime(self) -> datetime:
-        """Obtain the device's current date/time."""
+        """Obtain the device's current date/time.
+
+        This date/time is naive (i.e. not timezone aware) and in the device's "local"
+        time. Therefore, it may be quite different from the date/time for Briefcase and
+        caution should be used if comparing it to machine's "local" time.
+        """
+        datetime_format = "%Y-%m-%d %H:%M:%S"
         try:
-            # request datetime in seconds since Unix epoch
-            return datetime.fromtimestamp(int(self.run("shell", "date", "+%s")))
+            device_datetime = self.run("shell", "date", f"+'{datetime_format}'").strip()
+            return datetime.strptime(device_datetime, datetime_format)
         except (ValueError, subprocess.CalledProcessError) as e:
             raise BriefcaseCommandError("Error obtaining device date/time.") from e
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1293,10 +1293,10 @@ class ADB:
                     f"Unable to interrogate AVD name of device {self.device}"
                 ) from e
 
-    def has_booted(self):
+    def has_booted(self) -> bool:
         """Determine if the device has completed booting.
 
-        :returns True if it has booted; False otherwise.
+        :returns: True if it has booted; False otherwise.
         """
         try:
             # When the sys.boot_completed property of the device
@@ -1326,8 +1326,6 @@ class ADB:
         # checking that they are valid, then parsing output to notice errors.
         # This keeps performance good in the success case.
         try:
-            # Capture `stderr` so that if the process exits with failure, the
-            # stderr data is in `e.output`.
             return self.tools.subprocess.check_output(
                 [
                     os.fsdecode(self.tools.android_sdk.adb_path),
@@ -1349,8 +1347,7 @@ class ADB:
         """Install an APK file on an Android device.
 
         :param apk_path: The path of the Android APK file to install.
-
-        Returns `None` on success; raises an exception on failure.
+        :returns: `None` on success; raises an exception on failure.
         """
         try:
             self.run("install", "-r", apk_path)
@@ -1363,8 +1360,7 @@ class ADB:
         """Force-stop an app, specified as a package name.
 
         :param package: The name of the Android package, e.g., com.username.myapp.
-
-        Returns `None` on success; raises an exception on failure.
+        :returns: `None` on success; raises an exception on failure.
         """
         # In my testing, `force-stop` exits with status code 0 (success) so long
         # as you pass a package name, even if the package does not exist, or the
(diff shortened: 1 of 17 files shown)


mhsmith:
> * I'm operating under the assumption that Android's `date` command has supported `+%s` formatting for as long as we've needed it to.

The Briefcase Android template currently has a minimum API level of 24, so if you can confirm it works on an emulator with that version, that should be good enough.

rmartin16:
> > * I'm operating under the assumption that Android's `date` command has supported `+%s` formatting for as long as we've needed it to.
> 
> The Briefcase Android template currently has a minimum API level of 24, so if you can confirm it works on an emulator with that version, that should be good enough.

Thanks. I changed `AndroidSDK.DEFAULT_SYSTEM_IMAGE` to return `system-images;android-24;default;x86_64` instead of `system-images;android-31;default;x86_64` and the date is returned correctly for an emulator using that system image. I believe this effectively tests API 24 as the oldest supported Android.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1519,4 +1514,12 @@ def kill(self):
...
             self.run("emu", "kill")
         except subprocess.CalledProcessError as e:
-            raise BriefcaseCommandError("Error starting ADB logcat.") from e
+            raise BriefcaseCommandError("Error stopping the Android emulator.") from e
+
+    def datetime(self) -> datetime:
+        """Obtain the device's current date/time."""
+        try:
+            # request datetime in seconds since Unix epoch
+            return datetime.fromtimestamp(int(self.run("shell", "date", "+%s")))
```
freakboy3742, line 1523: Just confirming: is this using UTC epoch or local timezone epoch?

freakboy3742 (changes requested):
Looks good for the most part; one question about timezone, since it wasn't immediately obvious; it might not be revealed in testing because you're running locally UTC negative will work in your favor for testing purposes.

src/briefcase/integrations/android_sdk.py
rmartin16: hmm...so, if `datetime.fromtimestamp()` isn't given a timezone, then it returns a naive date/time for the platform's date/time configuration.

In practice, this assumes the device and the host machine are configured for the timezone....and daylight saving...and whatever else....

rmartin16: I suppose this requirement could be avoided if a specific format is requested from `date` on the device and an equivalent format is specified for `datetime.strptime()`

freakboy3742: Look, we can all agree that time was invented by the devil to punish programmers. :-)

We could definitely go to great lengths to ensure timezone compliant round trips - but I'm not sure that's actually required. We just need to make sure that the time that comes back from `adb shell date` is in the same timezone as the log file filter. If that's true with `fromtimestamp()`, then I'm 99% happy; the remaining 1% is to document the new `datetime()` endpoint to make sure nobody tries to use it for anything else in future.

rmartin16: > Look, we can all agree that time was invented by the devil to punish programmers. :-)

If that isn't the truth....

> We could definitely go to great lengths to ensure timezone compliant round trips - but I'm not sure that's actually required. We just need to make sure that the time that comes back from `adb shell date` is in the same timezone as the log file filter. If that's true with `fromtimestamp()`, then I'm 99% happy; the remaining 1% is to document the new `datetime()` endpoint to make sure nobody tries to use it for anything else in future.

Well...I guess as long as 99% of the time users' devices have date/time configurations that match their machines that are running Briefcase....then we're good here :/

This is why I had to commit https://github.com/beeware/briefcase/pull/1356/commits/c2e4b6b2c3c222b1a35eedbbb19e0ddb98fc87bf after pushing; I discovered that GitHub's CI runners are configured for UTC (as opposed to my EDT) and were returning a different date/time for the same value returned from `date` on the device.

I could play around with explicit date/time formats tomorrow to see what this takes.

rmartin16: [ed9c234](https://github.com/beeware/briefcase/pull/1356/commits/ed9c234f0e1a8ac06237a44d1aad41f03be5524c) will ensure this logging date/time is local to the device.

mhsmith (approved):
I can't reliably reproduce my original problem now, and I suspect it may actually have been an intermittent issue with it failing to find the PID even when the app is running. But anyway, this all looks fine to me.

rmartin16:
fwiw...I was only able to _not_ find a PID when I commented out `pid = adb.pidof(package, quiet=True)`...even when I invoked an `ImportError`, it always found the PID from my main machine.

Though, on a sufficiently slow machine (like my Pi 4b), every once in a while, it doesn't find a PID in 5 seconds and then runs `logcat_tail()` before the app even crashes. I also even noticed once that Briefcase "forcibly killed gradle" when trying to exit the output streamer. I highly suspect my Pi is not cpu limited in these cases....but disk limited since it's all running on a tiny micro sd card. Probably not anything that deserves changes right now....but some interesting data points from a slower machine.

freakboy3742 (approved):
Looks good to me. The failure mode you describe on RPi/slow machines is definitely a problem, but it's sufficiently edge case that I can definitely live with it for now. The only immediately obvious fix I can think of would be to increase the 5s window to something more, but 5s is already inconvenient for people with fast machines, and there's essentially no delay that will be long enough for a sufficiently slow machine. 

If we can find a different marker that allows us to differentiate between "The app has crashed" and "the app hasn't started... yet", that would be great - but in the meantime, this will be a significant improvement for the 95% case of app developers who have an app failing on startup and a device with a clock discrepancy.

PR #1372, opened 2023-07-22
Updates after dropping Python 3.6 and 3.7

hugovk:
Fix some things left over from the drop of EOL Python 3.6 and 3.7:

```python
# TODO: Py3.6 compatibility; os.fsdecode not required in Py3.7
```

Replace `os.fsdecode()` with `str()`, because they can be pathlib Paths.

Also need things like `archive.__str__.return_value = "/path/to/download.zip"` for the MagicMock objects.

---

```python
    # Consider to remove if block when we drop py3.7 support, only keep statements from else.
    # MagicMock below py3.8 doesn't have __fspath__ attribute.
```

The if/else blocks (they used to look like https://github.com/beeware/briefcase/commit/05c50b999e90d37ed8ebd67543c33c008bd91215) have already been cleaned up (presumably by pyupgrade), so remove these old comments.

## PR Checklist:

- [n/a] All new features have been documented

### DIFF
--- a/changes/1372.misc.rst
+++ b/changes/1372.misc.rst
@@ -1,0 +1,1 @@
+Updates after dropping Python 3.6 and 3.7.
--- a/src/briefcase/integrations/java.py
+++ b/src/briefcase/integrations/java.py
@@ -262,10 +262,9 @@ class JDK(ManagedTool):
 
         with self.tools.input.wait_bar("Installing OpenJDK..."):
             try:
-                # TODO: Py3.6 compatibility; os.fsdecode not required in Py3.7
                 self.tools.shutil.unpack_archive(
-                    os.fsdecode(jdk_zip_path),
-                    extract_dir=os.fsdecode(self.tools.base_path),
+                    str(jdk_zip_path),
+                    extract_dir=str(self.tools.base_path),
                 )
             except (shutil.ReadError, EOFError) as e:
                 raise BriefcaseCommandError(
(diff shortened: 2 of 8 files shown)


freakboy3742 (changes requested):
The cleanups of Py3.6/3.7 detritus are definitely required and appreciated; however, the usage of `os.fsdecode` is deliberate. This is based on advice from Brett Cannon, (provided in many places; [this is one example](https://discuss.python.org/t/make-pathlib-extensible/3428/79)), my understanding is that `os.fsdecode()` is preferred for rendering Pathlib objects, in particular because of `None` handling. 

If you've got a different understanding, let us know what I've missed (or misunderstood).

hugovk:
Ah right, I was mainly going on the comments, so will re-instate the `os.fsdecode` calls.

And keep the removal of the `# TODO: Py3.6 compatibility; os.fsdecode not required in Py3.7` comments?

freakboy3742:
> And keep the removal of the `# TODO: Py3.6 compatibility; os.fsdecode not required in Py3.7` comments?

Those comments can all be burned :-) You guessed correctly that they were all left in because pyuprade removed an if/else branch, but didn't pick up the comment explaining the branch.

hugovk:
Pushed! Let me know if you'd prefer a fixup/rebase.

freakboy3742 (approved):
Awesome - thanks for the 🧹 !

PR #1375, opened 2023-07-22
Disable `pip` warnings/checks during `requires` install

rmartin16:
## Changes
- Disables some unnecessary (IMO) checks and feedback during app `requires` installation; specifically: 
  - `pip` version checking
  - `pip` Python version compatibility

These are all spurious and unnecessary while building the app. I left them for `briefcase dev`, though, since Briefcase is literally installing in to the user's environment.

## PR Checklist:


rmartin16:
hmm....seems `pip` on `debian:bullseye` is too old for `--root-user-action`....

Although, that brings up the fact we're using old versions of `pip` in the containers. It may be best to have the Dockerfiles update `pip`...

Briefcase already requires at least `pip >= 23.1.1` and `--root-user-action` was introduced in `22.1`.

freakboy3742 (approved):
👍

PR #1376, opened 2023-07-22
Use wheels in `tox` instead of sdists

rmartin16:
## Changes
- Inspired by Hynek's recent [post](https://hynek.me/articles/turbo-charge-tox/), update `tox` to install Briefcase as a wheel instead of an sdist; it's much faster (although, in fairness, it was already pretty fast).

## PR Checklist:

### DIFF
--- a/changes/1376.misc.rst
+++ b/changes/1376.misc.rst
@@ -1,0 +1,1 @@
+When running tests or building docs in tox, Briefcase is now installed as a wheel instead of an sdist.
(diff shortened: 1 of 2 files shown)


freakboy3742 (commented):
This seems to make sense - but to confirm my own understanding: this won't give us any improvement in CI, because we're already using --installpkg with a pre-built wheel. This is entirely an optimization for end-users running tox locally, right? 

(To be clear - that's totally worth doing - I just want to make sure I haven't misunderstood what this is doing)

rmartin16:
> this won't give us any improvement in CI, because we're already using --installpkg with a pre-built wheel

That's correct.

freakboy3742 (approved):

PR #1387, opened 2023-07-28
Cache image prior to checking docker user mapping

### LINKED ISSUE #1368: Better user feedback when Linux system package target is not available in the Docker registry
### Describe the bug

If a user targets a platform for which a suitable Docker image is not available, the user feedback doesn't really communicate that.

```console
$ briefcase package --target fedora:37
Unable to find image 'fedora:37' locally
37: Pulling from library/fedora
docker: no matching manifest for linux/arm/v7 in the manifest list entries.
See 'docker run --help'.

Unable to determine if Docker is mapping users
```

While the above example is trying to pull an image for an architecture it doesn't provide, a simple misspelling creates the same situation.

```console
❯ briefcase package --target ubnutu
Unable to find image 'ubnutu:latest' locally
docker: Error response from daemon: pull access denied for ubnutu, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.

Unable to determine if Docker is mapping users
```

### Steps to reproduce

Run `briefcase package --target ubnutu`

### Expected behavior

I know the output explicitly says "unable to find image" but the "Unable to determine if Docker is mapping users" is in red and at least causes initial, if not entire, misdirection to the real problem.

Docker has different exit codes for when `docker run` fails versus when something _inside_ a container goes wrong.

Therefore, I think we can catch and interpret these [exit codes](https://docs.docker.com/engine/reference/run/#exit-status) differently:
```
    125: docker run itself fails
    126: contained command cannot be invoked
    127: if contained command cannot be found
    128 + n Fatal error signal n:
        130 = (128+2) Container terminated by Control-C
        137 = (128+9) Container received a SIGKILL
        143 = (128+15) Container received a SIGTERM
```

In this way, we could detect `125` for any `docker run` call and output a more generic message about the target image.

### Screenshots

_No response_

### Environment

- Python version: 3.10.12
- Software versions:
  - Briefcase: `0.3.16.dev19+g30ee4a80`

### Logs

[briefcase.2023_07_21-11_31_11.package.log](https://github.com/beeware/briefcase/files/12130279/briefcase.2023_07_21-11_31_11.package.log)

### Additional context

_No response_

rmartin16:
## Changes
- Closes #1368
- Simply ensuring the image is cached beforehand prints out a much better error message.
- While the docker return code checking may prove useful in the future, it feels like overkill right now.

## PR Checklist:

### DIFF
--- a/changes/1368.bugfix.rst
+++ b/changes/1368.bugfix.rst
@@ -1,0 +1,1 @@
+When the target Docker image name is not valid when building Linux System packages, the error message now mentions the invalid image name instead of an error for Docker user mapping.
--- a/changes/1368.bugfix.rst
+++ b/changes/1368.bugfix.rst
@@ -1,1 +1,1 @@
-When the target Docker image name is not valid when building Linux System packages, the error message now mentions the invalid image name instead of an error for Docker user mapping.
+Error reporting has been improved when the target Docker image name is invalid.
--- a/src/briefcase/integrations/docker.py
+++ b/src/briefcase/integrations/docker.py
@@ -271,13 +271,18 @@ See https://docs.docker.com/go/buildx/ to install the buildx plugin.
             "/host_write_test", host_write_test_path.name
         )
 
+        image_tag = "alpine" if image_tag is None else image_tag
+        # Cache the image first so the attempts below to run the image don't
+        # log irrelevant errors when the image may just have a simple typo
+        self.cache_image(image_tag)
+
         docker_run_cmd = [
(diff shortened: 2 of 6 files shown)


freakboy3742 (approved):
Looks like a good tweak to me; I've made one minor update adding some cosmetic logging. Turning off log streaming for the download is a good improvement, but it means the first line of output is Docker download details, rather than Briefcase telling you what it is doing.

PR #1390, opened 2023-07-29
Mount data directory at `/briefcase` in Docker containers

### LINKED ISSUE #1374: Briefcase data directory is still being mounted in to containers for `brutus` user when `root` is used
### Describe the bug

With #1331, the user in Docker containers can now be `root` as well as `brutus`; however, Briefcase always mounts the data directory in to `.cache` directory for `brutus`. This doesn't appear to be having adverse effects right now....but it should probably be reconciled.

### Steps to reproduce

1. Using Docker Desktop, run `briefcase open linux --target debian`
2. Run `whoami` to see current user
3. Run `ls -lh /home/brutus/.cache/`

### Expected behavior

I suppose it doesn't _actually_ hurt anything to leave this as is....but it seems a lot more consistent if we use a user-agnostic location to bind mount the Briefcase data directory given we're not always using the `brutus` user now.

### Screenshots

_No response_

### Environment

- Python version: 3.10.12
- Software versions:
  - Briefcase: `0.3.16.dev20+g8b020d6c`

### Logs

_No response_

### Additional context

_No response_

rmartin16:
## Changes
- Since the user in containers can be `root` now, the data directory is mounted to the user-agnostic `/briefcase`
- Closes #1374 

## PR Checklist:


src/briefcase/integrations/docker.py
```diff
@@ -504,8 +504,8 @@ def _dockerize_path(self, arg: str) -> str:  # pragma: no-cover-if-is-windows
         """
         if arg == sys.executable:
             return f"python{self.python_version}"
-        arg = arg.replace(os.fsdecode(self.host_bundle_path), "/app")
-        arg = arg.replace(os.fsdecode(self.host_data_path), str(self.docker_data_path))
```
rmartin16, line 508: I know we've touched on `os.fsdecode()` before but I wanted to make sure we have the same expectations for its use.

As stated before, it is safest to use `os.fsdecode()` to cast an arbitrary object to a string for use as a file path since some objects that have a `__str__()` will never be valid file paths. (It also manages special escaping for its inverse function `os.fsencode()`.)

When using `os.fsdecode()` on an object known to be a `Path`, it is really just a long way of calling `str()`. That is, `fsdecode(obj)` calls `fspath(obj)` calls `type(obj).__fspath__(obj)` which is just `str(self)` for `Path`.

So...do we want to use `os.fsdecode()` on objects that are known to be an instance of `Path`? And is my understanding correct?

freakboy3742: If we're absolutely 100% certain it's a `Path`, and won't ever be `None`, then yes - `str()` is safe. However, using `os.fsdecode()` regardless is one of those safety mechanisms by practice - if we use it even when we *strictly* don't have to, we're more likely to remember we have to do it when it *does* matter.

freakboy3742 (changes requested):
This makes sense as an approach, and the mechanics here make sense; however, don't we also need to make a change so that `BRIEFCASE_HOME` points to `/briefcase`? Otherwise we're going to be downloading a fresh set of tools and templates on every build. Previously, BRIEFCASE_HOME was inherently correct by virtue of being the user's home directory, but if that's not the case, we need to point Briefcase elsewhere.

src/briefcase/integrations/docker.py
rmartin16: Yeah; that's mostly my thought as well - it's safer to just use `fsdecode()`....although, there's so many exceptions to this rule currently....namely all the f-strings implicitly converting Paths to strings. ehh...it is what it is, though.

rmartin16:
> This makes sense as an approach, and the mechanics here make sense; however, don't we also need to make a change so that `BRIEFCASE_HOME` points to `/briefcase`?

I don't think so. Briefcase never actually runs inside the container.....so, nothing running in the container is inherently respecting `BRIEFCASE_HOME`. That said, if processes inside the container need to know where Briefcase's data directory is, they'll need to be informed directly. For instance, AppImage builds update `PATH` so linuxdeploy can find its plugins.

src/briefcase/integrations/docker.py
rmartin16: Since I git amended, I'll clarify here that I put the `os.fsdecode()` calls back.

freakboy3742:
> > This makes sense as an approach, and the mechanics here make sense; however, don't we also need to make a change so that `BRIEFCASE_HOME` points to `/briefcase`?
> 
> I don't think so. Briefcase never actually runs inside the container.....so, nothing running in the container is inherently respecting `BRIEFCASE_HOME`. 

🤦 Of course - The tools are being found by the path mapping process.

freakboy3742 (approved):

PR #1392, opened 2023-07-30
Add support to target non-native architectures

### LINKED ISSUE #1128: Linux / System / Add Support for Targeting Arbitrary Machine Architectures

Building a linux system package is currently limited to the architecture of the machine building the package. So, an `x86_64` machine can only build packages for `x86_64`. A likely desired scenario is building for `aarch64/arm64` on `x86_64`.

ref: https://github.com/beeware/briefcase/pull/1106#issuecomment-1457435428

The linux system package build process already supports targeting arbitrary distros with Docker; we should extend this to support arbitrary architectures via the `buildx` Docker builder.

https://docs.docker.com/build/building/multi-platform/

QEMU supports running `aarch64` OSes on `x86_64`.....just be prepared to demonstrate patience :)

### Additional context

_No response_

rmartin16:
## Changes
- Adds support for AppImage and Linux System builds to target non-native architectures
- Closed #1128

## Notes
- A while back, we merged #1153
- These changes now leverage `buildx` to create images for arbitrary machine architectures; using these images, the app can be built for a different arch as though it was that of the host machine
- However, as I've been learning, this is not without some ambiguity....so, while building/using multi-platform images is relatively straight-forward for a single host machine....supporting varied Docker installs may be a bit trial/error
- Support for multi-platform images
  - Docker Desktop ships with support for running images for any architecture by virtue of its LinuxKit VM
  - Docker Engine needs this support enabled
    - See current arch support: `docker run --privileged --rm tonistiigi/binfmt`
    - Add support for arch: `docker run --privileged --rm tonistiigi/binfmt --install arm64`
    - https://hub.docker.com/r/tonistiigi/binfmt
    - TODO: confirm this is recommended approach for a user's daily driver
    - TODO: does this even work with rootless Docker?
- Building for the correct arch
  - Docker seems to keep a memory of the last platform used to run a particular container
    - This means it will likely be necessary to _always_ specify the platform to use when running `docker`....which means Briefcase will need to be able to derive the appropriate default platform for a user's machine when one isn't provided.
    - Ideally, some Docker command can provide this information....since its more complicated than it should be
      - Maybe `docker system info --format '{{.OSType}}/{{.Architecture}}'`
      - But maybe it's not always accurate...: https://github.com/BretFisher/multi-platform-docker-build
    - Furthermore, this kinda pollutes the user's environment...definitely best if this could be avoided somehow...
  - Additionally, the `build` directory will need to be differentiated for the target architecture
    - TODO: include the architecture in the `bundle_path`

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/docker.py
+++ b/src/briefcase/integrations/docker.py
@@ -111,7 +111,12 @@ See https://docs.docker.com/go/buildx/ to install the buildx plugin.
         "Linux": "https://docs.docker.com/engine/install/#server",
     }
 
-    def __init__(self, tools: ToolCache, image_tag: str | None = None):
+    def __init__(
+        self,
+        tools: ToolCache,
+        platform: str | None = None,
+        image_tag: str | None = None,
+    ):
         """A wrapper for the user-installed Docker.
 
         :param tools: ToolCache of available tools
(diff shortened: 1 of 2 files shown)


rmartin16:
This is worthwhile but I haven't had time in a while and probably won't soon. Closing.

PR #1413, opened 2023-08-21
Move command-line args out of dev-mode Python invocation

jeamland:
For some reason NSApplication/CoreFoundation gets very confused if we pass certain command-line arguments to the outer Python interpreter when running in dev mode.

This patch removes all command-line arguments from the invocation, using environment variables to achieve the equivalent outcomes. The remaining argument, the actual code to start the application, appears to not cause problems.

### DIFF
--- a/changes/1413.bugfix.rst
+++ b/changes/1413.bugfix.rst
@@ -1,0 +1,1 @@
+The command line arguments used to configure the Python environment for ``briefcase dev`` no longer leak into the runtime environment on macOS.
--- a/src/briefcase/commands/dev.py
+++ b/src/briefcase/commands/dev.py
@@ -18,6 +18,25 @@ class DevCommand(RunAppMixin, BaseCommand):
     output_format = None
     description = "Run a Briefcase project in the dev environment."
 
+    # On macOS CoreFoundation/NSApplication will do its own independent parsing of argc/argv.
+    # This means that whatever we pass to the Python interpreter on start-up will also be
+    # visible to NSApplication which will interpret things like `-u` (used to make I/O
+    # unbuffered in CPython) as `-u [URL]` (a request to open a document by URL). This is,
+    # rather patently, Not Good.
+    # To avoid this causing unwanted hilarity, we use environment variables to configure the
+    # Python interpreter rather than command-line options.
+    DEV_ENVIRONMENT = {
+        # Equivalent of passing "-u"
+        "PYTHONUNBUFFERED": "1",
+        # Equivalent of passing "-X dev"
+        "PYTHONMALLOC": "debug",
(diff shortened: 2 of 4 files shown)


changes/1413.bugfix.rst
```diff
@@ -0,0 +1 @@
+Moved most command-line args out of the Python invocation used inside ``briefcase dev``.
```
freakboy3742, line 1:
```suggestion
The command line arguments used to configure the Python environment for ``briefcase dev`` no longer leak into the runtime environment on macOS.
```

src/briefcase/commands/dev.py
```diff
@@ -18,6 +18,18 @@ class DevCommand(RunAppMixin, BaseCommand):
     output_format = None
     description = "Run a Briefcase project in the dev environment."
 
+    DEV_ENVIRONMENT = {
```
freakboy3742, line 21:
```suggestion
    # On macOS, a document-based app reads the raw argc/argv that is used to start the
    # process. This means we can't use any command line arguments when invoking the
    # python executable, because those will leak into the running app as "documents"
    # that are to be opened on startup. These arguments *can* be passed in as command
    # line arguments, but are configured using environment variables.
    DEV_ENVIRONMENT = {
```

freakboy3742 (changes requested):
2 minor suggestions listed inline, but otherwise looks good.

freakboy3742 (approved):
👍

src/briefcase/commands/dev.py
```diff
@@ -18,6 +18,25 @@ class DevCommand(RunAppMixin, BaseCommand):
...
+    # To avoid this causing unwanted hilarity, we use environment variables to configure the
+    # Python interpreter rather than command-line options.
+    DEV_ENVIRONMENT = {
+        # Equivalent of passing "-u"
+        "PYTHONUNBUFFERED": "1",
+        # Equivalent of passing "-X dev"
+        "PYTHONMALLOC": "debug",
+        "PYTHONASYNCIODEBUG": "1",
+        "PYTHONFAULTHANDLER": "1",
+        "PYTHONWARNINGS": "default",
```
rmartin16, line 35: Curious....was there a reason that `PYTHONDEVMODE=1` wasn't used instead?

Setting just [PYTHONDEVMODE](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONDEVMODE) seems more likely to be forward-compatible with enabling this mode.

freakboy3742: ...we didn't realise the setting existed :-)

It's obvious now that I'm looking at the docs; no objection replacing these 4 settings with the single PYTHONDEVMODE=1

PR #1425, opened 2023-08-29
Improve subprocess stdout encoding management

### LINKED ISSUE #1407: Windows console output encoding corruption
### Describe the bug

Something is getting lost in translation when Briefcase streams `subprocess` calls.

For instance, if a user adds `print("Année")` to their app, then `briefcase dev` and `briefcase run` will ostensibly attempt to print `utf-8` characters to the console and you end up with something like `Ann├⌐e`. Since `é` is a two-byte char in UTF-8, it gets translated to `├` and `⌐` in cp437.

### Steps to reproduce

Add `print("Année")` to a Briefcase app and run `briefcase dev`.

### Expected behavior

Encodings should be managed properly.

This seems to be an outcome from Briefcase output streaming where Briefcase attaches a pipe to the subprocess' `stdout`, reads it in, and writes it to the console via Rich. One of these steps must be making an error about the encoding to use.

Interesting, if you just use Rich directly, it will say the console is UTF-8 while it is actually a msft code page...but still print the text correctly:
```console
❯ python -c 'from rich.console import Console; console=Console(); console.print("Année"); print(console.encoding)'
Année
utf-8

❯ python -c 'import sys, os; print(os.device_encoding(sys.__stdout__.fileno()))'
cp437
```

### Screenshots

_No response_

### Environment

- Operating System: Windows 11
- Python version: python 3.10.11
- Software versions:
  - Briefcase: `0.3.16.dev103+g48fc4baa.d20230820`

### Logs

<details>
<summary>Briefcase Log</summary>

```
Date/Time:       2023-08-19 18:48:32 
Command line:    C:\Users\user\github\beeware\briefcase\venv-3.10-briefcase\Scripts\briefcase dev --log

OS Release:      Windows 10
OS Version:      10.0.22621
Architecture:    AMD64
Platform:        Windows-10-10.0.22621-SP0

Python exe:      C:\Users\user\github\beeware\briefcase\venv-3.10-briefcase\Scripts\python.exe
Python version:  3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)]
Virtual env:     True
Conda env:       False

Briefcase:       0.3.16.dev103+g48fc4baa
Target platform: windows
Target format:   None

Environment Variables:
    ALLUSERSPROFILE=C:\ProgramData
    APPDATA=C:\Users\user\AppData\Roaming
    COMMONPROGRAMFILES=C:\Program Files\Common Files
    COMMONPROGRAMFILES(X86)=C:\Program Files (x86)\Common Files
    COMMONPROGRAMW6432=C:\Program Files\Common Files
    COMPUTERNAME=VM-WIN11
    COMSPEC=C:\Windows\system32\cmd.exe
    DRIVERDATA=C:\Windows\System32\Drivers\DriverData
    FPS_BROWSER_APP_PROFILE_STRING=Internet Explorer
    FPS_BROWSER_USER_PROFILE_STRING=Default
    HOMEDRIVE=C:
    HOMEPATH=\Users\user
    LOCALAPPDATA=C:\Users\user\AppData\Local
    LOGONSERVER=\\VM-WIN11
    NUMBER_OF_PROCESSORS=4
    ONEDRIVE=C:\Users\user\OneDrive
    OS=Windows_NT
    PATH=C:\Users\user\github\beeware\briefcase\venv-3.10-briefcase\Scripts;C:\Program Files\PowerShell\7;C:\Users\user\.pyenv\pyenv-win\shims;C:\Users\user\.pyenv\pyenv-win\versions\3.9.13;C:\Users\user\.pyenv\pyenv-win\versions\3.8.10;C:\Users\user\.pyenv\pyenv-win\versions\3.10.11;C:\Users\user\.pyenv\pyenv-win\versions\3.11.3;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Windows\System32\OpenSSH\;C:\Program Files\Git\cmd;;C:\Program Files\dotnet\;C:\Program Files\starship\bin\;C:\Program Files\PowerShell\7\;C:\Program Files\Docker\Docker\resources\bin;C:\Users\user\.pyenv\pyenv-win\bin;C:\Users\user\AppData\Local\Microsoft\WindowsApps;C:\Users\user\.dotnet\tools;C:\Users\user\.dotnet\tools;C:\Users\user\AppData\Local\Programs\Microsoft VS Code\bin
    PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL
    POWERSHELL_DISTRIBUTION_CHANNEL=MSI:Windows 10 Pro
    PROCESSOR_ARCHITECTURE=AMD64
    PROCESSOR_IDENTIFIER=AMD64 Family 25 Model 33 Stepping 0, AuthenticAMD
    PROCESSOR_LEVEL=25
    PROCESSOR_REVISION=2100
    PROGRAMDATA=C:\ProgramData
    PROGRAMFILES=C:\Program Files
    PROGRAMFILES(X86)=C:\Program Files (x86)
    PROGRAMW6432=C:\Program Files
    PSMODULEPATH=C:\Users\user\Documents\PowerShell\Modules;C:\Program Files\PowerShell\Modules;c:\program files\powershell\7\Modules;C:\Program Files\WindowsPowerShell\Modules;C:\Windows\system32\WindowsPowerShell\v1.0\Modules
    PUBLIC=C:\Users\Public
    PYENV=C:\Users\user\.pyenv\pyenv-win\
    PYENV_HOME=C:\Users\user\.pyenv\pyenv-win\
    PYENV_ROOT=C:\Users\user\.pyenv\pyenv-win\
    STARSHIP_SESSION_KEY=********************
    STARSHIP_SHELL=pwsh
    SYSTEMDRIVE=C:
    SYSTEMROOT=C:\Windows
    TEMP=C:\Users\user\AppData\Local\Temp
    TMP=C:\Users\user\AppData\Local\Temp
    USERDOMAIN=vm-win11
    USERDOMAIN_ROAMINGPROFILE=vm-win11
    USERNAME=user
    USERPROFILE=C:\Users\user
    VIRTUAL_ENV=C:\Users\user\github\beeware\briefcase\venv-3.10-briefcase
    VIRTUAL_ENV_DISABLE_PROMPT=1
    WINDIR=C:\Windows
    WSLENV=WT_SESSION::WT_PROFILE_ID
    WT_PROFILE_ID={574e775e-4f2a-5b96-ac1e-a2962a402336}
    WT_SESSION=f5d14929-bfdc-4438-b46d-323336b15bc0
    _OLD_VIRTUAL_PATH=C:\Program Files\PowerShell\7;C:\Users\user\.pyenv\pyenv-win\shims;C:\Users\user\.pyenv\pyenv-win\versions\3.9.13;C:\Users\user\.pyenv\pyenv-win\versions\3.8.10;C:\Users\user\.pyenv\pyenv-win\versions\3.10.11;C:\Users\user\.pyenv\pyenv-win\versions\3.11.3;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Windows\System32\OpenSSH\;C:\Program Files\Git\cmd;;C:\Program Files\dotnet\;C:\Program Files\starship\bin\;C:\Program Files\PowerShell\7\;C:\Program Files\Docker\Docker\resources\bin;C:\Users\user\.pyenv\pyenv-win\bin;C:\Users\user\AppData\Local\Microsoft\WindowsApps;C:\Users\user\.dotnet\tools;C:\Users\user\.dotnet\tools;C:\Users\user\AppData\Local\Programs\Microsoft VS Code\bin

Briefcase Log:
[18:48:29]                                                                                                                                                                dev.py:208
           [helloworld] Starting in dev mode...                                                                                                                           dev.py:208
                                                                                                                                                                   subprocess.py:705
           >>> Running Command:                                                                                                                                    subprocess.py:706
           >>>     'C:\Users\user\github\beeware\briefcase\venv-3.10-briefcase\Scripts\python.exe' -u -X dev -X utf8 -c 'import runpy,                             subprocess.py:707
           sys;sys.path.pop(0);sys.argv.extend([]);runpy.run_module("helloworld", run_name="__main__", alter_sys=True)'                                                             
           >>> Working Directory:                                                                                                                                  subprocess.py:714
           >>>     C:\Users\user                                                                                                                                   subprocess.py:715
           >>> Environment Overrides:                                                                                                                              subprocess.py:724
           >>>     PYTHONPATH=C:\Users\user\github\beeware\briefcase\tmp\helloworld\src                                                                            subprocess.py:726
           >>>     PYTHONMALLOC=default                                                                                                                            subprocess.py:726
           ===========================================================================                                                                                    run.py:160
           Ann├⌐e                                                                                                                                                  subprocess.py:675
```

</details>
### Additional context

_No response_

rmartin16:
## Changes
- Fixes #1407 
- Ensure `subprocess` calls to Python always use UTF-8
  - _All_ calls for `sys.executable` now use `-X utf8` so Python's output is always UTF-8
- Refactor text mode configuration to more holistically align with the desired intention
  - Ensure text mode is enabled the same way that `subprocess` enables it so there aren't logic holes allowing text mode to activate without an explicit encoding specified
  - Normalize `text` argument against setting/presence of `text`, `universal_newlines`, `encoding`, and `errors`
  - Discard `universal_newlines` argument altogether in favor of `text`
- Decode Windows stub app output as UTF-8
  - The embedded Python interpreter uses `utf8` mode
- Decode Gradle output using system's default encoding
  - Use `locale.getencoding()` to specify the system default encoding
  - Of note, this is necessary for Gradle since the logic in `briefcase.integrations.subprocess` uses the same encoding as Python's `stdout` and that encoding is likely different than the default encoding on Windows...
- Decode ADB output as UTF-8
  - As best I can tell, `adb logcat` is outputing UTF-8

## Notes
- Tested on:
  - Windows 10 using cp437/cp1252
  - Ubuntu 22.04 using ISO-8859-15

## PR Checklist:


rmartin16:
haha, this rabbit hole is fun....gradle's output is actually `cp1252` on Windows...

rmartin16:
No matter what I do, Gradle continues to write to `stdout` in cp1252....although, there seems to be _a lot_ of arcane encoding settings for Gradle and jvm...so, maybe I've just missed it.

Setting `encoding=locale.getpreferredencoding(False)` seems appropriate for Gradle.

rmartin16:
Running Briefcase on Linux without using a UTF-8 locale seems especially problematic. I tried with ISO-8859-1.

1. The • character (i.e. [\u2022](https://www.compart.com/en/unicode/U+2022)) we're putting in Rich's Progress Bar isn't supported...so, Rich crashes out and recommends setting `PYTHONIOENCODING=utf-8`. (This recommendation seems to arise from environments without `$LANG` set and Python defaults to ASCII). This more or less fixes Rich....but it also means everything written to `stdout` over 7-bit ASCII is a bit of a disaster....but Python doesn't crash at least....

2. Briefcase runs Docker assuming the system encoding will come back....but the Docker containers will almost certainly be using UTF-8. This can be overcome by forcing `encoding="UTF-8"` for all Docker calls. Although, something weird happens when you do this and certain actions, like running `pip` in the container, can cause the container to crash `with <Signals.SIGKILL: 9>`....

3. linuxdeploy (or the way Briefcase is invoking) doesn't play well with non-ASCII filenames...they turn in to mojibake....or it may actually because of the next problem...

4. When `briefcase.toml` is loaded via `tomllib.load()`, it makes a naked `decode()` call on the bytes of the file....which means it uses ISO-8859-1 instead of UTF-8 when the file is encoded with UTF-8. This is obviously fine if the file only contains ASCII...but if you name your project "Année", that won't be the case and things go downhill from here. This could be avoided if the sanitized App name was used....like with system packages.

Given how severe the situation is with using Linux without UTF-8 and the fact no one has brought this up.....I'm inclined to do nothing. This reinforces that UTF-8 has significantly pervaded Linux distros and going out of our way to support non-UTF-8 locales is unnecessary.

freakboy3742:
> Given how severe the situation is with using Linux without UTF-8 and the fact no one has brought this up.....I'm inclined to do nothing. This reinforces that UTF-8 has significantly pervaded Linux distros and going out of our way to support non-UTF-8 locales is unnecessary.

Broadly agreed. We would have heard someone make noise about this if it were more than a severe edge case.

If we're able to detect this situation and drop a warning in the console when it occurs (but not make any attempt to fix problems beyond that), I'd be fine with that. But if not... I'm also fine with that - we can handle this with a documentation note saying "srsly... use UTF-8 already".

rmartin16:
The more I look at Python's management of encoding, the more confused I become. This confused state is quite personified by the discussion in python/cpython#87718 which ends with "This is no agreement yet on what is the "current locale encoding".

Nonetheless, [`locale.getencoding()`](https://docs.python.org/3/library/locale.html#locale.getencoding) appears to be the API I'm after. Although, note that before Python 3.11, it seems necessary to use [`locale.getdefaultlocale()[1]`](https://docs.python.org/3/library/locale.html#locale.getdefaultlocale) instead...which definitely seems imperfect for the [reasons](https://github.com/python/cpython/issues/87723) it was deprecated in 3.11...

I'll add a new cached property to `ToolCache` to return this value.

It will be used to drive the decode for Gradle output and log a warning on Linux when UTF-8 is not being used.

If anyone has any insight in to whether this is "correct" or better methods to determine the "system encoding", please add them here.

rmartin16:
Of note, the `subprocess` module itself [uses](https://github.com/python/cpython/blob/d69805b38a1815e7aaadf49bdd019c7cca105ac6/Lib/subprocess.py#L378C36-L378C36) `locale.getencoding()` to decode data in the pipes. Although, it also respects UTF-8 mode....which starts to make me feel like Python's "UTF-8 mode" is basically a sledgehammer that ostensibly means "no more weird defaults with encoding....it's UTF-8 all the way down"...which after the initial pain of understanding this is almost certainly the better default at this point...

freakboy3742:
> If anyone has any insight in to whether this is "correct" or better methods to determine the "system encoding", please add them here.

/me takes a step back.

I definitely don't have any deeper insight on this; I'm happy to take your research and understanding as gospel in this case.

freakboy3742 (approved):
I think the only reasonable interpretation here is that computers were a mistake, and we should all give up and become goat farmers. :-)

This all seems consistent with what I understand about the morass of encoding on Linux and Windows. I've added some docstrings to aid with future code archaeology (mostly documenting the baked-in assumptions about `adb logcat`, and providing breadcrumbs back to this PR), but otherwise this seems like a solid collection of cleanups and simplifications (especially around coercing `universal_newlines` into the modern `text` form).

The only thing I can see that might be worth a follow up in future is to consolidate the Python invocations into a wrapper. "Invoke module using `sys.executable`" is clearly a very common pattern, but it would be really easy to accidentally neglect to add `-X utf8` to that invocation. A common abstraction API would avoid that problem, and would also provide a convenient place for "invoke this module in a venv" functionality.

PR #1454, opened 2023-09-15
Fail CI if a default encoding is used

rmartin16:
## Changes
- When CI runs, require an explicit encoding anywhere one is needed

## Notes
- Python 3.15 will finally default to UTF-8; however, even then, `PYTHONUTF8=0` or `-X utf8=0` can disable it
- This is dependent on https://github.com/cookiecutter/cookiecutter/pull/1937

## PR Checklist:

### DIFF
--- a/changes/1454.misc.rst
+++ b/changes/1454.misc.rst
@@ -1,0 +1,1 @@
+Explicit encodings are now enforced in CI.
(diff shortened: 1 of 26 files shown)


rmartin16:
more weirdness with cookiecutter releases....`2.3.1` is on [PyPI](https://pypi.org/project/cookiecutter/#history) but the [GitHub release](https://github.com/cookiecutter/cookiecutter/releases) was deleted...(i know it existed at one point because GitHub sent me the release notification...)

freakboy3742 (approved):
👍 

Worth noting that this won't do anything on Py3.8 or 3.9, as the `warn_default_encoding` flag wasn't added until 3.10; but it's a no-op, so it doesn't hurt, and we really only need one version to fail.

PR #1458, opened 2023-09-17
Consolidate Dependabot PRs

rmartin16:
## Changes
- Consolidates Dependabot PRs in to two categories:
  - GitHub Actions
  - PyPI packages

## Notes
- Saw this in another [repo](https://github.com/PhasecoreX/docker-user-image/pull/23#event-10390328989) and figured you might benefit from it
- There are quite a few [options](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) to configure it if you want to control the grouping a bit more
  - For instance, only group updates for minor and patch updates to packages (although, `setup.cfg` basically prevents all major version updates soo...)
- Unfortunately, I don't see a way to effectively tell Dependabot "separate out the update for XXX package in to an individual PR"
  - Instead, I think its necessary to explicitly exclude that package from the group and re-run dependabot

## PR Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,5 @@
-
 version: 2
+
 updates:
   - package-ecosystem: "github-actions"
     directory: "/"
@@ -8,6 +8,10 @@ updates:
       interval: "weekly"
       day: "sunday"
       time: "20:00"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
 
   - package-ecosystem: "pip"
     directory: "/"
@@ -16,3 +20,7 @@ updates:
       interval: "weekly"
       day: "sunday"
       time: "20:00"
+      groups:
+        pip:
+          patterns:
+            - "*"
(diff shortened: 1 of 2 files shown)


freakboy3742 (commented):
This definitely reduces the number of dependabot build jobs on a Monday; but it also means we can't easily isolate which package in a series of updates is the cause of a problem. 

When we rolled out the pre-commit updater, we explicitly broke apart the updater into per-package updates; is the implication here that we should roll back that change as well?

rmartin16:
I think it's up to you; other than having a bunch of emails to move out of my inbox each Sunday afternoon, I'm mostly indifferent.

I just saw this Dependabot functionality and figured I throw it out there since this only became possible a few months ago.

> is the implication here that we should roll back that change as well?

I'd be fine with that; I could leave the current way it works as a "on demand" workflow that could be kicked off manually to create individual PRs. Just a thought.

freakboy3742:
I think my inclination is to preserve the per-package updates for now. Managing the updates isn't *that* onerous, and it's easier to isolate problems if/when they occur. 

Thanks for the suggestion (and for keeping an eye on Dependants capabilities).

PR #1470, opened 2023-09-29
Enable debug cookiecutter logging for deep debug

rmartin16:
## Changes
- Configure logging for cookiecutter before generating a template via [`cookiecutter.log.configure_logger()`](https://github.com/cookiecutter/cookiecutter/blob/24020484395f3962156dfcee5e9e1d6f2013da07/cookiecutter/log.py#L19)
- This enables cookiecutter's debug output when Briefcase's "deep debug" is enabled via `-vv`
- The changenote is `misc` since we already have a `feature` note for the addition of `-vv` support

## Notes
- cookiecutter uses the notorious `logging` package for its logging
  - By default, it appears that cookiecutter [only configures](https://github.com/search?q=repo%3Acookiecutter%2Fcookiecutter%20configure_logger&type=code) logging handlers for its logging when using its CLI; therefore, cookiecutter doesn't enable any handlers for its logger when Briefcase generates a template.
  - Although, perhaps notably, they also don't configure a `NullHandler` by default....so, cookiecutter is likely to be subjected to a default handler if one is configured....but such is the life of using `logging`...
  - At any rate, since no handlers were configured before, this changes cookiecutter logging to output `INFO` and above by default.
  - Although, cookiecutter only appears to use `logger.debug()` and `logger.error()` right now...so, it shouldn't matter in reality.
  - Finally, cookiecutter deletes any existing handlers when `configure_logger()` is called....so, it is safe to call multiple times without concerns about duplicating handlers for the logger.
- Additionally, I couldn't call `configure_logger()` in `BaseCommand.__init__()` because `verbosity` won't be set until `parse_options()` is called...and `verbosity` can technically change (although, it doesn't currently).

## PR Checklist:

### DIFF
--- a/changes/1470.misc.rst
+++ b/changes/1470.misc.rst
@@ -1,0 +1,1 @@
+Cookiecutter's debug logging is now enabled when deep debug is activated via ``-vv``.
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -13,6 +13,7 @@ from argparse import RawDescriptionHelpFormatter
 from pathlib import Path
 
 from cookiecutter import exceptions as cookiecutter_exceptions
+from cookiecutter.log import configure_logger as configure_cookiecutter_logger
 from cookiecutter.repository import is_repo_url
 from platformdirs import PlatformDirs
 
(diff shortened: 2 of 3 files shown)


freakboy3742 (approved):
Looks great. I can't imagine we'll need to turn this on very often, but it's good to know we have the option.

PR #1475, opened 2023-10-03
Ensure that macOS binaries are cross platform, even if wheels aren't

### LINKED ISSUE #1217: macOS apps aren't portable across x86_64/M1 unless all binary wheels are universal2

Many Python libraries ship separate x86_64 and arm64 binary wheels for macOS.  For example, see https://pypi.org/project/numpy/1.24.3/#files .  Even if the project *does* ship universal2 wheels, [pip won't install them](https://github.com/pypa/pip/issues/11573) because [the platform tags are in the wrong order](https://github.com/pypa/packaging/issues/381).  You cannot even require pip to install `universal2` wheels; it will prefer the more specific platform regardless of configuration.

When building an application "for release" (not sure if that has a precise definition within Briefcase but really I mean "codesigned & notarized for public distribution"), briefcase should:

- ensure that your interpreter is `universal2` and error out with a clear error message explaining that distribution binaries should be built only with an appropriate interpreter.  (I have a strong bias towards recommending python.org's builds for this purpose, but YMMV)
- download all the official `universal2` wheels it can find
- discover all dependencies with separate `x86_64` and `arm64` wheels
- use [`delocate-fuse`](https://github.com/matthew-brett/delocate/blob/df86dddd7c94a93b5c03948b8c127ba0777e2a4d/delocate/fuse.py) to smash together the relevant wheel, synthesizing the platform tag.

Encrust has an incredibly janky solution to this problem that requires the building computer be an M1, because it just invokes `pip install` twice, once under `arch -x86_64`.  This isn't generally usable because x86_64 CI (ahem, github actions) is still pretty popular.

See https://github.com/glyph/Encrust/blob/91b51faf299ad5967d1853a0651c27666107a156/src/encrust/_architecture.py

It uses this hack because discovering the proper set of platform tags to pass to `pip download` is kind of annoying and fiddly, since you can't just say `universal2`, you need to exactly call out base SDK versions as well.  For example, `pip download  --only-binary :all: numpy --platform=macosx_10_12_x86_64` for x86_64, but `pip download  --only-binary :all: numpy --platform=macosx_11_0_arm64` for arm.

### Additional context

Delocate is open to a PR that would fix the retagging issue on their end, so it is probably best to do that part upstream. https://github.com/matthew-brett/delocate/issues/174

iOS has a similar problem. Given that Beeware is at the forefront of the ecosystem for iOS wheels and thus might have more influence there, perhaps we could establish some norms about building & uploading (i.e., do not support single-architecture uploads to PyPI _at all_, there would be no point)

freakboy3742:
Fixes #1217.

If a third-party package provides a universal2 wheel, any BeeWare macOS app using that binary will also be universal.

However, if a third party package is single platform (e.g., numpy and Pillow both publish single-platform wheels), then the resulting binary will only run on the platform where it was built. 

This modifies the macOS build process to:
1. Install all packages for the build architecture into a temporary app_packages folder
2. Inspect all the installed wheels, looking for any wheel that is marked as non-pure *and* non-universal
3. Perform a *second* install of *just* the non-pure, non-universal packages, into a second temporary app_packages folder. This a similar site-packages hack as the iOS backend to trick pip into installing binaries for a different platform.
4. Merge the contents of the two app_packages folders into the final "production" app-packages folder.

Since the support package now contains multiple pieces that aren't needed at runtime (an XCFramework, plus the platform site-packages), the support package for the macOS app backend is also installed into a temporary location, and the standard library is copied into the "production" location.

Also modifies the handling of verbosity. The macOS backend has 2 steps where it prints a lot of output - purging content (where all the `__pycache__` folders are removed, amongst other files), and signing (where each file being signed is listed). This output is useful for debugging, but can easily obscure warnings like potential signing problems. It also obscured warnings about discrepancies between platform files. Therefore, this PR modifies output so that:

* At default verbosity, a progress bar is shown as files are deleted/signed/merged, but the individual files are *not* displayed. Any warnings *will* be visible.
* At verbosity 1 (`-v`), the individual files being deleted/signed/merged will also be printed.
* At verbosity 2 (`-vv`), the deep console debug (showing the actual commands being executed, shell environment, and the raw console response) will also be printed

## PR Checklist:


changes/1217.bugfix.rst
```diff
@@ -0,0 +1 @@
+macOS apps are now guaranteed to be universal binaries when dependencies only provide single-architecture binary wheels.
```
mhsmith, line 1: Clarify that the intention is for ALL apps to be universal binaries.

```suggestion
macOS apps are now guaranteed to be universal binaries, even when dependencies only provide single-architecture binary wheels.
```

src/briefcase/platforms/macOS/utils.py
```diff
@@ -0,0 +1,184 @@
...
+            are in the install path that are non-universal and non-pure.
+        """
+        binary_packages = []
+        for distinfo in install_path.glob("**/*.dist-info"):
+            # Read the WHEEL file in the dist-info folder.
+            # Use this to determine if the wheel is "pure", and the tag
+            # for the wheel.
+            with (distinfo / "WHEEL").open("r", encoding="utf-8") as f:
+                wheel_data = email.message_from_string(f.read())
+                is_purelib = wheel_data["Root-Is-Purelib"] == "true"
```
mhsmith, line 50: The wheel spec doesn't explicitly say that this key is required.

```suggestion
                is_purelib = wheel_data.get("Root-Is-Purelib", "false") == "true"
```

changes/1217.bugfix.rst
mhsmith: What's the path forward when packages stop releasing x86_64 and universal2 wheels, and focus only on arm64? I guess we'll need to re-introduce an option to build single-platform apps again.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -37,12 +40,96 @@ class macOSMixin:
...
+            ):
+                try:
+                    self.tools[app].app_context.run(
+                        [
+                            sys.executable,
+                            "-u",
+                            "-X",
+                            "utf8",
+                            "-m",
+                            "pip",
```
mhsmith, line 91: The common code between this call of pip, and the one in commands/create.py, should be factored out.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -37,12 +40,96 @@ class macOSMixin:
...
+        )
+
+        # Now install dependencies for the architecture that isn't the host architecture.
+        other_arch = {
+            "arm64": "x86_64",
+            "x86_64": "arm64",
+        }[self.tools.host_arch]
+
+        # Create a temporary folder to hold wheels for other targets
+        other_app_packages_path = self.bundle_path(app) / f"app_packages.{other_arch}"
```
mhsmith, line 74: What stops these temporary folders from being included in the final app?

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -37,12 +40,96 @@ class macOSMixin:
...
+            universal_suffix="_universal2",
+        )
+
+        # Now install dependencies for the architecture that isn't the host architecture.
+        other_arch = {
+            "arm64": "x86_64",
+            "x86_64": "arm64",
+        }[self.tools.host_arch]
+
+        # Create a temporary folder to hold wheels for other targets
```
mhsmith, line 73:
```suggestion
        # Create a temporary folder targeting the other platform
```

changes/1217.bugfix.rst
freakboy3742: I thought about that - it's potentially even an issue right now, if you have a particularly progressive app developer who decides that they either (a) don't care about x86 users, or (b) are impacted by an upstream dependency that doesn't provide an x86 package. Both are entirely possible outcomes - given that there are projects already deprecated Python 3.8 support, I wouldn't think we're *too* far off this being a practical concern.

My inclination is to land this patch *without* support for single platform builds, and tackle that as a subsequent PR - I'm guessing a mac-specific app option `universal2_build` which is True by default (for now), but can be configured to False, which reverts the install process to "old style" single platform installs; at some point in the future, we change the default, then deprecate the option.

However, if you think this is a feature we can't ship without including, I'll work up the modifications.

src/briefcase/platforms/macOS/__init__.py
freakboy3742: Location. The final directory structure for an app is something like:

- build/helloworld/macos/app
   - app_packages.arm64
   - app_packages.x86_64
   - support
   - Hello World.app
     - Contents
         - MacOS
         - Resources
            - app
            - app_packages
            - support

changes/1217.bugfix.rst
mhsmith: I agree we can postpone this for now, but it would be worth creating an issue to track it.

mhsmith (approved):

PR #1476, opened 2023-10-03
Require CI to succeed on Python 3.12

rmartin16:
## Changes
- CI uses latest stable 3.12 release and tests are required to succeed now

## Notes
- Python 3.12 is [not available](https://apps.microsoft.com/store/search?publisher=Python%20Software%20Foundation) in Windows Store yet
- Running tests against Python 3.13 is [not an option](https://github.com/actions/python-versions/blob/main/versions-manifest.json) yet....and I'm not really sure when to expect it to be available...
- I'm not sure if there are bigger plans to incorporate m1 macOS testing yet...but feel free to supersede this PR

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,12 +44,12 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ "macos", "ubuntu", "windows" ]
-        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12-dev" ]
+        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12" ]
         include:
         - experimental: false
         # Allow dev Python to fail without failing entire job
-        - python-version: "3.12-dev"
-          experimental: true
+        #- python-version: "3.13-dev"
+        #  experimental: true
         # Run tests against the latest Windows Store Python
         - platform: "windows"
           python-version: "winstore3.11"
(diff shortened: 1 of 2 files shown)


rmartin16:
will probably just wait on the win store and `3.13-dev` unless a compelling reason to merge sooner presents

freakboy3742 (commented):
We're obviously waiting for the 3.12 images to become available, but as soon as those are ready, this looks good.

I'm not sure what Steve Dower's plans are regarding the windows store version of 3.12; I'm happy to deal with that as a separate PR if it doesn't land before the 3.12 runner becomes available.e

I'm still trying to work out what we can do with M1 runners; there doesn't appear to be a free tier, so I need to work out what the cost will be if I turn those on. For now, I'm happy to live without M1 runners on Briefcase.

[3.13.0a1 should land October 17](https://peps.python.org/pep-0719/), so it seems reasonable to assume the Python installation action will be updated a couple of days after that.

rmartin16:
> We're obviously waiting for the 3.12 images to become available, but as soon as those are ready, this looks good.

They landed this morning. These CI checks ran against Python 3.12.0.

> I'm still trying to work out what we can do with M1 runners; there doesn't appear to be a free tier, so I need to work out what the cost will be if I turn those on. For now, I'm happy to live without M1 runners on Briefcase.
> 
> [3.13.0a1 should land October 17](https://peps.python.org/pep-0719/), so it seems reasonable to assume the Python installation action will be updated a couple of days after that.

👍🏼

freakboy3742 (approved):

PR #1483, opened 2023-10-07
Ensure that dynamic lib merge targets are unique.

### LINKED ISSUE #1481: "Build app (ppb, macos-latest)" CI job is broken
This is related to #1475:
```
Installing app requirements... done
Collecting wrapt==1.15.0
  Downloading wrapt-1.15.0-cp310-cp310-macosx_11_0_arm64.whl (36 kB)
Installing collected packages: wrapt
Successfully installed wrapt-1.15.0
Installing binary app requirements for arm64... done
Merging app_packages.x86_64... done
Merging app_packages.arm64... done
Merging libraries...
error: /Applications/Xcode_14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: can't move temporary file: /Users/runner/work/briefcase/briefcase/tests/apps/verify-ppb/build/verify-ppb/macos/app/Hello PPB.app/Contents/Resources/app_packages/wrapt/_wrappers.cpython-310-darwin.so to file: /Users/runner/work/briefcase/briefcase/tests/apps/verify-ppb/build/verify-ppb/macos/app/Hello PPB.app/Contents/Resources/app_packages/wrapt/_wrappers.cpython-310-darwin.so.lipo (No such file or directory)

Unable to create fat library for wrapt/_wrappers.cpython-310-darwin.so
```

freakboy3742:
The candidate list for dylibs to be merged will not be unique; every dylib will be found twice. This may also be the cause of the CI failure.

Also ensures that the merge step only operates on single-platform wheels.

Fixes #1481 

## PR Checklist:

### DIFF
--- a/changes/1481.misc.rst
+++ b/changes/1481.misc.rst
@@ -1,0 +1,1 @@
+macOS dynamic libraries are now inspected to ensure that they are single platform before merging, and they are only merged once.
--- a/src/briefcase/platforms/macOS/__init__.py
+++ b/src/briefcase/platforms/macOS/__init__.py
@@ -95,7 +95,22 @@ class macOSInstallMixin(AppPackagesMergeMixin):
                     },
                 )
         else:
-            self.logger.info("All packages are pure python or universal.")
+            self.logger.info("All packages are pure Python, or universal.")
+
+        # If given the option of a single architecture binary or a universal2 binary,
+        # pip will install the single platform binary. However, a common situation on
+        # macOS is for there to be an x86_64 binary and a universal2 binary. This means
+        # you only get a universal2 binary in the "other" install pass. This then causes
+        # problems with merging, because the "other" binary contains a copy of the
+        # architecture that the "host" platform provides.
+        #
+        # To avoid this - ensure that the libraries in the app packages for the "other"
+        # arch are all thin.
+        #
+        # This doesn't matter if it happens the other way around - if the "host" arch
+        # installs a universal binary, then the "other" arch won't be asked to install
+        # a binary at all.
+        self.thin_app_packages(other_app_packages_path, arch=other_arch)
 
         # Merge the binaries
         self.merge_app_packages(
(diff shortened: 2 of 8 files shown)


freakboy3742:
@mhsmith Hold off on reviewing this one - I've found another edge case related to dylib merging as a result of CI on beeware/toga#1399.

freakboy3742:
Ok - I've now fixed the problem we are seeing in beeware/toga#1399's CI. 

The problem is that pip will install platform specific binaries in preference to universal2 binaries. As a result, if a package publishes both x86_64 and universal2 binaries (which is a common configuration - fontTools is at least one example that does this), and you build your app on x86_64, the original app_packages install the single platform binary, discovers the need for an arm64 binary, which then installs a universal2 binary. When you try to merge the two, you get an error, because 2 of the libraries being merged have an x86_64 slice.

This fix does a thinning pass first, ensuring that all the libraries in the arm64 folder only contain arm64 components. This does result in some double handling - you get a universal binary, strip it, then make a universal binary... but there's no easy way I can think of to either (a) force the installation of a universal binary in the first place, or (b) make a decision to use existing the universal binary, rather than merge. The latter *could* potentially be done, but it's going to involve most of the same steps anyway (because you're going to have to inspect every binary to see if it's universal or not, then work out if using one of the binaries is an option, and if so, which one...), so I opted for the "strip then merge" approach.

mhsmith (approved):
That all makes sense, but why didn't it fail in #1475's pre-merge checks? Is it because it uses a thread pool, so the order of the lipo commands was non-deterministic?

freakboy3742:
> That all makes sense, but why didn't it fail in #1475's pre-merge checks? Is it because it uses a thread pool, so the order of the lipo commands was non-deterministic?

That's my best guess as well. When I run locally, I haven't been able to reproduce this error; and #1475 clearly had at least *one* CI pass (the last one) where it wasn't an issue. I'm guessing that the slower x86_64 CI machines are *more* prone to this problem, which is why we're seeing it manifest on most(?) CI runs.

PR #1495, opened 2023-10-16
Support testing against Python 3.13

rmartin16:
## Changes
- Add support for CI to test Python 3.13
- Add support for `tox` to run using Python 3.13

## Issues
- [X] `coverage` needs support for Python 3.13 (added with v7.4.1)
- [x] Need a `pytest` release that includes https://github.com/pytest-dev/pytest/issues/11874

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,12 +46,12 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ "macos", "ubuntu", "windows" ]
-        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12" ]
+        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12", "3.13-dev" ]
         include:
         - experimental: false
         # Allow dev Python to fail without failing entire job
-        #- python-version: "3.13-dev"
-        #  experimental: true
+        - python-version: "3.13-dev"
+          experimental: true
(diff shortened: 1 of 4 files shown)


freakboy3742 (changes requested):
Content of the PR looks good; although it's missing the setup.cfg metadata declaring Py3.13 compatibility. 

On top of that, I think we should hold off merging until coverage and CI is passing. We have a higher threshold for failures on dev releases, but as we're still early in the 3.13 cycle, I think we can afford to hold off until there's basic support in place.

rmartin16:
Agreed; I was mostly just curious where support stands. At a minimum, we're waiting on a release of `coverage`.

rmartin16:
Pytest [8.0.1](https://github.com/pytest-dev/pytest/releases/tag/8.0.1) released with the needed fix. I also added the classifier for Python 3.13.

freakboy3742 (approved):
Looks fairly straightforward; Approved on the basis that it does what it says on the tin.

As a follow up to this - we should probably turn on macOS ARM64 testing as well. That will start to explode the size of the CI matrix to absurd levels - but we can probably cut back on doing *every* python version on *every* OS, and limit ourselves to "min and max versions" (or min, max and dev) on every OS, and every version on *one* OS (probably linux, given there are more available runners).

PR #1498, opened 2023-10-19
Bump JDK to 17.0.9+9

rmartin16:
## Changes
- Bump JDK to 17.0.9+9 (from 17.0.7+7 in v0.3.15)

## Notes
- hmmm....the 17.0.9+9 [release](https://github.com/adoptium/temurin17-binaries/releases/tag/jdk-17.0.9%2B9) has slowly been gaining the missing assets since it was originally published...
- Windows x64 is still missing...
## PR Checklist:

### DIFF
--- a/changes/1498.feature.rst
+++ b/changes/1498.feature.rst
@@ -1,0 +1,1 @@
+The Java JDK was upgraded to 17.0.9+9.
--- a/src/briefcase/integrations/java.py
+++ b/src/briefcase/integrations/java.py
@@ -19,8 +19,8 @@ class JDK(ManagedTool):
 
     # Latest OpenJDK as of September 2023: https://adoptium.net/temurin/releases/
     JDK_MAJOR_VER = "17"
-    JDK_RELEASE = "17.0.8.1"
-    JDK_BUILD = "1"
+    JDK_RELEASE = "17.0.9"
+    JDK_BUILD = "9"
     JDK_INSTALL_DIR_NAME = f"java{JDK_MAJOR_VER}"
 
     def __init__(self, tools: ToolCache, java_home: Path):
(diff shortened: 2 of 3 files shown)


rmartin16:
I was thinking....should Briefcase support automatically upgrading existing Briefcase-owned JDK installs?

This would effectively remove the need for the `briefcase upgrade java` command.....but thinking about it, there seems to be a rational basis to remove `briefcase upgrade` altogether; instead, Briefcase would pin specific versions of these third party tools (while supporting overrides via the appropriate env vars).

In practice, most of these tools don't have a lot of development or releases happening...so, it doesn't really matter. But at least in the abstract.....ensuring a specific version of the tool is used for a specific version of Briefcase seems like the most stable choice while allowing users to override this to use an arbitrary version for each tool.

freakboy3742:
Pinning tool versions to a Briefcase release is definitely an interesting idea - it would avoid the "I set up Briefcase 2 years ago and now Briefcase doesn't work" class of problem. That hasn't been a *big* problem to date, but I can imagine it might emerge over time. 

My concern would be the fight between different tool versions when people have multiple versions of Briefcase installed. The worst situation would be having someone constantly upgrading and downgrading because they're moving between 2 releases of Briefcase (which, coincidentally, will be the sort of thing that *we*, as developers, will be seeing all the time while PRs are in development).

rmartin16:
looks like the 17.0.9+9 release for Windows x64 went a little sideways...and they've published a new release with just Windows x64. since the GitHub release is called [jdk-17.0.9+9.1](https://github.com/adoptium/temurin17-binaries/releases/tag/jdk-17.0.9%2B9.1), the URLs to download Windows vs Linux or macOS are different.....so guess we wait for the next release....

https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.9%2B9/OpenJDK17U-jdk_x64_linux_hotspot_17.0.9_9.tar.gz
https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.9%2B9.1/OpenJDK17U-jdk_x64_windows_hotspot_17.0.9_9.zip

PR #1502, opened 2023-10-20
Restore debug logging for `-v` flag

### LINKED ISSUE #1501: The `-v` flag no longer prints debug output to the console
### Describe the bug

When Briefcase is started with the `-v` flag, debug output is no longer printed to the console. Furthermore, when `-vv` is used, both debug and deep debug output is printed.

The log file, though, doesn't appear to be affected and continues to always include debug logging and the `-vv` flag still includes deep debug.

### Steps to reproduce

Run `briefcase build -v`.

### Expected behavior

When the `-v` is used, debug output should be printed to the console.

### Screenshots

_No response_

### Environment

- Python version: 3.10.13
- Software versions:
  - Briefcase: v0.3.16

### Logs

_No response_

### Additional context

_No response_

rmartin16:
## Changes
- This is the immediate fix....unencumbered with subsequent changes I'm going to propose to mitigate this class of error going forward
- Closes #1501

## PR Checklist:

### DIFF
--- a/changes/1501.bugfix.rst
+++ b/changes/1501.bugfix.rst
@@ -1,0 +1,1 @@
+Debug logging was restored when the ``-v`` flag is used.
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -117,7 +117,7 @@ class Log:
     """Manage logging output driven by verbosity flags."""
 
     # Level of verbosity when debug output is shown in the console
-    DEBUG = 2
+    DEBUG = 1
     # Printed at the beginning of all debug output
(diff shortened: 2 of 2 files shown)


freakboy3742 (changes requested):
Given we've just done a release, my inclination is to hold off until we've got a true fix for the debug issue, rather than rush a hotfix into main. There's definitely some discussion to be had around the "full" fix, but I don't think the current main is sufficiently encumbered or the full fix that far off that a reversal is needed.

rmartin16:
agreed; this definitely isn't the fix once I sought more understanding from #1475 om #1503.

PR #1503, opened 2023-10-20
Formalize logging levels

### LINKED ISSUE #1501: The `-v` flag no longer prints debug output to the console

rmartin16:
## Changes
- Closes #1501
- Introduces a `verbose` logging level between `info` and `debug`
- Explicit references to the logging level are now abstracted by a `LogLevel` enum and properties `Log.is_verbose`, `Log.is_debug`, and `Log.is_deep_debug`
- Limited the `>>>` preface to subprocess details logging 

## PR Checklist:

### DIFF
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -126,8 +126,6 @@ class LogLevel(IntEnum):
 class Log:
     """Manage logging output driven by verbosity flags."""
 
-    # Printed at the beginning of all debug output
-    DEBUG_PREFACE = ">>> "
     # subdirectory of command.base_path to store log files
     LOG_DIR = "logs"
 
@@ -215,10 +213,10 @@ class Log:
                     style=style,
                 )
 
-    def debug(self, message="", *, prefix="", markup=False):
+    def debug(self, message="", *, preface="", prefix="", markup=False):
         """Log messages at debug level; included if verbosity >= 2."""
         self._log(
-            preface=self.DEBUG_PREFACE,
+            preface=preface,
             prefix=prefix,
             message=message,
             show=self.is_debug,
@@ -617,7 +615,7 @@ class Console:
         try:
             input_value = self.input(prompt, markup=markup)
             self.print.to_log(prompt)
-            self.print.to_log(f"{Log.DEBUG_PREFACE}User input: {input_value}")
+            self.print.to_log(f"User input: {input_value}")
             return input_value
         except EOFError:
             raise KeyboardInterrupt
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import operator
 import os
 import platform
@@ -6,6 +8,7 @@ import sys
 import traceback
 from contextlib import contextmanager
 from datetime import datetime
+from enum import IntEnum
 from pathlib import Path
 
 from rich.console import Console as RichConsole
@@ -113,19 +116,24 @@ class Printer:
         return cls.log.export_text()
 
 
+class LogLevel(IntEnum):
+    INFO = 0
+    VERBOSE = 1
+    DEBUG = 2
+    DEEP_DEBUG = 3
+
+
 class Log:
     """Manage logging output driven by verbosity flags."""
 
-    # Level of verbosity when debug output is shown in the console
-    DEBUG = 2
     # Printed at the beginning of all debug output
     DEBUG_PREFACE = ">>> "
     # subdirectory of command.base_path to store log files
     LOG_DIR = "logs"
 
-    def __init__(self, printer=Printer(), verbosity=0):
+    def __init__(self, printer=Printer(), verbosity: LogLevel = LogLevel.INFO):
         self.print = printer
-        # --verbosity flag: 0 for info, 1 for debug, 2 for super-verbose
+        # --verbosity flag: 0 for info, 1 for debug, 2 for deep debug
         self.verbosity = verbosity
         # --log flag to force logfile creation
         self.save_log = False
@@ -150,12 +158,12 @@ class Log:
         :param context: The name of the context to enter. This *must* be simple text,
             with no markup or other special characters.
         """
+        self.info()
+        self.info(f"Entering {context} context...")
+        old_context = self._context
+        self._context = f"{context}| "
+        self.info("-" * (72 - len(context)))
         try:
-            self.info()
-            self.info(f"Entering {context} context...")
-            old_context = self._context
-            self._context = f"{context}| "
-            self.info("-" * (72 - len(context)))
             yield
         finally:
             self.info("-" * (72 - len(context)))
@@ -208,15 +216,19 @@ class Log:
                 )
 
     def debug(self, message="", *, prefix="", markup=False):
-        """Log messages at debug level; included if verbosity>=2."""
+        """Log messages at debug level; included if verbosity >= 2."""
         self._log(
             preface=self.DEBUG_PREFACE,
             prefix=prefix,
             message=message,
-            show=self.verbosity >= self.DEBUG,
+            show=self.is_debug,
             markup=markup,
             style="dim",
         )
+
+    def verbose(self, message="", *, prefix="", markup=False):
+        """Log message at verbose level if debug enabled; included if verbosity >= 1."""
+        self._log(prefix=prefix, message=message, show=self.is_verbose, markup=markup)
 
     def info(self, message="", *, prefix="", markup=False):
         """Log message at info level; always included in output."""
@@ -231,12 +243,37 @@ class Log:
         self._log(prefix=prefix, message=message, markup=markup, style="bold red")
 
     @property
-    def is_deep_debug(self):
-        """Is extra verbosity enabled via -vv?
+    def verbosity(self) -> LogLevel:
+        return self._verbosity
 
-        Enables verbose output from third-party tools.
+    @verbosity.setter
+    def verbosity(self, value: int | LogLevel):
+        """Clamp verbosity between valid log levels."""
+        self._verbosity = LogLevel(max(min(LogLevel), min(value, max(LogLevel))))
+
+    @property
+    def is_verbose(self):
+        """Is verbose logging enabled via the -v flag?
+
+        This includes Briefcase logging that isn't relevant during normal operation.
         """
-        return self.verbosity >= 2
+        return self.verbosity >= LogLevel.VERBOSE
+
+    @property
+    def is_debug(self):
+        """Is debug logging enabled via the -vv flag?
+
+        This includes debug logging from third party tools.
+        """
+        return self.verbosity >= LogLevel.DEBUG
+
+    @property
+    def is_deep_debug(self):
+        """Is deep debug logging enabled via the -vvv flag?
+
+        This includes debug logging from third party tools.
+        """
+        return self.verbosity >= LogLevel.DEEP_DEBUG
 
     def capture_stacktrace(self, label="Main thread"):
         """Preserve Rich stacktrace from exception while in except block.
--- a/src/briefcase/platforms/macOS/__init__.py
+++ b/src/briefcase/platforms/macOS/__init__.py
@@ -708,16 +708,10 @@ password:
                 raise BriefcaseCommandError(
                     "Can't notarize an app with an ad-hoc signing identity"
                 )
-            self.logger.info(
-                "Signing app with ad-hoc identity...",
-                prefix=app.app_name,
-            )
+            self.logger.info("Signing app with ad-hoc identity...", prefix=app.app_name)
             self.logger.warning(
-                (
-                    "Because you are signing with the ad-hoc identity, this "
-                    "app will run, but cannot be re-distributed."
-                ),
-                prefix=app.app_name,
+                "Because you are signing with the ad-hoc identity, this "
+                "app will run, but cannot be re-distributed."
             )
         else:
             # If we're signing, and notarization isn't explicitly disabled,
@@ -726,13 +720,16 @@ password:
                 notarize_app = True
 
             self.logger.info(
-                f"Signing app with identity {identity_name}...", prefix=app.app_name
+                f"Signing app with identity {identity_name}...",
+                prefix=app.app_name,
             )
 
             if notarize_app:
                 team_id = self.team_id_from_identity(identity_name)
 
         self.sign_app(app=app, identity=identity)
+
+        dist_path: Path = self.distribution_path(app)
 
         if app.packaging_format == "app":
             if notarize_app:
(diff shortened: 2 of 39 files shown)


rmartin16:
Looking at #1475 a bit more, it seems as though the intention was to effectively introduce a new level of logging between `info` and `debug`; although, in doing so, it redefined _all_ `debug` logging as `deep debug`. Therefore, any logging previously considered `debug` is only enabled when `deep debug` logging is also enabled.

Given that was ostensible objective, this PR (as well as #1501) restores the ability to include `debug` logging without needing to include `deep debug` logging....but it doesn't currently allow macOS packaging to use this new level of logging....

rmartin16:
I think the logging desires of #1475 might be best achieved by adding `quiet` to `subprocess.run()` the same way it was added to `subprocess.check_output()`. In that way, macOS packaging could drive the `quiet` argument based on the current level of verbosity.

Or perhaps, even, just let `Subprocess` drive the `quiet` argument using the logging verbosity...

freakboy3742:
> Looking at #1475 a bit more, it seems as though the intention was to effectively introduce a new level of logging between `info` and `debug`; although, in doing so, it redefined _all_ `debug` logging as `deep debug`. Therefore, any logging previously considered `debug` is only enabled when `deep debug` logging is also enabled.

I think that's a correct interpretation of the intent. 

The underlying problem was that some of the macOS commands generate important warning-level signals that the user really needs to see (as they indicate problems that don't *stop* progress, but *might* need to be addressed), but the macOS logs also contained extremely verbose tracking of progress. 

As a practical example: if you build an app using numpy or pillow, the merge of the x86 and arm wheels reveals a couple of files that are different between the two platforms. As it happens, this isn't *actually* a problem - but it's a potential indicator of a problem, and you should be confirming that the differences aren't actually a problem. However, the merge process then output dozens of lines talking about the individual files being merged, and then listed every file being signed, and the warning got completely obscured from the user's screen unless you did a deep dive on the logs... and ain't nobody got time to do that...:-)

Given that the per-file output isn't *really* needed since we have a really good progress bar, the intention was to downgrade the per-file output to "verbose mode"; that way, the default output will clearly show the warning, but only show a progress bar for merging and signing, so you're more likely to see the warning. 

However, we then needed a way to differentiate -v output from the output level that dumps the underlying subprocess detail. That's a deeper level of detail still that isn't needed if you just want to confirm that the right list of files has been merged/signed.

freakboy3742:
> I think the logging desires of #1475 might be best achieved by adding `quiet` to `subprocess.run()` the same way it was added to `subprocess.check_output()`. In that way, macOS packaging could drive the `quiet` argument based on the current level of verbosity.

I'm not sure I agree. `quiet` exists specifically to work around the verbosity level, not an implementation of verbosity. It's only used in 1 situation IIRC - when you're rapidly polling ADB for process status. In that instance, having dozens of calls logging the same command line, env etc is just noise, even if you *have* asked for "deep debug" that displays the subprocess detail. In an ideal world, we'd have the first invocation outputting the command line in the logs, followed by something like "[42 identical calls]" that let you know you were hammering the same thing over and over - but I'm not sure that's worth the effort to set up.

> [edit] additionally....given that signing is multi-threaded....the debug output is indecipherable for any practical purposes...

In what way? I agree that the output *previously* didn't work well with subprocessing because each line of output didn't include enough context to say what which file it related to; but I though I had fixed that in #1475 by including the filename every time a line is output. Unless you're seeing some sort of output where lines from different threads are being interwoven... but I haven't seen that myself.

I agree that the output is indecipherable when you turn on *deep* debug - but I'm not sure how we'd address that short of some sort of locking around the output of the subprocess preamble...

src/briefcase/console.py
```diff
@@ -208,16 +213,20 @@ def _log(
...
-            show=self.verbosity >= self.DEBUG,
+            show=self.is_debug,
             markup=markup,
             style="dim",
         )
 
+    def info_verbose(self, message="", *, prefix="", markup=False):
+        """Log message at info level if debug enabled; included if verbosity >= 1."""
+        self._log(prefix=prefix, message=message, show=self.is_debug, markup=markup)
+
```
freakboy3742, line 229: I'm wondering if the issue here isn't that debug should be "deep_debug", and "info_verbose" should be "debug"...

rmartin16:
I'll be able to respond more fully tomorrow...but I'm mostly interested in what you'd like to see at this point.

There seems to be a need to introduce a logging level between info and debug....so, that being the case, do we want to re-define what `-v` and `-vv` mean? while also introducing `-vvv` I think?

In this way, we might have:
- `-v`: verbose
- `-vv`: debug
- `-vvv`: deep debug

rmartin16:
For the time being, we can ignore the current implementation in this PR; I think I implemented it too hastily before fully understanding #1475.

I'd like your thoughts on my considerations and especially ones I've failed to consider at all.

Here's my thoughts to introduce a new logging level between `info` and `debug` called `verbose`:
- Zero `-v` flags remains info mode
  - Only calls to `Log.info()` are shown in the console
- Redefine `-v` as verbose mode
  - This would introduce `Log.verbose()` and would replace the logging in #[1475] that is gated by a `verbosity >= 1`
- Redefine `-vv` as debug mode
  - This would include all calls to `Log.debug()` including subprocess details
- Introduce `-vvv` as deep debug mode
  - This increases the verbosity for third party tools own logging
  - This would not inherently necessitate creating a `Log.deep_debug()` because we don't have any actual Briefcase logging at this level
  - Since this level of verbosity is really quite disconnected from Briefcase logging altogether....creating a dedicated flag for tool logging verbosity may make sense

To reiterate, I really don't want to merge debug in to deep debug because the third party tool logging is very noisy and I'm usually most interested in the subprocess details.

I suppose another alternative is to leave logging as is and instead summarize the warnings and errors for a user once the process completes. I was going to implement something like this for linuxdeploy logging....but it doesn't really seem as worthwhile now with deprecating AppImage. I'm also not sure if that's reasonably an option for the use-case here.

> > [edit] additionally....given that signing is multi-threaded....the debug output is indecipherable for any practical purposes...
> 
> In what way? I agree that the output _previously_ didn't work well with subprocessing because each line of output didn't include enough context to say what which file it related to; but I though I had fixed that in #1475 by including the filename every time a line is output. Unless you're seeing some sort of output where lines from different threads are being interwoven... but I haven't seen that myself.
> 
> I agree that the output is indecipherable when you turn on _deep_ debug - but I'm not sure how we'd address that short of some sort of locking around the output of the subprocess preamble...

Our terminology is becoming skewed; I was referring to the original debug mode that includes printing subprocess detail. Nonetheless, when subprocess details are included, all of the threads are printing over each other and you can't make heads or tails of it....this is also probably where I started thinking that `quiet` might be useful because printing the subprocess detail in this case isn't useful.

It's also probably pertinent to keep in mind that using `quiet` or gating logging behind a `if verbosity >=` check in the codebase means the logging doesn't appear anywhere including in the created logfile. So, we should definitely limit this to truly useless logging (like the rapid `adb.pid_exists()` calls).

Bringing in comments from #1501:
> I don't think it's entirely surprising that a change to DEBUG didn't trigger anything in the tests; log output isn't generally asserted in tests unless either the output is the point, or output is the only signal that something has been done.

Yes...but only be accident. Had the subprocess tests been properly setting `logger.verbosity` to `1` instead of `2`, then changing `Log.DEBUG` in #1475 would have caused all those tests to start failing.

> However, there's definitely some unintended damage from the DEBUG=1->2 change.
> 
> For me, I think the underlying issue is that we've got multiple competing terminology and descriptions of debug level - there's the call we use to do the debug (logger.info, logger.debug etc), and the constant that describes log level. We've also got DEBUG_PREFIX, which is really a "subprocess prefix" rather than anything related to debug.

While it's true that `DEBUG_PREFIX` is largely used to print subprocess detail, it is not regulated to that. Debug logging also includes dropping bread crumbs during Tool verification and android skin installation.

> My inclination is that rather than having `if verbosity >=` calls in the general codebase, verbosity should be checked exactly once

Completely agree; they are removed in this implementation and will be in any final implementation.

> and we should be using the appropriate `logger` call. That likely means adding a `logger.deep_debug` that is only used by subprocess and selected "way-too-much-detail" logging options (I can't think of any in particular candidates, but at least conceptually it's not *explicitly* a subprocess feature).

Included in my design above with the extra details for third party tool logging; FWIW, I think it might also be reasonable to create a new flag altogether to enable verbose third tool logging so it isn't wrapped up in the verbosity flag definition.

freakboy3742:
> Here's my thoughts to introduce a new logging level between `info` and `debug` called `verbose`:

This all makes sense to me. My only hesitation is that the naming doesn't map to traditional info/warning/debug logging labels - but those labels also don't have enough fidelity for what we're describing here, and I think the names you've proposed make sense in the context.

> To reiterate, I really don't want to merge debug in to deep debug because the third party tool logging is very noisy and I'm usually most interested in the subprocess details.

That definitely makes sense - I'd forgotten that there is a debug level "deeper" than showing the subprocess details.
 
> I suppose another alternative is to leave logging as is and instead summarize the warnings and errors for a user once the process completes. I was going to implement something like this for linuxdeploy logging....but it doesn't really seem as worthwhile now with deprecating AppImage. I'm also not sure if that's reasonably an option for the use-case here.

That's an interesting idea - although it would need a lot more housekeeping to make sure those details aren't lost. It would definitely work for the macOS case; but once we've reduced the default verbosity, the need to use an approach like this is also massively reduced. At least for now, I don't think it's worth the effort.

> > However, there's definitely some unintended damage from the DEBUG=1->2 change.
> > For me, I think the underlying issue is that we've got multiple competing terminology and descriptions of debug level - there's the call we use to do the debug (logger.info, logger.debug etc), and the constant that describes log level. We've also got DEBUG_PREFIX, which is really a "subprocess prefix" rather than anything related to debug.
> 
> While it's true that `DEBUG_PREFIX` is largely used to print subprocess detail, it is not regulated to that. Debug logging also includes dropping bread crumbs during Tool verification and android skin installation.

I think the conflict here is that for me, the `>>>` prefix is mentally linked to subprocesses - it's a visual marker that this content comes from calling an external tool. Using that prefix for tool verification isn't adding any *additional* signal to the log that can't already be derived from the `dim` setting.

Perhaps the fix here is to use `dim` for debug-level output, but only include the `>>>` prefix when a subprocess is involved?
 
> > and we should be using the appropriate `logger` call. That likely means adding a `logger.deep_debug` that is only used by subprocess and selected "way-too-much-detail" logging options (I can't think of any in particular candidates, but at least conceptually it's not _explicitly_ a subprocess feature).
> 
> Included in my design above with the extra details for third party tool logging; FWIW, I think it might also be reasonable to create a new flag altogether to enable verbose third tool logging so it isn't wrapped up in the verbosity flag definition.

An interesting idea; however, I'm having difficulty of thinking of a situation where we'd need to differentiate -vvv from "only show the debug level output from subprocesses". By the time you're running -vvv, you're already saying "something has gone badly wrong, give me everything so I can work out what broke". I don't think this is an area where having complicated interacting flags will be beneficial :-)

rmartin16:
Reworked based on our discussion:
- Formalized and abstracted logging levels
- Limited the `>>>` preface to subprocess details logging
  - The "dim" style for other debug logging is only visible in the console....but not in the logfile so it isn't as obvious this isn't "normal" output
  - However, I ultimately decided that doesn't really matter
- I tweaked some of the macOS packaging logging but I stopped short of going further...
  - First, the log file is a disaster....there is simply so much clobbered subprocess details output that it's really difficult to suss out the meaningful information
    - Given this, I'm definitely still of the opinion that logging needs to be somehow atomized/serialized....or the subprocess detail logging for these threaded processes should be disabled
  - Second, I think the requirements install logging could be improved
    - I think this is mostly a consequence of the output from pip being seamlessly combined with Briefcase's output....so, it can be hard to tell what Briefcase is doing and what pip is doing
  - Third, I noticed that ad-hoc signing happens twice....
```
[helloworld] Created build/helloworld/macos/app

[helloworld] Ad-hoc signing app...
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.0% • 00:12

[helloworld] Built build/helloworld/macos/app/Hello World.app

[helloworld] Signing app with ad-hoc identity...
Because you are signing with the ad-hoc identity, this app will run, but cannot be re-distributed.
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.0% • 00:12

[helloworld] Building DMG...
Building Hello World-0.0.1.dmg... done

[helloworld] Packaged dist/Hello World-0.0.1.dmg
```
  - Finally, once we're good with this, I will probably roll out additional use of verbose logging
    - For example, there are at least a few other places that Briefcase just lists a bunch of files its doing something with

Example log file: [briefcase.2023_10_25-13_02_58.package.log](https://github.com/beeware/briefcase/files/13168747/briefcase.2023_10_25-13_02_58.package.log)

rmartin16:
> * Finally, once we're good with this, I will probably roll out additional use of verbose logging
>    * For example, there are at least a few other places that Briefcase just lists a bunch of files its doing something with

I'll return to this later; want to get this merged first.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -711,16 +708,10 @@ def package_app(
...
+            self.logger.info("Signing app with ad-hoc identity...", prefix=app.app_name)
             self.logger.warning(
-                (
-                    "Because you are signing with the ad-hoc identity, this "
-                    "app will run, but cannot be re-distributed."
-                ),
-                prefix=app.app_name,
+                "Because you are signing with the ad-hoc identity, this "
+                "app will run, but cannot be re-distributed."
             )
```
rmartin16, line 715: This use of `prefix` looked a bit strange to me since we're not starting a new action or anything....we're just dropped a warning relevant to the immediately preceding log.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -745,11 +739,9 @@ def package_app(
                 )
                 self.notarize(self.binary_path(app), team_id=team_id)
 
-            with self.input.wait_bar(
-                f"Archiving {self.distribution_path(app).name}..."
-            ):
+            with self.input.wait_bar(f"Archiving {dist_path.name}..."):
```
rmartin16, line 742: To clarify since the diff looks like more may have changed than it did....I just added a Wait Bar here to better align with `zip` packaging and because I initially thought Briefcase froze when it got here.

freakboy3742 (approved):
This all looks great. As you note, there's opportunities to clean up other output to align with these new values, but as an initial beachhead (that does the unintended damage from my PR), this is great.

The only suggestion I've got is that we should probably clarify the contract for the `-v` option in the docs - but there isn't an existing documentation section for these options. I'll push an update to include this shortly.

PR #1518, opened 2023-10-31
Docs tweaks

rmartin16:
## Changes
- Stop checking hyperlinks to GitHub issues and PRs for Briefcase itself
  - Given the number of these links in the docs grows a lot for each release, `tox -e docs-lint` spends most of its time now checking links to the Briefcase repo.
  - This seems unnecessary....especially if you consider that a typo in the PR/issue number is still likely to be valid
- Updates how the platform support table is created so everything looks uniform

## PR Checklist:


docs/reference/platforms/index.rst
```diff
@@ -23,48 +23,60 @@ Platform support
 +-----+-------------------------------------+
 
 
-.. |Gradle| replace:: **Gradle**
```
rmartin16, line 26: curious....did you intentionally use non-breaking spaces for these? I feel like you did and hence, I maintained them. Although, without them, the whole table does fit better on the page.

rmartin16:
This [comment](https://github.com/beeware/briefcase/pull/1360#pullrequestreview-1601753261) talks with about switching to MyST. What's the blocker here? Just actually doing it?

docs/reference/platforms/index.rst
```diff
@@ -23,48 +23,60 @@ Platform support
...
-.. |iOS| replace:: **Xcode**
+.. |iOS| replace:: **Xcode project**
 .. _iOS: ./iOS/xcode.html
 
-.. |System| replace:: **Native System Packages**
+.. |AppImage| replace:: **AppImage**
+.. _AppImage: ./linux/appimage.html
+
+.. |Flatpak| replace:: **Flatpak**
+.. _Flatpak: ./linux/flatpak.html
```
freakboy3742, line 36: So - the reason these replace calls exist at all is so that we can make the headings bold inside a link. The Bold was intended to indicate the platform default. As long as you don't need the bold, you don't *strictly* need the replace, either.

I've tweaked to retain the replace, but remove the bolds.

freakboy3742 (approved):
All looks good. I've flagged two points of interest inline, but otherwise this looks good.

As for the MyST question - yes, it's mostly a "someone needs to do it" thing, combined with the fact that I haven't actually used MyST in practice, so I'm not sure how much of what I'm seeing about MyST is grass-is-greener optimism. 

ReST... works... but it has limitations. But we know what those are, and we've got workarounds when we hit them. MyST is somewhat of an unknown to me personally. There's plenty of social proof (other projects that have adopted it, including RTD). However, on the basis that *all* technology sucks, it's just a question of *how* it sucks, I'm a *little* wary of just switching.

That said - the other big reason to switch is that the rest of the ecosystem has very much converged on Markdown rather than ReST, and it's been made *very* clear that GitHub isn't going to improve it's ReST support any time soon, so switching to a Markdown-based documentation approach has a certain amount of "running downhill" associated with it.

I guess I would like to see a speculative port to see how disruptive a switch would be, how much we can automate rather than needing manual transformation, and how much we'll need to audit after an automated transformation. 

If, hypothetically, *someone* were so inclined :-)

docs/conf.py
```diff
@@ -115,6 +115,21 @@
...
+    r"^./linux/appimage.html$",
+    r"^./linux/flatpak.html$",
+    r"^./linux/system.html$",
+    r"^./macOS/app.html$",
+    r"^./macOS/xcode.html$",
+    r"^./web/static.html$",
+    r"^./windows/app.html$",
+    r"^./windows/visualstudio.html$",
+    r"^https://github.com/beeware/briefcase/issues/\d+$",
+    r"^https://github.com/beeware/briefcase/pull/\d+$",
```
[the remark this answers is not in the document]
rmartin16: I always forget to note when something is "intentionally provocative" and I fully expect push back...as I did here.

The only alternative I really have in mind is setting [`linkcheck_workers`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_workers) to something larger than five (5). It may at least run faster then.....as long as GitHub doesn't start returning 429s or something.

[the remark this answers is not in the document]
freakboy3742: Move fast and break things, right? :-)

In this case, I think the risk profile is pretty small. The release notes are updated once per release; it's not like a "normal" documentation page where links *could* be updated in any pull request (or an external documentation source could change). I think I can live with the ignore, since it will speed things up for *most* docs builds, and as releases build up, the list of links is going to get absurd - to the point where even N-fold parallelisation won't help (especially when you consider that we get a handful of links added every Monday because of dependabot).

docs/reference/platforms/index.rst
rmartin16: Ahhh, I see; the **bold** represents defaults. That apparently never clicked with me. I thought the differences in appearance in the table were being driving by the differences in how the links were constructed....and I rationalized that the link constructions were different because of some sphinx/rest thing I didn't understand....

Well, I may not have changed anything if I knew that before 😆

Issue #1523, opened 2023-11-03
Consolidate template repositories


rmartin16:
### What is the problem or limitation you are having?

Each cookiecutter template for Briefcase currently has its own repository....but these repositories are remarkably similar and could potentially benefit from consolidating in to a single repository.

https://github.com/beeware/briefcase-android-gradle-template
https://github.com/beeware/briefcase-iOS-Xcode-template
https://github.com/beeware/briefcase-linux-appimage-template
https://github.com/beeware/briefcase-linux-flatpak-template
https://github.com/beeware/briefcase-linux-system-template
https://github.com/beeware/briefcase-macOS-app-template
https://github.com/beeware/briefcase-macOS-Xcode-template
https://github.com/beeware/briefcase-template
https://github.com/beeware/briefcase-web-static-template
https://github.com/beeware/briefcase-windows-app-template
https://github.com/beeware/briefcase-windows-VisualStudio-template

### Describe the solution you'd like

With the release of cookiecutter v2.2.0, the ability to nest multiple templates within a single repository [became](https://github.com/cookiecutter/cookiecutter/pull/1770) a reality. Therefore, we are no longer technically limited to individual repositories.

One motivation for this is maintenance simplification. Twice recently, we had to go through every single template repo and create an identical PR to apply the same update to each one. The similar and shared nature of these repos strongly suggests to me they benefit more from sharing a repo than using their own.

One foreseeable complication is `briefcase-windows-app-template` and `briefcase-macOS-app-template` since these build and contain their stub binary. Given that `briefcase-macOS-app-template` has already [ballooned](https://github.com/beeware/briefcase/issues/933) to 221MB, this binary eventually needs to live outside of the repo; this would provide a convenient opportunity to do this.

I know a desire to ship the templates as [packages](https://github.com/beeware/briefcase/issues/1347#issuecomment-1624496296) has been expressed; although, as also expressed, using a consolidated repo for the templates may make even more sense at that point. (I'm not sure I'd like to go so far as putting the templates in the Briefcase repo, though....that's gonna complicate CI quite a bit.)

### Describe alternatives you've considered

Status quo.

### Additional context

I created this meta-issue to formalize this discussion and ideally create a rough sketch/design of the needed work. Before that, though, I'd like to capture the constraints and considerations for this move.....so, please comment them if you've got em.

mhsmith:
[...]

Issue #1166, opened 2023-04-07
Add a test mode for the web backend


freakboy3742:
### What is the problem or limitation you are having?

`--test` mode is supported on `briefcase run` on every platform *except* web. We need the ability to run test suites on web so that an app can validate runtime behavior on web.

### Describe the solution you'd like

A default Briefcase `helloworld` project should produce test output from `briefcase run web --test`.

### Describe alternatives you've considered

Don't test on web. YOLO. :-)

### Additional context

There are 2 major technical problems to be solved:
1. Passing in a different app startup arg to the PyScript startup (i.e., running `python -m tests.myapp` rather than `python -m myapp`)
2. Reading test output outside the browser

(1) could be solved by:
 - re-writing index.html for test mode
 - making index.html reference an external configuration source (e.g., a variable in pyscript.toml) when looking for the startup module
 - Making the startup script something that is a file on disk, rather than something embedded in index.html

(2) is more complex - we need to be able to capture anything written to stdout by Python in the browser, and redirect it to a location that Briefcase can see. Some options:
 - Installing a handler in Python to redirect stdout to a socket, and adding a separate server to Briefcase to read that content
 - A headless browser or browser test framework (e.g., Playwright, Selenium) that outputs the javascript console to stdout

LunaMeadows:
https://gist.github.com/LunaMeadows/233d4d1386a6e464fb1d22c6befc1c13

I am having difficulties fully understanding the code and how to get part 1 done however I do believe I have the base needed for the 2nd part. Using Selenium, I was able to start Selenium and then start a logging loop that will constantly check for the latest logs. If a delay was needed it would grab all the logs from after the last run and print them out. I am not sure how the logger stuff works so I will leave that to someone who has more knowledge of the code but this should be a starting point. 

I didn't do a headless at this point because in order to test the console output I needed an interface to access the dev console. Once stuff is being printed from python tho it should be able to be done headless from the testing I did. 

I attempted to figure out how to do it with playwright but couldn't find much documentation on reading the console output as easily as you can with Selenium.

Issue #430, opened 2020-06-22
Changing default configuration directory path to standard location


jtrakk:
Currently briefcase uses `~/.briefcase/` for saving configuration by default.

Some people, myself included, don't like the clutter and lag (especially on less-expensive hardware) caused by having large numbers of files in user home. The popular FreeDesktop standard [specifies](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html) a preferred location:

> $XDG_CONFIG_HOME defines the base directory relative to which user specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

Using this standardized location also means that there's one obvious place to look for each type of file (config, data, cache), which is a nice consistency for users like me so I don't have to remember a different location for each application.

Would the Briefcase project be willing to conform to this standard by default?

To help developers find the right location on different platforms, the tiny but widely used [`appdirs` module](https://pypi.org/project/appdirs/) provides a simple interface to the configuration, data, and cache directories.

freakboy3742:
Closing as a duplicate of #374. In short; yes, we'd be in favour of adopting the FreeDesktop specification.

PR #1524, opened 2023-11-04
Add support for third-party GUI framework plugins

rmartin16:
## Changes
- Allow new projects to be created using arbitrary third-party GUI frameworks using plugins
  - Any package can register a class as a GUI bootstrap using the `briefcase.bootstraps` entry point
  - To work with `beeware/briefcase-template`, the bootstrap must implement `BaseGuiBootstrap` interface
  - Alternatively, a registered bootstrap can implement an arbitrary interface to work in tandem with a different template specified via `--template`
- Drop support for PySide2 as its compatibility with the modern Python ecosystem has diminished

## Dependencies
- https://github.com/beeware/briefcase-template/pull/82
- https://github.com/beeware/.github/pull/68
- https://github.com/beeware/.github/pull/1542
- https://github.com/beeware/briefcase-android-gradle-template/pull/73
- https://github.com/beeware/briefcase-iOS-Xcode-template/pull/29
- https://github.com/beeware/briefcase-linux-appimage-template/pull/45
- https://github.com/beeware/briefcase-linux-flatpak-template/pull/30
- https://github.com/beeware/briefcase-linux-system-template/pull/17
- https://github.com/beeware/briefcase-macOS-app-template/pull/29
- https://github.com/beeware/briefcase-macOS-Xcode-template/pull/34
- https://github.com/beeware/briefcase-web-static-template/pull/13
- https://github.com/beeware/briefcase-windows-VisualStudio-template/pull/26
- https://github.com/beeware/briefcase-windows-app-template/pull/26

BRIEFCASE_REPO: https://github.com/rmartin16/briefcase
BRIEFCASE_REF: run-app-in-ci-testing
BRIEFCASE_TEMPLATE_REPO: https://github.com/rmartin16/briefcase-template
BRIEFCASE_TEMPLATE_REF: gui-plugin-support

## PR Checklist:


rmartin16:
PoC [success](https://github.com/beeware/briefcase/actions/runs/6757431499/job/18368005222)!!
```
[verify-toga] Starting app...
Install path: /home/runner/work/briefcase/briefcase/tests/apps/verify-toga/build/verify-toga/ubuntu/jammy/verify-toga-0.0.1/usr
Pre-initializing Python runtime...
PYTHONPATH:
- /usr/lib/python3.10
- /usr/lib/python3.10/lib-dynload
- /home/runner/work/briefcase/briefcase/tests/apps/verify-toga/build/verify-toga/ubuntu/jammy/verify-toga-0.0.1/usr/lib/verify-toga/app
- /home/runner/work/briefcase/briefcase/tests/apps/verify-toga/build/verify-toga/ubuntu/jammy/verify-toga-0.0.1/usr/lib/verify-toga/app_packages
Configure argc/argv...
Initializing Python runtime...
Running app module: verify_toga
(verify-toga:3530): dbind-WARNING **: 22:12:10.048: AT-SPI: Error retrieving accessibility bus address: org.freedesktop.DBus.Error.ServiceUnknown: The name org.a11y.Bus was not provided by any .service files
WARNING: Can't find icon resources/verify_toga; falling back to default icon
>>> successfully started...exiting <<<
```
## Notes
- A display was necessary to do the test (as well as installing gtk first)
  - The requirements here aren't as stringent as Toga so `xvfb` was enough
    - `xvfb-run briefcase run linux system`
- As I noticed with Toga testbed CI, you can't let `actions/setup-python` run or the app won't work (for Linux)
  - This may end up requiring some kind of refactoring in `app-build-verify.yml`....

## Next Steps
- Formalizing the plugin's interface
  - As I said, I'm imagining the plugin could ultimately manage everything about `briefcase new` to produce a template path and context dict.
  - However, I thinking of limiting the scope for this first go to:
    - Registering a module via an entry point
    - Declaring the name of the GUI toolkit for the app
    - Declare a function that returns a dictionary to be merged with the existing `context` dict for cookiecutter
- Establishing a plan for the "built-in" GUI frameworks
  - We could go ahead and break these off in to their separate packages and everything that goes with that
  - Or we could leave them mostly as-is and effectively become built-in plugins

Thoughts welcome on this approach....especially on this high-level design and scope.

rmartin16:
PoC [success](https://github.com/beeware/briefcase/actions/runs/6757431499/job/18368005222)!!
```
[verify-toga] Starting app...
Install path: /home/runner/work/briefcase/briefcase/tests/apps/verify-toga/build/verify-toga/ubuntu/jammy/verify-toga-0.0.1/usr
Pre-initializing Python runtime...
PYTHONPATH:
- /usr/lib/python3.10
- /usr/lib/python3.10/lib-dynload
- /home/runner/work/briefcase/briefcase/tests/apps/verify-toga/build/verify-toga/ubuntu/jammy/verify-toga-0.0.1/usr/lib/verify-toga/app
- /home/runner/work/briefcase/briefcase/tests/apps/verify-toga/build/verify-toga/ubuntu/jammy/verify-toga-0.0.1/usr/lib/verify-toga/app_packages
Configure argc/argv...
Initializing Python runtime...
Running app module: verify_toga
(verify-toga:3530): dbind-WARNING **: 22:12:10.048: AT-SPI: Error retrieving accessibility bus address: org.freedesktop.DBus.Error.ServiceUnknown: The name org.a11y.Bus was not provided by any .service files
WARNING: Can't find icon resources/verify_toga; falling back to default icon
>>> successfully started...exiting <<<
```
## Notes
- A display was necessary to do the test (as well as installing gtk first)
  - The requirements here aren't as stringent as Toga so `xvfb` was enough
    - `xvfb-run briefcase run linux system`
- As I noticed with Toga testbed CI, you can't let `actions/setup-python` run or the app won't work (for Linux)
  - This may end up requiring some kind of refactoring in `app-build-verify.yml`....

## Next Steps
- Formalizing the plugin's interface
  - As I said, I'm imagining the plugin could ultimately manage everything about `briefcase new` to produce a template path and context dict.
  - However, I thinking of limiting the scope for this first go to:
    - Plugin registers a module via an entry point
    - Interface specifies a module variable as the name of the GUI toolkit for the app
    - Interface specifies a function that returns a dictionary to be merged with the existing `context` dict for cookiecutter
- Establishing a plan for the "built-in" GUI frameworks
  - We could go ahead and break these off in to their separate packages and everything that goes with that
  - Or we could leave them mostly as-is and effectively become built-in plugins

Thoughts welcome on this approach....especially on this high-level design and scope.

freakboy3742:
> ## POC Design
> * This all fundamentally assumes that cookiecutter will be responsible for creating the app project

That seems like a reasonable assumption, unless we *fully* divest all responsibility for template app generation to the backend.

> * A GUI framework plugin is declared via the `briefcase.wizard.frameworks` entry point
>   
>   * Such a plugin currently defines:
>     
>     * `GUI_TOOLKIT_NAME`: a string defining the GUI toolkit such as `Toga`
>       
>       * This value is passed in the cookiecutter `context` dict as `gui_framework`
>     * `app_source`: a string defining the content of `app.py` for the app project
>   * This is a very preliminary interface for such a plugin

The GUI framework is probably needed for identification purposes, but I'm not 100% convinced it's needed in the template context. If we have it in the template context, then we're still going to have "if toga..." logic in the app template, which won't actually make our lives that much easier. If the framework plugin exposes all the places that are currently `{% if gui_framework == %}` template blocks, that will make for a much cleaner template that is more closely aligned with what the template actually needs to describe (the basic shape of a Briefcase project) rather than needing to encode framework/platform specifics.
     
>     * I am potentially imagining the ability for a plugin to manage the entire control flow for the `briefcase new` command and would be responsible for providing a cookiecutter template path and context dict
>     * Of course, much of this could be deferred to Briefcase's current functionality by default

I can see how this could work, but it would essentially require picking the framework *first*, so the right framework backend would run the app generation logic. 

I'm also a little wary of YAGNI here. Yes, we *could* allow any backend to completely customise the app generation workflow... but is that actually required? Asking additional questions, maybe - but changing the logic of how an app_name is selected? I'm having difficulty imagining why this would be required.
 
> ## Notes
> * Insofar as actually running the app in CI, I'm not sure if manifesting an entire display is necessary....
>   
>   * Or if just running the app without drawing it is sufficient...
>   * It looks like Toga goes through quite the effort to draw the testbed in a display...

I'm not sure "without drawing" is actually a test. We need to validate that a GUI window can be displayed by the running app. 

> * Adding support for GUI framework plugins could remove the need for beeware/briefcase-template altogether
>   
>   * This would mean that any Framework plugin must wholly define the cookiecutter template for the app
>   * Although, this would mean we basically create several "briefcase-template" repos....just each one would be dedicated to a specific GUI framework....but that's exactly what I'm suggesting we undo for the output format repos...sooo...

An interesting idea, but I wonder if that just bifurcates our maintenance overhead. For example - let's say we add a new top-level app property. Right now, we add that to the template once. If we give each framework update its own base template, we'd need to update that property on *every* base template. 

I think the better approach is to provide the mechanism to cross-mix the two. Provide a base app template that covers the common structure, but also provides the extension points so that frameworks can inject platform-specific details as required.

> ## Next Steps
> * Formalizing the plugin's interface
>   
>   * As I said, I'm imagining the plugin could ultimately manage everything about `briefcase new` to produce a template path and context dict.
>   * However, I thinking of limiting the scope for this first go to:
>     
>     * Plugin registers a module via an entry point
>     * Interface specifies a module variable as the name of the GUI toolkit for the app
>     * Interface specifies a function that returns a dictionary to be merged with the existing `context` dict for cookiecutter
>       
>       * Such an approach would include a dict entry for the source code to put in to the app; however, I don't think that content could contain jinja template syntax to be replaced....although, it might be reasonable to say a full replacement cookiecutter template should be used in that case...

I agree that it would be good to avoid meta-templating... but I'm not sure how we can avoid at least *some* meta templating, because we need to include the app name in the generate app content.

> * Establishing a plan for the "built-in" GUI frameworks
>   
>   * We could go ahead and break these off in to their separate packages and everything that goes with that
>   * Or we could leave them mostly as-is and effectively become built-in plugins
> 
> Thoughts welcome on this approach....especially on this high-level design and scope.

My initial inclination is to provide them in the Briefcase repo, and register them as part of installing Briefcase. As you've noted in #1523, the one thing we don't need is *more* repositories :-)

rmartin16:
> If the framework plugin exposes all the places that are currently `{% if gui_framework == %}` template blocks, that will make for a much cleaner template that is more closely aligned with what the template actually needs to describe (the basic shape of a Briefcase project) rather than needing to encode framework/platform specifics.

Thanks for this bit; it's definitely helped evolve my thinking on this.

To that end, a lot of the plugin needs to provide content for `pyproject.toml`. Since the TOML datatypes have more or less a one-to-one correspondence with Python datatypes, we could just expect the plugin to return them.

For instance, `requires` is just a list, so the plugin could just return `["PySide6-Essentials~=6.5", "PySide6-Addons~=6.5"]` as a Python `list` and Briefcase could format that in to the TOML.

Alternatively, the plugin could return a string of the same information and Briefcase could insert it verbatim in to the TOML. The advantage of this approach is metadata that will be difficult to capture in a Python datatype.

For instance, consider `system_requires` under the `linux.system.debian` section:
```
system_requires = [
    # Needed to compile pycairo wheel
    "libcairo2-dev",
    # Needed to compile PyGObject wheel
    "libgirepository1.0-dev",
]
```

These comments are valuable to end-users to understand why it's included in their rolled out project.

Therefore, my inclination is to allow plugins to do either and return the format that makes sense for them.

@freakboy3742, do you have any opinions or additional insight for this before I move forward? Thanks.

freakboy3742:
> Therefore, my inclination is to allow plugins to do either and return the format that makes sense for them.
> 
> @freakboy3742, do you have any opinions or additional insight for this before I move forward? Thanks.

`str | list` (or `str | dict[str, str]`, or whatever is appropriate) seems entirely workable, but I wonder if it's worth the effort to support both. The use case for returning `str` is "this definition is complex and needs explanatory notes"; that implies the use case for `list` is "this definition is simple and self evident" - which also means it's the case that will benefit least from the "strong typing" benefit of returning a list. 

The only other thought I've had is that we could switch to [tomlkit](https://tomlkit.readthedocs.io/en/latest/), which is a style-preserving TOML library. It allows you to annotate comments onto nodes in the TOML document... but it (a) only works on TOML, and (b) would require all plugins to *also* use tomlkit. However, this gets us to the point of supporting str, native types *and* tomlkit "DOM" nodes as inputs... which is even more complexity for even less gain. 

Honestly - returning a simple `str` seems like the best, and most flexible option.

rmartin16:
I pivoted this PR to just focus on adding support for third-party GUI framework plugins. A subsequent PR will build on this to support running apps from `app-build-verify` in CI.

rmartin16:
A few initial discussion points:

- Entry point name: `briefcase.wizard.frameworks`
  - Need to confirm this name. It originated in https://github.com/beeware/briefcase/issues/1288.
  - I'm not sure I'm a fan of "wizard" per se....
  - Perhaps along the lines of `briefcase.platforms`....this could be `briefcase.gui_frameworks`
- New directory hierarchy
  - I created a top-level directory named `plugins`....but it does seem a bit weird given this is the first-party repo
  - I'm considering just getting rid of it and moving the `frameworks` directory up under `src/briefcase`
- GUI framework plugin implementation
  - I wanted a way to define a prototype for the plugin...hence the `BaseGuiPlugin` class
  - I didn't add `abstractmethod` to every method because I didn't want to force every subclass to implement every single method
- Defaulting for `briefcase-template` context fields
  - I intentionally left meaningful defaults out of `briefcase-template`....because otherwise any changes here would be picked up by third-party plugins....and that doesn't seem appropriate...they should fully define their requirements

freakboy3742:
> A few initial discussion points:
> 
> * Entry point name: `briefcase.wizard.frameworks`
>   
>   * Need to confirm this name. It originated in [Create new project with Positron template (Django or Static) #1288](https://github.com/beeware/briefcase/issues/1288).
>   * I'm not sure I'm a fan of "wizard" per se....
>   * Perhaps along the lines of `briefcase.platforms`....this could be `briefcase.gui_frameworks`

FWIW, I don't *hate* wizard as a name - my biggest objection is that it's not clear that it's the "new project" wizard (not that we have other wizards, but in theory we *could*).

> * New directory hierarchy
>   
>   * I created a top-level directory named `plugins`....but it does seem a bit weird given this is the first-party repo
>   * I'm considering just getting rid of it and moving the `frameworks` directory up under `src/briefcase`

Agreed that it would make more sense in the `src` folder, rather than something standalone. 

Whatever name we choose, I think it should align with the name we use for the entry point.

I'm a little hesitant on "frameworks", because (a) it's arguable whether a GUI library is a "framework", and (b) what we're doing here is a little more that just the framework. The "Positron" use case is "Toga, but with a very specific project layout and bootstrap". "Toga" is a GUI framework; "Positron" is then a layer on top of Toga. But we need to be able to distribute "Positron" independent of "Toga", and the two need to be able to co-exist. For that matter, Toga itself might end up offering multiple plugins - one for a "simple" app, one for a Document app, one for a system tray app, and so on.

I'm wondering if "bootstrap" (`briefcase.bootstraps` / `src/briefcase/bootstraps` or `src/bootstraps`) might be a good name - talking about a "Toga bootstrap" or a "Positron bootstrap" seems a reasonable description of what its doing in the context of a Briefcase project lifecycle, without being tied to the implementation (wizard) or being tied to a the name of a project/framework. Having multiple "Toga bootstraps" with different options wouldn't be out of place either.

> * GUI framework plugin implementation
>   
>   * I wanted a way to define a prototype for the plugin...hence the `BaseGuiPlugin` class
>   * I didn't add `abstractmethod` to every method because I didn't want to force every subclass to implement every single method

Honestly, I'd be surprised if there *wasn't* a base class :-) Using the "default empty implementation" approach seems a reasonable approach.

rmartin16:
> I'm wondering if "bootstrap" (`briefcase.bootstraps` / `src/briefcase/bootstraps` or `src/bootstraps`) might be a good name - talking about a "Toga bootstrap" or a "Positron bootstrap" seems a reasonable description of what its doing in the context of a Briefcase project lifecycle, without being tied to the implementation (wizard) or being tied to a the name of a project/framework. Having multiple "Toga bootstraps" with different options wouldn't be out of place either.

This is a good choice to me; getting away from both "wizard" and "framework" sheds a lot of out-of-place connotations.

This PR is stable at this point. I may add some more tests or tweak things with more testing. I now need to get `briefcase-template` tests working again.....I don't think I even know it had tests...

I've been using this [example plugin](https://github.com/rmartin16/briefcase-plugin-test) for my testing....which, IMO, also nicely demonstrates how easy it'll be to extend the default Toga project.

I'm not super happy with the "bootstrap prompt" code....it isn't _too_ terrible but it's certainly a little awkward.

src/briefcase/bootstraps/base.py
```diff
@@ -0,0 +1,142 @@
...
+        "linux_flatpak_runtime_version",
+        "linux_flatpak_sdk",
+        "windows_requires",
+        "ios_requires",
+        "ios_supported",
+        "android_requires",
+        "android_supported",
+        "web_requires",
+        "web_supported",
+        "web_style_framework",
```
freakboy3742, line 53: This one should hopefully disappear - #1166 includes code that makes the style framework used by web deployments a feature of the GUI toolkit.

However that raises another question - details in the overall comment.

freakboy3742 (changes requested):
The broad strokes of this look really promising.

One idea/suggestion: What happens when a bootstrap needs an option that isn't on this list, or doesn't want to include a setting that is? Say they need to inject a Dockerfile fragment on Linux, or omit the `web_style_framework` setting completely - this setup won't let the bootstrap do that. Unless our API blesses a specific setting, they can't include it; and if we bless it, it *must* exist.

I'm wondering if it might be better to make the backends provide the entire configuration block for a platform, rather than explicitly enumerating each setting we think is important. And then, if the bootstrap doesn't define a configuration endpoint (e.g., PySide on iOS/Android/Web), the template outputs "not supported". That way we've only got 1 API point per deployment target (6... or 12 if we explicitly include the various Linux options), rather than one for each possible setting; and the expansion path for the future will be "one API endpoint for each new platform we support", rather than "one per configuration item that we think is important".

It might also be worth passing in the Briefcase version somewhere, so that third-party plugins can adapt to settings that need to exist (or not exist) on different Briefcase versions. I guess this *could* be extracted from the context that is passed to constructor, but for that particular detail, it might be better to be explicit.

rmartin16:
Those are good points.

I was also considering if there should be a plugin method for asking additional questions....although, technically, a plugin could drop in more user prompts anywhere.

And was thinking about adding entirely new sections to the TOML....probably most useful if a plugin is also being used for platform/format.

Mostly note to myself....but I'll want to confirm how Briefcase responds to commands for a section if a plugin leaves that section blank....probably want to avoid Briefcase bombing out too badly.

freakboy3742:
> Those are good points.
> 
> I was also considering if there should be a plugin method for asking additional questions....although, technically, a plugin could drop in more user prompts anywhere.

True - but having an obvious "this is where the questions go" point would (a) be good ergonomics from the perspective of the user, and (b) means there's a clear and obvious point in workflow where the question will be asked. Tying questions to the "linux RHEL system" output generation is going to lead to some... interesting side effects.
 
> And was thinking about a plugin adding entirely new sections to the TOML....probably most useful if a plugin is also being used for a custom platform/format as well.

That's a good point - not much point encouraging people to write their own plugins if we can't support them out of the box.

> Mostly note to myself....but I'll want to confirm how Briefcase responds to commands for a section if a plugin leaves that section blank....probably want to avoid Briefcase bombing out too badly.

AFAIK this should work. If a section doesn't exist, it won't cause a problem unless you ask for it - and even then, I *think* most cases it will fill it with an empty configuration. I've done this for some of the template tests - no need for a full Windows configuration when you're testing the macOS Xcode template.

rmartin16:
The "new project" templating is now almost entirely generalized.

Only project content that Briefcase requires in the `pyproject.toml` cannot be overridden with a GUI bootstrap plugin. However, any table in the TOML can be amended to and any of the output format tables can be arbitrarily written or omitted altogether allowing for an entirely custom `pyproject.toml` to be created.

The source code layout is still the same with a bootstrap plugin but the content for `__main__.py` and `app.py` can be completely custom.

setup.cfg
```diff
@@ -112,6 +112,12 @@ where = src
 [options.entry_points]
 console_scripts =
     briefcase = briefcase.__main__:main
+briefcase.bootstraps =
+    Toga = briefcase.bootstraps.toga:TogaGuiBootstrap
+    PySide2 = briefcase.bootstraps.pyside2:PySide2GuiBootstrap
+    PySide6 = briefcase.bootstraps.pyside6:PySide6GuiBootstrap
+    PursuedPyBear = briefcase.bootstraps.pursuedpybear:PursuedPyBearGuiBootstrap
+    Pygame = briefcase.bootstraps.pygame:PygameGuiBootstrap
```
rmartin16, line 120: I ultimately removed the class attribute `name` from the bootstrap class definition. That means that the name defined here for the entry point is also used for display purposes. In general, I'd expect the entry point name and the GUI display name to be the same....and that makes keeping track of them easier....but we can separate the concepts if that's a better design choice.

src/briefcase/bootstraps/base.py
```diff
@@ -0,0 +1,115 @@
...
+        "pyproject_table_linux_system_suse",
+        "pyproject_table_linux_system_arch",
+        "pyproject_table_linux_appimage",
+        "pyproject_table_linux_flatpak",
+        "pyproject_table_windows",
+        "pyproject_table_iOS",
+        "pyproject_table_android",
+        "pyproject_table_web",
+        "pyproject_extra_content",
+    ]
```
rmartin16, line 47: I'm not a huge fan of having to define the context fields here....since this needs to stay in sync with the methods below as well as the list in `cookiecutter.json`....but they need to be enumerated somewhere so they can be arbitrarily iterated by Briefcase.

Furthermore, though, this allows a third-party plugin to override the supported fields and implement entirely different methods that'd Briefcase run to build `context`.

rmartin16, line 47: Interested in feedback of prefacing many of these with the word `pyproject`.

My goal was to be as explicit as possible about what these fields are used for. A potential concern is this will create complications if a "new project" template doesn't use `pyproject.toml` in the future....since "pyproject" wouldn't really be an appropriate preface then. However, I think using a new preface (and therefore, all entirely new context fields) makes sense then to clearly communicate these fields are being used completely differently now (even if they are conceptually doing the same thing for the project).

src/briefcase/bootstraps/pursuedpybear.py
```diff
@@ -0,0 +1,150 @@
+from briefcase.bootstraps.base import BaseGuiBootstrap
+
+
+class PursuedPyBearGuiBootstrap(BaseGuiBootstrap):
+    def app_source(self):
+        return """\
+import importlib.metadata
```
rmartin16, line 7: To ensure this is clear....the app source for all these third-party GUI toolkits were doing a special import for `import.metdata` that pivoted on Python 3.8. Since Briefcase only supports >=3.8 now, I changed it just to import `importlib.metadata` always now.

src/briefcase/bootstraps/pursuedpybear.py
```diff
@@ -0,0 +1,150 @@
...
+    def pyproject_table_linux(self):
+        return """
+requires = [
+]
+"""
+
+    def pyproject_table_linux_system_debian(self):
+        return """
+system_requires = [
+# ?? FIXME
```
rmartin16, line 77: As much as possible, I tried to re-create the existing behavior (such as including these `FIXME` comments)....so, if we want to make more targeted changes to the templates based on the GUI, now is definitely a good time to recommend those.

src/briefcase/platforms/__init__.py
```diff
@@ -1,12 +1,11 @@
...
-    from importlib_metadata import entry_points
-except ImportError:  # pragma: no-cover-if-lt-py310
+import sys
+
+if sys.version_info >= (3, 10):  # pragma: no-cover-if-lt-py310
     from importlib.metadata import entry_points
+else:  # pragma: no-cover-if-gte-py310
+    # Before Python 3.10, entry_points did not support the group argument;
+    # so, the backport package must be used on older versions.
+    from importlib_metadata import entry_points
```
rmartin16, line 8: I only updated this to be compatible with something that `pyupgrade` can understand and handle for us.

src/briefcase/commands/new.py
```diff
@@ -427,29 +450,63 @@ def build_app_context(self):
...
+            try:
+                bootstrap_class = bootstraps[bootstrap_choice]
+            except KeyError:
+                bootstrap_class = bootstraps[bootstrap_choice.split(" ")[0]]
+
+            bootstrap = bootstrap_class(context=context)
+
+            if hasattr(bootstrap, "additional_context"):
+                if (additional_context := bootstrap.additional_context()) is not None:
+                    gui_context.update(additional_context)
```
rmartin16, line 488: I know the preference is to catch `AttributeError` instead of using `hasattr()`....but it made the code quite cumbersome.

```python
try:
    additional_context = bootstrap.additional_context
except AttributeError:
    pass
else:
    if (value := additional_context()) is not None:
        gui_context.update(value)
```
I felt it necessary to do all this to avoid squashing an `AttributeError` from the plugin.

So, I just went with `hasattr()`.

src/briefcase/commands/new.py
```diff
@@ -427,29 +450,63 @@ def build_app_context(self):
...
+
+            for context_field in bootstrap.fields:
+                if (context_value := getattr(bootstrap, context_field)()) is not None:
+                    gui_context[context_field] = context_value
```
rmartin16, line 492: To be explicit, returning `None` from the plugin is intended to be special case since it means "default to the cookiecutter template".

src/briefcase/commands/new.py
```diff
@@ -427,29 +450,63 @@ def build_app_context(self):
...
+            options=bootstrap_choices,
+        )
+
+        gui_context = {}
+
+        if bootstrap_choice != "None":
+            try:
+                bootstrap_class = bootstraps[bootstrap_choice]
+            except KeyError:
+                bootstrap_class = bootstraps[bootstrap_choice.split(" ")[0]]
```
rmartin16, line 482: "There must be a better way"

I didn't _want_ to do a bunch of refactoring here....so, I forced all this to work with what existed and behave the way it did before. There's just a bunch of cascading effects if doesn't behave like before.

I think a big improvement would be string matching on the input choice; so, instead of requiring the input to be the index of the choice, allow an input of "To" to match on Toga. This is obviously non-trivial and I decided not to bite it off here.

That said, I'm open to ideas for improving.

tests/bootstraps/test_base.py
rmartin16, line 1: So, I wanted to add some more tests here for the bootstraps....but I'm not sure how.

I was going to run `NewCommand` and verify the created files are what's expected....but that puts a dependency on the `briefcase-template` repo such that changes there could just break Briefcase's tests.

Then maybe it's reasonable to confirm the context passed to cookiecutter is what's expected....but that's already being tested in `tests/commands/new/test_build_context.py`.

src/briefcase/commands/new.py
```diff
@@ -427,28 +450,68 @@ def build_app_context(self):
...
```
rmartin16, line 483: "There must be a better way"

I didn't _want_ to do a bunch of refactoring here....so, I forced all this to work with what existed and behave the way it did before. There's just a bunch of cascading effects if doesn't behave like before (e.g. notably, this ordering of GUIs is necessary for all of CI work).

I think a big improvement would be string matching on the input choice; so, instead of requiring the input to be the index of the choice, allow an input of "To" to match on Toga, for instance. This is obviously non-trivial and I decided not to bite it off here.

That said, I'm open to ideas for improving.

rmartin16:
This and https://github.com/beeware/briefcase-template/pull/82 are ready for review. Except where it made sense, I tried to stick to existing behavior. So, if you want to see something changed around creating a new Briefcase project, now may be the time.

src/briefcase/platforms/__init__.py
freakboy3742: Ack.

freakboy3742 (changes requested):
I haven't done a fine-toothed check of the individual bootstrap templates yet, but the rest of this is starting to look fairly good. I've raised a couple of questions inline.

src/briefcase/commands/new.py
rmartin16: > If I'm understanding the problem, that _should_ address the testing issue - the list of plugins should be predictable in a CI setting. If I've missed something, let me know.

Not just predictable; the first 6 choices must be exactly known prior to CI running.

For instance, the list could be predictably alphabetical...but that isn't good enough for CI since it [injects](https://github.com/beeware/.github/blob/81b95eff757fcf485c1a8a2dfbdad41c2b97855c/.github/workflows/app-build-verify.yml#L145) the index for the option in to `stdin` to choose which GUI toolkit to use. Therefore, the list must be exactly what we have here.

This is also further complicated by the fact that "None" is last. In order for Briefcase to test the "None" case, no other plugins can be installed in to the environment....otherwise, option 6 is no longer "None". Technically, this is already a problem for running `pytest` outside of `tox`.

That's why I suggested that string matching on the name of choice instead of only the index of the option may improve this situation.

Now, admittedly...my nitpicking here over "predictable" vs "exactly known" is only relevant if other plugins are installed in to the environment since that's the only thing that could change this list from what it'd be by default. Right now, that isn't a problem for CI. However, the next step I'm planning for this work is to install a dedicated "Briefcase Automation" plugin in to CI so we can test running the helloworld apps there.

Sooo, explicitly controlling the listing of the GUI toolkits options like this is the most stable way to present this list....in that it ensures:
- The built-in plugins are always the first 5 choices in a specific order 
- Any additional plugin that's installed will be option 6
- As long as no other plugins are installed, "None" will be option 6

> If the issue is converting "3" into whichever bootstrap is third - IIRC we can pass a dictionary into input_select, and the selected option that is returned is the key from the selected choice. That's used for the Android device selector - you select item "3", but get back the device ID for the target you select. That should allow us to use the class name of the bootstrap as the key.

You're thinking of the [other](https://github.com/beeware/briefcase/blob/a0b30978380962d2eb0ac717128f9f0cf496d3da/src/briefcase/console.py#L669) "input select" method :) which takes me back actually....because one of my first PRs was an [attempt](https://github.com/beeware/briefcase/commit/e54a496406e34a1e3edf8f2ca430801e50aaebda#diff-c8b948515fc59ab4b4fe3edc9076b0c51e8c3f70ffc05e72e7e2b7a0eacdaab3) to corral input solicitation...but that was ultimately toned down.

Nonetheless, I'll see about switching this over to `briefcase.console.select_option()` which does allow the options to be a dictionary.

src/briefcase/bootstraps/base.py
```diff
@@ -0,0 +1,115 @@
...
+        "pyproject_table_linux_system_rhel",
+        "pyproject_table_linux_system_suse",
+        "pyproject_table_linux_system_arch",
+        "pyproject_table_linux_appimage",
+        "pyproject_table_linux_flatpak",
+        "pyproject_table_windows",
+        "pyproject_table_iOS",
+        "pyproject_table_android",
+        "pyproject_table_web",
+        "pyproject_extra_content",
```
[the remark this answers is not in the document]
rmartin16: Good call. Consistently using `extra` also aligns with `dockerfile_extra_content`.

src/briefcase/commands/new.py
```diff
@@ -427,28 +450,68 @@ def build_app_context(self):
...
 
+    def build_gui_context(self, context: dict[str, str]) -> dict[str, str]:
+        """Build context specific to the GUI toolkit."""
+        bootstraps = get_gui_bootstraps()
+        bootstrap_choices = [
+            "Toga",
+            "PySide2       (does not support iOS/Android deployment)",
+            "PySide6       (does not support iOS/Android deployment)",
+            "PursuedPyBear (does not support iOS/Android deployment)",
+            "Pygame        (does not support iOS/Android deployment)",
```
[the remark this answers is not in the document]
rmartin16: yeah....i was waffling between whether to include this in scope...

src/briefcase/bootstraps/base.py
```diff
@@ -0,0 +1,115 @@
...
+class BaseGuiBootstrap(ABC):
+    """Definition for a plugin that defines a new Briefcase app."""
+
+    fields: list[str] = [
+        "app_source",
+        "app_start_source",
+        "pyproject_table_briefcase_additional",
+        "pyproject_table_briefcase_app_additional",
+        "pyproject_requires",
+        "pyproject_test_requires",
```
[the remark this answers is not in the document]
rmartin16: ehh...no real reason other than the organic evolution of these changes. I think it does make sense to get rid of these, though.

src/briefcase/commands/new.py
rmartin16: Unfortunately, `briefcase.console.select_option()` isn't going to work because it doesn't respect the order of the options from the caller.

rmartin16: scratch that...I remembered that the Android device list always has the "running devices" first...so I must have missed something. A closer looks shows that only choices in dictionaries are sorted; a list of tuples is not.

src/briefcase/commands/new.py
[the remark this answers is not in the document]
rmartin16: I'm hesitant to add some sort of meaningful attributes to describe platform support at this time. Instead, I just added an attribute for annotating the display name of the bootstrap when presenting the user with the options.

tests/bootstraps/test_base.py
rmartin16: REF: https://github.com/beeware/briefcase-template/pull/82#discussion_r1390656804

I got an idea. What if I create a new workflow in `beeware/.github` called `app-create-verify.yml`. This workflow would be responsible for creating new projects for the built-in GUIs and validating that they are coherent and correct.

In this way, both this repo and `beeware/briefcase-template` could run them in CI. This would avoid the situation where changes are made in one repo that breaks the CI in the other.

thoughts?

setup.cfg
freakboy3742: The place where I see this breaking is when the bootstrap isn't *just* a GUI framework - e.g., a Positron project with a Django backend. A guess a name like "Positron-Django" could work, but having the flexibility to say "Positron (Django server)" would be nice.

freakboy3742 (changes requested):
I've flagged a couple of minor things, and I want to take it for a full spin with fresh eyes tomorrow, but I think this is nearing merge-ready.

setup.cfg
rmartin16: Can you clarify what you mean exactly?

If you're concerned about more interestingly named entry points....that doesn't seem to be to problem for me.

For example:
```
briefcase.bootstraps =
    Toga = briefcase.bootstraps.toga:TogaGuiBootstrap
    PySide2 = briefcase.bootstraps.pyside2:PySide2GuiBootstrap
    PySide6 = briefcase.bootstraps.pyside6:PySide6GuiBootstrap
    PursuedPyBear = briefcase.bootstraps.pursuedpybear:PursuedPyBearGuiBootstrap
    Pygame = briefcase.bootstraps.pygame:PygameGuiBootstrap
    Pygame (Testing) = briefcase.bootstraps.pygame:PygameGuiBootstrap
```
```
What GUI toolkit do you want to use for this project?

  1) Toga
  2) PySide2                  (does not support iOS/Android deployment)
  3) PySide6                  (does not support iOS/Android deployment)
  4) PursuedPyBear            (does not support iOS/Android deployment)
  5) Pygame                   (does not support iOS/Android deployment)
  6) PursuedPyBear Automation (does not support iOS/Android deployment)
  7) PySide2 Automation       (does not support iOS/Android deployment)
  8) PySide6 Automation       (does not support iOS/Android deployment)
  9) Pygame (Testing)         (does not support iOS/Android deployment)
  10) Pygame Automation        (does not support iOS/Android deployment)
  11) Toga Automation
  12) None
```
Note that I also have a plugin installed with a bunch of entry points.

tests/bootstraps/test_toga.py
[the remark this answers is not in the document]
rmartin16: whoops...thanks. removed.

setup.cfg
freakboy3742: Huh. TIL. I thought plugin names needed to be Python identifiers.

tests/bootstraps/test_base.py
freakboy3742: Agreed that this seems worthwhile. The testing need is the same on both.

The only thing I'm wondering is whether we need to modify `briefcase new` to add command line options to assist driving this use case. The current approach of using a file as pre-canned inputs... works... but isn't especially robust.

freakboy3742 (changes requested):
Everything here at the moment works really well. The only open questions at this point are removing PySide2, and the testing strategy.

src/briefcase/bootstraps/pyside2.py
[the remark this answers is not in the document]
rmartin16: I think this makes sense. Unless I'm misunderstanding their git [history](https://code.qt.io/cgit/pyside/pyside-setup.git/log/?h=5.15.2) and active PySide2 development is happening on a different branch, it definitely appears to be on life-support. In their dev meeting [notes](https://wiki.qt.io/Qt_for_Python_Development_Notes), they explicitly mentioned PySide2 once...and it was about whether it should be updated to support a new feature...

tests/bootstraps/test_base.py
rmartin16: > The only thing I'm wondering is whether we need to modify briefcase new to add command line options to assist driving this use case. The current approach of using a file as pre-canned inputs... works... but isn't especially robust.

I think we can avoid adding all that to Briefcase itself...injecting in to `stdin` does work just fine, after all and this use-case is really only relevant to CI.

I'm thinking about introducing a new CI action to "create a new Briefcase app" with the relevant inputs that are then converted to values that can be injected in to `stdin` for Briefcase. This will allow us to abstract downstream repos from the specific values that Briefcase needs to create the app while still allowing them to control the particulars of the app to test.

rmartin16:
What's left:

- [ ] Merge #1542
- [ ] Finish #430
- [ ] Add `app-create-verify.yml` to briefcase-template
- [ ] Add `app-create-verify.yml` to briefcase
- [ ] Remove testing apps and PySide2 CI from briefcase-android-gradle-template
- [ ] Remove testing apps and PySide2 CI from briefcase-iOS-Xcode-template
- [ ] Remove testing apps and PySide2 CI from briefcase-linux-appimage-template
- [ ] Remove testing apps and PySide2 CI from briefcase-linux-flatpak-template
- [ ] Remove testing apps and PySide2 CI from briefcase-linux-system-template
- [ ] Remove testing apps and PySide2 CI from briefcase-macOS-Xcode-template
- [ ] Remove testing apps and PySide2 CI from briefcase-macOS-app-template
- [ ] Remove testing apps and PySide2 CI from briefcase-template
- [ ] Remove testing apps and PySide2 CI from briefcase-web-static-template
- [ ] Remove testing apps and PySide2 CI from briefcase-windows-VisualStudio-template
- [ ] Remove testing apps and PySide2 CI from briefcase-windows-app-template

rmartin16:
### Status of the dependency among PRs

All of the PRs are dependent on each other and likely will fail if they are not all on the same branch. The CI workflows in https://github.com/beeware/.github/pull/68 are currently hardcoded to always use the relevant pieces of all this development; this will obviously all need to be replaced prior to merging and I've tried to clearly mark them all in the PRs. But it should mean that the CI running for these PRs is accurate insofar as meaning everything's compatible.

As for merging in to `main`, these three PRs all need to go first and at once:
- https://github.com/beeware/briefcase-template/pull/82
- https://github.com/beeware/briefcase/pull/1524
- https://github.com/beeware/.github/pull/68

Once those repos look good, the template PRs can go.

freakboy3742 (approved):
Except for the beeware/.github update, this is all looking good. Marking as approved, conditional on the changes that are needed as part of the landing strategy.

In terms of the landing strategy - this is what I understand the landing strategy will be:

1. Land beeware/briefcase#1542, to expose the `--config` option.
2. Address the feedback on beeware/.github#68. That should (I think) clear up the issues currently causing test failures on briefcase and briefcase-template; but there will still be test failures on the .github PR because the briefcase install tests won't work in their current state.
3. Rerun CI on briefcase and briefcase-template. With the fixes from (1), they should pass clean.
4. Push the updates to briefcase, briefcase-template and .github to reset the references to in-progress beeware/.github repos. We know this will break CI (because the repos are co-dependent), so we don't wait for a green board.
5. Merge briefcase, briefcase-template and .github in rapid succession. There's a minor risk that .github *might* fail because we couldn't see a green board prior to merging Briefcase, but we can hot fix that as needed.
6. Update the template repos to reset the beeware/.github references, wait for green CI, then merge.
7. Open beverage of choice :-)

Have I missed or misunderstood anything in this?

rmartin16:
> Have I missed or misunderstood anything in this?

Notwithstanding the inevitable issues from merging CI changes (especially this extensive), yeah, this is the plan. Definitely feel free to ping me to be around during the merges.

rmartin16:
note to self: add changenote for dropping PySide2 support.

freakboy3742:
> > Have I missed or misunderstood anything in this?
> 
> Notwithstanding the inevitable issues from merging CI changes (especially this extensive), yeah, this is the plan. Definitely feel free to ping me to be around during the merges.

Ack. I'm guessing #1542 won't be merged until early next week (Monday my time, Sunday for you); but once that's there, I'll ping you before I start merging anything. If we're lucky, that will be Monday morning after the dependabot updates; but failing that, I'll put it as my "first thing in the morning" set of merges on a day that works for you.

freakboy3742 (approved):
Approved for final ship.

PR #1525, opened 2023-11-05
Update the origin remote URL for a cached template before using it

### LINKED ISSUE #1158: Better Handling When the Repo in A User's Cookiecutter Cache Uses a `origin` Remote Other than Expected
### Describe the bug

The current usage and handling of the cookiecutter cache operates irrespective of where the `origin` remote actually lives. This can create situations where the cache exists for the repo but the `origin` remote is not what's expected or even explicitly requested.

### Steps to reproduce

1. Delete the cookiecutter cache for `briefcase-android-gradle-template`: `rm -rf ~/.cookiecutters/briefcase-android-gradle-template`
3. Add a template to `pyproject.toml` of a forked repo, e.g.: `template = "https://github.com/rmartin16/briefcase-android-gradle-template"`
4. Run `briefcase create android`
5. Remove template specification from `pyproject.toml`
6. Run `briefcase create android` again
7. Briefcase will reuse the existing cache for the `rmartin16` repo instead of the `beeware` repo

### Expected behavior

In the least, Briefcase should warn the user the cache is using an `origin` remote that doesn't match what's being requested.

The more complete change would probably update the `origin` remote, fetch, and then checkout the head for the specified template ref.

### Screenshots

_No response_

### Environment

- Operating System: Windows 10 and pop os 22.04
- Python version: 3.10
- Software versions:
  - Briefcase: `0.3.14.dev75+g9bb5444c.d20230331`

### Logs

_No response_

### Additional context

_No response_

rmartin16:
## Changes
- The URL for the `origin` remote for cached templates is always updated before fetching now
- Closes #1158

## Notes
- Tested without internet connectivity and `set-url` doesn't require a connection (as expected)
- I'm certainly not a git expert so definitely open to overlooked concerns with doing this
  - At least for my common case, i.e. using a fork of the template, this fixes my issue
- I haven't experimented with just letting cookiecutter take the remote address to the git repo directly....but I wonder if they have fixed the problems that necessitated all this git inspection and branch checking...

## PR Checklist:

### DIFF
--- a/changes/1158.bugfix.rst
+++ b/changes/1158.bugfix.rst
@@ -1,0 +1,1 @@
+When a custom Briefcase template from a git repository is used to create an app, Briefcase now ensures that git repository is always used.
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -844,6 +844,10 @@ Did you run Briefcase in a project directory that contains {filename.name!r}?"""
                 repo = self.tools.git.Repo(cached_template)
                 # Raises ValueError if "origin" isn't a valid remote
                 remote = repo.remote(name="origin")
+                # Ensure the existing repo's origin URL points to the location
+                # being requested. A difference can occur, for instance, if a
+                # fork of the template is used.
+                remote.set_url(new_url=template, old_url=remote.url)
                 try:
                     # Attempt to update the repository
                     remote.fetch()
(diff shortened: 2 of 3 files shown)


freakboy3742 (approved):
Seems like the right approach to me, and it definitely seems to work in my testing.

rmartin16:
I was doing some additional thinking.

A consequence of just changing the remote location of `origin` is all the history and refs of the previous remote location remain in the local repo clone. Then, the `git fetch` is going to merge in all the history and refs from the new remote location.

In practice, I wouldn't expect this to matter too much since the two remote locations with the same name should have a similar history. Although, this may behave strangely if the two locations have different histories for the same ref. Additionally, it would allow checkouts for refs on the old `origin`.

These issues are especially niche given our use-case so I'm not too worried. Another approach would be to just delete the whole locally cached repo if the requested template URL if different than the current `origin` remote's URL.

I may play around with this a bit more....I'm still curious if cookiecutter has improved its cache management in the last 4-5 years such that we don't need to do all this mucking around in there.

PR #1538, opened 2023-11-13
Remove Windows Store Python testing from CI

rmartin16:
- The underlying API to download this Python is no longer available to GitHub runners
- RE: https://github.com/beeware/.github/pull/65

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
       id: coverage
       if: always() && contains('success,failure', steps.test.outcome)
       # coverage reporting must use the same Python version used to produce coverage
-      run: tox -qe coverage$(tr -d "." <<< "${{ matrix.python-version }}")
+      run: tox -qe coverage$(tr -dc "0-9" <<< "${{ matrix.python-version }}")
 
   coverage:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -103,7 +92,7 @@ jobs:
       id: coverage
       if: always() && contains('success,failure', steps.test.outcome)
       # coverage reporting must use the same Python version used to produce coverage
-      run: tox -qe coverage$(echo '${{ matrix.python-version }}' | tr -dc '0-9')
+      run: tox -qe coverage$(tr -d "." <<< "${{ matrix.python-version }}")
```
freakboy3742, line 95: Won't `3.13-dev` cause a problem here?

freakboy3742 (changes requested):
Mostly looks good; one question about the coverage target change.

.github/workflows/ci.yml
rmartin16: Yes....but that's a problem for whoever adds 3.13 support ;)

I'll restore the slightly more complicated `tr`

freakboy3742 (approved):

PR #1542, opened 2023-11-16
Add the ability to override app configurations at the command line

### LINKED ISSUE #1115: Provide command line option to override app settings

freakboy3742:
Adds a `-C` / `--config` option that allows a user to override the value of a setting at the command line.

e.g.:
```
briefcase create -C template=\"https:/example.com/template\"
```
would generate an app using the values in the pyproject.toml, except for using a different template.

This is primarily useful for internal testing setups, such as generating a test app inside a template repository, or running CI for the Apple support package in CI. However, it could also be useful to pass in other values, such as a version number.

It's likely only of any practical use in the `create` command, but it's been implemented in all "app facing" commands in case it might be useful.

Fixes #1115.

/cc @rmartin16

## PR Checklist:


src/briefcase/commands/base.py
```diff
@@ -103,6 +104,26 @@ def split_passthrough(args):
...
+    """
+    overrides = {}
+    if config_overrides:
+        for value in config_overrides:
+            try:
+                overrides.update(tomllib.loads(value))
+            except ValueError:
+                raise BriefcaseCommandError(
+                    f"Unable to parse configuration override {value}"
+                )
```
rmartin16, line 123: I think Rich's stacktrace handles this regardless but can you raise from the `ValueError`?
```suggestion
            except ValueError as e:
                raise BriefcaseCommandError(
                    f"Unable to parse configuration override {value}"
                ) from e
```

src/briefcase/commands/base.py
```diff
@@ -687,6 +720,14 @@ def add_default_options(self, parser):
 
         :param parser: a stub argparse parser for the command.
         """
+        parser.add_argument(
+            "-C",
+            "--config",
+            dest="config_overrides",
+            action="append",
+            metavar="KEY=VALUE",
+            help="Override the value of the app configuration item KEY with VALUE.",
```
rmartin16, line 729: nitpick: no period at the end to match the other options (although, I also shouldn't have added one to `-v`...)
```suggestion
            help="Override the value of the app configuration item KEY with VALUE",
```

src/briefcase/commands/run.py
```diff
@@ -248,6 +248,7 @@ def __call__(
         no_update: bool = False,
         test_mode: bool = False,
         passthrough: list[str] | None = None,
+        config_overrides: list[str] | None = None,
```
rmartin16, line 251: needs to be passed to `finalize()`?

src/briefcase/commands/update.py
```diff
@@ -68,6 +68,7 @@ def __call__(
         update_resources: bool = False,
         update_support: bool = False,
         test_mode: bool = False,
+        config_overrides: list[str] | None = None,
```
rmartin16, line 71: needs to be passed to `finalize()`?

rmartin16 (approved):
Well timed; this is meeting my [needs](https://github.com/beeware/.github/actions/runs/6892256834/job/18749136654#step:19:188) of creating specific projects in CI.
```
briefcase create linux AppImage \
    --config 'template="../../"' \
    --config 'requires=["toga-gtk==0.3.1", "PyGobject<3.46.0"]'
```
I guess I hadn't considered this before but...apparently attributes can have spaces.
```
briefcase package -C '"zzz template"={"asdf"=42}'
```
I gave up trying to get a newline character in to the value for `-C`....but that shouldn't matter since you can pass multiple values for `-C`.

👍🏼

src/briefcase/commands/base.py
mhsmith: That should be unnecessary: exceptions raised within an except block are automatically chained unless you say `from None`.

mhsmith:
@rmartin16: I'm guessing from your comments that you didn't intend to approve this yet?

src/briefcase/commands/base.py
```diff
@@ -103,6 +104,26 @@ def split_passthrough(args):
...
+    :param config_overrides: The values passed in as configuration overrides. Each value
+        *should* be a "key=<valid TOML>" string.
+    :returns: A dictionary of app configuration keys to override and their new values.
+    :raises BriefcaseCommandError: if any of the values can't be parsed as valid TOML.
+    """
+    overrides = {}
+    if config_overrides:
+        for value in config_overrides:
+            try:
+                overrides.update(tomllib.loads(value))
```
mhsmith, line 119: I can't see any mention of multi-level config values, like `--config 'linux.requires=...'`, which is [valid TOML](https://toml.io/en/v1.0.0#keys).

I think this code might actually work with that as long as you only had one of them. But then if you added `--config linux.something_else=...`, the second `linux` would overwrite the first instead of merging with it.

We don't necessarily need to support that syntax in this PR, but if we don't support it, we should at least display a useful error when the user tries to use it.

rmartin16:
I meant to....but selecting "comment" instead probably would have been a better choice perhaps. I was mostly trying to communicate "don't feel required to send this back to me with changes".

rmartin16 (commented):
updating status to comment

src/briefcase/commands/base.py
freakboy3742: Interesting - the very next option defined (`--verbosity`) *did* have a period. I've corrected both.

src/briefcase/commands/base.py
freakboy3742: Might as well be explicit.

src/briefcase/commands/base.py
freakboy3742: Good call. Right now, it's probably easier to reject multi-level keys completely. We can always add them later if there's a reasonable use case.

freakboy3742: Actually - in the process of looking at this, it occurred to me that this approach avoids validation of any provided value... and when I went looking to see how I could do validation, I found a *much* cleaner way to handle the overrides, in a way that both validates *and* avoids the issue @rmartin16 found with not passing the overrides to some commands.

freakboy3742:
I've updated the PR so that any command-line overrides go through the same validation process as if they were specified in pyproject.toml. This ends up being a lot less invasive on commands, in addition to the validation benefit. 

I've also beefed up the validation that is done on the keys, ensuring that deep keys are rejected.

src/briefcase/commands/run.py
freakboy3742: Yes, but not any more with the revised approach that I've implemented.

src/briefcase/commands/update.py
freakboy3742: Again, Yes, but not any more with the revised approach that I've implemented.

mhsmith:
@rmartin16: FYI, if you want to cancel an approval, you have to choose "Dismiss review" under "Changes approved" at the bottom of the page.

mhsmith (approved):

PR #1547, opened 2023-11-20
Set runtime icon for PyGame apps

rmartin16:
## Changes
- The default runtime icon for PyGame apps is now replaced with Brutus

## PR Checklist:

### DIFF
--- a/src/briefcase/bootstraps/pygame.py
+++ b/src/briefcase/bootstraps/pygame.py
@@ -9,6 +9,7 @@ class PygameGuiBootstrap(BaseGuiBootstrap):
 import importlib.metadata
 import os
 import sys
+from pathlib import Path
 
 import pygame
 
@@ -18,14 +19,14 @@ WHITE = (255, 255, 255)
 
 
 def main():
-    # Linux desktop environments use app's .desktop file to integrate the app
-    # to their application menus. The .desktop file of this app will include
-    # StartupWMClass key, set to app's formal name, which helps associate
+    # Linux desktop environments use an app's .desktop file to integrate the app
+    # in to their application menus. The .desktop file of this app will include
+    # the StartupWMClass key, set to app's formal name. This helps associate the
     # app's windows to its menu item.
     #
-    # For association to work any windows of the app must have WMCLASS
-    # property set to match the value set in app's desktop file. For PPB this
-    # is set using environment variable.
+    # For association to work, any windows of the app must have WMCLASS property
+    # set to match the value set in app's desktop file. For PyGame, this is set
+    # using the SDL_VIDEO_X11_WMCLASS environment variable.
 
     # Find the name of the module that was used to start the app
     app_module = sys.modules["__main__"].__package__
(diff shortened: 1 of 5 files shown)


src/briefcase/bootstraps/pygame.py
```diff
@@ -34,6 +35,13 @@ def main():
 
     os.environ["SDL_VIDEO_X11_WMCLASS"] = metadata["Formal-Name"]
 
+    # Set the app's runtime icon
+    pygame.display.set_icon(
+        pygame.image.load(
+            pkg_resources.resource_filename(app_module, f"resources/{app_module}.png")
+        )
+    )
```
rmartin16, line 43: curious....is there a reason we fetch all of this information at runtime?

When we roll this template out, cookiecutter will have these pieces of information to inject directly in to the source code.

Was there anything preventing doing this originally?

src/briefcase/bootstraps/pygame.py
```diff
@@ -8,6 +8,7 @@ def app_source(self):
         return """\
 import importlib.metadata
 import os
+import pkg_resources
```
freakboy3742, line 11: pkg_resources is a [deprecated API](https://setuptools.pypa.io/en/latest/pkg_resources.html) (and is setuptools specific to boot); preference is to use `importlib.metadata`.

src/briefcase/bootstraps/pygame.py
freakboy3742: What's your precedent for that statement? Toga and PySide don't set an icon; PPB uses a hard-coded path from template data. 

They all use importlib metadata to get the app's *formal* name (Toga's app bootstrap doesn't expose the code, but that's how it's done internally); that's essentially to maintain a single source of metadata truth - the app's formal name can change but the code doesn't need to. However, I think you're right that we know where the resource path is, so we can hard code it. 

The only thing to be wary of is relative paths and CWD resolution. If you specify a file with a relative path, the anchor that is used to resolve the absolute path can vary depending on the platform, and whether you're in `dev` or `run`. PPB fixes this internally; I don't know what PyGame does, but I wouldn't bet money on it doing it in a way that would be consistent. FWIW, retrieving the resource path with importlib is one way to do this... but so is `__file__`.

freakboy3742 (changes requested):

src/briefcase/bootstraps/pygame.py
rmartin16: Thanks; I can't keep track of what in packaging is deprecated and what's not...

src/briefcase/bootstraps/pygame.py
rmartin16: You're right; I was a little too flippant in my assessment that "a lot of work is being done to access something we already know" when it should definitely be dynamically sourced from the metadata.

rmartin16:
I got rid of `pkg_resources`. I evaluated using `importlib.resources` but its support from Python 3.8 through 3.12 breaks in 3.9 (and possibly 3.11) when they rewrote the API. So, instead of trying to ensure a consistent API for `importlib.resources` (via PyPI's `importlib_resources`), I just used `__file__`. The only (breaking) downside I'm aware of with `__file__` is it doesn't work with zipped packages...but that shouldn't be a concern here.

Nonetheless, when Python 3.8 support is dropped, it should be fairly straightforward to use `importlib.resources` here instead.

freakboy3742 (approved):
All looks good to me. I'm not sure there's a *huge* need to move to importlib for resources given that `__file__` works, but we can cross that bridge in a year or so.

PR #1460, opened 2023-09-19
Add contributors to AUTHORS

rmartin16:
Should we add contributors to the `AUTHORS` file? Or should we just get rid of it?

Quick one-liner to update `AUTHORS` file:
```bash
sed -i '9,$d' AUTHORS && git shortlog --summary --numbered | grep -iEv 'dependabot|Brutus \(robot\)' | awk 'NR>1 {$1=""; print "   "$0}' >> AUTHORS
```

## PR Checklist:

- [X] All new features have been tested
- [X] All new features have been documented
- [X] I have read the **CONTRIBUTING.md** file
- [X] I will abide by the code of conduct

PR #1549, opened 2023-11-21
Introduce package for Briefcase Automation

rmartin16:
## Changes
- Creates a new package for Briefcase automation in CI
- This package facilitates the effort in https://github.com/beeware/.github/pull/69

## PR Checklist:

### DIFF
--- a/automation/pyproject.toml
+++ b/automation/pyproject.toml
@@ -1,0 +1,28 @@
+[build-system]
+requires = [
+    # keep versions in sync with ../pyproject.toml
+    "setuptools==69.0.0",
+    "setuptools_scm==8.0.4",
+    "setuptools_dynamic_dependencies @ git+https://github.com/beeware/setuptools_dynamic_dependencies",
+]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "x-briefcase-automation"
+description = "A Briefcase plugin for CI automation."
+readme = "README.md"
+license.text = "New BSD"
+classifiers = ["Private :: Do Not Upload"]
+dynamic = ["version", "dependencies"]
+
+[project.entry-points."briefcase.bootstraps"]
+"Toga Automation" = "automation.bootstraps.toga:TogaAutomationBootstrap"
+"PySide6 Automation" = "automation.bootstraps.pyside6:PySide6AutomationBootstrap"
+"Pygame Automation" = "automation.bootstraps.pygame:PygameAutomationBootstrap"
+"PursuedPyBear Automation" = "automation.bootstraps.pursuedpybear:PursuedPyBearAutomationBootstrap"
+
+[tool.setuptools_scm]
+root = "../"
+
+[tool.setuptools_dynamic_dependencies]
+dependencies = ["briefcase == {version}"]
--- a/automation/src/automation/bootstraps/pursuedpybear.py
+++ b/automation/src/automation/bootstraps/pursuedpybear.py
@@ -1,0 +1,83 @@
+import sys
+
+import tomli_w
+
+if sys.version_info >= (3, 11):
+    import tomllib
+else:
+    import tomli as tomllib
+
+from automation.bootstraps import BRIEFCASE_EXIT_SUCCESS_SIGNAL, EXIT_SUCCESS_NOTIFY
+from briefcase.bootstraps import PursuedPyBearGuiBootstrap
+
+
+class PursuedPyBearAutomationBootstrap(PursuedPyBearGuiBootstrap):
+    def app_source(self):
+        return f"""\
+import importlib.metadata
+import os
+import sys
+
+import ppb
+
+
+class {{{{ cookiecutter.class_name }}}}(ppb.Scene):
+    def __init__(self, **props):
+        super().__init__(**props)
+        self.updates: int = 0
+
+        self.add(
+            ppb.Sprite(
+                image=ppb.Image("{{{{ cookiecutter.module_name }}}}/resources/{{{{ cookiecutter.app_name }}}}.png"),
+            )
+        )
+
+    def on_update(self, event, signal):
+        self.updates += 1
+        # quit after 2 seconds since on_update is run 60 times/second
+        if self.updates > 120:
+            print("{EXIT_SUCCESS_NOTIFY}")
+            print("{BRIEFCASE_EXIT_SUCCESS_SIGNAL}")
+            signal(ppb.events.Quit())
+
+
+def main():
+    # Linux desktop environments use an app's .desktop file to integrate the app
+    # in to their application menus. The .desktop file of this app will include
+    # the StartupWMClass key, set to app's formal name. This helps associate the
+    # app's windows to its menu item.
+    #
+    # For association to work, any windows of the app must have WMCLASS property
+    # set to match the value set in app's desktop file. For PPB, this is set
+    # using the SDL_VIDEO_X11_WMCLASS environment variable.
+
+    # Find the name of the module that was used to start the app
+    app_module = sys.modules["__main__"].__package__
+    # Retrieve the app's metadata
+    metadata = importlib.metadata.metadata(app_module)
+
+    os.environ["SDL_VIDEO_X11_WMCLASS"] = metadata["Formal-Name"]
+
+    ppb.run(
+        starting_scene={{{{ cookiecutter.class_name }}}},
+        title=metadata["Formal-Name"],
+    )
+"""
+
+    # The constraint of pysdl2-dll==2.0.22 is required for ppb==1.1.0;
+    # the libraries in later versions of pysdl2-dll are not compatible.
+
+    def pyproject_table_linux_flatpak(self):
+        table = tomllib.loads(super().pyproject_table_linux_flatpak())
+        table.setdefault("requires", []).append("pysdl2-dll==2.0.22")
+        return f"\n{tomli_w.dumps(table)}"
+
+    def pyproject_table_windows(self):
+        table = tomllib.loads(super().pyproject_table_windows())
+        table.setdefault("requires", []).append("pysdl2-dll==2.0.22")
+        return f"\n{tomli_w.dumps(table)}"
+
+    def pyproject_table_macOS(self):
+        table = tomllib.loads(super().pyproject_table_macOS())
+        table.setdefault("requires", []).append("pysdl2-dll==2.0.22")
+        return f"\n{tomli_w.dumps(table)}"
--- a/automation/src/automation/bootstraps/pursuedpybear.py
+++ b/automation/src/automation/bootstraps/pursuedpybear.py
@@ -1,12 +1,3 @@
-import sys
-
-import tomli_w
-
-if sys.version_info >= (3, 11):
-    import tomllib
-else:
-    import tomli as tomllib
-
 from automation.bootstraps import BRIEFCASE_EXIT_SUCCESS_SIGNAL, EXIT_SUCCESS_NOTIFY
 from briefcase.bootstraps import PursuedPyBearGuiBootstrap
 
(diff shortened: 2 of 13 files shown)


rmartin16:
For posterity:
```
 [verifyapp] Built build/verifyapp/ios/xcode/build/Debug-iphonesimulator/Verify App.app

[verifyapp] Starting app on an iPhone SE (3rd generation) running iOS 16.2 (device UDID 1A0F79A9-B187-44AF-9000-8558A7179437)
Booting simulator... done
The application /Applications/Xcode_14.2.app/Contents/Developer/Applications/Simulator.app cannot be opened for an unexpected reason, error=Error Domain=NSOSStatusErrorDomain Code=-600 "procNotFound: no eligible process with specified descriptor" UserInfo={_LSLine=387, _LSFunction=_LSAnnotateAndSendAppleEventWithOptions}
Opening simulator...

Unable to open iPhone SE (3rd generation) simulator running 16.2

Log saved to /Users/runner/work/briefcase/briefcase/apps/verifyapp/logs/briefcase.2023_11_21-20_53_13.run.log
```

[briefcase.2023_11_21-20_53_13.run.log](https://github.com/beeware/briefcase/files/13431256/briefcase.2023_11_21-20_53_13.run.log)

rmartin16:
I'll pause here for comments on how I've incorporated the briefcase automation plugin in to the Briefcase repo.

This approach of a dedicated top-level directory seemed appropriate given the options I could muster. For instance, we could use the Toga approach or nested pyprojects within `src` but that seems particularly invasive and disruptive just to introduce a small package needed for CI testing. Definitely open to other ideas.

Something I'd also like to explore is how to make an extra dependency group for this automation package; that may help simplify getting it installed in to CI. However, I'm not sure how a reference to this dependency could be reliable in all cases....

Another alternative altogether I've considered is _not_ creating a new package....but to include these automation  bootstraps within Briefcase directly....such that they would even be shipped with releases. However, they would not be exposed as entry points by default. Instead, they could be dynamically loaded at some point during testing.

freakboy3742 (changes requested):
A couple of relatively minor suggestions, but on the whole I like the approach. A standalone install in the same repo makes sense to me - this won't be useful to anyone outside of Briefcase, so I don't see any reason to include it in the main package.

Making it an optional install on Briefcase is an interesting idea... but I'm not sure how that would work without either publishing the package on PyPI, or installing via a GitHub reference. AFAIK, you wouldn't be able to reference the dependency as local files unless the code is included in the published package. But I guess the current setup in .github is already installing the automation package via GitHub, which would just mean the configuration can be self-contained in the Briefcase repo.

automation/pyproject.toml
```diff
@@ -0,0 +1,24 @@
...
+description = "A Briefcase plugin for CI automation."
+license = {text = "BSD-3-Clause"}
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "License :: OSI Approved :: New BSD",
+    "Operating System :: OS Independent",
+]
+dependencies = [
+    "briefcase >= 0.3.17.dev100",
+    "toml",
```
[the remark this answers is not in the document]
rmartin16: yeah...i was in the middle of trying to get this new package to work and then realized there's several TOML packages in play (`toml`, `tomllib`, `tomli`, and `tomli_w`) and didn't want to spend time then trying to understand it. Nonetheless, I've updated this to use the TOML packages that Briefcase is already using.

automation/pyproject.toml
```diff
@@ -0,0 +1,24 @@
...
+version = "0.1.0"
+description = "A Briefcase plugin for CI automation."
+license = {text = "BSD-3-Clause"}
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "License :: OSI Approved :: New BSD",
+    "Operating System :: OS Independent",
+]
+dependencies = [
+    "briefcase >= 0.3.17.dev100",
```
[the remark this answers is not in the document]
rmartin16: Ok....this took me a minute. Your hyperlink returns a 404; so, I was frantically searching the internet for what on earth this is....and then I realized this all some brand new stuff for toga 🤣 Anyway, now that I understand, I got it working.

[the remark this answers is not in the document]
rmartin16: hmm....my first attempt at this is hampered by GitHub's PR/CI tactics.

The version of Briefcase [installed](https://github.com/beeware/briefcase/actions/runs/6985085377/job/19008965506#step:9:84) in to CI is `briefcase-0.3.17.dev118+g169210e9`.

The version of Briefcase wanted by the Automation plugin is `briefcase==0.3.17.dev117+g10ed6d70`.

I'm pretty certain this is stemming from GitHub creating a new merge commit for CI for the PR and that's what `actions/checkout` is fetching....therefore, the Briefcase artifact is one commit ahead.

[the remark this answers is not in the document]
rmartin16: hmm....this is non-trivial...but not impossible, of course. If we're going to put some kind of constraint on the version of Automation matching the version of Briefcase, then the two will always need to be sourced from the same place.

Given that Briefcase can be installed several different ways in to CI, each method will need to accommodate also installing Automation...or packaging Automation as well so it can be installed later.

Probably won't have time tonight but I'll return to this.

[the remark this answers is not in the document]
freakboy3742: Apologies for the run-around on this - for posterity, the link should have been to https://github.com/beeware/setuptools_dynamic_dependencies.

As for the version mismatch - I'm intrigued that this doesn't happen in Toga... I'm guessing the reason is that we're explicitly building wheels and installing those wheel, rather than installing from the repo directly.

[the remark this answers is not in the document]
rmartin16: ### Scenarios for how Briefcase is installed
- CI event for beeware/briefcase PR
  - Briefcase is packaged from CI's PR merge commit and wheel is installed
  - Automation should be installed from CI's PR merge commit
- CI event for beeware/briefcase that is not a PR
  - Briefcase is packaged for the current git ref and wheel is installed
    - The most common situation for this is CI running after a PR is merged in to `main`
    - Although, the git ref could also be a tag such as when releases are cut
  - Automation should be installed from the same git ref
- CI event for other repo
  - Briefcase is installed from a git ref relevant for the event
  - Automation should be installed for the same git ref

### Strategy
So, while it is relatively straightforward to say "use the same git ref to install Automation"...the ref used to install Briefcase may not be readily available arbitrarily later in the CI run. So, it'll probably be easiest to package and/or install Automation at the points in the CI when Briefcase is packaged and installed.

automation/pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+    "setuptools_dynamic_dependencies @ git+https://github.com/beeware/setuptools_dynamic_dependencies",
+]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "x-briefcase-automation"
+description = "A Briefcase plugin for CI automation."
+readme = "README.md"
+license.text = "New BSD"
+classifiers = ["Private :: Do Not Upload"]
```
rmartin16, line 15: couple notes:

1) Naming the package `x-briefcase-automation` avoids its wheel and sdist from being included in globs for things like `./dist/briefcase-*` and generally helps ensure that you only reference this package's artifacts intentionally and explicitly.

2) Classifier to ensure this package is never accepted by PyPI as an upload.

automation/pyproject.toml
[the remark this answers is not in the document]
rmartin16: This strategy turned out not to be too invasive and actually pretty straightforward. One [complication](https://github.com/beeware/.github/pull/69/files#r1406966084) from running `app-build-verify` for PRs for versioned branches of template repos where the automation plugin will not be available (or supported).

automation/pyproject.toml
freakboy3742: TIL about the Private trove classifier.

rmartin16:
This can be merged independently of the other changes; that'll help confirm `.github`'s CI is still working properly....especially since I just noticed that I need to invoke `python-package-create` differently for the `briefcase` repo in `.github`'s CI.

automation/src/automation/bootstraps/pursuedpybear.py
```diff
@@ -0,0 +1,83 @@
...
+        title=metadata["Formal-Name"],
+    )
+"""
+
+    # The constraint of pysdl2-dll==2.0.22 is required for ppb==1.1.0;
+    # the libraries in later versions of pysdl2-dll are not compatible.
+
+    def pyproject_table_linux_flatpak(self):
+        table = tomllib.loads(super().pyproject_table_linux_flatpak())
+        table.setdefault("requires", []).append("pysdl2-dll==2.0.22")
```
freakboy3742, line 72: Why is this only part of the automation, and not the core ppb bootstrap?

freakboy3742 (changes requested):
One minor Q about the PPB automation, and the workflow CI references, and this otherwise looks good to go.

To be clear about the landing process, though - once the CI references are fixed, we can land this; but the effect will be that the automation bootstrap will *exist*, but won't be used. We then land the .github change, and that enables the use of the automation plugin on current/future Briefcase PRs. The current branch-based CI references exist to validate that the "end state" when we're using the plugin, the plugin will actually *work*. 

Have I got that right?

automation/src/automation/bootstraps/pursuedpybear.py
rmartin16: Mostly just because I wasn't sure if we wanted to officially endorse this for new projects....but I'm probably only really being conservative because I don't know anything about the ppb project.

That said, it seems fairly harmless to add this to the actual bootstrap.

rmartin16:
> To be clear about the landing process, though - once the CI references are fixed, we can land this; but the effect will be that the automation bootstrap will _exist_, but won't be used. We then land the .github change, and that enables the use of the automation plugin on current/future Briefcase PRs. The current branch-based CI references exist to validate that the "end state" when we're using the plugin, the plugin will actually _work_.
> 
> Have I got that right?

Yep; that's spot on.

automation/src/automation/bootstraps/pursuedpybear.py
freakboy3742: Probably worth tagging @pathunstrom to see if she has any thoughts on the matter.

/me turns on the bat signal...

tox.ini
```diff
@@ -127,12 +127,13 @@ commands =
...
+[testenv:package{,-with-automation}]
 skip_install = True
 passenv = FORCE_COLOR
 deps =
     build==1.0.3
     twine==4.0.2
 commands =
-    python -m build --outdir dist/ .
+    python -m build . --outdir dist/
+    with-automation: python -m build automation/ --outdir dist/
```
freakboy3742, line 138: Actually - I take that back: the -with-automation target has been removed from the GitHub CI (it didn't show up in my earlier review because the CI workflow doesn't appear in the patch any more); but it doesn't appear to be used by the .github repos either. Am I missing something?

freakboy3742 (changes requested):

tox.ini
rmartin16: I [introduced](https://github.com/beeware/.github/pull/69/files#diff-cda3ef2c3e868ac57017648b0c5053246f4d1e56164f94dec64b10975fc73283R22) the ability to pass additional tox factors to the `python-package-create` workflow with beeware/briefcase#1460. And passing unknown inputs to a variable causes CI to fail; so, I'll need to create another PR here that'll need to be merged with beeware/briefcase#1460.

freakboy3742: Sorry - is #1460 the right reference there?

rmartin16: my bad - too many tabs open - I meant https://github.com/beeware/.github/pull/69

automation/src/automation/bootstraps/pursuedpybear.py
freakboy3742: Via Discord: @pathunstrom has indicated [the patch is on the right track](https://discord.com/channels/836455665257021440/836455665257021443/1181375691401461761); we'll likely need to revisit it when PPB can address the problem at their end, but for now, it works.

freakboy3742 (approved):
👍

PR #1550, opened 2023-11-22
Migrate to a pure PEP621 project configuration

freakboy3742:
Inspired by beeware/toga#2229, a full PEP621 configuration update for Briefcase.

There's no functional difference; the one technical benefit (aside from keeping up with modern standards) is that because we're using `setuptools_scm`, we don't need to maintain MANIFEST.in any more. The source discovery process is also much simpler - it requiring *no* explicit setuptools configuration, as we're entirely relying on one of the "default" project layout patterns.

I've audited the generated tarballs, and they have identical content, except for:
* the missing setup.cfg and MANIFEST.in files, 
* a missing `not-zip-safe` file in the egg-info. This file was needed in the days of eggs and `easy-install`, but is [now considered deprecated](https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html).
* minor changes to the METADATA file in the wheel:
  * A different format for authors and license
  * Slightly different URL definitions
  * No `platform: any` declaration. This has been subsumed by wheel tagging.

## PR Checklist:


pyproject.toml
```diff
@@ -1,7 +1,168 @@
...
+    "mobile",
+    "windows",
+    "tvOS",
+    "watch",
+    "watchOS",
+    "wearos",
+    "web",
+    "pyscript",
+    "pyodide",
+]
```
rmartin16, line 37: no linux... 😢 

but more seriously....adding Flatpak (and even AppImage if we dare) may be a good idea....going further, deb, RPM, pkg....less committed to these, though

pyproject.toml
```diff
@@ -1,7 +1,168 @@
...
 build-backend = "setuptools.build_meta"
 
+[project]
+dynamic = ["version"]
+name = "briefcase"
+description = "Tools to support converting a Python project into a standalone native application."
+readme = "README.rst"
+requires-python = ">= 3.8"
+license.file = "LICENSE"
```
rmartin16, line 14: I know you're pretty steeped in this and have probably considered the aspects around this.

But do you know how this will be displayed in PyPI? Curious because currently it at least says "New BSD". Unfortunately, the available trove classifier remains ambiguous with just "BSD License".

I didn't go all the way down the rabbit hole but I think [PEP-639](https://peps.python.org/pep-0639/) finally added support to explicitly say we're using bsd-3-clause.

rmartin16 (commented):
initial comments. i'll probably repeat the build artifact review so 2 sets of eyes looked at it.

pyproject.toml
freakboy3742: O_O 

No idea how linux fell off this list. But yes, will add Flatpak et al.

pyproject.toml
freakboy3742: I hadn't contemplated PyPI presentation. From the look of some high profile packages that are using PEP621, ([attrs](https://pypi.org/project/attrs/), [cryptography](https://pypi.org/project/cryptography/), [setuptools_scm](https://pypi.org/project/setuptools-scm/)), they're using `license.text` (or just `license`), not `license.file`; so that might be the better way to go. Looks like this approach puts the same License marker in the METADATA, and just having a LICENSE file in the repo is enough for wheels to include that file and include it as license content.

PEP639 would definitely alter this, but it is still in draft.

My survey of projects also pointed out that the way I handled the "Homepage" link isn't quite right. Fixes incoming.

rmartin16: Uploaded (with these changes) here: https://test.pypi.org/project/briefcase-rmartin16/0.3.17.dev112/

freakboy3742: All looks suitably pretty to me...

rmartin16: Yeah....they don't seem to respect ordering of URLs, though.... 😐 🤷🏼‍♂️

freakboy3742: Alphabetical is *an* ordering... :-)

pyproject.toml
```diff
@@ -1,7 +1,174 @@
...
+    "sphinx-copybutton == 0.5.2",
+    "sphinxcontrib-spelling == 8.0.0",
+]
+
+[project.urls]
+Homepage = "http://beeware.org/briefcase"
+Funding = "https://beeware.org/contributing/membership/"
+Documentation = "http://briefcase.readthedocs.io/en/latest/"
+Tracker = "https://github.com/beeware/briefcase/issues"
+Source = "https://github.com/beeware/briefcase"
```
rmartin16, line 118: Warning-zealous PyCharm is reminding me the `http://` scheme is insecure...

tox.ini
```diff
@@ -114,10 +131,8 @@ commands =
 skip_install = True
 passenv = FORCE_COLOR
 deps =
-    check_manifest
-    build
-    twine
+    build==1.0.3
+    twine==4.0.2
 commands =
-    check-manifest -v
```
rmartin16, line 121: This might just be for my own edification....but I was under the impression that this ensures the sdist matches what's in git. Is this enforced by default now? Or no longer a concern? Just not sure I fully understand this transition...

pyproject.toml
freakboy3742: Well that definitely needs to be updated...

tox.ini
freakboy3742: It's enforced by way of setuptools_scm. It controls the list of files included in the sdist to match everything that is in source control. Essentially we get a dynamic MANIFEST.in for free.

pyproject.toml
freakboy3742: I swear... it doesn't matter how many times you audit for this, you find new ones. A whole bunch of http references found throughout the project.

rmartin16: Gotta get this IDE that nanny's you all day ;)

rmartin16 (approved):
Compared the metadata in the sdist and wheel before and after. LGTM.

mhsmith:
I think this has [broken Toga's CI](https://github.com/beeware/toga/actions/runs/6960869033/job/18941360266), though I don't know how.

mhsmith:
See the "Install dependencies" step:
```
Collecting git+https://github.com/beeware/briefcase.git
  Cloning https://github.com/beeware/briefcase.git to /tmp/pip-req-build-ogtlw61a
  Running command git clone --filter=blob:none --quiet https://github.com/beeware/briefcase.git /tmp/pip-req-build-ogtlw61a
  Resolved https://github.com/beeware/briefcase.git to commit 1687c5e4590c43580fec7e594a880081c43603ea
Building wheels for collected packages: UNKNOWN
  Building wheel for UNKNOWN (pyproject.toml): started
  Building wheel for UNKNOWN (pyproject.toml): finished with status 'done'
  Created wheel for UNKNOWN: filename=UNKNOWN-0.3.17.dev114+g1687c5e4-py3-none-any.whl size=2660 sha256=1e18f288143411c2f7d189deb5d892b0735114a1a254be3db9462014cce8ed7b
  Stored in directory: /tmp/pip-ephem-wheel-cache-vtk5u_v1/wheels/e4/98/7a/33d68b092d9141[556](https://github.com/beeware/toga/actions/runs/6960869033/job/18941360266#step:4:557)f79cf6c5d442c85bec3e2d91effd31094
Successfully built UNKNOWN
Installing collected packages: UNKNOWN
Successfully installed UNKNOWN-0.3.17.dev114+g1687c5e4
```

rmartin16:
👀 

I can't replicate locally....i wonder if setuptools is too old in CI...

rmartin16:
Indeed...pip v22.0.2 ships with Ubuntu 22.04 and it will install briefcase as UNKNOWN. It should be upgraded first.

freakboy3742:
Hrm - that's worth a release note and a FAQ. PR incoming.

freakboy3742:
No - wait - this only affects source installs, right? If you're installing a wheel, an older pip will work fine (or, at least as well as it ever has). This is only an issue for dev installs and/or GitHub installs - am I correct?

rmartin16:
Right; the wheel is the same....but the build process is pretty different now.

PR #1554, opened 2023-11-24
Use Pixel 7 Pro skin for new AVDs

rmartin16:
## Changes
- I loved my Pixel 3....but those giant bezels make it look particularly dated at this point
- New AVDs now use the Pixel 7 Pro skin and look extra fresh

## Notes
- I'm not aware of special considerations for using a different skin....on my machine, this just works
- Switched to the Pixel 7 Pro skin since the hello world app looks a bit better than Pixel 8 Pro

## PR Checklist:

### DIFF
--- a/changes/1554.feature.rst
+++ b/changes/1554.feature.rst
@@ -1,0 +1,1 @@
+New virtual devices for the Android emulator are created using the Pixel 7 Pro skin.
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -168,7 +168,7 @@ class AndroidSDK(ManagedTool):
 
     @property
     def DEFAULT_DEVICE_SKIN(self) -> str:
-        return "pixel_3a"
+        return "pixel_7_pro"
 
     @property
     def DEFAULT_SYSTEM_IMAGE(self) -> str:
--- a/tests/integrations/android_sdk/AndroidSDK/test__create_emulator.py
+++ b/tests/integrations/android_sdk/AndroidSDK/test__create_emulator.py
@@ -160,7 +160,7 @@ def test_create_emulator_with_defaults(
     )
 
     # The emulator skin was verified
-    android_sdk.verify_emulator_skin.assert_called_once_with("pixel_3a")
+    android_sdk.verify_emulator_skin.assert_called_once_with("pixel_7_pro")
 
     # avdmanager was invoked
     mock_tools.subprocess.check_output.assert_called_once_with(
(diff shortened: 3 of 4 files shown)


rmartin16:
Pixel 7 Pro
image

Pixel 8 Pro
image

rmartin16:
hmm....I guess I'd be curious if an AVD created in Android Studio for Pixel 7 Pro has the same issues on those machines...? I'm also curious which version of the emulator Briefcase is using.

AFAICT, these skins aren't doing anything to the emulator; it just loads them at runtime. You can even [create](https://developer.android.com/studio/run/managing-avds#skins) arbitrary skins.

Given these issues, though, maybe it'd be better to just offer a list of some of the Pixel models and let users choose the one they like? Can also note the Pixel 3 may have better stability if a newer model causes problems.

rmartin16:
> On the second run, it launched successfully... but then showed the odd "white bar" behavior that you're reporting you saw for the Pixel 8 skin

This reminds me of when I bought the first Pixel that moved the camera in to the screen; this required Android to change how the "screen boundaries" of an app were handled. Since I bought the phone on release, many apps needed some time to adjust to this new layout....but before those apps were updated, they looked a lot like how the helloworld app is displaying with that white bar.

I don't know enough about Android UI dev to comment intelligently....but I wonder if Toga's Android UI handling needs tweaks to handle this better?

freakboy3742 (approved):
Now that I've got working skins, I've tried re-running on M1, and the Pixel 7 skin works great - no white bar. 

I then updated my x86 machine's emulator to the same version, and the emulator crashed, complaining about not supporting Vulkan (?!). But when I re-created the emulator image, it worked fine. 

I also noticed another detail that is present in your screenshots, but I missed the significance. The Pixel 8 screenshot (and the behavior I *was* seeing on the 7 image) doesn't just have a white bar at the top - it's the missing the *controls* at the bottom. I originally thought this was just the analog of the Apple "no home button" physical design change, but I now see the buttons should still be there. That suggests to me that the problem isn't Toga or the app not handling the titlebar correctly - it's an issue with the emulator not displaying the skin correctly.

I think I'm happy to call this a Android Studio QA glitch - recent skins with recent emulators on recent images look like they work, but combinations of new skins with old emulators and/or old images are a crap shoot.

I'm happy to approve on that basis; however, I'd like @mhsmith to weigh in just to make sure that we haven't missed something important here. I'll leave to him to do the final approve/merge.

mhsmith:
>  I originally thought this was just the analog of the Apple "no home button" physical design change

They have actually [done that as well](https://support.google.com/android/answer/9079644), and the newer emulator images are defaulting to the no-buttons mode. But apart from that, I agree with your analysis.

PR #1561, opened 2023-12-04
Fall back to system allocated port if port in use for web server

rmartin16:
## Changes
- Instead of just erroring if port 8080 is already in use, warn the user and use a system-allocated port

## PR Checklist:


freakboy3742 (approved):
👍

PR #1566, opened 2023-12-07
Added documentation for optional color configurations on Android.

bbusenius:
This PR is to add documentation for the new color settings added to [briefcase-android-gradle-template in PR #77](https://github.com/beeware/briefcase-android-gradle-template/pull/77). The new color settings are:

```
accent_color
primary_color
primary_color_dark
```

## PR Checklist:

### DIFF
--- a/docs/reference/configuration.rst
+++ b/docs/reference/configuration.rst
@@ -177,6 +177,14 @@ specific.
 Optional values
 ---------------
 
+``accent_color``
+~~~~~~~~~~~~~~~~
+
+A hexadecimal RGB color value (e.g., ``#D81B60``) for a subtle secondary color
+to be used throughout an application to call attention to key elements. This
+setting is only used if the platform allows color modification, otherwise it
+is ignored.
+
 ``author``
 ~~~~~~~~~~
 
@@ -290,6 +298,22 @@ A longer description of the purpose of the application. This description can be
 multiple paragraphs, if necessary. The long description *must not* be a copy of
 the ``description``, or include the ``description`` as the first line of the
 ``long_description``.
+
+``primary_color``
+~~~~~~~~~~~~~~~~~
+
+A hexadecimal RGB color value (e.g., ``#008577``) to use as the primary color
+for the application. On Android this will be used to color the app bar in the
+main window. This setting is only used if the platform allows color
+modification, otherwise it is ignored.
+
+``primary_color_dark``
+~~~~~~~~~~~~~~~~~~~~~~
+
+A hexadecimal RGB color value (e.g., ``#008577``) used alongside the primary
+color. On Android this is used to color the status bar at the top of the
+screen. This setting is only used if the platform allows color modification,
+otherwise it is ignored.
 
 ``requires``
 ~~~~~~~~~~~~


bbusenius:
I took a stab at writing a `newsfragment`. I hope that wasn't too audacious. I wasn't sure if that minuscule work should rise to the level of a feature but since it's new functionality, that was my best guess. Let me know if you want me to change anything.

changes/1566.feature.rst
```diff
@@ -0,0 +1 @@
+Briefcase Android Gradle Template now allows for setting accent_color, primary_color and primary_color_dark. These settings can now be used in pyproject.toml for Android projects.
```
freakboy3742, line 1: Even though it's *currently* Android-only, the feature is a generic configuration item, so it should be framed in those terms.
```suggestion
Apps can now specify a primary color (for both light and dark modes), and an accent color. If the platform allows apps to customize color use, these colors will be used to style the app's presentation.
```

docs/reference/configuration.rst
```diff
@@ -291,6 +299,22 @@ multiple paragraphs, if necessary. The long description *must not* be a copy of
 the ``description``, or include the ``description`` as the first line of the
 ``long_description``.
 
+``primary_color``
+~~~~~~~~~~~~~~~~~
+
+A hexadecimal RGB color value (e.g., ``#008577``) to use as the primary color
+for the application. On Android this will be used to color the app bar in the
+main window. This setting is only used if the platform allows color
+modification, otherwise it is ignored.
```
freakboy3742, line 308: Notes about the specific platform interpretations should be in the relevant platform guide, unless they're indicative of behavior on all platforms.
```suggestion
A hexadecimal RGB color value (e.g., ``#008577``) to use as the primary color
for the application. This setting is only used if the platform allows color
modification, otherwise it is ignored.
```

docs/reference/configuration.rst
```diff
@@ -291,6 +299,22 @@ multiple paragraphs, if necessary. The long description *must not* be a copy of
...
+
+``primary_color_dark``
+~~~~~~~~~~~~~~~~~~~~~~
+
+A hexadecimal RGB color value (e.g., ``#008577``) used alongside the primary
+color. On Android this is used to color the status bar at the top of the
+screen. This setting is only used if the platform allows color modification,
+otherwise it is ignored.
```
freakboy3742, line 316: As above
```suggestion
A hexadecimal RGB color value (e.g., ``#008577``) used alongside the primary
color. This setting is only used if the platform allows color modification,
otherwise it is ignored.
```

freakboy3742 (changes requested):
Looks good! This definitely qualifies as a "feature" to my mind - it's a notable capability that now exists that didn't previously.

A couple of review suggestions inline; it's probably also worth moving the Android-specific notes to the Android configuration guide. That document already mentions ``splash_background_color``; it would make sense to me to expand that section to clarify how the other app colors will be applied.

bbusenius:
That all makes sense. I made the suggested changes and moved the Android specific language to the Android configuration section of `docs/reference/platforms/android/gradle.rst`. Feel free to let me know if there's anything else you want me to change.

freakboy3742 (changes requested):

docs/reference/platforms/android/gradle.rst
```diff
@@ -102,6 +102,22 @@ You can specify a background color for the splash screen using the
 
 Android projects do not support installer images.
 
+Colors
+======
+
+In addition to ``splash_background_color``, Android projects allow for
+configuration of an ``accent_color``, ``primary_color``, and
+``primary_color_dark``.
```
freakboy3742, line 110: There's no need to list all the things that can be configured just before you list all the things that can be configured :-)
```suggestion
Android allows for some customization of the colors used by your app:
```

docs/reference/platforms/android/gradle.rst
```diff
@@ -102,6 +102,22 @@ You can specify a background color for the splash screen using the
...
+
+* ``accent_color`` is used subtly throughout an Android app to call attention
+  to key elements. It's used on things like form labels and inputs.
```
freakboy3742, line 113:
```suggestion
* ``accent_color`` is used as a subtle highlight throughout your app to call attention
  to key elements. It's used on things like form labels and inputs.
```

freakboy3742 (changes requested):
Almost there! A couple more revisions; also - if we're adding a whole section about configurating color (which I agree is a good idea), we can remove the reference to the splash background on L100-101.

docs/reference/platforms/android/gradle.rst
bbusenius: I made all those changes. Thanks for helping me get this in shape!

PR #1572, opened 2023-12-13
 Colorize ADB logcat by default

matan-h:
This PR adds the [ADB argument](https://developer.android.com/tools/logcat#formatmodify) for colored output. (`-v color`)
it automatically makes the ADB logcat to print with color highlighting (e.g. errors in red, info in green and so on) which I think provide better user experience 

## PR Checklist:

### DIFF
--- a/changes/1572.feature.rst
+++ b/changes/1572.feature.rst
@@ -1,0 +1,1 @@
+The output from an Android device while running an app is now colorized
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -29,6 +29,9 @@ from briefcase import __version__
 # Regex to identify settings likely to contain sensitive information
 SENSITIVE_SETTING_RE = re.compile(r"API|TOKEN|KEY|SECRET|PASS|SIGNATURE", flags=re.I)
 
+# Regex to remove ANSI color from text. from stackoverflow.com/a/38662876
+ANSI_ESCAPE = re.compile(r"(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]")
+
 
 class InputDisabled(Exception):
     def __init__(self):
@@ -114,7 +117,8 @@ class Printer:
     @classmethod
     def export_log(cls):
         """Export the text of the entire log."""
-        return cls.log.export_text()
+
+        return ANSI_ESCAPE.sub("", cls.log.export_text())
 
 
 class RichLoggingStream:
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1606,6 +1606,8 @@ Activity class not found while starting app.
                     self.device,
                     "logcat",
                     "--format=tag",
+                    "-v",
+                    "color",
                     "-t",
                     since.strftime("%m-%d %H:%M:%S.000000"),
                     "-s",
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1576,6 +1576,8 @@ Activity class not found while starting app.
                 self.device,
                 "logcat",
                 "--format=tag",
+                "-v",
+                "color",
                 "--pid",  # This option is available since API level 24.
                 pid,
             ]
--- a/tests/integrations/android_sdk/ADB/test_logcat.py
+++ b/tests/integrations/android_sdk/ADB/test_logcat.py
@@ -20,6 +20,8 @@ def test_logcat(mock_tools, adb):
             "exampleDevice",
             "logcat",
             "--format=tag",
+            "-v",
+            "color",
             "--pid",
             "1234",
             "EGL_emulation:S",
--- a/tests/integrations/android_sdk/ADB/test_logcat_tail.py
+++ b/tests/integrations/android_sdk/ADB/test_logcat_tail.py
@@ -21,6 +21,8 @@ def test_logcat_tail(mock_tools, adb):
             "exampleDevice",
             "logcat",
             "--format=tag",
+            "-v",
+            "color",
             "-t",
             "11-10 09:08:07.000000",
             "-s",


rmartin16:
I definitely like this idea; the readability of `adb` output is greatly improved IMO.

Although, (since we often forget about the logfile), this does inject a bunch of ANSI codes in to the log; for instance:
```
           ESC[33mW/native.stderr: s_glBindAttribLocation: bind attrib 2 name inShadowParamsESC[0m                                                                       subprocess.py:681
           ESC[33mW/native.stderr: s_glBindAttribLocation: bind attrib 0 name positionESC[0m                                                                             subprocess.py:681
           ESC[33mW/native.stderr: s_glBindAttribLocation: bind attrib 1 name colorESC[0m                                                                                subprocess.py:681
           ESC[33mW/native.stderr: s_glBindAttribLocation: bind attrib 0 name inPositionESC[0m                                                                           subprocess.py:681
           ESC[33mW/native.stderr: s_glBindAttribLocation: bind attrib 1 name inColorESC[0m                                                                              subprocess.py:681
           ESC[33mW/native.stderr: s_glBindAttribLocation: bind attrib 2 name inTextureCoordsESC[0m                                                                      subprocess.py:681
           ESC[34mD/MainActivity: onPrepareOptionsMenu() startESC[0m                                                                                                     subprocess.py:681
           ESC[34mD/MainActivity: onPrepareOptionsMenu() completeESC[0m                                                                                                  subprocess.py:681
           ESC[34mD/MainActivity: onPrepareOptionsMenu() startESC[0m                                                                                                     subprocess.py:681
           ESC[34mD/MainActivity: onPrepareOptionsMenu() completeESC[0m                                                                                                  subprocess.py:681
           ESC[33mW/native.stderr: s_glBindAttribLocation: bind attrib 0 name positionESC[0m                                                                             subprocess.py:681
           ESC[33mW/native.stderr: s_glBindAttribLocation: bind attrib 1 name colorESC[0m                                                                                subprocess.py:681

```
This is arguably tolerable, though....if you think the color output is an improvement. I didn't look for too long....but it may be possible to strip any ANSI codes when `export_text()` is called.

@matan-h, as for the CI failure, if you add a file called `1572.feature.rst` in the `changes` directory that says something like "The output from an Android device while running an app is now colorized.", CI will continue on. Also, probably worth adding this to `logcat_tail()` as well.

matan-h:
Thank you, @rmartin16. 
about the `CI failure` I've added the file, and currently waiting the `tox`

About the `export_text`, which function do you mean? 
do you mean changing the [`export_log`](https://github.com/beeware/briefcase/blob/60d0a9766b068a3b84aed5853f6d7eb988dce4e7/src/briefcase/console.py#L115) code? Or the original rich [`export_text`](https://github.com/Textualize/rich/blob/fd981823644ccf50d685ac9c0cfe8e1e56c9dd35/rich/console.py#L2126) ?

About the ANSI colors, It's not hard to remove them: there is [this](https://stackoverflow.com/a/38662876) Stack Overflow answer which use regex.

You can try it if you have [sd](https://github.com/chmln/sd):  `adb logcat -v color |head -1000|sd "(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]" ""`

rmartin16:
> About the `export_text`, which function do you mean? do you mean changing the [`export_log`](https://github.com/beeware/briefcase/blob/60d0a9766b068a3b84aed5853f6d7eb988dce4e7/src/briefcase/console.py#L115) code? Or the original rich [`export_text`](https://github.com/Textualize/rich/blob/fd981823644ccf50d685ac9c0cfe8e1e56c9dd35/rich/console.py#L2126) ?
> 
> About the ANSI colors, It's not hard to remove them: there is [this](https://stackoverflow.com/a/38662876) Stack Overflow answer which use regex.
> 
> You can try it if you have [sd](https://github.com/chmln/sd): `adb logcat -v color |head -1000|sd "(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]" ""`

I was referring to Briefcase calling `export_text()` in `console.Printer.export_log()`. Ideally, I think, we'd be able to defer this to Rich....but I didn't see a straightforward way to accomplish this either with the creation of a Rich `Console` or when exporting the logged text.

matan-h:
Thanks, I've added a commit that strip ANSI codec from the log file in the `export_log` function.
I agree that since rich has its own ecosystem of ANSI colors and terminal detection, it wouldn't need to use the regex to remove them.

matan-h:
The same CI :  [`Build app (toga, ubuntu-22.04) / Verify App Build`](https://github.com/beeware/briefcase/actions/runs/7200706517/job/19615434007?pr=1572#logs) is falling with timeout (30m) again. Is it something I did? It stopped with the `adb logcat` command.
[This](https://github.com/beeware/briefcase/actions/runs/7200706517/job/19615434007?pr=1572#step:23:517) is the last logcat from the CI (with nice colors ):
```log
I/python.stdout: >>> successfully started...exiting <<<
I/python.stdout: >>>>>>>>>> EXIT 0 <<<<<<<<<<
D/MainActivity: onResume() start
I/python.stdout: Toga app: onResume
D/MainActivity: onResume() complete
```

freakboy3742:
> The same CI : [`Build app (toga, ubuntu-22.04) / Verify App Build`](https://github.com/beeware/briefcase/actions/runs/7200706517/job/19615434007?pr=1572#logs) is falling with timeout (30m) again. Is it something I did? It stopped with the `adb logcat` command. [This](https://github.com/beeware/briefcase/actions/runs/7200706517/job/19615434007?pr=1572#step:23:517) is the last logcat from the CI (with nice colors ):
> 
> ```
> I/python.stdout: >>> successfully started...exiting <<<
> I/python.stdout: >>>>>>>>>> EXIT 0 <<<<<<<<<<
> D/MainActivity: onResume() start
> I/python.stdout: Toga app: onResume
> D/MainActivity: onResume() complete
> ```

My guess will be that the ANSI codes are affecting how the test shuts down.

The test that is failing is the only one that will be running a live Android test (we only run Android tests on Toga, on Ubuntu). Testing on Android (and iOS for that matter) is a little odd, because we don't have direct access to verify whether the process we're testing has exited. As a result, we have to use a logging sentinel - that's the `>>>>>>>>>> EXIT 0 <<<<<<<<<<` line in stdout. We monitor the logs, and when we see the exit sentinel, we use that as an indicator that the test has finished.

If that exit pattern ([see `briefcase/commands/run.py`](https://github.com/beeware/briefcase/blob/main/src/briefcase/commands/run.py#L16)) doesn't match, then Briefcase  won't ever identify that the process has exited, and thus the test won't finish, until it times out and is killed. 

The extra ANSI content at the start and end of the line is probably enough to throw that match off.

Running any Briefcase test on android (`briefcase run android --test`) should be sufficient to reproduce the problem locally.

matan-h:
I was able to reproduce it locally.
The test passes if I added `line = re.sub(ANSI_ESCAPE,'',line)` to `run.py/LogFilter.__call__`.
What is the preferred way to reuse this regex: does the regex need to be in another place and import it in both files?

freakboy3742:
> I was able to reproduce it locally. The test passes if I added `line = re.sub(ANSI_ESCAPE,'',line)` to `run.py/LogFilter.__call__`. What is the preferred way to reuse this regex: does the regex need to be in another place and import it in both files?

The regex is only used internally in that one LogFilter utility; however, the logging utility needs to work on essentially every platform, not just Android. It also presumably needs to preserve the escape sequence for *display* purposes, and only strip it for *filtering* match purposes, so the place where you apply that filter will be significant.

matan-h:
Oh, my place is also removing the colors from the user view. 
I find better place to strip the line: `run.py/LogFilter.test_filter.filter_func` doing `recent = re.sub(ANSI_ESCAPE,'',recent)`

matan-h:
> The regex is only used internally in that one LogFilter utility; however, the logging utility needs to work on essentially every platform, not just Android. 

Right now I use it in two places: one of them, in `console.py/Printer.export_log` and the other in 
`commands/run.py/LogFilter.test_filter.filter_func`

rmartin16:
I was able to look at this a bit longer today.

To begin, the current infrastructure in Briefcase is obviously a bit sensitive to third-party tools including ANSI sequences or other control codes in their output. To that end, we've normally just disabled streaming altogether for such output.

In the case of actually running apps, we can't just disable output streaming because Briefcase needs to actively sift through the app's output. So, we can do this if we accommodate these non-printable characters in the streamed output.

This is such an implementation:
```diff
diff --git a/src/briefcase/commands/run.py b/src/briefcase/commands/run.py
index 4977b341..1ad6c883 100644
--- a/src/briefcase/commands/run.py
+++ b/src/briefcase/commands/run.py
@@ -6,6 +6,7 @@ from abc import abstractmethod
 from contextlib import suppress
 
 from briefcase.config import AppConfig
+from briefcase.console import sanitize_text
 from briefcase.exceptions import BriefcaseCommandError, BriefcaseTestSuiteFailure
 from briefcase.integrations.subprocess import StopStreaming
 
@@ -49,7 +50,7 @@ class LogFilter:
         """
         # Compute the clean line
         if self.clean_filter:
-            filtered = self.clean_filter(line)
+            filtered = self.clean_filter(sanitize_text(line))
 
             # If the clean filter says the line can be dumped, return without yielding
             if filtered is None:
@@ -73,7 +74,7 @@ class LogFilter:
         # recent history to the most recent 10 clean lines, and build a single
         # string that is the tail of the recent clean lines.
         if included:
-            self.recent_history.append(clean_line)
+            self.recent_history.append(sanitize_text(clean_line))
             self.recent_history = self.recent_history[-10:]
             tail = "\n".join(self.recent_history)
 
diff --git a/src/briefcase/console.py b/src/briefcase/console.py
index 18532c98..1327819f 100644
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -13,6 +13,7 @@ from enum import IntEnum
 from pathlib import Path
 
 from rich.console import Console as RichConsole
+from rich.control import strip_control_codes
 from rich.highlighter import RegexHighlighter
 from rich.markup import escape
 from rich.progress import (
@@ -29,6 +30,11 @@ from briefcase import __version__
 # Regex to identify settings likely to contain sensitive information
 SENSITIVE_SETTING_RE = re.compile(r"API|TOKEN|KEY|SECRET|PASS|SIGNATURE", flags=re.I)
 
+# 7-bit and 8-bit C1 ANSI sequences
+ANSI_ESCAPE_RE = re.compile(
+    r"(?:\x1B[@-Z\\-_]|[\x80-\x9A\x9C-\x9F]|(?:\x1B\[|\x9B)[0-?]*[ -/]*[@-~])"
+)
+
 
 class InputDisabled(Exception):
     def __init__(self):
@@ -37,6 +43,15 @@ class InputDisabled(Exception):
         )
 
 
+def sanitize_text(text: str) -> str:
+    """Remove control codes and ANSI escape sequences from a line of text.
+
+    This is useful for extracting the plain text from the output of third party tools
+    that may be including markup for display in the console.
+    """
+    return ANSI_ESCAPE_RE.sub("", strip_control_codes(text))
+
+
 class RichConsoleHighlighter(RegexHighlighter):
     """Custom Rich highlighter for printing to the console.
 
@@ -99,7 +114,7 @@ class Printer:
         """
         if show:
             cls.to_console(*messages, **kwargs)
-        cls.to_log(*messages, stack_offset=stack_offset, **kwargs)
+        cls.to_log(*map(sanitize_text, messages), stack_offset=stack_offset, **kwargs)
 
     @classmethod
     def to_console(cls, *renderables, **kwargs):
diff --git a/src/briefcase/integrations/android_sdk.py b/src/briefcase/integrations/android_sdk.py
index 64871909..bc4f1685 100644
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1575,6 +1575,8 @@ Activity class not found while starting app.
                 "-s",
                 self.device,
                 "logcat",
+                "-v",
+                "color",
                 "--format=tag",
                 "--pid",  # This option is available since API level 24.
                 pid,
```

So, first...Rich does provide a mechanism to parse ANSI codes.
```python
from rich.text import Text

plain_text = Text.from_ansi('\x1b[1;32mMy\x1b[0m\x1b[1;31mawesome\x1b[0m\x1b[1;33mtext\x1b[0m').plain
```
However, this is way more than we need here since it actually interprets the codes and converts them to internal Rich styles. Therefore, I just opted for the regex to identify ANSI sequences.

More to the point, my implementation above solves two problems:
- Output put in the log file is sanitized first
- Log filters use the sanitized text for filtering

As a side note, we should also sanitize the environment variables for the log file.

As for log filtering, my primary concern with this approach is the functionality for the filter to return text that should actually be displayed in the console (via the `clean_output` argument). If we pass the filters sanitized text, then they can obviously only return sanitized text; currently, several of the filters explicitly disable this while others explicitly enable it.

To avoid only letting log filters return sanitized text, we could push down the use of `sanitize_text()` in to the filters themselves. Maybe it'd be best to drop support for actually modifying output during log filtering and only support dropping or including individual lines of output. Then, log filtering could probably more easily deal with the rich-text version of the output alongside the plain-text version.

@freakboy3742, what are you thoughts on this since you initially implemented the log filtering and all?

freakboy3742:
> @freakboy3742, what are you thoughts on this since you initially implemented the log filtering and all?

To my reading, I don't consider ANSI removal to be part of "cleaning". "Cleaning" is stripping the content that is noise from the perspective of the end user - e.g., on macOS, iOS and Android, a single "raw" log line has a *bulky* preamble that takes up half or more of an 80 column terminal. Cleaned output should strip the preamble, but leave everything else. If "cleaning" strips ANSI codes as well... why are we generating them in the first place? :-)

This obviously complicates the filtering process though - because we need to be able to handle both "with ANSI" and "without ANSI" versions of console output (because disabling color usage needs to remain an option for accessibility reasons). We also need to ensure that the patterns match if there are any variations in the ANSI codes (e.g., using a different color palette).

rmartin16:
So, I agree that removing markup isn't a part of cleaning...although, I guess I was baking in an assumption we wouldn't be attempting to parse "rich" text. That said, this is only a problem currently for macOS and iOS since Android (and other platforms) filtering doesn't change what's printed to the console. So, strictly speaking, we could enable coloring Android adb output with these changes without currently changing current functionality.

I guess I see three ways forward:
1) Support log filtering for "rich" text
2) Always run log filtering on sanitized output and do not enable "rich" output for platforms where the output is modified (i.e. macOS and iOS)
3) Do nothing

Option (1) sounds a bit risky from a maintenance perspective. Option (2) is more precarious but seems reasonable. Option (3) is simply status quo.

I am curious if Rich might provide an interface via `rich.text.Text` (or similar) to interact with "rich" text; that might help simplify Option (1).

freakboy3742:
> So, I agree that removing markup isn't a part of cleaning...although, I guess I was baking in an assumption we wouldn't be attempting to parse "rich" text. That said, this is only a problem currently for macOS and iOS since Android (and other platforms) filtering doesn't change what's printed to the console. 

That's not true: [Android filters its logs](https://github.com/beeware/briefcase/blob/main/src/briefcase/platforms/android/gradle.py#L359)

> I guess I see three ways forward:
> 
> 1. Support log filtering for "rich" text
> 2. Always run log filtering on sanitized output and do not enable "rich" output for platforms where the output is modified (i.e. macOS and iOS)
> 3. Do nothing
> 
> Option (1) sounds a bit risky from a maintenance perspective. Option (2) is more precarious but seems reasonable. Option (3) is simply status quo.

I'm not sure (2) is *that* precarious. ANSI codes are clearly identifiable - that's why they're a useful spec. It should be straightforward to remove ANSI codes from any arbitrary line of text, as long as the line is *complete* - i.e., we don't need to worry about partial codes. 

We *do* need to be clear if filters operate on pre- or post-ANSI removed content - but that's a resolvable issue, IMHO.

> I am curious if Rich might provide an interface via `rich.text.Text` (or similar) to interact with "rich" text; that might help simplify Option (1).

Possibly - although I suspect all it would do is allow us to modify regexes from including ANSI byte sequences like `\033[31m` with `[red]`.

rmartin16:
> > So, I agree that removing markup isn't a part of cleaning...although, I guess I was baking in an assumption we wouldn't be attempting to parse "rich" text. That said, this is only a problem currently for macOS and iOS since Android (and other platforms) filtering doesn't change what's printed to the console.
> 
> That's not true: [Android filters its logs](https://github.com/beeware/briefcase/blob/main/src/briefcase/platforms/android/gradle.py#L359)

Android does use a filter; however, the output of the filter does not replace the output of what's being streamed (via `clean_output=False`). The app output that's actually printed to the console is only actually being replaced/modified for macOS and iOS.

Because of this, though, we could just update the Android filter itself to strip the ANSI codes and run all the same regexes. (And we'd also need to strip the ANSI codes for the exit filter but that doesn't affect anything.)

As for my specific options, I think their interpretations may have gotten mixed up. Nonetheless, I'm not entirely opposed to regexes that account for ANSI codes...but there are certainly possible implementations here that'd achieve the goal of colored `adb` output without them.

That said, if the Android filtering was updated in the future to actually modify what's printed to the console, then the regexes would definitely need to accommodate the ANSI codes then (...this is mostly what I meant by precarious...since someone would kinda just stumble in to this realization then).

freakboy3742:
Closing this PR, as the feature has now been implemented by #1676.

PR #1599, opened 2024-01-12
Add the ability to define app permissions

### LINKED ISSUE #547: Add support for platform-specific app permissions

Some platforms (most notably mobile platforms, but desktop platforms are moving in a similar direction) have explicit permissions for accessing certain hardware features, such as:

 * Geolocation
 * Network access
 * File system access
 * Camera access
 * Accelerometer
 * Push notifications

At present, Briefcase provides no way to customise these permissions when rolling out an app.

We should add a configuration parameter that can be used to inject permissions into the app template. Suggested format:

```
[tool.briefcase.app.helloworld.iOS.permissions]

NSCameraUsageDescription = "Helloworld will use your camera to take pictures of cats"
```

In this example asking for access to the camera on iOS, the key is the platform-specific permission, and the value is a short explanation for why the permission is required. The explanatory text is required on iOS and macOS; it does not appear to be required on Android, but could still serve as useful documentation.

These definitions would then be passed to the app template, and injected as required.

Two "stretch" goals that might be worth considering:

1. **"Universal" permission aliases** Every platform has permissions for accessing the camera, geolocation, and so on. It may be worth establishing a "Briefcase alias" for these universal permissions, which is mapped to the platform specific name (or names) as required. That way, you could ask for `camera` permission, rather than needing to know the platform-specific name of each permission. 

2. **Extracting required permissions from library metadata** Manually defining permissions is simple from an implementation perspective, but doesn't lead to a good *user* experience. It is reasonable to expect that first-time app developers won't understand why their app doesn't work when they're calling a camera API, but haven't manually added permissions. Ideally, we would be able to determine the permissions required based on the libraries that are used. For example, a `toga-camera` library would declare in library metadata that it requires the `camera` permission, and that would be automatically injected into the app template. The explanatory text associated with that permission would need to be a placeholder; however, we can provide placeholder text, and warn the user that placeholder text has been provided and should be overridden (in a similar way to how default icons are currently used and a warning provided).

However, regardless of whether we hit these stretch goals, a framework for platform-specific permissions will be required. This is because OS release schedules are decoupled from Briefcase release schedules; a new OS may introduce new permissions (or alter permission semantics); and we need to be able to support those permissions without requiring a new Briefcase release.

The current alternative is to:

1. manually modify the generated project generated by `briefcase create`
2. develop a custom platform template that includes the required permissions 

(1) isn't persistent between app builds; (2) requires more effort than should be required.

freakboy3742:
Adds the ability for an app to declare permissions, as well as some related extension points.

Adds initial cross-platform permissions for `camera`, `microphone`, `fine_location`, `coarse_location`, `background_location`, and `photo_library`. This is done by adding entries to `pyproject.toml` like:

```
permission.camera = "You can use the camera to take photos of cats".
```

The cross-platform definitions are converted into platform-specific interpretations:

* iOS - adds keys to the `Info.plist`
* macOS - adds keys to the `Info.plist` and `Entitlements.plist`
* Android - adds `<uses-permission>` and `<uses-feature>` content to `AndroidManifest.xml`
* Flatpak - adds `finish-args` to the `manifest.yml`

In addition, the platform-specific mechanisms (`Info.plist` etc) have been exposed so that apps can define platform-specific modifications and extensions that aren't covered by the cross-platform capabilities.

* iOS - `info` definitions
* macOS - `info` and `entitlement` definitions
* Android - `permission` and `feature` definitions, plus extension points thats allow injection of arbitrary content into `AndroidManifest.xml`
* Flatpak - `finish_arg` definitions, plus `modules_extra_content` to allow injection of arbitrary build rules into the Flatpak manifest.
 
Requires the use of updated templates:
* iOS: beeware/briefcase-iOS-Xcode-template#30
* Android: beeware/briefcase-android-gradle-template#79
* macOS app: beeware/briefcase-macOS-app-template#35
* macOS Xcode: beeware/briefcase-macOS-Xcode-template#35
* Flatpak: beeware/briefcase-linux-flatpak-template#33

The tests (including app build tests) all pass because the extra template context generated by Briefcase is ignored if the template doesn't support the key. The template PRs are all based on *this* PR, so they're exercising the new template variables - although they're not actually doing anything with them (other than the defaults injected on macOS, Android and Flatpak)

Fixes #547.

## PR Checklist:

### DIFF
--- a/docs/reference/configuration.rst
+++ b/docs/reference/configuration.rst
@@ -487,7 +487,7 @@ If the platform requires, the value may be displayed to the user as part of an
 authorization dialog. This description should describe *why* the app requires the
 permission, rather than a generic description of the permission being requested.
 
-The use of cross-platform may also imply other settings in your app. See the individual
+The use of permissions may also imply other settings in your app. See the individual
 platform backends for details on how cross-platform permissions are mapped.
 
 Document types
--- a/docs/reference/configuration.rst
+++ b/docs/reference/configuration.rst
@@ -462,7 +462,7 @@ Permissions
 ===========
 
 Applications may also need to declare the permissions they require. Permissions are
-specified as sub-attributes of a ``permissions`` property, defined at the level of an
+specified as sub-attributes of a ``permission`` property, defined at the level of an
 project, app, or platform. Permission declarations are *cumulative*; if an application
 defines permissions at the global level, application level, *and* platform level, the
 final set of permissions will be the *merged* set of all permissions from all levels,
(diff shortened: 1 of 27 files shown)


docs/reference/configuration.rst
```diff
@@ -458,6 +458,38 @@ sources from all levels, starting from least to most specific.
...
+whatever files are used to define permissions on that platform.
+
+Permissions can also be configured by adding platform-specific configuration items. See the documentation for the the platform backends to see the options that are available.
+
+The value for each permission is a short description of why that permission is required.
+If the platform requires, the value may be displayed to the user as part of an
+authorization dialog. This description should describe *why* the app requires the
+permission, rather than a generic description of the permission being requested.
+
+The use of cross-platform may also imply other settings in your app. See the individual
```
mhsmith, line 490:
```suggestion
The use of permissions may also imply other settings in your app. See the individual
```

docs/reference/platforms/android/gradle.rst
```diff
@@ -117,9 +117,85 @@ Android allows for some customization of the colors used by your app:
...
-``tool.briefcase.app.<appname>.android`` section of your ``pyproject.toml``
-file.
+The following options can be added to the ``tool.briefcase.app.<appname>.android``
+section of your ``pyproject.toml`` file.
+
+``android_manifest_attrs_extra_content``
+----------------------------------------
+
+Additional attributes that will be added verbatim to add to the ``<manifest>``
+declaration of the ``AndroidManifest.xml`` of your app.
```
mhsmith, line 127: There's at least one too many "add"s here, and everywhere else.

mhsmith (approved):

PR #1603, opened 2024-01-13
Allow running Linux System apps for foreign targets

rmartin16:
## Changes
- Adds support for running apps from within Docker for other distributions
- The strategy uses `socat` to run a TCP proxy as a spoofed X display to the current X display
  - Most X displays run through a UNIX socket; so, the TCP proxy will normal connect there
  - Although, in the case of SSH X11 forwarding, the X display will be a TCP port provided by a SSH channel
  - X authentication is managed via `xauth` and is assumed to be using MIT cookies

## TODO
- [x] Update bootstraps for new run requirements
- [x] Bump Docker server minimum version to 20.10

## PR Checklist:


freakboy3742 (commented):
I haven't run the code to check if it works - but if it does... OMGYES. This looks remarkably straightforward and elegant, and it could only be a benefit for testing purposes.

rmartin16:
After coming to a better understanding of the ways an X11 server can be exposed, this approach seems like our best option.

An alternative straightforward approach is to run the container on the host network and allow the app to connect to the X11 server via SSH. However, this requires a few things of the host:
- Running an SSH daemon
- SSH daemon is configured for X11 forwarding
- X11 connections are enabled beyond the typical default permissions via a command like `xhost +local:docker`

Alternatively, there's many Docker wrappers that'll spawn a VNC server for the container or even ones that'll stand up a dedicated X11 server for a container.

All of these dependencies on the host machine seem particularly undesirable.

That said, this approach does not work with Docker Desktop...since it actually runs containers inside a Linux VM and you can't just bind-mount a UNIX socket from the host in to a VM and have things work out.

Additionally, **a key piece that allows this to work is the fact the user in our containers masquerades as the host user**; if the UID was different, then the host's X11 server would reject the connection with typical permissions configured.

rmartin16:
The tests are driving me insane...but this implementation is stable if you have any general thoughts before a more thorough review.

src/briefcase/bootstraps/toga.py
```diff
@@ -71,6 +71,9 @@ def pyproject_table_linux_system_debian(self):
     "libcairo2-dev",
     # Needed to compile PyGObject wheel
     "libgirepository1.0-dev",
+    # Needed to run the app
+    "gir1.2-gtk-3.0",
+    # "gir1.2-webkit2-4.0",
```
freakboy3742, line 76: minor nit - we should include a comment saying why this is commented out.
```suggestion
    # Needed if your app uses WebView
    # "gir1.2-webkit2-4.0",
```

src/briefcase/bootstraps/toga.py
```diff
@@ -112,6 +117,9 @@ def pyproject_table_linux_system_suse(self):
     "cairo-devel",
     # Needed to compile PyGObject wheel
     "gobject-introspection-devel",
+    # Needed to run the app
+    "gtk3", "typelib-1_0-Gtk-3_0",
+    # "libwebkit2gtk3", "typelib-1_0-WebKit2-4_1",
```
freakboy3742, line 122: Minor nit: these should be broken out by line, and explain why webkit is commented out:
```suggestion
    "gtk3", 
    "typelib-1_0-Gtk-3_0",
    # Needed if your app uses WebView
    # "libwebkit2gtk3", 
    # "typelib-1_0-WebKit2-4_1",
```

src/briefcase/platforms/linux/system.py
```diff
@@ -809,23 +794,24 @@ def run_app(
...
-        # Start the app in a way that lets us stream the logs
-        app_popen = self.tools.subprocess.Popen(
-            [os.fsdecode(self.binary_path(app))] + passthrough,
-            cwd=self.tools.home_path,
-            **kwargs,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.STDOUT,
-            bufsize=1,
-        )
+        with self.tools[app].app_context.run_app_context(kwargs) as kwargs:
```
freakboy3742, line 797: This is the only part that doesn't make sense to me; details in the overall comment.

freakboy3742 (commented):
A couple of minor nits, and one architectural question; but otherwise, this makes sense to me.

I still haven't been able to run this, because my current virtualisation setup won't really allow it. I'm running Linux in a VM, so I can't run Docker on Linux; and I can't run system apps on macOS. So, this review still comes with a "..but I haven't tried it" qualifier.

I agree that avoiding the requirement of an SSH daemon and port forwarding is preferable. I'd rather not get into the game of becoming network administrators for Briefcase users, especially given the level of experience that our user base generally has.

The Docker Desktop limitation is unfortunate - and I wonder if that might be what make this infeasible. I don't have a good feel for whether Linux users with Docker are using Docker Desktop or Docker Engine, but the Docker tutorial seems to direct Linux users towards Docker Desktop. Have you got any feel for whether the Desktop limitation will be a serious constraint in practice? Also - if this is a constraint, it feels like we need some sort of protection identifying if the user is on Docker Desktop so they're not left stranded with (I presume) some sort of gnarly bind error.

In terms of the specifics of the PR - the only part that doesn't make sense to me is the use of a context manager to process keyword args. Why isn't the interface a straight up call to `Popen` on the app context, with the X variable transformation being a straightforward function returning an updated kwargs set? There's no entry or exit logic, no setup or teardown to perform... I'm not sure I see what "context" we actually get.

rmartin16:
> A couple of minor nits, and one architectural question; but otherwise, this makes sense to me.
> 
> I still haven't been able to run this, because my current virtualisation setup won't really allow it. I'm running Linux in a VM, so I can't run Docker on Linux; and I can't run system apps on macOS. So, this review still comes with a "..but I haven't tried it" qualifier.

hmm....nested virtualization on macOS does seem limited right now....although, QEMU on x86 should work...I'm sure that'd be a fun rabbit hole for ya 😉 

> I agree that avoiding the requirement of an SSH daemon and port forwarding is preferable. I'd rather not get into the game of becoming network administrators for Briefcase users, especially given the level of experience that our user base generally has.
> 
> The Docker Desktop limitation is unfortunate - and I wonder if that might be what make this infeasible. I don't have a good feel for whether Linux users with Docker are using Docker Desktop or Docker Engine, but the Docker tutorial seems to direct Linux users towards Docker Desktop. Have you got any feel for whether the Desktop limitation will be a serious constraint in practice?

Hard to say; anyone familiar with Docker would not understand using Docker Desktop on Linux outside of probably development workflows that explicitly use Docker Desktop for its features that specifically target those workflows. That said, Docker seems to like the idea of pushing a more platform-agnostic form of Docker....so, maybe it'll become predominant over time. Although, Docker knows bind-mounting a socket is a missing feature with Desktop...so, maybe they'll add a proxy to support that in the future.

> Also - if this is a constraint, it feels like we need some sort of protection identifying if the user is on Docker Desktop so they're not left stranded with (I presume) some sort of gnarly bind error.

When I added support a while back for rootless Docker and Docker Desktop, I could not find a reliable way to determine how Docker is running. Docker created a very generalized infrastructure that could be used to create all kinds of front-ends (and back-ends) for Docker.....and then they made Docker Desktop as one of them.

That said, Docker does seem to use potentially reliable names of their default contexts:
```console
❯ docker context list
NAME                TYPE                DESCRIPTION                               DOCKER ENDPOINT                                    KUBERNETES ENDPOINT   ORCHESTRATOR
default *           moby                Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                                              
desktop-linux       moby                Docker Desktop                            unix:///home/russell/.docker/desktop/docker.sock                         
rootless            moby                Rootless mode                             unix:///run/user/1000/docker.sock   

❯ docker context show
default
```
We could at least try to use this as a hint when someone is using Docker Desktop. (Although, maybe this isn't a unreliable as I'm making it out to be....)

If we try to catch the error afterwards, that'll probably be difficult. Docker Desktop doesn't complain about trying to bind-mount the socket....it does as instructed but the bind-mount is unusable as a socket. So, for GTK anyway, it just errors on finding a display to use.

> In terms of the specifics of the PR - the only part that doesn't make sense to me is the use of a context manager to process keyword args. Why isn't the interface a straight up call to `Popen` on the app context, with the X variable transformation being a straightforward function returning an updated kwargs set? There's no entry or exit logic, no setup or teardown to perform... I'm not sure I see what "context" we actually get.

The context manager cleans up the `xauth` file that's created for the container.

rmartin16:
### Misconceptions

I did some more research today and it turns out the internet is a bit full of cargo culting and misinformation.

Initially, I mentioned that a Docker container can use the host's X server via SSH; this is not true....at least, not in the way it's being presented around the internet. And while I kept seeing this, it bothered me because my machine wasn't even running an SSH daemon...but I figured the X server and clients must be vendoring SSH implementations or something and were making it work.

My suspicion for all this confusion is that 1) "X11 forwarding" is commonly discussed with SSH and 2) a lot of people are trying to run graphical applications within Docker on a remote machine. Therefore, all the different pieces are jumbled together with all the copy/pasting happening everywhere.

### Reality

Consider the widely recommended command to run a graphical application in Docker, e.g.:
```bash
docker run --rm -it --net host --env DISPLAY ubuntu bash -c "apt update && apt install -y x11-apps && xeyes"
```
The trick here is understanding what `--net host` is actually doing and how the X server accepts connections.

Superficially, it gives the container the same access to the network that the host has. But...on modern Linux, the network isn't just the network 🤔. Specifically, `--net host` configuration puts the container in the same "network namespace" as the host machine; while normally, Docker will create an isolated network namespace for a container to use.

Network namespaces manage access-control not only for network resources but also for "abstract UNIX domain sockets". These are relevant because not only does Xorg's X server expose a socket via the files at `/tmp/.X11-unix/X#`, but those filepaths are also used to define abstract sockets with the same behavior.

So, while my initial thought was `--net host` was allowing the container to talk to the X server with TCP, most (or all) distros disable TCP communication with their X server. Instead, this setting is actually allowing the container access to the X server's abstract socket (which you can see via `netstat -x`) and the `DISPLAY` environment variable obviously tells it what display to use.

### Our Situation

My implementation avoids using `--net host` and instead directly exposes the relevant UNIX socket for the X server (combined with some X authentication stuff I'm leaving out so we can focus on even just getting a connection).

This approach allows both Docker Engine as well as rootless Docker to use the X server....however, this fails with Docker Desktop because a socket cannot be bind-mounted in to a container in Docker Desktop. Furthermore, though, `--net host` _only_ works with Docker Engine because neither Docker Desktop nor rootless Docker can actually use the same network namespace as the host....in general, `--net host` seems mostly like a no-op for these two Docker implementations.

Therefore, I really only see one solution right now for Docker Desktop: configure a proxy for the X server's socket.

Such a proxy could relatively straightforwardly be set up with `socat` or even `ssh` back to localhost. However, to perform this on an arbitrary machine programmatically starts to become a little much perhaps...

note: an example of someone explaining this properly: https://blog.cykerway.com/posts/2021/07/08/run-gui-applications-in-docker.html

rmartin16:
For posterity, this will expose the X11 socket via TCP:

Run on the host and leave running:
```bash
DISPLAY_NUM=$(echo $DISPLAY | cut -d ':' -f 2 | cut -d '.' -f 1)
DOCKER_BRIDGE_IP=$(ifconfig docker0 | awk '/inet / {print $2}')
socat -ddd TCP-LISTEN:$((6000+$DISPLAY_NUM)),reuseaddr,fork,bind=$DOCKER_BRIDGE_IP ABSTRACT-CONNECT:/tmp/.X11-unix/X$DISPLAY_NUM
```
In a separate terminal, start the container in Docker Desktop:
```bash
docker run --rm -it --net host --env DISPLAY=172.20.0.1:1 ubuntu bash -c "apt update && apt install -y x11-apps && xeyes"
```
Viola
image

freakboy3742:
> For posterity, this will expose the X11 socket via TCP:
> 
> Run on the host and leave running:
> 
> ```shell
> DISPLAY_NUM=$(echo $DISPLAY | cut -d ':' -f 2 | cut -d '.' -f 1)
> DOCKER_BRIDGE_IP=$(ifconfig docker0 | awk '/inet / {print $2}')
> socat -ddd TCP-LISTEN:$((6000+$DISPLAY_NUM)),reuseaddr,fork,bind=$DOCKER_BRIDGE_IP ABSTRACT-CONNECT:/tmp/.X11-unix/X$DISPLAY_NUM
> ```

That... doesn't seem so bad - and it would seem to lend itself to being something that Briefcase manages as aPopen that is cleaned up when the process exits...

Also, throwing another spanner in the works: Wayland.

(you're welcome :-)

rmartin16:
> That... doesn't seem so bad - and it would seem to lend itself to being something that Briefcase manages as aPopen that is cleaned up when the process exits...

Yeah....although, I'm trying to replicate this on Fedora and running in to issues with sharing the socket over TCP through the `docker0` bridge interface. Potentially resolvable...

> Also, throwing another spanner in the works: Wayland.

True; however, Wayland runs a Xwayland process that mimics an X environment including these sockets. So, I imagine for the foreseeable future, this will continue being the case....that said, my initial testing with Fedora 39 isn't going well with wayland sooooo.....more fun I'm sure

rmartin16:
I was able to resolve my networking woes....but it required a much deeper understanding of networking with Docker Desktop.

### Background

The original design of Docker includes the `docker0` bridge interface; by default, all containers are attached to it and it facilitates communication among the containers as well as the host (in limited ways anyway) and the broader internet. Abstractions on top of Docker, e.g. docker-compose, will create separate networks for each container and in many ways obviate `docker0`....since arguably, having any random container be able to talk to any other random container by default doesn't really scream "isolated and secure".

At any rate, Docker Desktop (notoriously at this point) runs its containers in a Linux VM; therefore, the default `docker0` bridge interface is a virtual interface inside of the VM....so, this is why you can't bind a server to the `docker0` interface on the host and have it available inside the container on the `docker0` interface....because they are two completely different and isolated interfaces. (It was only working on my development machine because my host `docker0` interface is using a different subnet than the `docker0` in Docker Desktop and my host's iptables were routing the requests....that said, even when I changed the subnet for `docker0` in Docker Desktop on Fedora, I still couldn't access the host...so, this seems unreliable at best).

### Accessing network-based services on the host

Finally, I discovered Docker's solution for this. In Docker Desktop, a DNS server is running in the VM and it is mapping `host.docker.internal` to the network assigned IP of the host machine. Docker [provides](https://docs.docker.com/desktop/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host) this facility specifically to support development workflows where a server is running on the host and containers need access to it.

However, it is only available by default when using Docker Desktop; so, when just using Docker Engine, the `docker run` command needs to be [augmented](https://docs.docker.com/engine/reference/commandline/run/#add-host) with `--add-host host.docker.internal:host-gateway`. This adds an entry to the `/etc/hosts` file in the container that directs `host.docker.internal` to the IP of `docker0`. (Although, if you add this while using Docker Desktop, it is the IP assigned to the host machine from the network.)

### Conclusion

The X socket can be exposed via TCP but it must be binded to an address available to the container. For Docker Desktop, this will need to be the host's network assigned IP. For Docker Engine, this would need to be the IP of the default Docker bridge (unless we create a dedicated bridge interface for this purpose). Alternatively, we could just bind to 0.0.0.0.

Additionally, the minimum Docker version must be bumped to [20.10](https://docs.docker.com/engine/release-notes/20.10/#20100) (released dec 2020) to ensure `host.docker.internal` is available.

rmartin16:
I modified the implementation to use a TCP proxy for the host's X socket. However, my initial testing is showing that X authorization is seemingly being ignored entirely now. While the proxy was open on my main machine, I could open windows on my host from a VM running on entirely different machine 👀 sooo....I'll need to figure out what's going on...or accept this is just really insecure, I guess.

rmartin16:
### Yet more evolved thoughts on this design

1) If an existing TCP server for the X display exists, it should be used
    - The X11 standard reserves ports 6000 + <$DISPLAY> for this; so, for `$DISPLAY=:10`, the port is 6010
    - Such a TCP server should be exposed to the Docker container and specified in the container's `DISPLAY` env var
    - The most likely scenario for this is an SSH connection with X11 forwarding configured
    - However, such an existing TCP server would likely be bound to the loopback interface...therefore, it would likely be necessary to setup a TCP proxy bound to the other interfaces back to the loopback interface
    - Doing so may require spoofing a new X display....which complicates authentication but that can be worked around with a little parsing of the `xauth` commands
2) Otherwise, if a `/tmp/.X11-unix/X#` socket is available, use that
    - Initially, I had the TCP proxy connecting to the abstract socket, but those aren't portable beyond Linux; so, just connect the proxy to the file-based UNIX socket

### Notes
- The use case of proxying an existing TCP connection for X is debatable
  - Although, I'd like to see this all working even for X11 forwarding
  - In theory, it shouldn't be too must more difficult than the proxy for the UNIX socket....I'll have a better idea once I have it working
- TCP proxy cannot just be bound to the loopback interface
  - From my testing and research, the proxy will likely have to be exposed to the network so the Docker container can access it
  - For Docker Engine and rootless Docker, it would be possible to just bind the proxy to the `docker0` interface
  - But for Docker Desktop, AFAICT, the host's actual network interface is attached to a network defined for the VM; from there, a network bridge is created in Docker to expose this network from the VM.
    - So, unless a service on the host is binding to the "real" network interface, it won't be available in the container
    - This is obviously in contrast to Docker Engine where the network interface, `docker0`, is literally the same thing inside of a Docker container and outside

rmartin16:
Finally got this in to a working state for an X display using a TCP transport or a UNIX socket.

The process for all this now is a bit more encumbered, though, now...
- Allocate an X display to spoof for the proxy
  - There's no formal process for this; so, choose the first display number that doesn't have an existing TCP server running or a UNIX socket already in use
  - This largely uses openSSH's [implementation](https://github.com/openssh/libopenssh/blob/05dfdd5f54d9a1bae5544141a7ee65baa3313ecd/ssh/channels.c#L3559) as inspiration
- Start TCP proxy server for the spoofed X display
  - If a TCP server is running for the current X display, proxy to it
  - Else if a UNIX socket is running for the current X display, proxy to it
  - The TCP server for the spoofed display is binded to 0.0.0.0 so Docker Desktop containers can access
    - This will technically expose the X display to the entire local network, but the xauth file should limit arbitrary connections....as long as the user hasn't effectively disabled X authentication altogether on the host
    - That said, none of this is especially secure....but it's only set up temporarily and is just for development
- Write an xauth file for connections to the spoofed X display to use
  - Extract the X11 cookie configured for the current display
  - Use that cookie to create an authentication entry for the spoofed display
  - Re-write this xauth entry to support any hostname in requests
  - This also uses openSSH's [implementation](https://github.com/openssh/libopenssh/blob/05dfdd5f54d9a1bae5544141a7ee65baa3313ecd/ssh/clientloop.c#L294) for inspiration
- Finally, configure the Docker container to use the spoofed display by default
  - `DISPLAY=host.docker.internal:<spoofed display number>`
  - `XAUTHORITY=/path/to/xauth_file/for/spoofed/display`

This implementation makes a bunch of assumptions....but those assumptions are basically along the lines of "this assumes you are using a modern Linux distro with sane configuration".

I'll need to clean it up a bit more and finish the tests....but this is mostly what I was trying to accomplish.

Another general issue is the continued use of `which` to check for dependency existence late in to the process. I'm kinda inclined to create some sort of "linux distro" abstraction as a tool. That way, the dependencies and all the details about each Linux for the Linux System build process could be better contained as well. I haven't formalized this idea yet, though.

rmartin16:
@freakboy3742, if you're recovered and feeling better, I'd appreciate an additional high level review to make sure we're on the same page before I wrap things up.

Notable changes:
 - Added support for apps running inside a container to connect to the host's X11 server
   - As detailed in the previous post, this creates a TCP proxy to the existing X display via a spoofed X display and hijacks the x auth for the current display so any connections to the spoofed display use it
   - The primary use-case for this additional complexity over just creating a TCP proxy to the UNIX socket is SSH X11 forwarding....but also just in case there is an existing TCP server running for the current X display
 - This connection to the X server is primarily facilitated by the `x11_passthrough()` context manager
   - I added this to the `Docker` class instead of `DockerAppContext` since this is really applicable to any Docker container....not just ones running for a specific app
 - Abstracted the dockerization of subprocess arguments from `DockerAppContext` to `Docker`
   - In line the rationale for adding `x11_passthrough()` to `Docker`, all of these transformations are generic to running containers in general
   - That said, some of the dockerization is specific to an app; so, a `dockerize_args()` wrapper still exists in `DockerAppContext`

freakboy3742 (commented):
> if (or when) you're recovered and feeling better

Rumors of my demise have been exaggerated... but only slightly 😷

With the caveat that I only *barely* understand the network mechanics at play, and I'm severely limited in my ability to test the resulting code - this looks fine to me. All the X11 passthrough code is abstracted in a way that makes sense (to me, at least); other than a couple of cosmetic things, I think this is headed in the right direction.

src/briefcase/integrations/docker.py
```diff
@@ -378,9 +380,282 @@ def check_output(self, args: list[SubprocessArgT], image_tag: str) -> str:
...
+        :returns: augmented keyword args for the call to subprocess
+        """
+        if not (DISPLAY := self.tools.os.getenv("DISPLAY")):
+            raise BriefcaseCommandError(
+                "The DISPLAY environment variable must be set to run an app in Docker"
+            )
+
+        # Create a TCP proxy for a spoofed X display
+        proxy_process, proxy_display_num = self._x11_tcp_proxy(DISPLAY)
+        xauth_file_path = self.tools.base_path / f".docker.xauth.{proxy_display_num}"
```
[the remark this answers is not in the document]
rmartin16: IIRC, I moved this out of `/tmp/` because of Docker Desktop; it maintains a superset of directories that bind mounts can use to expose directories/files to containers. I think a recent local install of Docker Desktop for me did not include `/tmp/`....so, it just errored when Briefcase tried to start the container.

I agree, though, about using Briefcase's `tools` directory. I think Docker Desktop includes the user's `home` directory by default.....so, there might be that but I'll investigate further.

rmartin16:
> > if (or when) you're recovered and feeling better
> 
> Rumors of my demise have been exaggerated... but only slightly 😷

Glad to hear you're coming out the other side :)

> With the caveat that I only _barely_ understand the network mechanics at play, and I'm severely limited in my ability to test the resulting code - this looks fine to me. All the X11 passthrough code is abstracted in a way that makes sense (to me, at least); other than a couple of cosmetic things, I think this is headed in the right direction.

There is definitely a lot going on among several disparate systems with a lot of leveraging of opaque mechanisms. I've tried to reveal them in comments but feel free to point out gaps and I can try to fill them in.

I was thinking, once I've finalized the implementation, of dropping some directions in the Discord to test this out if any one's willing...

freakboy3742:
> > With the caveat that I only _barely_ understand the network mechanics at play, and I'm severely limited in my ability to test the resulting code - this looks fine to me. All the X11 passthrough code is abstracted in a way that makes sense (to me, at least); other than a couple of cosmetic things, I think this is headed in the right direction.
> 
> There is definitely a lot going on among several disparate systems with a lot of leveraging of opaque mechanisms. I've tried to reveal them in comments but feel free to point out gaps and I can try to fill them in.

The bit I'm missing is more at the level of "here's a 101 primer on how X11 tunnelling works", rather than a quick comment on a single line of code. I think I have a *broad* understanding of how the pieces are working here; but I wouldn't want to have to sit an accreditation exam on the topic :-)

It *might* be worth writing up what you know/have learned as an internal topic guide. It doesn't need to be exceedingly  detailed or polished, but having a "this is how it's meant to work" guide that we can use as a reference could be helpful for future maintenance. [Toga has one of these sections](https://toga.readthedocs.io/en/stable/background/internals/index.html) already - this would be as a good a time as any to add one for Briefcase.
 
> I was thinking, once I've finalized the implementation, of dropping some directions in the Discord to test this out if any one's willing...

Definite +1 on this. Having Linux users hit it with a stick will definitely reveal all the weird-ass network configurations that will cause this headaches.

src/briefcase/integrations/docker.py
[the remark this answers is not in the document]
rmartin16: Indeed, by default, Docker Desktop limits bind mounts to directories/files under the user's `home` directory.

> By default the `/home/<user>` directory is shared. If your project is outside this directory then it must be added to the list, otherwise you may get `Mounts denied` or `cannot start service` errors at runtime.

<sup>source: https://docs.docker.com/desktop/settings/linux/#file-sharing</sup>

So, if this temporary file is moved out of the Briefcase tools directory, we are limited to somewhere in the user's `home` directory....which probably just means the root of the user's home directory.

Otherwise, we'd need to drop an error and try to get users to add the `/tmp` directory to their Docker Desktop settings.

rmartin16:
Quite frustratingly spent most of the day chasing my tail trying to figure out why this seemingly wouldn't work on Fedora.

Well, when you change the Docker context (for example, switch from Docker Engine to Docker Desktop), you need to re-run `briefcase create --target ubuntu` so the proper `Dockerfile` is used and in turn the proper container is used....or more specifically, so you're appropriately either `root` or `brutus` inside the container.....since `xauth` will not use an `XAUTHORITY` file owned by someone else... 🙃 

Additionally, though, this revealed an interesting difference between Fedora and Ubuntu.

On Ubuntu, the X server doesn't require authorization for the logged in user.
```console
❯ xhost
access control enabled, only authorized clients can connect
SI:localuser:russell
```
AFAICT, the user is inferred by the owner of the process that connects to the UNIX socket for the X display. Although, since we're using a TCP proxy...._anything_ that sends data to the proxy ends up interpreted as coming from the logged in user....part of the fun insecurity of X in general...

At any rate, Fedora doesn't seem to start user sessions with the same permissions for the logged in user.
```console
❯ xhost
access control enabled, only authorized clients can connect
```
The absence of any entries on Fedora just means all X server access must be cookie authenticated.

tests/integrations/docker/test_DockerAppContext__run.py
```diff
@@ -37,12 +39,14 @@ def test_simple_call(mock_docker_app_context, tmp_path, sub_stream_kw, capsys):
...
+@pytest.mark.usefixtures("mock_docker")
+@pytest.mark.usefixtures("mock_docker_app_context")
+def test_interactive(mock_tools, my_app, tmp_path, sub_kw, capsys):
     """Docker can be invoked in interactive mode."""
-    mock_docker_app_context.run(["hello", "world"], interactive=True)
+    mock_tools[my_app].app_context.run(["hello", "world"], interactive=True)
 
     # Interactive means the call to run is direct, rather than going through Popen
-    mock_docker_app_context.tools.subprocess._subprocess.run.assert_called_once_with(
+    mock_tools.subprocess._subprocess.run.assert_called_once_with(
```
rmartin16, line 49: Still have quite a bit of work to do....but was curious on your thoughts on this use of fixtures?

Several times in these tests, I'll add a fixture to the parameter list just because I need its code to run....not because I actually need a symbol for it; well, `pytest.mark.usefixture()` to the rescue.

Furthermore, though, the way I've implemented it for the `Docker` tests, it basically allows us to add tools to `mock_tools` quite nicely, IMO. I think this approach could be useful for removing blocks of code that are repeated in many related tests that effectively make a tool available or do some needed set up.

(just beware...you can't use `pytest.mark.usefixture` if the function is already also decorated with `pytest.fixture`...)

freakboy3742: Hrm... I think I'm a "meh". I've kind of internalised the "a fixture is something that runs, but doesn't require you to use it" thinking, so I'm not sure what we gain out of having a separate `use_fixtures` decorator. I don't fundamentally object to the pattern - I just don't see what we get from using it - especially since it moves fixture naming to be a string, rather than a symbol.

rmartin16:
@freakboy3742, would like to make sure we're on the same page with this part of my refactor.

So, after this PR, if you want to run a subprocess in a container, everything basically runs through `Docker.dockerize_args()`.

As a result of this, many of the signatures are also now just `func(*args, **kwargs)` because the signature is especially unwieldy and 5 or so methods are using it.

Additionally, it seems as though the primary purpose of `dockerize_args()` was to focus on the "args" that composed the actual command line to run. As a part of this refactor, the "args" now represent all of the args and kwargs that are necessary to drive `subprocess`. I primarily did this because several of the command line args to configure `docker` are taken from arguments that normally drive `subprocess` itself, for instance, `env` and `cwd`. Trying to manage these two groups of arguments that are intimately intertwined was proving difficult...hence why I ultimately combined them.

At any rate, this abstraction makes sense, IMO....but it means that the signature for something like `DockerAppContext.check_output()` is a bit divorced from its concrete definition now. That said, there isn't too far to go to `Docker.dockerize_args()` to figure out one's options....and I tried to communicate that kwargs for `subprocess` are also in the set of options.

**So, do you mostly agree with this state of the re-factor?**

And if so, how extensively do you think something like `DockerAppContext.check_output()` needs to be tested? For instance, should each of the possible inputs for `Docker.dockerize_args()` be tested again for `check_output()`, `run()`, and `Popen()`? My thoughts are no...because the tests for `Docker.dockerize_args()` itself will confirm those work. Instead, I'd just make sure that `dockerize_args()` is called as expected and maybe inspect some logging output.

Thanks.

src/briefcase/integrations/docker.py
```diff
@@ -378,9 +381,377 @@ def check_output(self, args: list[SubprocessArgT], image_tag: str) -> str:
...
+
+    def _x11_tcp_proxy(self, DISPLAY: str) -> tuple[subprocess.Popen, int]:
+        """Starts a TCP proxy as a spoofed X display.
+
+        The TCP server's port is chosen as 6000 + the display number as defined in the
+        X11 standard. In this way, when X clients attempt to make a connection for the
+        chosen display number, this TCP port will be automatically evaluated by the
+        client for consideration as an X display.
+
+        The proxy is binded to 0.0.0.0 and as such will be reachable on all network
```
freakboy3742, line 561:
```suggestion
        The proxy is bound to 0.0.0.0 and as such will be reachable on all network
```

src/briefcase/integrations/docker.py
```diff
@@ -378,9 +381,377 @@ def check_output(self, args: list[SubprocessArgT], image_tag: str) -> str:
...
+                break
+        else:
+            raise BriefcaseCommandError("Failed to allocate X11 display")
+
+        return display_num
+
+    def _x11_is_display_tcp(self, display_num: int) -> bool:
+        """Is a TCP server running for the X display?"""
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock.settimeout(3)
```
freakboy3742, line 658: If I'm reading this right, the pathological case could take ~150s to resolve that there's no available display. I presume this isn't likely in practice... what *is* a likely delay? Would it be worth putting the display discovery in a waitbar?

src/briefcase/integrations/docker.py
```diff
@@ -378,9 +381,377 @@ def check_output(self, args: list[SubprocessArgT], image_tag: str) -> str:
...
+        approach is similar to that of OpenSSH's strategy for reserving an X display
+        for X11 forwarding: simply iterate over candidate display numbers using the
+        first one that is not already is use.
+
+        This check for whether a display number is already in use is limited to
+        verifying a TCP server or UNIX socket does not already exist for the display.
+        While other transport protocols are possible, any modern system should be using
+        at least one of these transports in tandem with anything else.
+        """
+        for num in range(20, 167 + 1, 3):
```
freakboy3742, line 646: Are these magic numbers literally the same as those used by OpenSSH? Is there any rhyme or reason to this selection?

freakboy3742 (changes requested):
tl;dr - I agree the direction of this refactor makes sense. Yes, it means the literal prototype of check_output et al is a little obscured, but not in a way that I think is problematic - if anything, this approach means it's a lot more obvious what is going on (as `dockerize_args` is a "complete" wrapper for invoking a process).

I also agree that a full external interface test for `check_output` etc isn't required. The complex logic is all in `dockerize_args`; as long as that is battle tested, and there's a notional test of check_output to confirm that things being passed into `check_output` are being passed through to `dockerize_args`, I'd be happy. 

The only caveat I'd put on that is that if the tests being done for `dockerize_args` could be parameterised in a meaningful way and rolled out over all the "public" interfaces, then it might be worth having the full interface tests on the basis that it's "just another parameterised rollout"... but I suspect verifying the output won't be that simple in practice.

Otherwise, this is looking really solid. I mean, it's incredibly complex, but it's also doing incredibly complex things - but the abstraction is fairly clean and makes sense under the circumstances. A couple of minor tweaks and clarification inline, but otherwise this is looking pretty close.

src/briefcase/integrations/docker.py
rmartin16: This one's kinda funny....my IDE has had this squiggly-underlined since I wrote it. It's mostly a personal disagreement I have with current rules of English. To my benefit of my viewpoint, though, there is at least one relevant example: hung vs hanged...although, a bit more graphic. Nonetheless, I'll relent.

rmartin16:
Ok; the implementation should be complete at this point....so, we can get down to the nitty gritty. (I still need to update the bootstraps but that isnt too big of a deal.)

With that, whenever you'd like to start a more thorough review should be fine at this point. I'll address your current comments as well.

After experimenting more in WSL, I realized that X authentication with cookies isn't implemented....furthermore, I'm not sure if any authentication is implemented. This means the attempt to write an xauth database file for the spoofed display would always fail. To accommodate this, I added a warning when xauth handling fails and Docker is launched with basically just `DISPLAY`....this works in WSL anyway.

I also still need to write up some documentation and clean up the top PR post to reflect the current implementation.

src/briefcase/integrations/docker.py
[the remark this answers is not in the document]
rmartin16: I updated this to use the `build` directory in the CWD...which should be the projects `build` directory.

src/briefcase/integrations/docker.py
rmartin16: These numbers are mostly arbitrary....with a goal of minimizing the time to identify an available display number.

As for OpenSSH, they run a similar [loop](https://github.com/openssh/libopenssh/blob/05dfdd5f54d9a1bae5544141a7ee65baa3313ecd/ssh/channels.c#L3559). By default (via `sshd_config`), the loop starts at display number 10 and incrementing by 1 each time, uses the first port that is available on all network interfaces. Additionally, by default, they will actually iterate this process up to nearly 1000 times to find a display 👀 

I'm definitely open to thoughts on my implementation; I've been mostly focused on getting things operational with the intention to fine tune after.

src/briefcase/integrations/docker.py
rmartin16: Interestingly...when I first implemented this, I forgot to include a `break`; so, for days, it was always connecting to all 50 ports and returning the last one when the loop ended. The loop would run so fast, I never even noticed.

So, setting a timeout at all is just to ensure this doesn't block indefinitely on a port that's completely misbehaving.

The truly pathological case where all 50 ports block for 3 seconds....that'd be interesting. I think that would mean "something" is listening....but not responding. Because if nothing is listening on that port, the underlying OS should immediately reject the connection attempt (I'd expect anyway).

Sooo, I'm not completely against adding a Wait Bar....but setting up and tearing down the Wait Bar would probably take longer than the 99.9% case where the first connection works and is returned.

src/briefcase/integrations/docker.py
freakboy3742: Don't get me started on the US need to use "drug" as a the past participle of drag. :-)

src/briefcase/integrations/docker.py
freakboy3742: Ok - if the "full timeout" case is borderline impossible, I'm OK without a waitbar here.

src/briefcase/integrations/docker.py
freakboy3742: Ok - I guess I'm just looking for a comment that explicitly says these numbers are arbitrary (with maybe an explanation of the broad strategy), rather than some deep magick in how X11 uses ports.

rmartin16: Gotcha; after further consideration, I removed the particularly arbitrary step-size of 3, increased the range from 50 candidates to 250, and re-positioned the range to start at 50.

My initial consideration with a larger step size was to hopefully get out of a range collision with another process allocating displays sequentially. However, if there was a collision, the more likely case may be multiple instances of Briefcase running where a larger step-size isn't going to help anyway....unless it varied by user.

I should also mention this all has an inherent race condition where two instances of Briefcase (or something else altogether) could allocate the same display. Given the expected use-cases, this seems especially unlikely to actually happen, though.

src/briefcase/bootstraps/toga.py
```diff
@@ -71,6 +71,10 @@ def pyproject_table_linux_system_debian(self):
     "libcairo2-dev",
     # Needed to compile PyGObject wheel
     "libgirepository1.0-dev",
+    # Needed to run the app in Docker
```
freakboy3742, line 74: Why "in docker"? Won't this be a requirement regardless - just one that virtually every Debian install will have by default?

rmartin16: > Won't this be a requirement regardless

It's only a requirement to _run_ the app....not to build it. If you aren't using Docker, then Briefcase will tell you the `system_requires_runtimes` list needs to be installed first.

But I guess this brings up a larger point....

By adding these runtime requirements to `system_requires`, we're really co-opting its purpose which to specify the build-time requirements.

Maybe we should just have the `Dockerfile` install both...

rmartin16: Instead of duplicating the runtime requirements in to `system_requires`, I just updated the Dockerfile invocation to install both `system_requires` and `system_runtime_requires`. This makes so much more sense in retrospect.

src/briefcase/bootstraps/toga.py
rmartin16: n/a. Updated Dockerfile handling to use both `system_requires` and `system_runtime_requires` when it creates the image.

src/briefcase/bootstraps/pyside6.py
```diff
@@ -79,34 +79,25 @@ def pyproject_table_linux_system_debian(self):
...
     "libsm6",
     "libice6",
     "libglib2.0-0",
     "libgl1",
-    "libegl1-mesa",
+    "libegl1",
     "libdbus-1-3",
-    "libgssapi-krb5-2",
 ]
 """
```
rmartin16, line 102: I went through the effort of de-duping this a bit. Several of these are implicitly included as dependencies by others and at least one was a dummy package whose usefulness expired long ago. I also got rid of the kerberos package....I htink that was just something about my machine.

docs/how-to/internal/x11passthrough.rst
```diff
@@ -0,0 +1,270 @@
...
+X Window System Background
+--------------------------
+
+Nearly all modern Linux distributions use `Version 11
+<https://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html>`__ of the X
+Window System Protocol (X11) as the foundation for their graphical displays.
+Notably, though, `Wayland <https://wayland.freedesktop.org/>`__ is gaining
+traction as the successor to X11. Due to reasons of entrenchment, Wayland will
+likely continue providing their X11 compatibility layer to support X
+applications well in to the future.
```
freakboy3742, line 19:
```suggestion
Linux distributions use either `X Windows <https://www.x.org/>`__(sometimes called X or
X11) or `Wayland <https://wayland.freedesktop.org/>`__ to manage their graphical
displays. X11 is the older of the two; Wayland maintains compatibility with the `X11
protocol <https://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html>`__ for
backwards compatibility.
```

docs/how-to/internal/x11passthrough.rst
```diff
@@ -0,0 +1,270 @@
...
+``HOSTNAME`` is absent, it is assumed to be the machine the client is running
+on. The ``SCREENNUMBER`` is largely historical since all monitors are normally
+collapsed together in to a single screen now. Therefore, an expected setting
+for ``DISPLAY`` is ``:0`` for many installations.
+
+While security is relatively weak for X11, there are basic facilities to
+mitigate unauthenticated access. The ``XAUTHORITY`` environment variable can
+specify a file path to an ``xauth`` database; if ``XAUTHORITY`` is not set,
+there will normally be a system default file path configured foe each user. The
+``xauth`` database, itself, is protected by file system access controls and
```
freakboy3742, line 39:
```suggestion
there will normally be a system default file path configured for each user. The
``xauth`` database itself is protected by file system access controls and
```

docs/how-to/internal/x11passthrough.rst
```diff
@@ -0,0 +1,270 @@
...
+any process owned by the logged in user access to the display. This is
+configured by ``xhost``.
+
+Example ``xhost`` configuration:
+
+.. code-block:: console
+
+  $ xhost
+  access control enabled, only authorized clients can connect
+  SI:localuser:russell
```
freakboy3742, line 62:
```suggestion
  SI:localuser:brutus
```

docs/how-to/internal/x11passthrough.rst
```diff
@@ -0,0 +1,270 @@
...
+   $ xauth list
+   jupiter/unix:  MIT-MAGIC-COOKIE-1  9e9a67185b1fdc0c46e00dc400559873
+   #ffff#6a757069746572#:  MIT-MAGIC-COOKIE-1  9e9a67185b1fdc0c46e00dc400559873
+
+Along with cookie-based authentication, it is also possible to add entities to
+an allowlist for a display. For instance, some distros are configured to allow
+any process owned by the logged in user access to the display. This is
+configured by ``xhost``.
+
+Example ``xhost`` configuration:
```
freakboy3742, line 56:
```suggestion
An ``xhost`` configuration authorising a user named ``brutus`` might look like:
```

docs/how-to/internal/x11passthrough.rst
```diff
@@ -0,0 +1,270 @@
...
+
+While security is relatively weak for X11, there are basic facilities to
+mitigate unauthenticated access. The ``XAUTHORITY`` environment variable can
+specify a file path to an ``xauth`` database; if ``XAUTHORITY`` is not set,
+there will normally be a system default file path configured foe each user. The
+``xauth`` database, itself, is protected by file system access controls and
+will contain "cookies" for individual displays that assigned by the X server
+for clients to use to facilitate authentication.
+
+An example ``xauth`` database:
```
freakboy3742, line 43:
```suggestion
An ``xauth`` database authorising a host named ``jupiter`` might look like:
```

docs/how-to/internal/x11passthrough.rst
```diff
@@ -0,0 +1,270 @@
...
+Docker Desktop VM, it is much different than Docker Engine and abstract sockets
+on the host are not available to the container.
+
+Along with not being possible to expose abstract sockets on the host to a
+container running via Docker Desktop, it is also not possible to expose
+arbitrary UNIX sockets either. Therefore, attempting to bind mount
+``/tmp/.X11-unix/X0``, for instance, in to a Docker Desktop container will not
+allow processes inside the container to successfully communicate with the
+socket. (The Docker team has added support to pass specific sockets such as the
+socket Docker, itself, uses as well as the SSH agent socket; but exposing
```
freakboy3742, line 139:
```suggestion
socket Docker itself uses, as well as the SSH agent socket; but exposing
```

docs/how-to/internal/x11passthrough.rst
```diff
@@ -0,0 +1,270 @@
...
+~~~~~~~~~
+
+The `socat <https://repo.or.cz/socat.git>`__ tool is a widely available program
+to relay bi-directional data transfer between independent data channels. It
+allows running a process on the host to listen on a network port and send any
+received data to a socket connected to an X display on the other side.
+
+Creating a TCP proxy for the X display effectively creates a spoofed X display.
+The proxy is configured to listen on the TCP port for an unallocated display;
+the port number will be 6000 + the number of the display. Additionally, the
```
freakboy3742, line 201: Is this a standard, a convention, or something specific to Briefcase?

docs/how-to/internal/x11passthrough.rst
```diff
@@ -0,0 +1,270 @@
...
+``xauth`` program allows for reading and writing the database. The database is
+relatively simple mapping of display information to a cookie. When a client
+wants to establish a connection for a display, it queries the database for the
+display and receives the cookie back.
+
+Since the proxy creates a spoofed display, a new ``xauth`` database needs to be
+created for the spoofed display using the authentication afforded to the user
+for the current display.
+
+Steps to create new database:
```
freakboy3742, line 224:
```suggestion
To create new database, you need to:
```

docs/how-to/internal/x11passthrough.rst
```diff
@@ -0,0 +1,270 @@
...
+ - Extract the cookie for the current display
+ - Create a new database file
+ - Add an entry for the spoofed display using the extracted cookie to the new
+   database
+ - Rewrite the hostname of the entry that was just created to be "FamilyWild"
```
freakboy3742, line 229: What's the significance of the FamilyWild name?

docs/how-to/internal/x11passthrough.rst
```diff
@@ -0,0 +1,270 @@
...
+An example ``xauth`` database:
+
+.. code-block:: console
+
+   $ xauth list
+   jupiter/unix:  MIT-MAGIC-COOKIE-1  9e9a67185b1fdc0c46e00dc400559873
+   #ffff#6a757069746572#:  MIT-MAGIC-COOKIE-1  9e9a67185b1fdc0c46e00dc400559873
+
+Along with cookie-based authentication, it is also possible to add entities to
+an allowlist for a display. For instance, some distros are configured to allow
```
freakboy3742, line 52:
```suggestion
an allowlist for a display. For instance, some distributions are configured to allow
```

docs/how-to/internal/x11passthrough.rst
```diff
@@ -0,0 +1,270 @@
...
+Such a UNIX socket connection will be configured for each display for the
+machine. As with many things in Linux, UNIX sockets are exposed in the machine
+via a file in the ``/tmp/.X11-unix/`` directory. The socket files are named for
+the number of the DISPLAY they are connected to; so, the file for Display 0 is
+``X0``. Therefore, any display can be found at ``/tmp/.X11-unix/X#``.
+
+Along with a UNIX socket connection, X servers can also listen on a TCP socket
+on the machine's network interfaces. However, since a network connection can
+easily be reached by other machines, listening on a TCP socket is normally
+disabled on most distros. That said, the X11 standard reserves port numbers
```
freakboy3742, line 84:
```suggestion
disabled on most Linux distributions. That said, the X11 standard reserves port numbers
```

docs/spelling_wordlist
```diff
@@ -19,6 +20,7 @@ DBus
 dev
 dialogs
 distributable
+distros
```
freakboy3742, line 23: Shouldn't be needed with other modifications above.
```suggestion
```

src/briefcase/integrations/docker.py
```diff
@@ -378,9 +385,406 @@ def check_output(self, args: list[SubprocessArgT], image_tag: str) -> str:
...
+        try:
+            # Create the xauth database file for the spoofed display
+            try:
+                self._x11_write_xauth_file(DISPLAY, xauth_file_path, proxy_display_num)
+            except XauthDatabaseCreationFailure:
+                self.tools.logger.warning(
+                    """\
+An X11 authentication database could not be created for the display.
+
+Briefcase will proceed but if access to the display is rejected, this may be why.
```
freakboy3742, line 532:
```suggestion
Briefcase will proceed, but if access to the display is rejected, this may be why.
```

src/briefcase/integrations/docker.py
```diff
@@ -378,9 +385,406 @@ def check_output(self, args: list[SubprocessArgT], image_tag: str) -> str:
...
+        network interface as a device on a virtual network interface bridge created by
+        the VM. Because of this, the proxy must be exposed to the network at large to
+        also be exposed within the container.
+
+        :param DISPLAY: value of DISPLAY environment variable
+        :returns: the Popen process for the proxy and the display number it is using
+        """
+        if not self.tools.shutil.which("socat"):
+            raise BriefcaseCommandError(
+                "Install socat to run an app for a targeted Linux distribution"
```
freakboy3742, line 597: It would be helpful to have distro-specific help here - I was able to guess that `apt install socat` worked on Ubuntu, but that won't necessarily be obvious to everyone, and I don't know if the package name will be the same on other host distros.

src/briefcase/integrations/docker.py
```diff
@@ -378,9 +385,406 @@ def check_output(self, args: list[SubprocessArgT], image_tag: str) -> str:
...
+    ) -> None:
+        """Writes a xauth file using the current user's X authorization.
+
+        :param DISPLAY: Value of DISPLAY environment variable
+        :param xauth_file_path: File path to write xauth database for target display
+        :param target_display_num: Number of targeted X display
+        """
+        if not self.tools.shutil.which("xauth"):
+            raise BriefcaseCommandError(
+                "Install xauth to run an app for a targeted Linux distribution"
```
freakboy3742, line 723: As with socat - would be helpful to have an installation hint here.

freakboy3742 (changes requested):
A couple of tweaks to the internal docs, and two related comments related to the socat/xauth error message - but otherwise, this is looking really solid. 

The socat/xauth comments came as a result of being able to actually test this on a real live Ubuntu machine. It worked perfectly. socat wasn't installed, and it returned the expected error message; once it was installed, the app started as expected. 

However - when I got the error, I wasn't 100% certain what needed to be installed. It wasn't too hard to guess `apt install socat`, but that might not be obvious to all, and the equivalent might not be correct on Fedora/Arch etc.

src/briefcase/integrations/docker.py
rmartin16: I'm planning to revisit this with the resolution of https://github.com/beeware/briefcase/issues/1626.

That whole refactor will avoid logging this error so late in the process and will provide an error message relevant to the host system. I just didn't want to do all that in this PR.

You all right if we wait? I'm planning to address it after this is merged.

freakboy3742: I had the same back-of-my head thought. If #1626 is the next thing on your todo list, I'm fine leaving this a future work.

tests/platforms/linux/system/test_run.py
```diff
@@ -16,51 +18,49 @@
...
-    command._stream_app_logs = mock.MagicMock()
-
-    return command
+    # Provide Docker
+    monkeypatch.setattr(
+        Docker, "_is_user_mapping_enabled", mock.MagicMock(return_value=True)
+    )
+    command.tools.docker = Docker(tools=command.tools)
 
+    # Make X11 passthrough a no-op  TODO:PR: should probably change this before im done
```
rmartin16, line 43: Just realized I forgot about this - I'll update these tests

docs/how-to/internal/x11passthrough.rst
rmartin16: Defined in the X11 standard.

https://www.x.org/releases/X11R7.7/doc/man/man1/Xorg.1.xhtml#heading5

docs/how-to/internal/x11passthrough.rst
rmartin16: It's the term the specs for `xauth` use.

https://gitlab.freedesktop.org/xorg/lib/libxau

> The Family is specified in two bytes to allow out-of-band values
(i.e. values not in the Protocol) to be used.  In particular,
two new values "FamilyLocal" and "FamilyWild" are defined.  FamilyLocal
refers to any connections using a non-network method of connection from the
local machine (Unix domain sockets, shared memory, loopback serial line).
In this case the host address is specified by the data returned from
gethostname() and better be unique in a collection of machines
which share NFS directories.  FamilyWild is currently used only
by xdm to communicate authorization data to the server.  It matches
any family/host address pair.

docs/how-to/internal/x11passthrough.rst
rmartin16: hmmm...."X Windows" sounds like msft has something to do with this? What about "X Window Manager"?

freakboy3742 (approved):
A couple of minor markup and wording tweaks; but otherwise this looks good to go!

Nice work - this is a complex piece of plumbing, but the end result is really elegant.

PR #1604, opened 2024-01-14
Ensure log does not contain ANSI sequences or control codes

rmartin16:
## Changes
- ANSI escape sequences and console control codes are stripped from text before being written to the log

## Notes
- This was originally discussed as part of https://github.com/beeware/briefcase/pull/1572
  - Relatedly, I think that PR just needs a little direction to keep going; that is, should the regexes accommodate the ANSI sequences? Or could they be stripped for the purposes of Android log filtering?

## PR Checklist:

### DIFF
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -412,8 +412,7 @@ class Log:
         # build log header and export buffered log from Rich
         uname = platform.uname()
         sanitized_env_vars = "\n".join(
-            f"\t{env_var}="
-            f"{sanitize_text(value) if not SENSITIVE_SETTING_RE.search(env_var) else '********************'}"
+            f"\t{env_var}={value if not SENSITIVE_SETTING_RE.search(env_var) else '********************'}"
             for env_var, value in sorted(command.tools.os.environ.items())
         )
         return (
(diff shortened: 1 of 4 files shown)


src/briefcase/console.py
```diff
@@ -399,7 +412,8 @@ def _build_log(self, command):
         # build log header and export buffered log from Rich
         uname = platform.uname()
         sanitized_env_vars = "\n".join(
-            f"    {env_var}={value if not SENSITIVE_SETTING_RE.search(env_var) else '********************'}"
+            f"\t{env_var}="
+            f"{sanitize_text(value) if not SENSITIVE_SETTING_RE.search(env_var) else '********************'}"
```
freakboy3742, line 416: Is this advisable? Environment variables are something that we really need to know verbatim - any manipulation could lead us to believe that one value has been set, when it's actually a different value. 

To be clear - I can't think of a reason why the user's PS1 setting could cause a problem, but then, that's likely a lack of imagination on my part.

freakboy3742 (changes requested):
+1 to the broad idea; one minor quibble inline.

src/briefcase/console.py
rmartin16: Yeah; that's fair....and it's really of marginal value to remove these anyway.

freakboy3742 (approved):
Looks good - and I'm going to need the cleansing utility to get #1606 working, so this is really timely :-)

Issue #485, opened 2020-09-09
Add ability for libraries to declare build-time requirements


freakboy3742:
Although Briefcase is aimed at "pure python" packaging, sometimes apps have requirements on platform native libraries. 

These may need to be installed as operating system packages (e.g., Linux having a dependency on GTK, Cairo and WebKit), or included in a build process (Android having a dependency on some Jetpack libraries). Although they're not currently needed, It's easy to imagine that similar requirements could exist on other platforms (e.g., a dependency on a CocoaPod for an iOS build).

The mechanism for including these requirements will be different on every platform, as will the format the dependency specification will take. 

However, the dependencies *are* to specific packages. If you need to package a Toga-GTK app, you will have a dependency on GTK. That dependency does not exist if your app is a Qt app. 

It should therefore be possible for a *library* to define the system libraries that are necessary (within a Briefcase context) for a build to complete successfully.

**Describe the solution you'd like**

It should be possible for a Python package with build-time dependencies to declare how to obtain those dependencies. This could possible be attached to package metadata in a custom field. Briefcase could then inspect those requirements and inject them as required into the build process.

**Describe alternatives you've considered**

Briefcase already includes `system_requires` to support requirements that must exist for the Linux build to complete. Analogous features could be added to support system requirements on other platforms. This level of coordination will always be required; However, `pyproject.toml`-level configuration means that any dependencies of your libraries must be manually (and correctly) added to your `pyproject.toml`.

freakboy3742:
I'm going to close this ticket, as what it describes is essentially the scope of [PEP 725](https://peps.python.org/pep-0725/).

In the meantime, as of #2592 (and Briefcase 0.4.0), any `system-requires` are validated as part of running `briefcase dev`, which is as close as we can get to honouring this requirement without a solution for PEP 725.

PR #1610, opened 2024-01-17
Allow Android apps to specify their own list of libraries

freakboy3742:
beeware/toga#2346 has raised the prospect that we might need to add an extra library to the list in the app template. Rather than make a special case of that one library for Toga, this PR adds a general feature, and includes the extra library as part of the Toga bootstrap.

Any existing project will inherit the list of requirements that was present in the Briefcase 0.3.16 Android template.

Requires the use of beeware/briefcase-android-gradle-template#82

Refs #485.

## PR Checklist:


freakboy3742:
It just occurred to me that `dependencies` clashes with the PEP621 name, so we probably want to pick a different name for Android configuration. 

Possible options:
* `android_dependencies`
* `gradle_dependences`
* `system_runtime_requires` (for symmetry with Linux?)

src/briefcase/platforms/android/gradle.py
```diff
@@ -170,6 +170,26 @@ def output_format_template_context(self, app: AppConfig):
             build = int(getattr(app, "build", "0"))
             version_code = f"{v[0]:d}{v[1]:02d}{v[2]:02d}{build:02d}".lstrip("0")
 
+        # The default runtime libraries included in an app. The default value is the
+        # list that was hard-coded in the Briefcase 0.3.16 Android template, prior to
+        # the introduction of customizable system requirements for Android.
+        try:
+            dependencies = app.system_runtime_requires
```
mhsmith, line 177: Shouldn't this be `app.dependencies`? Maybe this is why you couldn't import the navigation bar class.

src/briefcase/platforms/android/gradle.py
```diff
@@ -180,6 +200,7 @@ def output_format_template_context(self, app: AppConfig):
                 for path in (app.test_sources or [])
                 if (name := Path(path).name)
             ),
+            "dependencies": {"implementation": dependencies},
```
mhsmith, line 203: Since we won't be using any of the other dependency types in the foreseeable future, adding an extra `implementation` layer seems unnecessary, and loses the simplicity of having the cookiecutter context match directly with the pyproject.toml structure.

src/briefcase/platforms/android/gradle.py
```diff
@@ -170,6 +170,26 @@ def output_format_template_context(self, app: AppConfig):
...
+        # list that was hard-coded in the Briefcase 0.3.16 Android template, prior to
+        # the introduction of customizable system requirements for Android.
+        try:
+            dependencies = app.system_runtime_requires
+        except AttributeError:
+            self.logger.warning(
+                (
+                    "The Android configuration for this app does not contain "
+                    "a `dependencies` definition. Using a default value; you should "
+                    "add an explicit list of Android dependencies for the project."
```
mhsmith, line 183: Most users will have no clue how to do that, so the message should link to the documentation page with the default list.

src/briefcase/platforms/android/gradle.py
mhsmith: > It just occurred to me that `dependencies` clashes with the PEP621 name

`system_runtime_requires` seems like a good replacement, since it's conceptually doing the same kind of thing.

src/briefcase/platforms/android/gradle.py
freakboy3742: Unfortunately, it is needed. Cookiecutter doesn't let you pass a list as a context value - if you do, it's interpreted as a choice, and it gives you the "default" (the first item). [Passing a dictionary with a single key](https://github.com/cookiecutter/cookiecutter/issues/479#issuecomment-842224554) is how you work around this; in this case, it just happens that there is a convenient name available for that key, even if we're not going to use any of the alternative key values.

src/briefcase/platforms/android/gradle.py
freakboy3742: I wasn't using this branch when I was getting the import failure; I was manually modifying the build.gradle after calling create.

But yes - this should be `app.dependencies` (or whatever name we land on).

I agree that this is conceptually similar to `system_runtime_requires`; however, all the variables we've added recently map as closely as possible to the "native platform naming". The benefit here is that we don't need to answer questions about how a Briefcase name maps to a platform specific name (e.g., explaining that "`system_runtime_requires` maps to `build.gradle` `dependencies`) - the naming of the Android-specific customizations is at least *slightly* intuitive to Android users. My inclination is to go with `gradle_dependencies` on that basis.

freakboy3742: Scratch that - for consistency with existing Android options, it should be `build_gradle_dependencies`.

mhsmith: Yes, that makes sense.

docs/reference/platforms/android/gradle.rst
```diff
@@ -156,6 +157,22 @@ the ``AndroidManifest.xml`` of your app.
 Additional content that will be added verbatim just before the closing ``</activity>``
 declaration of the ``AndroidManifest.xml`` of your app.
 
+``base_theme``
+--------------
+
+The base theme for the application. Defaults to ``Theme.AppCompat.Light.DarkActionBar``
+
+``build_gradle_dependencies``
+-----------------------
```
mhsmith, line 166: RTD is currently falling here with the error "Title underline too short".

tests/commands/new/test_build_context.py
```diff
@@ -1235,6 +1243,14 @@ def main():
 requires = [
     "toga-android~=0.4.0",
 ]
+
+base_theme = "Theme.MaterialComponents.Light.DarkActionBar"
+
+build_gradle_dependencies = [
+    "com.google.android.material:material:1.11.0",
```
freakboy3742, line 1250: The dependency list gets a lot simpler under material, because appcompat and constraintlayout are both dependencies of material.

mhsmith (approved):

Issue #1573, opened 2023-12-13
Android emulator does not open until CTRL+C is pressed during "Starting emulator..."


rmartin16:
### Describe the bug

Several users in Discord have reported `briefcase run android` results in Briefcase getting stuck at the Wait Bar for `Starting emulator...`. When CTRL+C is pressed, Briefcase exits and the emulator opens.

This issue can be worked around by leaving the emulator open and running `briefcase run android` again and then choosing the running emulator in the list

### Steps to reproduce

Not sure yet...

### Expected behavior

Emulator opens while Briefcase is running.

### Environment

- Operating System: all reports have been Windows 10 22H2 so far
- Python version: 3.10, 3.11
- Software versions:
  - Briefcase: 0.3.16

### Logs

[briefcase.2023_12_06-23_21_25.run(1).log](https://github.com/beeware/briefcase/files/13666570/briefcase.2023_12_06-23_21_25.run.1.log)
[briefcase.2023_12_13-21_53_23.run(1).log](https://github.com/beeware/briefcase/files/13666572/briefcase.2023_12_13-21_53_23.run.1.log)
[briefcase.2023_12_04-23_10_36.run.log](https://github.com/beeware/briefcase/files/13666574/briefcase.2023_12_04-23_10_36.run.log)

JaxkDev:
Can add Python 3.8 + 23H2 to the list, seems to replicate 100% of the time...

f-bdolan:
Additionally, when it does start, my application is not installed

rmartin16:
> Additionally, when it does start, my application is not installed

@f-bdolan: correct; when you press CTRL+C and the emulator opens, your app will not have been installed. However, if you run `briefcase run android` again and select the "running emulator", Briefcase will install your app then.

f-bdolan:
Ah! Excellent, so there is a work-around.  For those of you just showing up:

If `briefcase run android` stalls on "starting":
1. Run `briefcase run android`
2. Pick an emulator (if you use beePhone it will have debugging options turned on)
3. Wait a moment to stall
4. Hit Ctrl-C
5. Leave the emulator open
6. Run `briefcase run android` again and choose the SAME emulator
7. It will install

Issue #1036, opened 2023-01-05
Android emulator acceleration messages may lead the user to install an incompatible version of HAXM


MarkusPiotrowski:
### Describe the bug

The Android emulator fails with the error messages `Failed to sync vcpu reg` when the latest version of Intel's Hardware Accelerated Execution Manager, version 7.8.0, is installed. Installing version 7.7.1 makes it running.

### Steps to reproduce

1. Download version 7.8.0 of HAXM here: https://github.com/intel/haxm/releases/
2. Install on a computer with an Intel processor capable of hardware acceleration (vt, vt-x)
3. Run the Android emulator in a beeware project

### Expected behavior

This report is for documentation only. I would expect that the Android emulator which is installed will work with the latest HAXM version. However, it's not even clear to me if this probably is a specific problem of my computer/setup. 

### Environment

- Operating System: Windows 10
- Python version: 3.8.10
- Software versions:
  - Briefcase: 0.3.11
  - Toga: 0.3.0.dev39
  - Android Emulator:
Pkg.UserSrc=false
Pkg.Revision=31.3.14
Pkg.Path=emulator
Pkg.Desc=Android Emulator
Pkg.BuildId=9322596
  
- Processor Intel Core i7-1165G7

### Logs

```
[xxxx] Starting emulator beePhone...
Starting emulator...
INFO    | Android emulator version 31.3.14.0 (build_id 9322596) (CL:N/A)
emulator: INFO: Found systemPath C:\Users\xxxx\AppData\Local\BeeWare\briefcase\Cache\tools\android_sdk\system-images\android-31\default\x86_64\
WARNING | unexpected system image feature string, emulator might not function correctly, please try updating the emulator.
INFO    | Duplicate loglines will be removed, if you wish to see each indiviudal line launch with the -log-nofilter flag.
INFO    | added library vulkan-1.dll
INFO    | configAndStartRenderer: setting vsync to 60 hz
[...]

PR #1630, opened 2024-02-03
Improve Android emulator startup failure messages

rmartin16:
## Changes
- Updates handling and management of information presented to the user when the Android emulator fails to start
- Related issues:
  - #1036
  - #1573

## UX
- Conceptually, I broke the errors in to two categories:
  - Briefcase detects the emulator closed
  - The user aborts the start up process
- In the first case, the emulator's output probably includes the recommended steps for the user.
- In the second case, Briefcase probably won't be able to retrieve the emulator's output because the emulator will likely still be running....and retrieving `stdout` from a running process without blocking is non-trivial....but potentially really valuable here

Presentation when Briefcase detects startup failure:
image

Presentation when the User aborts the startup:
image

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1378,13 +1378,16 @@ You can also start the emulator manually by running:
                         # Try again in 2 seconds...
                         self.sleep(2)
         except BaseException as e:
-            self.tools.logger.info("Emulator output log", prefix=self.name)
+            self.tools.logger.warning(
+                "Emulator output log for startup failure",
+                prefix=self.name,
+            )
             try:
                 # if the emulator exited, this should return its output immediately
                 self.tools.logger.info(emulator_popen.communicate(timeout=1)[0])
             except subprocess.TimeoutExpired:
                 self.tools.logger.info(
-                    "> Briefcase failed to retrieve emulator output "
+                    "Briefcase failed to retrieve emulator output "
                     "(this is expected if the emulator is running)"
                 )
 
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -8,7 +8,6 @@ import shutil
 import subprocess
 import sys
 import time
-from contextlib import suppress
 from datetime import datetime
 from pathlib import Path
 
@@ -1306,20 +1305,25 @@ In future, you can specify this device by running:
             for arg in map(str, emulator_popen.args)
         )
 
-        error_msg = (
-            "{prologue}"
-            + f"""
+        general_error_msg = f"""
+Review the emulator output above for:
+ - Troubleshooting or resolution steps such as enabling hardware acceleration
+ - Other errors or warnings that may be suggesting the cause of the startup failure
 
-Try starting the emulator manually by running:
+Ensure your Android SDK is up-to-date by running:
+
+    $ briefcase update {AndroidSDK.name}
+
+To review Google's general troubleshooting steps for the emulator, visit:
+
+    https://developer.android.com/studio/run/emulator-troubleshooting
+
+You can also start the emulator manually by running:
 
     $ {emulator_command}
-
-Resolve any problems you discover, then try running your app again. You may
-find this page helpful in diagnosing emulator problems.
-
-    https://developer.android.com/studio/run/emulator-acceleration#accel-vm
 """
-        )
+
+        failed_startup_error_msg = f"{{prologue}}\n{general_error_msg}"
 
         # The boot process happens in 2 phases.
         # First, the emulator appears in the device list. However, it's
@@ -1335,7 +1339,7 @@ find this page helpful in diagnosing emulator problems.
                 while adb is None:
                     if emulator_popen.poll() is not None:
                         raise BriefcaseCommandError(
-                            error_msg.format(
+                            failed_startup_error_msg.format(
                                 prologue="Android emulator was unable to start!"
                             )
                         )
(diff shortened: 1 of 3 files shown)


rmartin16:
As it relates to #1573, these changes may make it less likely for users to report when the issue occurs. However, such reports aren't really producing any valuable information anyway right now.

Given this, it'd be nice if we could add some instrumentation to include potentially valuable information about the emulator when it occurs.

The only idea I have for that is trying to retrieve any output to `stdout` from the emulator since it started. This is tricky but is possible with the output thread streamer using something like `stop_func=lambda: time.time() > [1 sec in future]`. That said, I don't have any reason to believe the output will contain anything useful....especially if the process is simply blocked on something after starting.

From the provided logs from users, we can see the emulator is in an `offline` state....I don't know if that's useful but it's about all we have to do on.

Open to ideas...

src/briefcase/integrations/android_sdk.py
```diff
@@ -1366,27 +1370,37 @@ def start_emulator(
...
-            # if the emulator exited, this should return its output immediately;
-            # if it is still running, this will quickly time out and print nothing.
-            with suppress(subprocess.TimeoutExpired):
+            self.tools.logger.info("Emulator output log", prefix=self.name)
+            try:
+                # if the emulator exited, this should return its output immediately
                 self.tools.logger.info(emulator_popen.communicate(timeout=1)[0])
+            except subprocess.TimeoutExpired:
+                self.tools.logger.info(
+                    "> Briefcase failed to retrieve emulator output "
```
freakboy3742, line 1387: Is this chevron intentional? If it's intentional, I'm not sure what it's communicating.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1366,27 +1370,37 @@ def start_emulator(
...
                                 )
                             )
 
                         # Try again in 2 seconds...
                         self.sleep(2)
         except BaseException as e:
-            # if the emulator exited, this should return its output immediately;
-            # if it is still running, this will quickly time out and print nothing.
-            with suppress(subprocess.TimeoutExpired):
+            self.tools.logger.info("Emulator output log", prefix=self.name)
```
freakboy3742, line 1381: I'd be inclined to bump this to warning or error. As currently rendered, it looks a bit like "normal output"; but if we're at this point, we're already in "something is in the process of breaking" mode.

It might also be worth highlighting that it's not just the output log - it's "output because of error".

freakboy3742 (changes requested):
The broad direction of this looks like a good improvement - anything that makes it easier to surface what might be going on in this case will be a good improvement. A couple of cosmetic suggestions inline; but I don't have any additional ideas on other ways to improve the error reporting. 

What's really frustrating is that the people who are experiencing this problem are the least likely to be able to help us diagnose it... but I'm willing to bet that as soon as we find someone who (a) experiences this problem and (b) has enough skills to know how to generate a useful log message, we'll be able to identify the problem and a fix really quickly.

The test failure is really weird... it reads like the specific machine that is being assigned for the job is lacking disk or memory space... but I would have though re-running the job would get you a new machine in the pool...

src/briefcase/integrations/android_sdk.py
rmartin16: I went through a few iterations of what to print in this situation because initially, at least, I concluded that whatever is printed here could be construed as "the emulator failed to retrieve the logs". So, then I experimented with some bracket markup and then eventually just added "Briefcase" as the explicit actor that was trying to retrieve the logs.

At any rate, I can remove it.

src/briefcase/integrations/android_sdk.py
rmartin16: Yeah; I think I'm hesitant to make it red error output because that will automatically become the first thing the user looks at and perceived as the most important....and it might be. Mostly, I wanted to direct the user to where I wanted them to start reading...at least in the case of CTRL+C. For the other failure case, it's just a wall of red text.

Anyway, I think changing the emulator output heading to a yellow warning and saying `Emulator Output Log for Startup Failure` instead.

freakboy3742 (commented):
I'm happy with this from a functionality point of view. The CI failure is the only thing standing in the way here. 

I'm wondering if this might be a configuration change that has been introduced in a recent update to the Ubuntu system image... if the user's home directory has a restriction on the amount of free space, that might explain what we're seeing here.

If you want to try and debug this, Toga has a (commented out) mechanism for getting a shell session on the CI machine.

rmartin16:
It's something with the runner [image](https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20240201.1) released on the 1st.

I re-ran the most recent successful CI; it used the latest runner image and failed as well.

https://github.com/beeware/briefcase/actions/runs/7763463029/job/21192659116

I may try to dig in later. I haven't found anything saying the available disk space was changed....intentionally anyway.

I suppose it's also possible the emulator is requiring more space for the qemu disks now...

freakboy3742:
Interestingly, the problem *doesn't* appear to be affecting Toga. [Fresh PRs are passing Android CI without any problem](https://github.com/beeware/toga/actions/runs/7771441929/job/21192786775?pr=2374)

rmartin16:
Unfortunately, re-running a previous run of CI uses the same workflows it ran the first time. If a workflow is updated (as we did in beeware/.github#88), you need a new commit; I'll push an update.

freakboy3742 (approved):
All looks good in CI, so 🚢 !

PR #1641, opened 2024-02-11
Update android.rst to include explanation on packaging apks

Aarontheissueguy:
While abb's are the primary distribution method of android apps. APK's are still needed to distritbute packages that are not hosted in the play store

## PR Checklist:


mhsmith:
Rather than duplicating the documentation of this option, which already exists in [reference/platforms/android/gradle.rst](https://briefcase.readthedocs.io/en/stable/reference/platforms/android/gradle.html), it would be better just to link to that page.

If you like, it would also be a good improvement to move the option's documentation into the "Additional options" section further down the page, in the same format as the Android-specific `briefcase run` options.

Finally, to get this PR passing the automated checks, please follow the "Details" links next to the red crosses below.

freakboy3742:
Closing due to lack of response from the original submitter.

PR #1652, opened 2024-02-17
Add arm64 macOS testing to CI

rmartin16:
## Changes
- Adds macOS arm64 runners to CI
- To limit the overall number of jobs, Python 3.9, 3.10, and 3.11 are only tested on Ubuntu

## PR Checklist:


rmartin16:
I also added arm64 macOS to the integration tests. With the couple of times I ran the full set of jobs, it is clear that GitHub is limiting the number of concurrent runners a repo/org can use beyond what they limit for the Linux runners. That said, the Linux jobs still take long enough that even the macOS jobs that have to wait end up finishing before or around the same time.

It's also interesting how long the unit tests run for 3.13-dev on Windows...they must be using a debug build of Python or something.

P.S. as a mental note....the biggest opportunity for speeding up the integration tests on Linux is the Docker builds. First, since the linux package from Briefcase is installed, it may be possible to alternatively just ask the installer to do a dry-run and not actually install everything. Second, caching the Briefcase images in GHCR.

.github/workflows/ci.yml
```diff
@@ -78,9 +91,11 @@ jobs:
...
     - name: Test
       id: test
-      env:
-        COVERAGE_FILE: ".coverage.${{ matrix.platform }}.${{ matrix.python-version }}"
-      run: tox -e py --installpkg dist/briefcase-*.whl
+      run: |
+        RUNNER_OS=$(cut -d- -f1 <<< ${{ matrix.platform }})
+        RUNNER_VERSION=$(cut -d- -f2 <<< ${{ matrix.platform }})
+        COVERAGE_FILE=".coverage.${RUNNER_OS}.${RUNNER_VERSION}.${{ matrix.python-version }}" \
+          tox -e py --installpkg dist/briefcase-*.whl
```
rmartin16, line 98: This nonsense is necessary because `coverage` doesn't just do a substring match on the coverage file names; it adds a period to the end of the substring match...so, the files literally need to start with `.coverage.ubuntu.`, for instance.

.github/workflows/ci.yml
```diff
@@ -40,18 +40,31 @@ jobs:
...
     needs: [ pre-commit, towncrier, package ]
-    runs-on: ${{ matrix.platform }}-latest
+    runs-on: ${{ matrix.platform }}
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
       matrix:
-        platform: [ "macos", "ubuntu", "windows" ]
-        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12", "3.13-dev" ]
+        platform: [ macos-12, macos-14, windows-latest, ubuntu-latest ]
```
freakboy3742, line 48: I'm not sure removing the quotes here (and below) is a good idea. YAML has an... interesting relationship with bare values; I've found it safer to unilaterally quote when you know the value is a string, even if it isn't *strictly* necessary (and it clearly isn't here, because CI is passing).

freakboy3742 (changes requested):
The mechanics of this all look good. One comment about value quoting inline; the other detail is that the [platform support chart](https://briefcase.readthedocs.io/en/stable/reference/platforms/index.html) can upgrade the support level for ARM64.

.github/workflows/ci.yml
rmartin16: This is definitely the thing I hate most about YAML....other than everything else. To add to this, I've also wondered how the GitHub substitutions are handled, i.e. `${{ xxx }}`, and whether any of these substitutions happen before a YAML processor runs...

freakboy3742 (approved):
👍

PR #1661, opened 2024-02-21
Add support to pass arguments through to Docker image build

rmartin16:
## Changes
- Allow passing arguments to `docker build`
- RE: https://github.com/beeware/.github/pull/99

## PR Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,9 +62,9 @@ dependencies = [
     # Dependencies required at runtime are set as ranges to ensure maximum
     # compatibility with the end-user's development environment.
     #
-    # Any dependency that is part of the "core python toolchain" (e.g. pip,
-    # wheel) specify a minimum version, but no maximum, because we always want
-    # the most recent version.
+    # Any dependency that is part of the "core python toolchain" (e.g. pip, wheel)
+    # specify a minimum version, but no maximum, because we always want the most
+    # recent version.
     #
     # limited to <=3.9 for the `group` argument for `entry_points()`
     "importlib_metadata >= 4.4; python_version <= '3.9'",
@@ -73,13 +73,15 @@ dependencies = [
     "setuptools >= 60",
     "wheel >= 0.37",
     "build >= 0.10",
-    # For the remaining packages: We set the lower version limit to the lowest
-    # possible version that satisfies our API needs. If the package uses semver,
-    # we set a limit to prevent the next major version (which will potentially
-    # cause API breakages). If the package uses calver, we don't pin the upper
-    # version, as the upper version provides no basis for determining API
-    # stability.
-    "cookiecutter >= 2.3.1, < 3.0",
+    #
+    # For the remaining packages: We set the lower version limit to the lowest possible
+    # version that satisfies our API needs. If the package uses semver, we set a limit
+    # to prevent the next major version (which will potentially cause API breakages).
+    # If the package uses calver, we don't pin the upper version, as the upper version
+    # provides no basis for determining API stability.
+    #
+    # See beeware/briefcase#1663 to bump <2.6.0 pin
+    "cookiecutter >= 2.3.1, < 2.6.0",
     "dmgbuild >= 1.6, < 2.0; sys_platform == 'darwin'",
     "GitPython >= 3.1, < 4.0",
     "platformdirs >= 2.6, < 5.0",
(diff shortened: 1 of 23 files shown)


src/briefcase/integrations/docker.py
```diff
@@ -904,7 +907,9 @@ def prepare(
                     self.tools.subprocess.run(
                         [
                             "docker",
+                            "buildx",
                             "build",
+                            "--load",
                             "--progress",
```
rmartin16, line 913: Added `buildx` so it is explicit that `buildx` is building the image. Added `--load` because it is only implied when using the default build container; if you use a custom one, the image isn't loaded in to the local Docker image cache.

pyproject.toml
```diff
@@ -79,7 +79,7 @@ dependencies = [
     # cause API breakages). If the package uses calver, we don't pin the upper
     # version, as the upper version provides no basis for determining API
     # stability.
-    "cookiecutter >= 2.3.1, < 3.0",
+    "cookiecutter >= 2.6.0, < 3.0",
```
freakboy3742, line 82: No objection to doing this bump if it's needed, but I'm interested what bug/feature has required it?

rmartin16: Using pre-2.6.0 and 2.6.0 produces two different files for `app.py` and `__main__.py`. I didn't track down the exact thing that changed but 2.6.0 introduced a way to control the jinja tag boundaries....so, thinking about it now, maybe a regression was introduced around how `{% %}` is replaced in a template....I should probably check how our other templates look now.

At any rate, CI was failing with 2.6.0: https://github.com/beeware/briefcase/actions/runs/7993409808/job/21829525533

rmartin16: We should hold off. There's all kinds on new newlines in `pyproject.toml` with 2.6.0.

rmartin16: well, actually, i guess it doesn't matter....2.6.0 is going to be installed regardless...so, I guess we may as well move forward and resolve this regression independently.

rmartin16: https://github.com/cookiecutter/cookiecutter/issues/2023

rmartin16: To explicitly lay it out, I think these are the options I see to move forward:
- Bump to require 2.6.0 as I have here
  - This would accommodate the whitespace changes in 2.6.0
  - Once/if these are resolved in `cookiecutter`, we can revert the changes and bump to that new version
- Lower the upper bound from `<3.0` to `<2.6.0`
  - This will obviously maintain the status quo
  - Once/if the issue is resolved, we can increase the upper bound again

Let me know what you think. As a note, CI for all PRs for Briefcase will fail until this is addressed.

rmartin16:
Another question that came up for me while I was implementing this:

Do you see value in keeping `LinuxSystemPassiveMixin` and `LinuxSystemMostlyPassiveMixin` as separate classes still given that `LinuxSystemRunCommand` now depends on the build system?

Some of the separation is already a little fuzzy to me....but I was also considering collapsing some of these forks in the code for whether the build system is the local machine or in Docker and just using the same logic for both via calls to `tools[app].app_context.run()`.

freakboy3742:
> Do you see value in keeping `LinuxSystemPassiveMixin` and `LinuxSystemMostlyPassiveMixin` as separate classes still given that `LinuxSystemRunCommand` now depends on the build system?

You're not alone in having this thought. I can't remember which PR I was working on where this case up, but I noticed the same thing. It was an optimisation in the past, but I'm fairly certain the use case for "mostly passive" no longer exists. I didn't get quite as far as doing the refactor, but if we *can* get rid of the "Mostly", I'd be happy to sign off on that.

pyproject.toml
freakboy3742: I guess a third option is "!=2.6.0", if we have reason to believe that 2.6.0 is an isolated buggy release and the issue will be fixed in 2.6.1.

My inclination for now is to use "< 2.6.0" until we know whether this is an intentional change (or, at least, a change that won't be rolled back), or a bug that will be resolved. If this is the "new normal", then we should set the 2.6.0 as the new minimum; if it's a bug that gets fixed, we can exclude the specific problem releases.

rmartin16: Good deal; I agree...especially since the issue is more widespread than I initially thought.
- https://github.com/beeware/briefcase/issues/1663

rmartin16:
CI looks good now 🦾 once this is merged, I will re-run CI for https://github.com/beeware/.github/pull/99 and re-base https://github.com/beeware/briefcase/pull/1649; both of those should then be ready to review/merge.

freakboy3742 (approved):
👍

PR #1662, opened 2024-02-22
Soft Ruff implementation

rmartin16:
## Changes
- Replace pyupgrade, isort, and flake8 with Ruff's linter

## Notes
- Another pitch to use Ruff.
- Other than 🚀, the benefit I see is using a single tool instead of several...and a seemingly well-received tool in the Python community.
- I'm considering this a soft implementation (right now anyway) because I didn't replace Black with Ruff's formatter.
  - This is primarily because I couldn't get Ruff to play nicely with a quirk we're taking advantage of.
    - Currently, Black ignores many lines that exceed its ~88 character limit....but we're currently using flake8 to catch excessively long lines over 119 characters
    - I haven't figured out how to get Ruff to do the same without adding a bunch of `noqa` or ignoring really long lines altogether
  - I also didn't want to spend too much more time with this if the team isn't interested right now.

## PR Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,15 +8,6 @@ repos:
       - id: check-docstring-first
       - id: end-of-file-fixer
       - id: trailing-whitespace
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v3.15.1
-    hooks:
-      - id: pyupgrade
-        args: [--py38-plus]
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.13.2
-    hooks:
-      - id: isort
   - repo: https://github.com/PyCQA/docformatter
     rev: v1.7.5
     hooks:
@@ -27,10 +18,11 @@ repos:
     hooks:
       - id: black
         language_version: python3
-  - repo: https://github.com/PyCQA/flake8
-    rev: 7.0.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.2.2
     hooks:
-      - id: flake8
+      - id: ruff
+        args: [--fix]
   - repo: https://github.com/codespell-project/codespell
     rev: v2.2.6
     hooks:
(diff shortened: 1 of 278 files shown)


freakboy3742:
> * Other than 🚀, the benefit I see is using a single tool instead of several...and a seemingly well-received tool in the Python community.
> * I'm considering this a soft implementation (right now anyway) because I didn't replace Black with Ruff's formatter.

Ok - so... count me as one of the people who's on the "not receiving it well" camp :-)

I have a lot of concerns about replacing community maintained, in-language tooling with VC-backed, out-of-language tooling. Ruff hasn't been around long enough for me to feel comfortable that they're a project that is going to exist after the VC money has run out; and if/when that happens, the oxygen will have gone out of the tools that Ruff replaces - the maintainer of flake8 has already mentioned that he's feeling demotivated that paid developers have taken the community's work and built on it. 

It's also going to be harder to *find* community contributors, because someone who knows and uses Python can't use those skills to debug and contribute to Rust. The skills are learnable, to be sure - and I feel a lot better about a tool being written in Rust than I would about the same tool in (say) C; but Rust is undeniably harder, so going to be a barrier.

Using a single tool is nice I guess; but the real benefit that keeps being pushed is "faster" - to which... I honestly don't care. I can't say that the speed of the *linting* step is the thing that I've historically been most concerned about. If the 10 second linting pass when you're about to commit is the thing slowing down your contribution cycle... I have questions about your development process. :-)

This isn't a hard No - but it's definitely a "I need to be convinced".

rmartin16:
> This isn't a hard No - but it's definitely a "I need to be convinced".

eh...I probably won't do that...I'm mostly ambivalent.

FWIW, I have similar concerns...but it also seems that sticking with tools which seem destined to languish at this point is concerning. This prognostication for the future of Python tooling is generally a bit disconcerting  😕 the implication kinda seems like the best outcome is astral becomes the long term de facto maintainer of tooling for Python...

As far as it being faster, it's marginal until you get to the huge codebases so yeah...I wish that wasn't the top line selling point either. (The 🚀 is definitely sarcastic 🙃 ....one of my friends even updated his phone to automatically replace "rust" with "rust (🚀)" lol).

So, anyway, I'm definitely fine holding off until the future is hopefully more discernible.

PR #1676, opened 2024-02-29
ADB log tailing color output

rmartin16:
## Changes
- Inspired by https://github.com/beeware/briefcase/pull/1572
- Configures ADB to use color-coordinated output

image

## Notes
- I finally took the time to figure out the regex to account for ANSI sequences in the output.
- The regex only handles ANSI sequences at the beginning and end of the output....but that's the only place I've seen it.

## PR Checklist:

### DIFF
--- a/changes/1676.feature.rst
+++ b/changes/1676.feature.rst
@@ -1,0 +1,1 @@
+When an app runs on an Android device or emulator, the logging output is now colored.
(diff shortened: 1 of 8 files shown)


rmartin16:
In all its glory in CI: https://github.com/beeware/briefcase/actions/runs/8104029040/job/22150011201?pr=1676#step:28:624

One thing I was considering is that there isn't any way to disable this...`adb` doesn't seem to respect `NO_COLOR`, at least. It might be possible to inspect if RIch disabled color...and do the same for `adb logcat`...

freakboy3742 (approved):
It's definitely colourful... :-)

Not much to fault here. I guess we'll learn pretty soon if ANSI sequences appear anywhere in ADB logs other than the start of the line; but we can fix that pretty quickly if we need to.

I'm also not overly concerned about testing how Rich interprets things - at some point, you have to be comfortable that the tools you're using are testing their own behavior, and I think Rich falls in that camp.

PR #1683, opened 2024-03-07
Add CI workflow to bump config file dependencies

rmartin16:
## Changes
- Implements https://github.com/beeware/.github/pull/106

## PR Checklist:

- [x] All new features have been tested
- [X] All new features have been documented
- [X] I have read the **CONTRIBUTING.md** file
- [X] I will abide by the code of conduct

PR #1686, opened 2024-03-07
Pin all deps in tox.ini

rmartin16:
## Changes
- With #1683 live, this pins all dependencies in tox.ini

## Notes
- As it relates to `towncrier`, I confirmed the changes to `docs/background/releases.rst` were the same for this version as the previous.
- Additionally, I noticed that the current use of `towncrier` in `tox` seemingly requires that Briefcase has been installed...otherwise, the version is indeterminable.

## PR Checklist:

### DIFF
--- a/changes/1686.misc.rst
+++ b/changes/1686.misc.rst
@@ -1,0 +1,1 @@
+All dependencies in tox.ini are now pinned to specific versions.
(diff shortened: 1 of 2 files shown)


freakboy3742 (approved):
The towncrier update makes me mildly nervous, because we only ever run towncrier when we're doing a release, so there's no real exercising of this until we actually cut a release. I'm not sure there's a better option, though - testing release notes is going to be near impossible, and tweaking a towncrier config for compatibility as part of a release isn't the end of the world.

PR #1687, opened 2024-03-08
Capture emulator `stdout` during startup

rmartin16:
## Changes
- An attempt to resolve https://github.com/beeware/briefcase/issues/1573
- One user in Discord has reported that reading `stdout` resolves the issue
- Introduces `PopenOutputStreamer` for encapsulating the output stream `Thread` and adds the ability to capture instead of print output
- Adds `Subprocess.stream_output_non_blocking()` alongside the existing `Subprocess.steam_output()` to facilitate callers streaming (and conditionally capturing output) while doing other tasks

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1411,7 +1411,7 @@ run Briefcase again. The running emulator can then be selected from the list.
 
             raise
         finally:
-            emulator_streamer.stop_flag.set()
+            emulator_streamer.request_stop()
 
         # Return the device ID and full name.
         return device, full_name
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1288,19 +1288,25 @@ In future, you can specify this device by running:
         if extra_args is None:
             extra_args = []
 
+        # Start the emulator
         emulator_popen = self.tools.subprocess.Popen(
-            [
-                os.fsdecode(self.emulator_path),
-                f"@{avd}",
-                "-dns-server",
-                "8.8.8.8",
-            ]
-            + extra_args,
+            [self.emulator_path, f"@{avd}", "-dns-server", "8.8.8.8"] + extra_args,
             env=self.env,
             stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
             bufsize=1,
             start_new_session=True,
+        )
+
+        # Start capturing the emulator's output
+        # On Windows, the emulator can block until stdout is read and the emulator will
+        # not actually run until the user sends CTRL+C to Briefcase (#1573). This
+        # avoids that scenario while also ensuring emulator output is always available
+        # to print in the console if other issues occur.
+        emulator_streamer = self.tools.subprocess.stream_output_non_blocking(
+            label="Android emulator",
+            popen_process=emulator_popen,
+            capture_output=True,
         )
 
         # wrap AVD name in quotes since '@' is a special char in PowerShell
@@ -1330,12 +1336,11 @@ You can also start the emulator manually by running:
         failed_startup_error_msg = f"{{prologue}}\n{general_error_msg}"
 
         # The boot process happens in 2 phases.
-        # First, the emulator appears in the device list. However, it's
-        # not ready until the boot process has finished. To determine
-        # the boot status, we need the device ID, and an ADB connection.
+        # First, the emulator appears in the device list. However, it's not ready until
+        # the boot process has finished. To determine the boot status, we need the
+        # device ID, and an ADB connection.
 
-        # Phase 1: Wait for the device to appear so we can get an
-        # ADB instance for the new device.
+        # Phase 1: Wait for the device to appear so we can get an ADB instance for it.
         try:
             with self.tools.input.wait_bar("Starting emulator...") as keep_alive:
                 adb = None
@@ -1388,14 +1393,7 @@ You can also start the emulator manually by running:
                 "Emulator output log for startup failure",
                 prefix=self.name,
             )
-            try:
-                # if the emulator exited, this should return its output immediately
-                self.tools.logger.info(emulator_popen.communicate(timeout=1)[0])
-            except subprocess.TimeoutExpired:
-                self.tools.logger.info(
-                    "Briefcase failed to retrieve emulator output "
-                    "(this is expected if the emulator is running)"
-                )
+            self.tools.logger.info(emulator_streamer.captured_output)
 
             # Provide troubleshooting steps if user gives up on the emulator starting
             if isinstance(e, KeyboardInterrupt):
@@ -1412,6 +1410,8 @@ run Briefcase again. The running emulator can then be selected from the list.
                 self.tools.logger.info(general_error_msg)
 
             raise
+        finally:
+            emulator_streamer.stop_flag.set()
 
         # Return the device ID and full name.
         return device, full_name
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -4,6 +4,7 @@ import contextlib
 import json
 import operator
 import os
+import queue
 import shlex
 import subprocess
 import sys
@@ -56,7 +57,7 @@ def is_process_dead(pid: int) -> bool:
     monitoring of a PID to identify when the process goes from existing to not existing.
 
     :param pid: integer value to be checked if assigned as a PID.
-    :return: True if PID does not exist; False otherwise.
+    :returns: True if PID does not exist; False otherwise.
     """
     return not psutil.pid_exists(pid)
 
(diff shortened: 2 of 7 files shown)


src/briefcase/integrations/android_sdk.py
```diff
@@ -1412,6 +1410,8 @@ def start_emulator(
                 self.tools.logger.info(general_error_msg)
 
             raise
+        finally:
+            emulator_streamer.stop_flag.set()
```
freakboy3742, line 1414: I'd be inclined to wrap this as a `streamer.stop()` method  (or similar), rather than reaching into the internal implementation so if we ever have more complex "kill the streamer" logic, we don't miss anything that is required.

src/briefcase/integrations/subprocess.py
```diff
@@ -653,52 +755,39 @@ def stream_output(
...
+        main thread monitors other activities; alternatively, the output of the process
+        can be captured to be retrieved later in the event of an error, for instance.
 
-        :param popen_process: popen process to stream stdout
-        :param filter_func: a callable that will be invoked on every line
-            of output that is streamed; see ``stream_output`` for details.
+        :param label: A description of the content being streamed; used for to provide
+            context in logging messages.
+        :param popen_process: A running Popen process with output to print
+        :param capture_output:
```
freakboy3742, line 775: Missing a description here

src/briefcase/integrations/subprocess.py
```diff
@@ -145,6 +146,107 @@ def inner(sub: Subprocess, args: SubprocessArgsT, *wrapped_args, **wrapped_kwarg
...
+                    for filtered_line in filtered_output:
+                        if self.capture_output:
+                            self.output_queue.put_nowait(filtered_line)
+                        else:
+                            self.logger.info(filtered_line)
+
+                    if stop_streaming:
+                        self.stop_flag.set()
+
+                if self.stop_flag.is_set():
```
freakboy3742, line 195: Is there a reason for evaluating `not is_set` then `is set` as separate if statements? Naïvely, it feels like this should be an if/else; but I can imagine there might be a complex interaction with `readline()` that I'm missing. If there is, there's a need for a docstring; if there isn't - it's an easy simplification :-)

src/briefcase/integrations/android_sdk.py
```diff
@@ -1412,6 +1410,8 @@ def start_emulator(
                 self.tools.logger.info(general_error_msg)
```
freakboy3742, line 1410: Will this warning message still be required? Or is it worth keeping just in case?

freakboy3742 (changes requested):
Broadly makes sense (I think...); a couple of nitpicks inline, but on the whole this looks like a good refactoring of a complex part of the streaming code.

The only other issue I noticed is a weird one - the code clearly has 100% coverage, but I can't see how the `capture_output = False` case is being exercised. I'm guessing it's being tested implicitly by another test somewhere, but given it's at the core of how this works, it would be worth having an explicit test.

To make sure I'm understanding what is going on here - the key to the problem is that the old Popen usage wasn't ever invoking `readline()`, which apparently causes Windows to lock up and not produce output. The code now *always* uses a background thread to poll and collect stdout (not just for the `stream_output()` case). The Android usage previously only used `communicate()` to gather output in the case of an error, and otherwise ignored output. The new implementation does continuous `readline()` polling to get output as it is generated, which should (hopefully 🤞) unblock Windows. 

Have I got that right?

rmartin16:
> The only other issue I noticed is a weird one - the code clearly has 100% coverage, but I can't see how the `capture_output = False` case is being exercised. I'm guessing it's being tested implicitly by another test somewhere, but given it's at the core of how this works, it would be worth having an explicit test.

Would you consider the `test_output()` test in `tests/integrations/subprocess/test_PopenOutputStreamer.py` to cover this case? Or maybe something more?

> To make sure I'm understanding what is going on here - the key to the problem is that the old Popen usage wasn't ever invoking `readline()`, which apparently causes Windows to lock up and not produce output. The code now _always_ uses a background thread to poll and collect stdout (not just for the `stream_output()` case). The Android usage previously only used `communicate()` to gather output in the case of an error, and otherwise ignored output. The new implementation does continuous `readline()` polling to get output as it is generated, which should (hopefully 🤞) unblock Windows.
> 
> Have I got that right?

Right....the theory is if we read `stdout` from the emulator, it'll stop blocking it from properly opening for users.

More broadly, this provides the option of creating a `Popen` and instead of only being able to call `Subprocess.stream_output()` to read `stdout` while blocking, there's now `Subprocess.stream_output_non_blocking()` to (obviously) not block while reading `stdout`.

src/briefcase/integrations/android_sdk.py
rmartin16: That makes sense; although, I'm inclined to call it `streamer.request_stop()` because `readline()` is blocking and as such, we can't guarantee setting this flag will actually ever cause the thread to exit (....you know, if its stuck in `readline()`).

src/briefcase/integrations/subprocess.py
rmartin16: This is intentional because I didn't want `filter_func()` to set the flag and then come back around for another `readline()` where the streamer could remain stuck if there's nothing in the `stdout` buffer. So, this way, whether the stop flag was set either while `readline()` was waiting or as a consequence of `filter_func()`, the thread would then exit without any further reading or printing/capturing.

I can add something to encompass this....unless you might see it differently.

freakboy3742:
> > The only other issue I noticed is a weird one - the code clearly has 100% coverage, but I can't see how the `capture_output = False` case is being exercised. I'm guessing it's being tested implicitly by another test somewhere, but given it's at the core of how this works, it would be worth having an explicit test.
> 
> Would you consider the `test_output()` test in `tests/integrations/subprocess/test_PopenOutputStreamer.py` to cover this case? Or maybe something more?

Ah - that was the case I was missing. It's relying on the default value of the argument.

> > Have I got that right?
> 
> Right....the theory is if we read `stdout` from the emulator, it'll stop blocking it from properly opening for users.

👍 
 
> More broadly, this provides the option of creating a `Popen` and instead of only being able to call `Subprocess.stream_output()` to read `stdout` while blocking, there's now `Subprocess.stream_output_non_blocking()` to (obviously) not block while reading `stdout`.

Yeah - that's a nice API addition. Not sure if we'll have any use for it, but it's a nice club to have in the bag just in case, and the rest of the refactoring around adding that API are all nice improvements as well.

src/briefcase/integrations/android_sdk.py
rmartin16: I think the general error message is definitely still useful; copied here:
```
Review the emulator output above for:
 - Troubleshooting or resolution steps such as enabling hardware acceleration
 - Other errors or warnings that may be suggesting the cause of the startup failure

Ensure your Android SDK is up-to-date by running:

    $ briefcase upgrade {AndroidSDK.name}

To review Google's general troubleshooting steps for the emulator, visit:

    https://developer.android.com/studio/run/emulator-troubleshooting

You can also start the emulator manually by running:

    $ {emulator_command}
```

Do you mean this one?
```
                self.tools.logger.info(
                    """
If the emulator opened after pressing CTRL+C, then leave the emulator open and
run Briefcase again. The running emulator can then be selected from the list.
"""
                )
```
Because I was considering removing this one.....but since I have rather little evidence this is a silver bullet, I figured we'd leave it in for now until we haven't heard more reports in a while. But I don't have a strong opinion since most warnings like this seem to go unread anyway...

freakboy3742: It was the latter one that I was questioning. Agreed that we probably want to keep it just in case it turns out this doesn't fix the problem - I wasn't sure of your level of confidence on this resolving the underlying issue. If it *does* resolve the issue it's worth keeping in mind for cleanup later.

src/briefcase/integrations/android_sdk.py
freakboy3742: Sure the exact name doesn't really concern me, it's more the conceptual wrapping of the behavior, rather than reaching into internals. Also `stop()` could easily be confused for a method on the core thread API (it doesn't AFAICT, but `stop()` would be an obvious conceptual match for `run()`... but it won't behave that way).

src/briefcase/integrations/subprocess.py
freakboy3742: That's exactly the sort of edge case I was thinking of. It makes perfect sense; we just need a docstring so that future travellers don't look at the code and thing "oh, I can clean this up!".

src/briefcase/integrations/android_sdk.py
rmartin16: RE: stopping threads

I always loved this blog post from raymond chen....although, he seemingly removed it...

https://web.archive.org/web/20190108101649/https://blogs.msdn.microsoft.com/oldnewthing/20150814-00/?p=91811

freakboy3742 (approved):
All looks good... now to cross our fingers and hope it actually fixes the windows emulator lockup problem 🤞

PR #1694, opened 2024-03-13
Remove `fsdecode()` calls for `subprocess` commands

rmartin16:
## Changes
- Removes calls to `os.fsdecode()` for command arguments for `subprocess` calls
- `subprocess` will call `fsdecode()` for `subprocess` command arguments for us
  - Windows: `subprocess` always [calls](https://github.com/python/cpython/blob/25684e71310642ffd20b45eea9b5226a1fa809a5/Lib/subprocess.py#L1460) `list2cmdlist` for Windows processes
  - POSIX: for `posix_spawn` [path](https://github.com/python/cpython/blob/7f418fb111dec325b5c9fe6f6e96076049322f02/Modules/posixmodule.c#L7489) or the `fork_exec` [path](https://github.com/python/cpython/blob/7f418fb111dec325b5c9fe6f6e96076049322f02/Modules/_posixsubprocess.c#L1066), everything has `fsdecode` called

## PR Checklist:


freakboy3742 (approved):
This all makes sense; the RTD build failure looks to be unrelated. I've pushed #1695 to correct this; once that lands we can rebase and get a clean board here. Approving on that basis; as soon as there's a clean CI run, I'm OK to merge.

Issue #738, opened 2022-05-11
Improve support for bootstrapping a development environment without an active network connection


freakboy3742:
When you run a briefcase command for the first time, it requires an active network connection, and will download multiple gigabytes of tools. This is particularly bad on macOS/iOS (which require Xcode) and Android (which needs the sdkmanager an associated tools). This is a problem in bandwidth-limited scenarios (like tutorial sessions or conference wifi).

It should be possible to give someone a USB drive populated with files, and bootstrap their briefcase installation.

**Describe the solution you'd like**

A new command:
`briefcase bootstrap <platform> <folder>` 

will inspect <folder>, and install all the tools needed for <platform>. This may not 100% remove the need for network activity, but it should be possible to dramatically lower the initial bandwidth/time requirement.

A parallel command/option will also be required to *create* the archive folder in the first place (`briefcase bootstrap --create <platform> <folder>`?)

**Describe alternatives you've considered**

No real options, other than requiring an active network connection.

**Additional context**

Support packages and templates can be copied directly from the source folder; it may be desirable to require them to be stored in zip format.

The Android sdkmanager has support for a `--package_file` argument that can be used to directly install packages from a zip file. It's not clear 

There is likely not much we can do about Xcode, as it can't be distributed outside the App Store.

jlgimeno:
It is possible to download the Xcode .xip file if one has an Apple Developer account.  I'd say that would be an edge case for the typical briefcase user.

freakboy3742:
See #388 for an other issue related to offline operation

mhsmith:
To build offline for Android, as well as sdkmanager packages, you would also need cached copies of all the necessary Gradle plugins and libraries. I think the only feasible way of doing that would be to actually do a build, and then save a copy of the `~/.gradle` directory.

cycleuser:
I also found a very clumsy solution to this problem.
I downloaded all the templates and put them in `python\Lib\site-packages\briefcase\templates`.
Then modify the code on `new.py` like:
```Python
        if template is None:
            # template = "https://github.com/beeware/briefcase-template"
            template = "./templates/briefcase-template"
```

针对这个问题，我也找到了一个很笨拙的解决方案。
就是将所有要用到的template都下载好，放在`python\Lib\site-packages\briefcase\templates`里面。
然后把`new.py`修改一下：

```Python
        if template is None:
            # template = "https://github.com/beeware/briefcase-template"
            template = "./templates/briefcase-template"
```

[...]

PR #1715, opened 2024-03-31
adding --offline to briefcase new

cycleuser:
tested it on Windows 11 and Ubuntu and found that it's fine, and it doesn't affect other commands, so I just downloaded the existing templates and put them in the templates folder under src\briefcase, and then modified the base py in the briefcase\commands folder under src\briefcase. briefcase\commands, then modified base.pyandnew.pyundersrc\briefcase\commands`.

Just tested it, and both on Windows 11 and Ubuntu 22.04, it works with `briefcase new --offline`.

I didn't pr directly at the beginning because I was afraid that it wouldn't pass the test on different platforms, but I just tested it on Windows 11 and Ubuntu and found that it's fine, and it doesn't affect other commands, so I just downloaded the existing templates and put them in the `templates` folder under `src\briefcase`, and then modified the `base` py in the `briefcase\commands` folder under `src\briefcase`. briefcase\commands`, then modified `base.py` and `new.py` under `src\briefcase\commands`.

On Ubuntu 22.04 it works as the screenshot below:

image
image

#738 

https://github.com/beeware/briefcase/issues/738

## PR Checklist:

- [x] All new features have been tested on Windows 11 and Ubuntu 22.04
- [x] All new features have been documented with a `about.md`.


cycleuser:
@freakboy3742 

You are right that CI really can't get through because of issues with long file names and such.

My original purpose of the fork is just to implement a dependency on GitHub for access to `briefcase new`.

Then saw what you said at the time about the PR.

Now it seems that this requirement, although realized with a simple modification, should not have warranted a separate PR for it.

As for your question about how much work is involved, this is really very minor and is not worthy to care about.

> I guess you could argue that this PR's title only claims to solve briefcase new... but then, all but one of the templates that are embedded aren't used by briefcase new.

I'm not going to argue anything because there was no intention of submitting a PR in the first place.

There's no point in arguing about that either and no value to guess about that,

After all, I just want the `briefcase new` process to be internet-independent, or at least not have to visit GitHub.

Now that it's realized, it's fine.

Issue #431, opened 2020-06-23
Provide better protection against manually installed requirements


freakboy3742:
A frequently reported error: A user creates a venv, manually installs a package into the venv, and runs `briefcase dev`. The app runs succcessfully. The user then runs `briefcase run`, and the app fails because the package can't be found.

This is 100% working-as-intended; the package that was installed manually needs to be added to the `requires` list to be packaged with the independent environment used in the running app.

However, it's sufficiently confusing that we should provided better protection against this mode of failure. 

Two possible mitigations:

1) Don't use the local venv when running `briefcase dev`. Tox, for example, creates an independent venv and runs tests in that venv. If `briefcase dev` used a similar approach, the environment where the app runs would be isolated from the venv where briefcase is running. This would have the added advantage that development tooling (like briefcase) wouldn't be present in the `dev` environment.

2) Raise a warning during `briefcase run` if there are packages in the development environment that aren't present in the `requires` list.  

The downside to (1) is a potentially slower development cycle, as starting the app will take a little more time (as the venv needs to be created/activated before the app runs)

The downside to (2) is the noise associated with packages that are in the dev environment legitimately - e.g., briefcase itself. It may be necessary to have a 'known dev packages' list in pyproject.toml to silence this warning.

jgirardet:
My answer is not 100% about it but it's the same idea.
We don't have yet a way to install dev packages. It can be done using `briefcase dev -d` but will run the app too.
A `briefcase dev -i` would install  all the depencies from pyrproject `requires`.
A require-dev in pyproject would solve the probleme of other dependencies (pytest, black,....) it's basically how poetry works and it's really straightforward.

About the real question (is it : "how to protect programmers from themselves?")  I'm not sure it's in the scope of briefcase.  Check every subdependency would be big work I think. Does it worth it ?.
And I don't mention the case where some one use a "dev" dependency in runtime code but forgot it.
Maybe should it be solved in #399 

If at the bare minimum we can have **one** source for dependencies, it's good.

freakboy3742:
I completely agree that having a mechanism to install dependencies is something worth adding, if only for testing. However, I'm not sure it addresses the problem I described.

The use case that is a problem is:
 * User decides to use `requests` in their app
 * User run `pip install requests`, and adds `import requests` to their app code
 * User runs `briefcase dev`; app runs fine.
 * User runs `briefcase run`; app fails because `requests` *isn't* in the `requires` list in `pyproject.toml`.

[...]

PR #1719, opened 2024-04-04
Eagerly update app requirements for `DevCommand`

rmartin16:
## Changes
- When requirements are re-installed in to the build for `BuildCommand`, the installation target directory is deleted prior to installing the app's requirements. Effectively, this means the transitive requirements are eagerly updated since the resolver will re-run and determine the latest version to install.
- To align with the behavior of `BuildCommand`, this updates `DevCommand` to eagerly update the app's transitive requirements when installing its direct requirements.

## PR Checklist:

### DIFF
--- a/changes/1719.feature.rst
+++ b/changes/1719.feature.rst
@@ -1,0 +1,1 @@
+The transitive requirements for an app are now eagerly updated when an app's requirements are installed for the ``briefcase dev`` command.
--- a/src/briefcase/commands/dev.py
+++ b/src/briefcase/commands/dev.py
@@ -97,6 +97,7 @@ class DevCommand(RunAppMixin, BaseCommand):
                             "-m",
                             "pip",
                             "install",
+                            "--upgrade-strategy=eager",
                             "--upgrade",
                         ]
                         + (["-vv"] if self.logger.is_deep_debug else [])
--- a/tests/commands/dev/test_install_dev_requirements.py
+++ b/tests/commands/dev/test_install_dev_requirements.py
@@ -26,6 +26,7 @@ def test_install_requirements_no_error(dev_command, first_app, logging_level):
             "-m",
             "pip",
             "install",
+            "--upgrade-strategy=eager",
             "--upgrade",
         ]
         + (["-vv"] if logging_level == LogLevel.DEEP_DEBUG else [])
@@ -58,6 +59,7 @@ def test_install_requirements_error(dev_command, first_app):
             "-m",
             "pip",
             "install",
+            "--upgrade-strategy=eager",
             "--upgrade",
             "package-one",
             "package_two",
@@ -93,6 +95,7 @@ def test_install_requirements_test_mode(dev_command, first_app):
             "-m",
             "pip",
             "install",
+            "--upgrade-strategy=eager",
             "--upgrade",
             "package-one",
             "package_two",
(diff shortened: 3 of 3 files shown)


rmartin16:
For some of the output formats, Briefcase defers installing the app's requirements to the build system.
- Android Gradle
  - Running `briefcase build android -r` doesn't seem to actually force a re-installation of the requirements...
  - Instead, some comparison must be being made to determine if the step can be skipped....perhaps similar to how Docker layer caching works
  - The implication therefore being requirements can only be updated with direct version pinning or re-creating the build cache
  - **Would it be possible to update the Gradle build to respect a flag to force the re-installation of requirements?**
- Linux Flatpak
  - Each build always re-installs all direct and transitive requirements matching the behavior of `pip install` in `BuildCommand`
- Web Static
  - Gathering/building wheels in `BuildCommand` seems to re-run the resolver each time for all requirements...

freakboy3742:
>   * Running `briefcase build android -r` doesn't seem to actually force a re-installation of the requirements...
>   * **Would it be possible to update the Gradle build to respect a flag to force the re-installation of requirements?**

It forces re-writing the requirements.txt file; but that doesn't touch transitive requirements (unless we were to modify Briefcase to output the requirements resolution, rather than just the *explicit* requirements. That might pair well with a solution to #476.

I don't know if the Chaquopy gradle plugin can be forced to do an eager update. AFAIK, it's essentially using a internally vendored fork of pip, so it *might* be possible - @mhsmith would need to answer that one.

freakboy3742 (commented):
I can see the edge case you're describing, but I'm not 100% convinced whether it's desirable or not. 

Consistency with `briefcase run` is definitely a compelling argument. However, I'm wary that `briefcase dev` is running in a virtual environment that might be shared with other tools. Eagerly updating a transitive dependency could easily have unintended consquences - especially if there's no way to opt out of that mode of update. 

For my money, this is almost papering over limitations that would be addressed by a combination of #476 and #431.  

Interestingly, one of the possible solutions for #431 is to make `dev` use an isolated virtual environment, rather than the one that is running `briefcase`. In that context, eager update would absolutely make sense (and be consistent with `run`), and would have the additional benefit that `pip install` leakage would no longer be an issue - if you don't put a requirement in the requires list, it won't be available. 

@mhsmith - Do you have any thoughts here?

mhsmith:
> I don't know if the Chaquopy gradle plugin can be forced to do an eager update. AFAIK, it's essentially using a internally vendored fork of pip, so it _might_ be possible - @mhsmith would need to answer that one.

This is actually happening at the level of the Chaquopy Gradle plugin – it only reruns pip if the requirements have changed. Any change at all to the requirements file should be sufficient, including a comment.

rmartin16:
> It forces re-writing the requirements.txt file;

> it only reruns pip if the requirements have changed.

This could offer a rationale for inserting a datetime or something similar in `requirements.txt` each time it's written so Gradle detects a change and re-installs the requirements.

> Eagerly updating a transitive dependency could easily have unintended consquences - especially if there's no way to opt out of that mode of update.

hmm....using the same python environment for disparate projects seems likely to create issues whether we do eager updates or not. But there's no real pressing issue to merge something like this in when other solutions could more holistically address this situation.

So, I'll will take a look at those references and see if anything looks doable right now.

rmartin16:
Closing in favor of an as of yet not designed solution that more holistically addresses dev environment management.

PR #1720, opened 2024-04-04
Use verbose logging for Linux file perms updating

rmartin16:
## Changes
- Instead of individually listing all files whose permissions are updated for Linux packaging by default, this logging is now verbose so it only appears in the log or when `-v` is used.

## PR Checklist:


freakboy3742 (approved):
Makes sense to me!

PR #1722, opened 2024-04-07
Consolidate text wrapping support for console printing

rmartin16:
## Changes
- This consolidates the text wrapping support in `Console` from its original implementation in `NewCommand`.

## Notes
- If there isn't objection, the next step is to apply this everywhere the formatting is implemented directly in the code.
- That's a lot of places, though....so, I'm not sure yet if I'll try to do that in this PR or not.
- Notably, this will likely require converting the liberal use of  multil-ine `f"""..."""`  string definitions in to strings without newlines so `Console.textwrap()` can format them.
- Additionally, I think it would be ideal to textwrap the boxed `**** WARNING ****` messages...but that'll also take some thought for implementation.

## PR Checklist:

### DIFF
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -1,7 +1,5 @@
 import argparse
-import shutil
 import sys
-import textwrap
 from argparse import RawDescriptionHelpFormatter
 
 from briefcase import __version__
@@ -18,6 +16,7 @@ from briefcase.commands import (
     UpgradeCommand,
 )
 from briefcase.commands.base import split_passthrough
+from briefcase.console import MAX_TEXT_WIDTH, Console
 from briefcase.platforms import get_output_formats, get_platforms
 
 from .exceptions import InvalidFormatError, NoCommandError, UnsupportedCommandError
@@ -36,19 +35,18 @@ COMMANDS = [
 ]
 
 
-def parse_cmdline(args):
+def parse_cmdline(args, console: Console = Console()):
     """Parses the command line to determine the Command and its arguments.
 
     :param args: the arguments provided at the command line
+    :param console:
     :return: Command and command-specific arguments
     """
     platforms = get_platforms()
-    width = max(min(shutil.get_terminal_size().columns, 80) - 2, 20)
 
-    briefcase_description = textwrap.fill(
+    briefcase_description = (
         "Briefcase is a tool for converting a Python project "
-        "into a standalone native application for distribution.",
-        width=width,
+        "into a standalone native application for distribution."
     )
 
     description_max_pad_len = max(len(cmd.command) for cmd in COMMANDS) + 2
@@ -59,15 +57,14 @@ def parse_cmdline(args):
 
     platform_list = ", ".join(sorted(platforms, key=str.lower))
 
-    additional_instruction = textwrap.fill(
+    additional_instruction = (
         "Each command, platform, and format has additional options. "
-        "Use the -h option on a specific command for more details.",
-        width=width,
+        "Use the -h option on a specific command for more details."
     )
 
     parser = argparse.ArgumentParser(
         prog="briefcase",
-        description=(
+        description=console.textwrap(
             f"{briefcase_description}\n"
             "\n"
             "Commands:\n"
@@ -80,7 +77,9 @@ def parse_cmdline(args):
         ),
         usage="briefcase [-h] <command> [<platform>] [<format>] ...",
         add_help=False,
-        formatter_class=lambda prog: RawDescriptionHelpFormatter(prog, width=width),
+        formatter_class=(
+            lambda prog: RawDescriptionHelpFormatter(prog, width=MAX_TEXT_WIDTH)
+        ),
     )
     parser.add_argument("-V", "--version", action="version", version=__version__)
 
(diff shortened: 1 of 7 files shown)


rmartin16:
FYI. I've seen this error twice today now.
```
flake8.exceptions.FailedToLoadPlugin: Flake8 failed to load plugin "pyflakes" due to source code string cannot contain null bytes.
```
https://github.com/beeware/briefcase/actions/runs/8591482948/job/23540572715?pr=1722#step:11:61

freakboy3742:
> FYI. I've seen this error twice today now.

Any idea of the cause of this one? It's evidently transient, because it doesn't survive a re-run; but it's not a mode of failure I've seen before, and it doesn't read like an obvious networking issue (which is the usual GitHub Actions CI issue). Maybe a glitchy NFS mount?

rmartin16:
hmm....I'm inclined to say it wasn't a low-level protocol error/blip....since the error was identical both times:
```
  File "/opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/flake8/plugins/pyflakes.py", line 10, in <module>
    import pyflakes.checker
  File "/opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/pyflakes/checker.py", line 12, in <module>
    import doctest
SyntaxError: source code string cannot contain null bytes
```
Other instance: https://github.com/beeware/briefcase/actions/runs/8590845652/job/23538952152#step:11:61

I wonder if GitHub has a corrupted version of `doctest` in their caches somehow....

src/briefcase/cmdline.py
```diff
@@ -36,19 +35,18 @@
 ]
 
 
-def parse_cmdline(args):
+def parse_cmdline(args, console: Console = Console()):
```
freakboy3742, line 38: The use of the default here means a Console() instance will always be created (as the default argument is evaluated when the function is defined). Passing in None, and constructing if the argument is None would be preferable.

src/briefcase/cmdline.py
```diff
@@ -36,19 +35,18 @@
...
     """Parses the command line to determine the Command and its arguments.
 
     :param args: the arguments provided at the command line
+    :param console:
```
freakboy3742, line 42: Documentation required here.

tests/console/Console/test_textwrap.py
```diff
@@ -0,0 +1,63 @@
...
+            "roll the image, make it flutter. We can change the focus to a soft blur or\n"
+            "sharpen it to crystal clarity. For the next hour, sit quietly, and we will\n"
+            "control all that you see and hear. We repeat: There is nothing wrong with your\n"
+            "television set.",
+        ),
+    ],
+)
+def test_textwrap(in_text, out_text):
+    """Text is formatted as expected."""
+    assert Console().textwrap(in_text) == out_text
```
freakboy3742, line 55: Possibly overkill, but would it be worth testing a range of output sizes here to confirm that breaking works as expected?

freakboy3742 (changes requested):
Broadly looks good; a couple of minor suggestions inline.

tests/console/Console/test_textwrap.py
freakboy3742: I was thinking more about the adaptation to the `width` argument in the line wrap - i.e., that wrapping with width of  30/50/80/120 yield different breaks.

rmartin16: ohh...ok; for the `width` argument tests; gotcha

src/briefcase/cmdline.py
rmartin16: 🤦🏼

src/briefcase/cmdline.py
rmartin16: good point; defaulting to `None` now

rmartin16:
Do you have any objections to applying this to the rest of the project?

This was prompted from my work on "build tracking"....so, I think I'll still probably delay actually updating all multi-line output from this PR but wanted to make sure we're on the same page for transitioning from the `f"""..."""` text blocks.


P.S. furthermore, I'm targeting to get this PR and my "linux abstractions" updates in before PyCon....since I'm expecting a release before then. just fyi. (third on that list is Toga typing but probably not targeting pre-pycon.)

freakboy3742:
> Do you have any objections to applying this to the rest of the project?

None at all. Anywhere that we're currently doing manual line breaks is definitely a candidate for being cleaned up in this way. The only place that won't be a drop-in replacement are the "banner" warnings (e.g., the warning when JAVA_HOME doesn't point at a JDK - but I'd argue those should be factored out into some sort of common "banner" mechanism as well (although that doesn't need to be part of this PR unless you're feeling enthused).

> P.S. furthermore, I'm targeting to get this PR and my "linux abstractions" updates in before PyCon....since I'm expecting a release before then. just fyi. (third on that list is Toga typing but probably not targeting pre-pycon.)

Agreed - I'm expecting a release of both Toga and Briefcase before PyCon. For planning purposes, I'd say we want to have a release candidate around May 1, so that any release issues get caught before we're live in the middle of a tutorial.

rmartin16:
All right; good deal. As a correction, I meant I'm targeting to get "build tracking" and "linux abstractions" in before PyCon....so, I'll shoot for May 1...that may turn out to be optimistic, though.

As for this PR, feel free to merge when you're comfortable with it and I'm return to the updating the rest of the text blocks later.

freakboy3742 (approved):
👍

Issue #1011, opened 2022-12-16
App Name Validation Can Allow Names that Produce an Invalid `pyproject.toml`


rmartin16:
### Describe the bug

The App name validation can allow characters that are not valid as "bare keys" in the table names of the `pyproject.toml` file.

For example, an app name of `this_is_a_dotless_lowercase_i_ı` produces table names like `[tool.briefcase.app.this_is_a_dotless_lowercase_i_ı.macOS]`.

TOML only allows ASCII in "[bare key](https://toml.io/en/v1.0.0#keys)" names. But "quoted keys" can use a much broader set of characters. So, `[tool.briefcase.app."this_is_a_dotless_lowercase_i_ı".macOS]` becomes valid TOML.

Given that Briefcase is simply using the [regex](https://peps.python.org/pep-0508/#names) from PEP-508 to validate names, the root of this may stem from much further upstream. Nonetheless, I think the underlying problem is they recommended using `re.IGNORECASE`; my guess is the underlying case folding is making `ı` in to a "normal" lowercase `i` and the the validation succeeds.

### Steps to reproduce

Use `ı` or also `İ` in the App Name when running `briefcase new`.

### Expected behavior

Such names should be rejected or the TOML should be created properly.

### Environment

- Operating System: pop os 22.04
- Python version: 3.10.8
- Software versions:
  - Briefcase: `0.3.12.dev251+g6d29ece1`

### Logs

[briefcase.2022_12_16-15_17_55.dev.log](https://github.com/beeware/briefcase/files/10249089/briefcase.2022_12_16-15_17_55.dev.log)

freakboy3742:
`this_is_a_dotless_lowercase_i_ı` is a valid module name in python (as in `import this_is_a_dotless_lowercase_i_ı` works if there's a `this_is_a_dotless_lowercase_i_ı.py` on your filesystem), so the app name validation is working; however, it's not valid TOML in a bare state. This suggests to me that we need a "safe toml key" filter that can determine if a string requires quoting (and add quotes if it is), then use that filter in the briefcase template anywhere a user-provided string is used in a key. This should be pretty straightforward - "is this string bare ascii" is a robust check, and we already have the `briefcase.integrations.cookiecutter` module defining a bunch of filters.

rmartin16:
[...]

PR #1762, opened 2024-04-29
Fix minor holes in PEP-508 name validation

rmartin16:
## Changes
- This resolves a hole in the recommended implementation of name validation from [PEP-508](https://peps.python.org/pep-0508/#names) to avoid allowing non-ASCII values.
```
The regex (run with re.IGNORECASE) is:

^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$
```
- The implementation of `re.IGNORECASE` is ostensibly coercing the case of the input because it will match on non-ASCII.
  - This matters for, at least, İ ([0x130](https://everythingfonts.com/unicode/0x0130)) and K ([0x212a](https://everythingfonts.com/unicode/0x212A)).
```pycon
>>> import re
>>> PEP508_NAME_RE = re.compile(r"^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$", re.IGNORECASE)
>>> 
>>> bool(PEP508_NAME_RE.match("helloworld"))
True
>>> bool(PEP508_NAME_RE.match("İstanbul"))
True
>>> bool(PEP508_NAME_RE.match("Kelvin"))
True
>>> bool(PEP508_NAME_RE.match("Æolia"))
False
>>> bool(PEP508_NAME_RE.match("jalapeño"))
False
>>> bool(PEP508_NAME_RE.match("Beyoncé"))
False
>>> bool(PEP508_NAME_RE.match("naïve"))
False
```

## Related
- https://github.com/beeware/briefcase/issues/1011

## PR Checklist:

### DIFF
--- a/src/briefcase/config.py
+++ b/src/briefcase/config.py
@@ -15,7 +15,11 @@ from briefcase.platforms import get_output_formats, get_platforms
 from .constants import RESERVED_WORDS
 from .exceptions import BriefcaseConfigError
 
-# PEP508 provides a basic restriction on naming
+# PEP 508 restricts the naming of modules. The PEP defines a regex that uses
+# re.IGNORECASE; but in in practice, packaging uses a version that rolls out the lower
+# case, which has very slightly different semantics with non-ASCII characters. This
+# definition is the one from
+# https://github.com/pypa/packaging/blob/24.0/src/packaging/_tokenizer.py#L80
 PEP508_NAME_RE = re.compile(r"^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9._-]*[a-zA-Z0-9])$")
 
 
--- a/src/briefcase/config.py
+++ b/src/briefcase/config.py
@@ -16,7 +16,7 @@ from .constants import RESERVED_WORDS
 from .exceptions import BriefcaseConfigError
 
 # PEP508 provides a basic restriction on naming
-PEP508_NAME_RE = re.compile(r"^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$", re.IGNORECASE)
+PEP508_NAME_RE = re.compile(r"^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9._-]*[a-zA-Z0-9])$")
 
 
 def is_valid_pep508_name(app_name):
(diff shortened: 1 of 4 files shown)


rmartin16:
PEP-508 says PyPI uses this regex...and indeed, it appears they [do](https://github.com/pypa/packaging/blob/32deafe8668a2130a3366b98154914d188f3718e/src/packaging/utils.py#L37).

However, `packaging.requirements.Requirement` doesn't [use](https://github.com/pypa/packaging/blob/32deafe8668a2130a3366b98154914d188f3718e/src/packaging/_tokenizer.py#L82) `re.IGNORECASE`....so, even if you get such a package on to PyPI, it seems likely `pip` or another tool will reject it at some point.

src/briefcase/config.py
```diff
@@ -16,7 +16,7 @@
 from .exceptions import BriefcaseConfigError
 
 # PEP508 provides a basic restriction on naming
```
freakboy3742, line 18:
```suggestion
# PEP 508 restricts the naming of modules. The PEP defines a regex that uses
# re.IGNORECASE; but in in practice, packaging uses a version that rolls out the lower
# case, which has very slightly different semantics with non-ASCII characters. This
# definition is the one from
# https://github.com/pypa/packaging/blob/24.0/src/packaging/_tokenizer.py#L80
```

tests/config/test_is_valid_app_name.py
```diff
@@ -34,6 +34,11 @@ def test_is_valid_app_name(name):
         "main",
         "socket",
         "test",
+        # İ and K (i.e. 0x212a) are valid ASCII when made lowercase and as such are
+        # accepted by the PEP-508 regex...but these should be rejected here to ensure
+        # compliance where app name is used downstream and only ASCII is accepted.
+        "İstanbul",
+        "Kelvin",
```
freakboy3742, line 41:
```suggestion
        # ı, İ and K (i.e. 0x212a) are valid ASCII when made lowercase and as such are
        # accepted by the official PEP 508 regex... but they are rejected here to ensure
        # compliance with the regex that is used in practice.
        "helloworld_ı",
        "İstanbul",
        "Kelvin",
```

freakboy3742 (changes requested):
That's an odd discrepancy, but I can't argue with the practicality. A couple of minor clarification suggestions, including preserving the symmetry of the dotless-i test case.

freakboy3742 (approved):

PR #1766, opened 2024-05-02
Ensure macOS apps pick up new icons after an update

freakboy3742:
With the switch to Toga using the binary's icon at runtime, a new bug emerges - macOS caches app icons. When using the app template, and you change the app icon, the app bundle on the filesystem will be updated, but starting the app will show the old icon.

If you re-create the app, the new icon is displayed.

The cache key appears to be the modification time on the .app bundle itself; if you touch the bundle as part of an update, the new icon is loaded.

This didn't affect the Xcode template because every run is a new build, which implicitly touches the .app bundle.

This PR force touches the .app bundle on the macOS backends to ensure that the icon cache is busted.

Includes a coupe of cosmetic whitespace updates for good measure.

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/create.py
+++ b/src/briefcase/commands/create.py
@@ -686,8 +686,8 @@ class CreateCommand(BaseCommand):
         # Briefcase v0.3.18 - splash screens deprecated.
         if getattr(app, "splash", None):
             self.logger.warning(
-                "Splash screens are now configured based on the icon. "
-                "The splash configuration will be ignored."
+                "\nSplash screens are now configured based on the icon. "
+                "The splash configuration will be ignored.\n"
             )
 
         for extension, doctype in self.document_type_icon_targets(app).items():
--- a/src/briefcase/commands/new.py
+++ b/src/briefcase/commands/new.py
@@ -717,6 +717,7 @@ To run your application, type:
 
     $ cd {context['app_name']}
     $ briefcase dev
+
 """
         )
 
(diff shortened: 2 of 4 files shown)


mhsmith:
FYI: I've seen something similar on Android, and I never discovered any workaround other than uninstalling and reinstalling the app. But I think it only affected the icon visible in the apps list in the launcher, not the running app itself. It's possible this has been fixed on newer Android versions.

src/briefcase/commands/create.py
```diff
@@ -686,8 +686,8 @@ def install_app_resources(self, app: AppConfig):
         # Briefcase v0.3.18 - splash screens deprecated.
         if getattr(app, "splash", None):
             self.logger.warning(
-                "Splash screens are now configured based on the icon. "
-                "The splash configuration will be ignored."
+                "\nSplash screens are now configured based on the icon. "
+                "The splash configuration will be ignored.\n"
             )
```
rmartin16, line 691: As an FYI, as a consequence of `str.splitlines()` in `Log()`, a single trailing newline is effectively ignored. However, a trailing newline in the output text can be achieved with two trailing newlines in the string.

```pycon
>>> "\nline of text\n".splitlines()
['', 'line of text']
>>> "\nline of text\n\n".splitlines()
['', 'line of text', '']
```
Same as the "extra" newline in the multiline strings.
```pycon
>>> """line 1
... line 2
... line 3
... 
... """
'line 1\nline 2\nline 3\n\n'
```

freakboy3742: Good catch - I'd completely forgotten we'd made this change. I've cleaned this up in #1769.

It would be nice if there was a way to automate catching stylistic regressions like this, but I'm not sure I have any ideas on how we'd do that.

PR #1781, opened 2024-05-07
Add support for console apps

### LINKED ISSUE #556: Generating a command-line executable without the GUI features

The Beeware tutorials only mention how to build and distribute a GUI app. It's unclear how to generate a command-line program. 

I wonder if I could opt out of the GUI requirements and build a standalone CLI executable that includes all the implicit Python dependencies, i.e., a single-file executable. Then with the `package` command this executable can be included in an installer.

### LINKED ISSUE #1184: Add support for creating macOS installers

Briefcase currently produces bare `.app` bundles, or `.dmg` bundles. They can be executed as-is, or dragged into the Applications folder.

However, macOS also supports the creation of installer packages. These aren't needed for simple apps, but if an app requires a post install script, or there are multiple resources files that need to be installed, an installer may be required.

Briefcase should support the creation of installers as a packaging format.

`briefcase package macOS -p installer`

Ideally, this would also allow for the addition of a user-supplied post-install script to the installer.

Do nothing, and don't support pre/post install scripts as a use case.

### Additional context

[This StackOverflow post](https://stackoverflow.com/a/11487658) includes a fairly detailed rundown of building an installer.

If an installer with a post-install script can be generated, it could provide a path for #556 on macOS. A command line app is an installed app that puts itself on the command line. The Python.org macOS installer is an example of this, adding `/Library/Frameworks/Python.framework/Versions/3.X/bin` to the path. Installing an app bundle in `~/Applications`, then adding `~/Applications/My App.app/Contents/MacOS/myapp` to the path would provide a packaged command-line app for macOS.

### LINKED ISSUE #1729: `briefcase run` fails after `briefcase create` when using stub app templates
### Describe the bug

When running an app on macOS or Windows, `briefcase run` will fail if a `build` step hasn't been executed. 

On Windows, the error manifests as:
```
SystemExit: stub: No module named stub
```
On macOS, it is 
```
The application cannot be opened for an unexpected reason, error=Error Domain=RBSRequestErrorDomain Code=5 "Launch failed." UserInfo={NSLocalizedFailureReason=Launch failed., NSUnderlyingError=0x158e246f0 {Error Domain=NSPOSIXErrorDomain Code=153 "Unknown error: 153" UserInfo={NSLocalizedDescription=Launchd job spawn failed}}}
```

### Steps to reproduce

```
$ briefcase new -Q app_name=foobar
$ cd foobar
$ briefcase create
$ briefcase run
```

### Expected behavior

App should start without error. The `run` should imply the need for a `build` on the macOS/Windows app templates, but because the template and binary both exist, the need for a build isn't detected.

### Screenshots

_No response_

### Environment

- Operating System: macOS, Windows
- Python version: All
- Software versions:
  - Briefcase: 0.3.17+

### Logs

```
  File "D:\obj\windows-release\37amd64_Release\msi_python\zip_amd64\runpy.py", line 183, in _run_module_as_main
  File "D:\obj\windows-release\37amd64_Release\msi_python\zip_amd64\runpy.py", line 136, in _get_module_details
runpy._Error: No module named stub

During handling of the above exception, another exception occurred:

SystemExit: stub: No module named stub

Application quit abnormally (Exit code -1)!
stub: No module named stub

Unable to start app helloworld.

```

### Additional context

The briefcase dev is normal, but the briefcase run is abnormal.

If you invoke `briefcase run` on a clean project, the app works.

If you invoke `briefcase build` before invoking `briefcase run`, the app works.

This only affects the `app` targets; the `Xcode`/`VisualStudio` templates do the right thing, because they require an explicit compilation step to create the binary.

freakboy3742:
Adds support for packaging console apps with Briefcase.

In order to support console apps, it is also necessary to introduce a `.pkg` app packaging format for macOS. This generates an installer that is able to perform the post-installation step of adding a symlink to `/usr/local/bin`. The PKG format is the default and only packaging format allowed for macOS console apps. 

This PR also renames the macOS `app` packaging format to `zip` - this is for consistency with Windows, to match the actual output format, and to avoid confusion about "packaging an app as an app that is a zip" descriptions.

To facilitate easier refactoring of the PKG format, the handling of signing identities on macOS has been modified to introduce a single `SigningIdentity` object that incorporates the id, identity name and team ID into a single object. 

To enable generation of product GUIDs on Windows, a new `dns_uuid5` template filter has been added. This will convert a domain name like `feature.appname.example.com` into a GUID using the UUID5 algorithm in the DNS namespace.

When running console apps (with `dev` or `run`), any output streaming is disabled. This is so that the user gets raw, unbuffered stdout - and especially stdin - access. However, when running a console app in test mode, streaming *is* enabled, as we need to be able to process test output looking for the sentinel.

This branch also adds a `BRIEFCASE_DEBUG` environment variable to the runtime environment if running in debug (i.e., `-vv` mode. One of the major changes to the stub binaries is that they no longer generate the "preamble" console output (detailing the PYTHONPATH etc) unless `BRIEFCASE_DEBUG` is enabled. This allows us to generate console apps that only generate output that the user actually requests, while retaining the ability to get deep debugging information if necessary.

Lastly, a new bootstrap (and a testing bootstrap) for console apps has been added.

This code can be tested with the macOS Xcode backend, the windows VisualStudio backend, the linux System backend the Flatpak backend. It requires template modifications to generate console-compatible apps. To test this branch, add:
```
template_branch = "console-app"
console_app = True
```
to your `pyproject.toml`, and run briefcase with:
```
$ briefcase run macOS Xcode
$ briefcase package macOS Xcode
```

The macOS App template and Windows App template have been partially updated. The logic to generate and handle 2 stub binaries has been added; however, those PRs cannot be completed until the macOS Xcode and Windows VisualStudio template changes have been merged. Inadvertently, this also fixes #1729; by moving the process of renaming the stub from the create step to the build step, we're able to clearly identify if the a build has been performed.

The landing strategy for this: 
- [x] Land #1850, adding the UUID cookiecutter template.
- [x] Do a preliminary review of this Briefcase branch
- [x] Test and land the non-app templates using this branch
- [x] Use this branch to manually generate stubs for the two app templates
- [ ] Test and land the remaining templates
- [ ] Do a final review and land this PR.

As an aside: This PR increases the need to address #993, as the app repos have just doubled the number of binaries that are stored. See #1849 for the in-progress work to extract the stub binaries.

* Fixes #556
* Fixes #1184
* Fixes #1729

* Refs beeware/briefcase-macOS-Xcode-template#41
* Refs beeware/briefcase-linux-system-template#26
* Refs beeware/briefcase-linux-flatpak-template#43
* Refs beeware/briefcase-windows-visualStudio-template#29
* Refs beeware/briefcase-macOS-app-template#54
* Refs beeware/briefcase-windows-app-template#48

---
## Update: 4 Jun

The Xcode template updates have been landed, so `briefcase run macOS Xcode` and `briefcase run Windows VisualStudio` will work without the `template_branch` setting.

The app templates have been updated with new stub binaries, so `briefcase run` will work as long as the `template_branch = "console-app"` setting exists.

## PR Checklist:


src/briefcase/platforms/macOS/__init__.py
```diff
@@ -766,93 +864,234 @@ def package_app(
...
+                            "BundleOverwriteAction": "upgrade",
+                            "RootRelativeBundlePath": self.binary_path(app).name,
+                        }
+                    ],
+                    components_plist,
+                )
 
-                try:
-                    icon_filename = self.base_path / f"{app.installer_icon}.icns"
+        # Console apps are installed in /Library/Formal Name, and include the
```
michelcrypt4d4mus, line 978: Can't console apps also be installed in `~/Library/`?

michelcrypt4d4mus:
Just poking around here for a bit and sorry if this is a dumb question as i've never really looked at the briefcase code base before but let's say i was trying to create an app with a pre or post-install script - it's not immediately clear to me how to do that. do I just create a `preinstall` and/or `postinstall` script somewhere (where?) in my project?

freakboy3742:
> let's say i was trying to create an app with a pre or post-install script - it's not immediately clear to me how to do that. do I just create a `preinstall` and/or `postinstall` script somewhere (where?) in my project?

This isn't a customisation point that this PR supports - but if you were to use a custom app template (e.g.,, a derivative of the [Xcode template](https://github.com/beeware/briefcase-macOS-Xcode-template/pull/41) that is required to use this branch), you could customise the existing `postinstall`, or add an additional `preinstall` script. Those two names are "special" to the pkgbuild tool; the location in the `scripts` folder is hard-coded by Briefcase.

michelcrypt4d4mus:
ah ok, i hadn't seen that repo before. why not let the user configure their own postinstall script location in `pyproject.toml`? seems like that could be a useful feature.

michelcrypt4d4mus:
i tried installing this branch of `briefcase` as per the docs and added
```
template_branch = "console-app"
console_app = True
```
to my `pyproject.toml`. `briefcase run macOS Xcode` worked fine but `briefcase package macOS Xcode` failed:
```
[myapp] Building PKG...
Installing license... errored

Log saved to /Users/uzor/workspace/my_app/briefcase_app/myapp/logs/briefcase.2024_05_21-17_35_53.package.log

  File "/Users/uzor/workspace/my_app/briefcase_app/myapp/.venv/bin/briefcase", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/uzor/workspace/my_app/lib_repos/briefcase/src/briefcase/__main__.py", line 29, in main
    command(**options)
  File "/Users/uzor/workspace/my_app/lib_repos/briefcase/src/briefcase/commands/package.py", line 151, in __call__
    state = self._package_app(
            ^^^^^^^^^^^^^^^^^^
  File "/Users/uzor/workspace/my_app/lib_repos/briefcase/src/briefcase/commands/package.py", line 99, in _package_app
    state = self.package_app(app, **full_options(state, options))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/uzor/workspace/my_app/lib_repos/briefcase/src/briefcase/platforms/macOS/__init__.py", line 879, in package_app
    self.package_pkg(
  File "/Users/uzor/workspace/my_app/lib_repos/briefcase/src/briefcase/platforms/macOS/__init__.py", line 932, in package_pkg
    (installer_path / "resources").mkdir(exist_ok=True)
  File "/opt/homebrew/Cellar/python@3.12/3.12.3/Frameworks/Python.framework/Versions/3.12/lib/python3.12/pathlib.py", line 1311, in mkdir
    os.mkdir(self, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/uzor/workspace/my_app/briefcase_app/myapp/build/myapp/macos/xcode/installer/resources'
```

because there's no `installer/` dir in `macos/xcode/`. is it an issue that i didn't originally `briefcase create` with these settings or something?

and while i'm here - what does it mean to do `briefcase run macOS Xcode` vs. `briefcase run macOS app`? I would have assumed that the `Xcode` option just tells briefcase to create a scaffolding `xcodeproj` etc, not actually run anything.

src/briefcase/platforms/macOS/__init__.py
michelcrypt4d4mus: couldn't you soft link `/usr/local/bin/app` to the `~/Library/app` executable in the `postinstall` the same way it does now? i guess come to think of it couldn't you just soft link straight to `/Applications/app`?

just asking because it seems like most of my installed apps are working with `~/Library` via `Application Support/` etc.  instead of the system wide `/Library` and apple is only getting more and more strict about what is allowed to ever touch system wide stuff.

michelcrypt4d4mus:
i think you may just need to change the line
```python
(installer_path / "resources").mkdir(exist_ok=True)
```
to add the `parents=True` option?

freakboy3742:
If you generated the project before you added the `template_branch` setting, then that will be the cause of the problem. The template is only used by the `create` command; once the template has been rolled out, the template isn't used again.

The difference between `macOS Xcode` and `macOS app` is that the Xcode template contains the full Xcode definition for building the stub app, and the app template contains a pre-compiled binary. *Most* projects can just use the pre-compiled binary, but obviously you need to generate the pre-compiled binary from somewhere - the Xcode project is also used to generate that binary.

michelcrypt4d4mus:
1. what's the best practice for recreating my app with a different template? Right now I'm looking at moving `pyproject.toml` out of the way temporarily, running `briefcase create`, and then moving it back... but that seems a bit fraught.
2. re:  `macOS Xcode` and `macOS app` is the right way to think about it that they are basically the same but `macOS Xcode` leaves behind more artifacts in the form of the full `xcodeproj` etc?

michelcrypt4d4mus:
nevermind - i sorted this out with `briefcase create macOS Xcode`. might want to update the PR description to add that step if other people might be testing this.

michelcrypt4d4mus:
is there any mechanism for injecting custom build options into `project.pbxproj`? for instance i have customized the template with `INFOPLIST_KEY_LSBackgroundOnly = YES;` in the `buildSettings`:
```
0D354FDD2551BFBD119178D1 /* Debug */ = {
			isa = XCBuildConfiguration;
			buildSettings = {
				ARCHS = "$(ARCHS_STANDARD)";
				ASSETCATALOG_COMPILER_APPICON_NAME = "App Server";
				ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME = AccentColor;
				CLANG_CXX_LANGUAGE_STANDARD = "gnu++0x";
				CLANG_CXX_LIBRARY = "libc++";
				CODE_SIGN_ENTITLEMENTS = AppServer/app.entitlements;
				CODE_SIGN_IDENTITY = "-";
				CODE_SIGN_STYLE = Automatic;
				COMBINE_HIDPI_IMAGES = YES;
				ENABLE_HARDENED_RUNTIME = YES;
				GCC_C_LANGUAGE_STANDARD = gnu99;
				INFOPLIST_FILE = AppServer/Info.plist;
				INFOPLIST_KEY_LSBackgroundOnly = YES;
				LD_RUNPATH_SEARCH_PATHS = (
					"$(inherited)",
					"@executable_path/../Frameworks",
				);
				MACOSX_DEPLOYMENT_TARGET = 11.0;
				PRODUCT_BUNDLE_IDENTIFIER = com.appserver.app;
				PROVISIONING_PROFILE_SPECIFIER = "";
				STRIP_INSTALLED_PRODUCT = NO;
			};
			name = Debug;
		};
```

If I could somehow specify that in the briefcase config so that `briefcase create` didn't overwrite it that would be awesome.

freakboy3742:
> nevermind - i sorted this out with `briefcase create macOS Xcode`. might want to update the PR description to add that step if other people might be testing this.

The PR description *does* say this:
Screenshot 2024-05-22 at 9 16 00 AM


> is there any mechanism for injecting custom build options into `project.pbxproj`? for instance i have customized the template with `INFOPLIST_KEY_LSBackgroundOnly = YES;` in the `buildSettings`:

There's no mechanism to directly manipulate the `project.pbxproj` file; however, you can inject values into the `Info.plist` using the [`info`](https://briefcase.readthedocs.io/en/stable/reference/platforms/macOS/app.html#info) key in your configuration. In your case, it sounds like adding `info."LSBackgroundOnly" = true` would do what you want.

mhsmith:
> * Do a preliminary review of this Briefcase branch
> * Test and land the non-app templates (except for Windows VisualStudio) using this branch
> * Use this branch to manually generate stubs for the two app templates

Why not VisualStudio in step 2?

docs/reference/platforms/macOS/app.rst
```diff
@@ -24,11 +24,15 @@ By default, apps will be both signed and notarized when they are packaged.
 Packaging format
 ================
 
-Briefcase supports two packaging formats for a macOS ``.app`` bundle:
+Briefcase supports three packaging formats for a macOS ``.app`` bundle:
```
mhsmith, line 27:
```suggestion
Briefcase supports three packaging formats for a macOS app:
```

docs/reference/platforms/macOS/app.rst
```diff
@@ -24,11 +24,15 @@ By default, apps will be both signed and notarized when they are packaged.
...
 
-1. A DMG that contains the ``.app`` bundle (the default output of ``briefcase package
-   macOS``, or by using ``briefcase package macOS -p dmg``); or
-2. A zipped ``.app`` folder (using ``briefcase package macOS -p app``).
+1. A DMG that contains the ``.app`` bundle (using ``briefcase package macOS -p dmg``);
+   or
```
mhsmith, line 30:
```suggestion
1. A DMG that contains the ``.app`` bundle (using ``briefcase package macOS -p dmg``).
```

docs/reference/platforms/macOS/app.rst
```diff
@@ -24,11 +24,15 @@ By default, apps will be both signed and notarized when they are packaged.
...
+2. A zipped ``.app`` folder (using ``briefcase package macOS -p zip``).
+3. A ``.pkg`` installer (using ``briefcase package macOS -p pkg``).
+
+The ``.pkg`` format is the *required* format for console apps. ``.dmg`` format is the
+default format GUI apps.
```
mhsmith, line 35:
```suggestion
``.pkg`` is the *required* format for console apps. ``.dmg`` is the
default format for GUI apps.
```

docs/reference/platforms/macOS/xcode.rst
```diff
@@ -21,11 +21,15 @@ By default, apps will be both signed and notarized when they are packaged.
 Packaging format
 ================
 
-Briefcase supports two packaging formats for a macOS Xcode project:
+Briefcase supports three packaging formats for a macOS Xcode project:
```
mhsmith, line 24: This text should always be identical to the same section in app.rst. Could it be shared using an `.. include` directive?

docs/reference/configuration.rst
```diff
@@ -237,6 +237,13 @@ on an app with a formal name of "My App" would remove:
 3. Any ``.exe`` file in ``path`` or its subdirectories.
 4. The file ``My App/content/extra.doc``.
 
+``console_app``
+~~~~~~~~~~~~~~~
+
+A Boolean describing if the app is a console app, or a GUI app. Defaults to ``False``
+(producing a GUI app). This setting has no effect on platforms that do not support
+a console mode (e.g., web or mobile platforms).
```
mhsmith, line 245: It would be useful to document exactly what effects this setting has.

freakboy3742:
> Why not VisualStudio in step 2?

It's a chicken and egg thing. The VisualStudio template requires the UUID change in this PR, so we can't land the template change without also landing this Briefcase branch. The same problem doesn't exist for the Xcode template because that template doesn't depend on any new Briefcase features.

If it would help, I could extract the UUID piece into a standalone PR; it's nicely self contained, and would then allow running the Visual Studio template against a mainline Briefcase.

docs/reference/platforms/macOS/xcode.rst
freakboy3742: Unfortunately, it's *just* different enough to make this impossible - the command to invoke packaging is different.

src/briefcase/platforms/macOS/__init__.py
freakboy3742: That approach won't work if there's more than one user.

freakboy3742:
> If it would help, I could extract the UUID piece into a standalone PR; it's nicely self contained, and would then allow running the Visual Studio template against a mainline Briefcase.

I've opened #1850 to include just the UUID piece; I've modified the landing instructions to reflect the extra step of landing that PR.

docs/reference/platforms/macOS/xcode.rst
mhsmith: In that case, the same updates should be applied to this file.

src/briefcase/platforms/macOS/__init__.py
mhsmith: The Anthropic statement of work says we would install to `/Applications`. That seems to make more sense, because regardless of whether it's a console app or not, it is still an application and not a library.

If this changes, the postinstall script will need to be updated to match.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -765,93 +863,234 @@ def package_app(
...
+                dist_path.with_suffix(""),
+                format="zip",
+                root_dir=self.binary_path(app).parent,
+                base_dir=self.binary_path(app).name,
+            )
+
+    def package_pkg(
+        self,
+        app: AppConfig,
+        notarize_app: bool,
```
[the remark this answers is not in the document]
freakboy3742: True - that's a leftover of a feature that I started looking at, then backed out of. 

It is *possible* to sign and notarise installers; however, the signing process requires a separate signing identity to the app distribution signing identity. 

This is a solvable problem - one that we should open as a feature ticket as soon as this PR lands - but it's a little fiddly, and AFAICT, it's not needed to use the installer in practice

My inclination is to leave the argument in, even though it isn't used, because it *should* be used; but I won't argue too hard if you think it should be removed.

src/briefcase/platforms/macOS/__init__.py
freakboy3742: I picked the `/Library` location because macOS will surface any `.app` that exist in `/Applications` or `/Users/xxx/Applications` in the Launchpad, and while you would be able to launch a command line app, it wouldn't display a terminal as a result of being launched, so it would look like a no-op.

Putting console apps in `/Library` seemed broadly consistent with how tools like Xcode handle console-only tooling. It's a location that is clearly a "system" location; but doesn't result in command line apps surfacing in GUI representations.

It's also worth noting that the install location varies depending on the type of app. A GUI app will install into `/Applications`; a console app installs into `/Library`.

mhsmith: From in-person discussion, the reason to use /Library is that anything in /Applications will appear in the Launchpad menu in the dock, but a command-line tool can't be launched that way.

It looks like this also applies to subdirectories of /Applications – for example, the official Python installer creates a directory `/Applications/Python 3.x` containing `IDLE.app` and `Python Launcher.app`, both of which appear in the Launchpad.

src/briefcase/platforms/macOS/__init__.py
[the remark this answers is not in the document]
mhsmith: OK, let's keep it but leave a comment in the code explaining the situation.

To test this, we agreed that you'll sign a pkg and send it to me to make sure I can install it.

src/briefcase/platforms/macOS/__init__.py
mhsmith: OK, that makes sense, but since it's a change to the spec, let's remember to mention to Anthropic next time we send them an update.

src/briefcase/platforms/macOS/__init__.py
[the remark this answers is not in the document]
freakboy3742: I had the sudden realization that I have a son with an M1 laptop (🤦), so I’ve just tested the installer on his machine. It works, but it triggers the “unknown developer” handling. It’s installable, but you’ve got to do the “explicit allow” dance through the Settings panel. So, it would appear that notarisation is a necessary step. Any preferences on rolling that change into this PR, vs a follow up?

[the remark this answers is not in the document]
mhsmith: I've finished testing on all three platforms and found no other problems, so let's start merging now and do it as a follow-up.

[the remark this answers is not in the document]
mhsmith: I've approved the Xcode and VisualStudio PRs.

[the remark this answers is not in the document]
freakboy3742: PRs for the app templates have been updated with new stubs; they're now ready for review.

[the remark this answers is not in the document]
michelcrypt4d4mus: > _So, it would appear that notarisation is a necessary step._

FWIW technically i don't think this is called "notarization"... that's a whole _other_ process with a whole _other_ developer certificate that you do before submitting to the app store. to get the installer to work without "unknown developer" issues (or at least none beyond right click -> install anyway) you can just run `pkgbuild` with the  `--sign "Developer ID Installer: Nasir Jones (XYZXYZXYZ)"`.

my understanding - which is not 100% but i did manage to get a briefcase app into an installer successfully - is that there may be certificate types that can be used for both packaging and notarizing but there are definitely certificate types that will only do one or the other.

Screenshot 2024-06-04 at 1 08 27 AM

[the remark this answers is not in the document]
freakboy3742: Ok - technically *signing* is a required step; Notarization is an extra step that you might as well do once you're signing (especially given that we've got notarization ready to go). 

However, the key detail is that you can't use the "Developer ID Application" certificate to sign (and thus notarize) the `pkg` archive, so additional certificate/identity handling is required.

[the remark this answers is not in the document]
michelcrypt4d4mus: [this issue](https://github.com/electron/notarize/issues/60) has some more of the details on the difference between signing and notarizing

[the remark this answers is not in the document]
michelcrypt4d4mus: > _the key detail is that you can't use the "Developer ID Application" certificate to sign_

correct

> _Notarization is an extra step that you might as well do once you're signing_

it could very well be but tbh i'm not sure... as far as i can tell (and i could be wrong) to get an app notarized requires sending it to apple and getting something back from them. i don't know if there are any kind of limits on how many packages an app developer can sign but i'd be a bit wary of just automatically submitting to apple for notarization every time the package is built (at least not without fully understanding the apple notarization process).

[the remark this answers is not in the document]
freakboy3742: Yes, it involves sending code to Apple and getting a response back. Briefcase already manages this process.

And yes, there is a rate limit. You don't want to put notarisation in a CI process, but there's no issue with having lots of apps, as long as you don't try to notarise them all on the same day.

mhsmith:
With the macOS app template I got this error:

```
[consoletest] Building PKG...
Installing license... errored

Log saved to /Users/msmith/git/beeware/apps/consoletest/logs/briefcase.2024_06_04-21_50_50.package.log

  File "/Users/msmith/.venv/bw-38/bin/briefcase", line 8, in <module>
    sys.exit(main())
  File "/Users/msmith/git/beeware/briefcase/src/briefcase/__main__.py", line 29, in main
    command(**options)
  File "/Users/msmith/git/beeware/briefcase/src/briefcase/commands/package.py", line 151, in __call__
    state = self._package_app(
  File "/Users/msmith/git/beeware/briefcase/src/briefcase/commands/package.py", line 99, in _package_app
    state = self.package_app(app, **full_options(state, options))
  File "/Users/msmith/git/beeware/briefcase/src/briefcase/platforms/macOS/__init__.py", line 878, in package_app
    self.package_pkg(
  File "/Users/msmith/git/beeware/briefcase/src/briefcase/platforms/macOS/__init__.py", line 931, in package_pkg
    (installer_path / "resources").mkdir(exist_ok=True)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/pathlib.py", line 1288, in mkdir
    self._accessor.mkdir(self, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/installer/resources'
(bw-38) msmith@mbp consoletest % ll /Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/
total 16
drwxr-xr-x  3 msmith  staff   96 Jun  4 21:50 Console test.app
-rw-r--r--  1 msmith  staff  334 Jun  4 21:49 Entitlements.plist
drwxr-xr-x  2 msmith  staff   64 Jun  4 21:50 app_packages.x86_64
-rw-r--r--  1 msmith  staff  563 Jun  4 21:49 briefcase.toml
drwxr-xr-x  6 msmith  staff  192 Jun  4 21:50 support
```

I added `installer_path.mkdir(exist_ok=True)`, but then I got this:

```
[consoletest] Building PKG...
Installing license... done
Copying app into products folder... done
Writing component manifest... done

>>> Running Command:
>>>     pkgbuild --root /Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/installer/root --component-plist /Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/installer/components.plist --install-location '/Library/Console test' --scripts /Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/installer/scripts /Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/installer/packages/consoletest.pkg
>>> Working Directory:
>>>     /Users/msmith/git/beeware/apps/consoletest
pkgbuild: error: Cannot write package to "/Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/installer/packages/consoletest.pkg". (The file “scripts” couldn’t be opened.)
pkgbuild: Reading components from /Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/installer/components.plist
pkgbuild: Adding component at Console test.app
>>> Return code: 1
Building app package... done

>>> Running Command:
>>>     productbuild --distribution /Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/installer/Distribution.xml --package-path /Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/installer/packages --resources /Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/installer/resources '/Users/msmith/git/beeware/apps/consoletest/dist/Console test-0.0.1.pkg'
>>> Working Directory:
>>>     /Users/msmith/git/beeware/apps/consoletest
productbuild: error: Specified distribution "/Users/msmith/git/beeware/apps/consoletest/build/consoletest/macos/app/installer/Distribution.xml" not found.
>>> Return code: 1
Building Console test-0.0.1.pkg... done

[consoletest] Packaged dist/Console test-0.0.1.pkg
(bw-38) msmith@mbp consoletest % echo $?
0
```

There are at least two problems here:
* None of the things in the `installer` directory of the Xcode template are present in the app template, including resources/welcome.html, scripts/postinstall, and Distribution.xml.
* Two consecutive commands failed, but Briefcase continued running and returned a successful exit code.

mhsmith:
With the Windows app template, `briefcase run` and `briefcase package` worked fine, but after running the installer:

* It created a Start menu shortcut to "C:\Program Files\Jane Developer\Console test\Console test.exe", but the file is actually called consoletest.exe. Unlike on macOS, this shortcut is not useless because it would open a terminal window, so it's probably worth fixing it rather than removing it.

* The PATH is not updated, because the .wxs file updates aren't present in the app template.

freakboy3742:
> * None of the things in the `installer` directory of the Xcode template are present in the app template, including resources/welcome.html, scripts/postinstall, and Distribution.xml.

Apologies - I had those modifications locally, but managed to not commit them. They're in the template now.

> * Two consecutive commands failed, but Briefcase continued running and returned a successful exit code.

Looks like I forgot to add `check=True` to two the `pkgbuild`  and `productbuild` command invocations. Fixed.

> With the Windows app template, `briefcase run` and `briefcase package` worked fine, but after running the installer:
> 
> * It created a Start menu shortcut to "C:\Program Files\Jane Developer\Console test\Console test.exe", but the file is actually called consoletest.exe. Unlike on macOS, this shortcut is not useless because it would open a terminal window, so it's probably worth fixing it rather than removing it.

Agreed - I've added this to the app template, and opened beeware/briefcase-windows-VisualStudio-template#32 to add the same change to the VS project.

> * The PATH is not updated, because the .wxs file updates aren't present in the app template.

Fixed.

mhsmith (approved):

PR #1784, opened 2024-05-08
Update android reference docs

JaxkDev:
Small typo as found in discord.

## PR Checklist:


mhsmith:
Thanks, but I think the correct fix here is to add spaces to the `.. code-block` rather than removing them from the following text, because all of it should be indented within the "JSON dictionary" bullet point.

PR #1807, opened 2024-05-20
Added docs for externally supported entry points

oddbookworm:
Documenting external projects that have Briefcase entry points

Docs don't exist yet

N/A

## PR Checklist:


src/briefcase/commands/new.py
```diff
@@ -378,6 +378,7 @@ def select_option(
         default: str | None,
         options: Sequence[str],
         override_value: str | None,
+        external_hooks: str | None = None,
```
freakboy3742, line 381: This doesn't need to be a configurable argument. An external hook is only needed for one specific template selection, so it would be better to put the extra content directly into the intro prompt.

oddbookworm: I pushed a commit to remove this argument and put it directly into the intro. Let me know what you think of the new version

freakboy3742 (approved):
I've done a bit of editing on the specific text (including a major addition around the plugin interface itself), but the original PR was a great starting point. Thanks for the PR!

Issue #1235, opened 2023-04-25
Deprecate Briefcase config keys that conflict with PEP621 representations.


freakboy3742:
### What is the problem or limitation you are having?

#1219 introduced support for using PEP621 metadata keys. However, no changes were made to Briefcase's own representations; the PEP621 keys were honoured, but the names weren't used.

In addition, some of the keys in PEP621 allow a much richer specification that Briefcase allows. Briefcase should allow at least as much breath of representation as PEP621 for these keys. 

### Describe the solution you'd like

* The singular `author` and `author_email` keys should be deprecated in favour of a list of `authors`
* The singular `url` should be deprecated in favor of `urls`, with the `Homepage` subway filling the role of the current `url`.
* The handling of LICENSE should be expanded to allow for `text` and `file` subkeys (and the filename should be used in Linux system backends instead of hardcoding the name of the file.

Optional changes worth considering:
* Adding a `changelog` key analogous to `license`, so that Linux can specify a file that is the changelog
* Normalising from `requires` to `dependencies`
* Normalising `test_requires`. There's no obvious mapping here; the simple approach would be just using `test_dependencies` as a name; or we could adopt some form of `optional-dependencies` on a per-backend config basis.

### Describe alternatives you've considered

Do nothing :-)

mhsmith:
The license metadata of PEP 621 has now been replaced by PEP 639. The [packaging user guide](https://packaging.python.org/en/latest/specifications/pyproject-toml) contains the current recommendation.

freakboy3742:
#2146 is tracking the license metadata issue; we can solve that without necessarily solving the rest of PEP621 compliance.

mhsmith:
Is this issue proposing that we replace the `tool.briefcase` section in the [briefcase-template](https://github.com/beeware/briefcase-template/blob/main/%7B%7B%20cookiecutter.app_name%20%7D%7D/pyproject.toml) with a `project` section? Or is it proposing that we keep the `tool.briefcase` section, but redefine the meaning of some of its keys? 

Replacement seems better, since in the long term it would save us the work of maintaining two different ways of giving the same information.

freakboy3742:
[...]

PR #1812, opened 2024-05-20
Support specifying the license file in the pyproject.toml file

MarieRoald:
Briefcase can now read the `project.license.file`-key in the `pyproject.toml` file. Without this PR, the license file name is hard-coded to "LICENSE". This is relevant for linux builds, where the license file is included in the built package. With this PR, it still defaults to "LICENSE", but if `project.license.file` is set in `pyproject.toml`, then that file is used instead.

This is related to the LICENSE file handling in https://github.com/beeware/briefcase/issues/1235

## PR Checklist:


freakboy3742 (changes requested):
Summary of in-person discussion: This is great; however, we probably want to:
* Preserve the PEP621 structure of the `license` field as a dictionary
* If the `file` sub-field is used, use that as the file name (rather than a separate `license_file` field)
* If the `text` sub-field is used, output that text as a `LICENSE` file - possibly with a warning if the text doesn't appear to be a full license text
* If `license` is a bare string (which will be the legacy behavior from older Briefcase projects), issue a warning and convert it to `license = {"file": "LICENSE"}`, since we know the license will be a file.

freakboy3742 (approved):
I've tweaked the error and warning text, so we're good to go! Thanks for the contribution

PR #1816, opened 2024-05-20
Document that users should change default Android cert password

MaggieFero:
Update docs to clarify that SHA-1 is now breakable enough that the Android password matters

Updated android.rst to clarify that recent improvements to intentional SHA-1 collision development bringing the cost of an attack under USD$10k mean that you reallllly actually should change the password from default, but in a hopefully-user-friendly FUD-free way. I updated the existing description and added an admonition to emphasize that you really ought to update it.

Previously the docs said that you didn't need to change the password. 

## PR Checklist:

- [ N/A ] All new features have been documented


MaggieFero:
Note that Claire reviewed the draft documentation here for me in person, but I forgot to get her GitHub.

freakboy3742 (approved):
Thanks for the fix!

docs/how-to/code-signing/android.rst
```diff
@@ -77,8 +77,13 @@ used as an upload key for the Google Play store, we call the key "upload-key".
...
 Android keystores <https://developers.google.com/android/guides/client-auth>`__.
-You can change the password if you like. It is more important to limit who
-has access to the keystore file than to change the password.
+You should change the password. It is most critical to limit who
+has access to the keystore file, but changing the password is also important.
+
+.. admonition:: Android Keystore Password
+
+  Previously, it has been considered safe to use the default password. However,
+  recent advancements in SHA-1 attacks make that inadvisable.
```
mhsmith, line 86: If an attacker has the file and it uses the default password, then they can already access the key; they don't need to break any hashes. If an attacker doesn't have the file, then surely the password doesn't matter. So what's the significance of SHA-1 here?

MaggieFero: If the attacker doesn't have the file, the password enables them to get the SHA-1 fingerprint of the file. As of 2020 in a research environment SHA-1 fingerprint collisions could be generated accurately for $11k, and since then the decreases in relevant hardware prices make it even cheaper. 

I'm hoping to hint which direction somebody should go in to learn more without trying to make *everybody* catch up on it. If you've got a better idea for a hint (or if this project prefers not to leave hints), I'm very open to suggestions! This is my first contribution for this project 😅

MaggieFero: Noting here that we chatted about this in person and then I did some extra research, and we're pretty sure this attack still requires access to the keystore file. It's still important to update the password, but maybe not bad enough to warrant the admonition, since (in its current form) it's likely to confuse people who aren't super into this specific topic. 

Accordingly, I'll remove the admonition but keep the revised wording earlier on.

MaggieFero:
...I tagged the wrong person in that commit message, and it was actually @mhsmith I was chatting with in person. Unfortunately, it was such a small change I made it in the GitHub web UI and I'm not sure amending the message is worth a force push?

mhsmith (approved):
Thanks!

PR #1822, opened 2024-05-21
Add eol python warnings

### LINKED ISSUE #1669: Add EOL Python warnings

This idea is a bit half-baked but I'm thinking of adding warnings for users running Briefcase with an EOL Python.

When users are running an EOL Python:
- they may not know it's EOL since they're just using whatever the `python/python3` command finds
- they will likely be using outdated versions of Briefcase and Toga

On several occasions after a user explains something weird happening, I notice they are running Python 3.7 and using a newer Python fixes the problem.

I guess if we really wanted to, we could replace wheels on PyPI...but I'm going to focus on future releases.

We already have (at least rough) [EOL dates](https://devguide.python.org/versions/) for Python versions. So, we can include these and check against the current date when Briefcase runs and warn the user (or more).

This might become a little problematic around newer Python versions. Python 3.13 is scheduled for EOL around 2029/10 but there is no such schedule for 3.14, 3.15, etc....unless you extrapolate the current scheduling. Therefore, it may be difficult to warn someone in the year 2035 that Python 3.14 is EOL and they shouldn't be using Briefcase v0.3.18. Buuuut....I also think excluding such a pathological case from this warning is ok 😆 

Keep trying to let people know they are using old Python versions.

### Additional context

_No response_

kaleJohn:
Added a validation function for the system's python version, warning if it is past it's end of life date, based on the dates from here: https://devguide.python.org/versions/.

Fixes #1669 

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -10,6 +10,7 @@ import subprocess
 import sys
 from abc import ABC, abstractmethod
 from argparse import RawDescriptionHelpFormatter
+from datetime import date
 from pathlib import Path
 from typing import Any
 
@@ -178,12 +179,12 @@ class BaseCommand(ABC):
         self.data_path = self.validate_data_path(data_path)
         self.apps = {} if apps is None else apps
         self.is_clone = is_clone
-
         self.tools = tools or ToolCache(
             logger=logger,
             console=console,
             base_path=self.data_path / "tools",
         )
+        self.validate_python_version()
 
         # Immediately add tools that must be always available
         Subprocess.verify(tools=self.tools)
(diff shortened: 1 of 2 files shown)


src/briefcase/commands/base.py
```diff
@@ -305,6 +306,20 @@ def validate_locale(self):
...
             )
 
+    def validate_python_version(self):
+        """Validate the system's python version is not past its EOL date."""
+        python_version = sys.version_info
+        today = date.today()
+        EOL_year = today.year - 2024  # increment this each year past 10-2024
+        if today.month > 10:
+            EOL_year += 1
+        if python_version[0] < 3 or (
```
freakboy3742, line 316: version < 3 can't ever happen; Briefcase can't be installed on Python 2.

src/briefcase/commands/base.py
```diff
@@ -305,6 +306,20 @@ def validate_locale(self):
 """
             )
 
+    def validate_python_version(self):
+        """Validate the system's python version is not past its EOL date."""
+        python_version = sys.version_info
+        today = date.today()
+        EOL_year = today.year - 2024  # increment this each year past 10-2024
```
freakboy3742, line 313: A short description of the math being used here would be helpful - why is 2024 important? Should this number be updated every year? Also, in any description, it's more helpful to say "October" than "10" because it's clear we're referring to a month.

freakboy3742 (changes requested):
Thanks for the PR! I've dropped a couple of minor suggestions inline (mostly comments so that it's obvious to future code explorers what is going on). 

You'll also note that the tests are currently failing. That's because of a drop of coverage - there's no code testing the 2 if statements that have been added. To test these, you'll need to mock the behavior of sys.platform and date so that you can test (for example) what would happen if you ran this branch of code in January 2040.

freakboy3742:
Thanks for the contribution - the missing pieces have been added by #2305.

Issue #1472, opened 2023-10-02
Print notarization log on failure


wiktor-k:
### What is the problem or limitation you are having?

When notarization fails it's not clear what went wrong.

### Describe the solution you'd like

Automatically fetch the notarization log (e.g. `xcrun notarytool log 7d9249fb-f9ac-4bb1-b296-256ebabff66b --apple-id ... --team-id ...`) and display to the user.

### Describe alternatives you've considered

Display a help message on notarization error printing the command the the user can run to learn more.

### Additional context

Just for the record a sample log:

```sh
xcrun notarytool log 7d9249fb-f9ac-4bb1-b296-256ebabff66b --apple-id ... --team-id ...
```

```json
{
  "logFormatVersion": 1,
  "jobId": "7d9249fb-f9ac-4bb1-b296-256ebabff66b",
  "status": "Invalid",
  "statusSummary": "Archive contains critical validation errors",
  "statusCode": 4000,
  "archiveFilename": "pks-2.0.0.dmg",
  "uploadDate": "2023-09-29T07:40:29.653Z",
  "sha256": "40125eba88640082a9f1098d9bb133cc584cf4c59c44e25c8e0e1431f9dd2eac",
  "ticketContents": null,
  "issues": [
    {
      "severity": "error",
      "code": null,
      "path": "pks-2.0.0.dmg/pks.app/Contents/Resources/app_packages/PySide6/Qt/lib/QtWebEngineCore.framework/Versions/A/QtWebEngineCore",
      "message": "The signature of the binary is invalid.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087735",
      "architecture": "x86_64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "pks-2.0.0.dmg/pks.app/Contents/Resources/app_packages/PySide6/Qt/lib/QtWebEngineCore.framework/Versions/A/QtWebEngineCore",
      "message": "The signature of the binary is invalid.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087735",
      "architecture": "arm64"
    }
  ]
}
```

[...]

PR #1825, opened 2024-05-21
ISSUE-1472: Added xcrun debugging tips to error message for failed signing attempts

ethompsy:
This implements a bit of guidance for failed signing attempts on MacOS to smooth the issue described in #1472. Considering that we would need to run `xcrun notarytool history` to first find the `submission-id` of the failed signing attempt before running `xcrun notarytool log <submission-id>` to get the logs, all within the except of a try/except, it seems most sensible to just inform the user of how they could track down the error themselves.

Refs #1472 

## PR Checklist:

### DIFF
--- a/src/briefcase/platforms/macOS/__init__.py
+++ b/src/briefcase/platforms/macOS/__init__.py
@@ -686,10 +686,18 @@ password:
                         store_credentials = True
                     else:
                         raise BriefcaseCommandError(
-                            f"""Unable to submit {filename.relative_to(self.base_path)} for notarization.
-                                To examine this failure use `xcrun notarytool history` to find the submission-id,
-                                then use `xcrun notarytool log <submission-id>` for a log of the error.
-                            """
+                            f"""\
+Unable to submit {filename.relative_to(self.base_path)} for notarization.
+To find the cause of this failure, get the submission ID by running:
+
+    xcrun notarytool history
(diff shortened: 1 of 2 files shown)


src/briefcase/platforms/macOS/__init__.py
```diff
@@ -686,7 +686,10 @@ def notarize(self, filename, team_id):
                         store_credentials = True
                     else:
                         raise BriefcaseCommandError(
-                            f"Unable to submit {filename.relative_to(self.base_path)} for notarization."
+                            f"""Unable to submit {filename.relative_to(self.base_path)} for notarization.
+                                To examine this failure use `xcrun notarytool history` to find the submission-id,
+                                then use `xcrun notarytool log <submission-id>` for a log of the error.
+                            """
```
freakboy3742, line 692:
```suggestion
                            f"""\
Unable to submit {filename.relative_to(self.base_path)} for notarization.
To find the cause of this failure, get the submission ID by running:

    xcrun notarytool history
    
Then run: 

    xcrun notarytool log <submission-id>
    
to generate a full log of the error.
"""
```

freakboy3742 (approved):
I think it would still be worth doing the extra work to automate this if we can; but as an immediate workaround, providing the manual commands is a good idea.

PR #1839, opened 2024-05-26
Ensure all files are unique for check-wheel-contents

rmartin16:
## Changes
- [hynek/build-and-inspect-python-package](https://github.com/hynek/build-and-inspect-python-package) runs [`check-wheel-contents`](https://pypi.org/project/check-wheel-contents/) and it requires the wheel contains zero duplicated files
- RE: https://github.com/beeware/.github/pull/137

## PR Checklist:

### DIFF
--- a/changes/1839.misc.rst
+++ b/changes/1839.misc.rst
@@ -1,0 +1,1 @@
+Files with non-unique content were made unique so the packaged wheel passes ``check-wheel-contents``.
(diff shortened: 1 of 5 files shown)


freakboy3742 (approved):
I'm intrigued to know what this check is actually be protecting against... but no objection to complying.

rmartin16:
fwiw, their documentation says (1) "you copied a file when you meant to rename it" or (2) "you built a wheel, renamed a file, and re-created the wheel without deleting the `build` directory first".

Number (2) sounds like a historical problem....but number (1) seems reasonable I guess. I'd argue, though, it is late in the process to be catching this when you're building a wheel but...

PR #1849, opened 2024-05-30
Manage stub binaries as external resources

### LINKED ISSUE #933: Application Template Generation Seems Frozen for a Large Template git Repository with Slow Internet Connection
**Describe the bug**
Running `briefcase create` for a large template git repository over a slow connection feels frozen.

For some reason, the `briefcase-macOS-app-template` repo is 165MB.

For completeness:
```
~/github/beeware 
❯ du --max-depth 1 -h | grep -i template | sort -h
212K	./Python-Linux-template
236K	./Python-Django-template
240K	./Python-OSX-template
288K	./Python-Windows-template
360K	./briefcase-windows-msi-template
668K	./Python-Android-template
892K	./Python-macOS-template
1.2M	./briefcase-web-static-template
1.4M	./briefcase-linux-flatpak-template
1.4M	./Python-tvOS-template
1.5M	./briefcase-template
1.6M	./podium-template
2.0M	./Python-iOS-template
2.8M	./briefcase-windows-VisualStudio-template
3.0M	./briefcase-linux-appimage-template
3.4M	./briefcase-macOS-Xcode-template
3.5M	./briefcase-windows-app-template
4.0M	./briefcase-iOS-Xcode-template
5.1M	./briefcase-android-gradle-template
165M	./briefcase-macOS-app-template
```

**To Reproduce**
Run `briefcase create macOS app` with a slow internet connection….or at least one where GitHub is seemingly throttling the connection.

**Expected behavior**
Long operations should provide the user with feedback that activity is happening.

Alternatively, ensure the recommended template repos aren't bloated.

**Environment:**
 - Operating System: macOS 12
 - Python version: 3.10.7
 - Software versions: 
   - Briefcase: `0.3.12.dev8+g2435b9b9`

freakboy3742:
The macOS and Windows app templates include a stub binary; however these binaries are currently managed as artefacts in the macOS/Windows app template. As the template is git repository, it contains the full history of all changes to the template; as a result, on first use, the user must download *all* historical stub binaries as part of the template. 

This change modifies the handling of stub binaries so that they are external artefacts, analogous to the support package. They will be stored on S3, same as the support packages, and cached on use; however, as they're managed as external resources, the base template can be made a lot smaller.

This change by itself won't reduce the size of the template checkout, as cookiecutter does a full repo clone. #1855 adds a modification that does a blobless clone, which will reduce the size of the initial repository clone.

This change requires a template update to take full effect; but, conveniently, CI will pass without any changes, because the stub binary that is unpacked from a downloaded archive has the same name as the stub binary that is generated by the current app template, and the current app template doesn't have the marker that indicates a stub needs to be downloaded.

Refs beeware/briefcase-macOS-app-template#58
Refs beeware/briefcase-windows-app-template#51

Fixes #933.

## PR Checklist:


mhsmith (approved):

mhsmith:
As we discussed yesterday, the way this will work is:

* Update the `VisualStudio` or `Xcode` repository.
* Apply a tag in the `app` repository.
* This triggers a GitHub Action which builds an app from the `VisualStudio` or `Xcode` repository, and uploads the stub executables to S3 with a build number from the tag.
* Test the executables manually using the `stub_binary` setting.
* Update briefcase.toml in the `app` repository with the build number from the tag.

This means that in the `app` repository, the tag for each build will always be on the commit *before* that build was adopted, e.g. the tag for build 6 will always be on a commit that says `stub_binary_revision = 5`.

Also, there's no direct link between the stubs and the source code they were built from. You'd have to work it out from the tag timestamps, which I think are only stored on GitHub and not in the Git repository itself.

These could both be fixed if the tags and the actions were moved to the `VisualStudio` or `Xcode` repository instead.

freakboy3742:
> These could both be fixed if the tags and the actions were moved to the `VisualStudio` or `Xcode` repository instead.

I thought about that; I wasn't sure if it was odd to have something that only the app template uses being controlled by a different repo. However, your point about having a tag on the exact code being used is a good one that I hadn't considered. I'll put up PRs to migrate the binary tagging over to the Xcode/VS template repo.

PR #1851, opened 2024-05-30
Update warning for pyproject.toml license spec

rmartin16:
## Changes
- Rework a bit the license warning from https://github.com/beeware/briefcase/pull/1812 to include additional information
- This change also drops a warning for each time the `license` field is specified in `pyproject.toml`

## Notes
- The presentation of the original warning was a little jarring to me since it spread across my entire terminal. Additionally, though, it wasn't using the "WARNING" banner typically reserved for messages like this.
- Moreover, I didn't really understand the change I was being asked to make.

## PR Checklist:

### DIFF
--- a/changes/1851.misc.rst
+++ b/changes/1851.misc.rst
@@ -1,0 +1,1 @@
+The warning for specifying the license in pyproject.toml was updated to include additional information.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,6 +102,7 @@ dev = [
     "pre-commit == 3.5.0 ; python_version < '3.9'",
     "pre-commit == 3.7.1 ; python_version >= '3.9'",
     "pytest == 8.2.1",
+    "pytest-xdist == 3.5.0",
     "setuptools_scm == 8.1.0",
     "tox == 4.15.0",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,8 @@ dev = [
     "pre-commit == 3.5.0 ; python_version < '3.9'",
     "pre-commit == 3.7.1 ; python_version >= '3.9'",
     "pytest == 8.2.1",
-    "pytest-xdist == 3.5.0",
+    # Having xdist in the pytest environment causes some weird failures on Windows with Py3.13
+    "pytest-xdist == 3.6.1 ; python_version < '3.13'",
     "setuptools_scm == 8.1.0",
     "tox == 4.15.0",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,6 @@ dev = [
     "pre-commit == 3.5.0 ; python_version < '3.9'",
     "pre-commit == 3.7.1 ; python_version >= '3.9'",
     "pytest == 8.2.1",
-    "pytest-xdist == 3.6.1",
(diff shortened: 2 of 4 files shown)


freakboy3742 (approved):
Looks good to me; the Win-3.13 failure is a weird one though. I've seen a that pop up recently, but it's always been transient...

freakboy3742:
The Windows failure definitely isn't transient; it's persisted through multiple re-runs, include full reruns after a rebase.

In a desparate search for anything that could be causing this, I noticed that pytest-xdist is installed; I removed it, and the problem went away. I don't know *why* xdist would be the problem, other than the fact that multi-threading a test suite could cause some interesting cleanup side effects.

Tagging @rmartin16 in case you have strong opinions to the contrary; my inclination is to drop xdist for the test suite if it's going to be more stable.

rmartin16:
> my inclination is to drop xdist for the test suite if it's going to be more stable.

Would `tox -e py-fast` still work?

Moreover, though, what if you roll back the recent version bump for pytest-xdist? They reworked how process forking works in the latest version.

freakboy3742:
> > my inclination is to drop xdist for the test suite if it's going to be more stable.
> 
> Would `tox -e py-fast` still work?

I guess not - so there'd need to be other cleanups.
 
> Moreover, though, what if you roll back the recent version bump for pytest-xdist? They reworked how process forking works in the latest version.

Well that'd do it... I'll try and see.

freakboy3742:
> > Moreover, though, what if you roll back the recent version bump for pytest-xdist? They reworked how process forking works in the latest version.
> 
> Well that'd do it... I'll try and see.

Looks like it doesn't - 3.5.0 generates the same error.

So - my inclination is to drop xdist. While parallel support is nice to have, it doesn't yield *that* much of a performance benefit; plus, dropping the extra dependency is one less cause of error, and one less dependency to keep up to date. Thoughts?

rmartin16:
> > > Moreover, though, what if you roll back the recent version bump for pytest-xdist? They reworked how process forking works in the latest version.
> > 
> > 
> > Well that'd do it... I'll try and see.
> 
> Looks like it doesn't - 3.5.0 generates the same error.

well, my guess is `coverage` is doing something weird....at least, if my previous investigation showing this is unclosed file is coming from `coverage` is correct.

> So - my inclination is to drop xdist. While parallel support is nice to have, it doesn't yield _that_ much of a performance benefit; plus, dropping the extra dependency is one less cause of error, and one less dependency to keep up to date. Thoughts?

I do normally use `tox -e py-fast`...since it is about 3x faster (although, `tox -e py` only takes 18s for me...).

One more plea perhaps...what if we just don't install `pytest-xdist` for 3.13? And if you problem still exists when 3.13 is released, we drop it completely then.

freakboy3742:
> One more plea perhaps...what if we just don't install `pytest-xdist` for 3.13? And if the problem still exists when 3.13 is released, we drop it completely then.

I can live with that for now. It's also possible that this might be fixed in 3.13-b2, or with a coverage bump at some point; if we get close to 3.13-final and this is still lingering, we can re-evaluate then.

freakboy3742:
@rmartin16 Looks like the py3.13 exclusion on xdist has worked; If you're happy, feel free to merge.

PR #1855, opened 2024-06-04
Perform a shallow clone of templates on first use

freakboy3742:
Cookiecutter's default behavior is to perform a full clone of any repository passed as a URL. This is proving problematic for the macOS and Windows app templates, because the full history contains a lot of old binaries. Unfortunately, cookiecutter doesn't appear to have any setting to perform a shallow clone. However, it *can* operate on a local file path, and we already have tools to manage branches and remotes in place. 

This PR makes 3 changes:
* It moves all git repo handling to be handled by Briefcase (via GitPython), rather than via cookiecutter. This means than cookiecutter is only ever invoked with a path to a local filesystem checkout
* The initial checkout of a template repository is a shallow clone
* The template cache directory is now managed as a Briefcase cache, rather than using the `~/.cookiecutters` directory.

This is a partial fix for #933; as it reduces the size of the template clone that is required. The full fix is #1849, as that will only download the stub for the Python version that is in use.

Refs #933.

## PR Checklist:


src/briefcase/commands/base.py
```diff
@@ -876,16 +871,39 @@ def update_cookiecutter_cache(self, template: str, branch="master"):
...
+            if cached_template.exists():
+                # There is a pre-existing cache of the template. Attempt to update it;
+                # if the fetch fails, use the existing state of the cache. Any other
+                # failure is surfaced to the user.
+                try:
+                    repo = self.tools.git.Repo(cached_template)
+                except self.tools.git.exc.GitError as e:
+                    raise BriefcaseCommandError(
+                        f"Template repository is in a weird state. Delete {cached_template} and retry."
+                    ) from e
```
rmartin16, line 889: I think that if Briefcase's own cache for a template is corrupted, it should just delete it for the user instead asking the user to delete a directory deep in to hidden file system structures.

rmartin16 (commented):
I like this idea but I'm not sure I understand how it's going to work currently.

If we do a shallow clone of the repo, then all Briefcase can know about the repo is the last commit of the default branch. Therefore, it will not be able to checkout the necessary branch for the version of Briefcase....not to mention an arbitrary branch from the CLI.

For instance, trying to use `v0.3.18` of the template:
```
❯ rm -rf build/ ~/.cache/briefcase/templates/briefcase-linux-system-template

❯ briefcase create -vvv --config='template_branch="v0.3.18"'

[helloworld] Finalizing application configuration...
Targeting pop:jammy (Vendor base debian)
Determining glibc version... done
Targeting glibc 2.35
Targeting Python3.10

Using app template: https://github.com/beeware/briefcase-linux-system-template.git, branch v0.3.18
Cloning template 'https://github.com/beeware/briefcase-linux-system-template.git'...
DEBUG git.cmd: Popen(['git', 'clone', '-v', '--depth=1', '--', 'https://github.com/beeware/briefcase-linux-system-template.git', '/home/russell/.cache/briefcase/templates/briefcase-linux-system-template'], cwd=/home/russell/tmp/beeware/helloworld, stdin=None, shell=False, universal_newlines=True)
DEBUG git.repo.base: Cmd(['git', 'clone', '-v', '--depth=1', '--', 'https://github.com/beeware/briefcase-linux-system-template.git', '/home/russell/.cache/briefcase/templates/briefcase-linux-system-template'])'s unused stdout:
DEBUG git.cmd: Popen(['git', 'remote', 'set-url', '--', 'origin', 'https://github.com/beeware/briefcase-linux-system-template.git', 'https://github.com/beeware/briefcase-linux-system-template.git'], cwd=/home/russell/.cache/briefcase/templates/briefcase-linux-system-template, stdin=None, shell=False, universal_newlines=False)
DEBUG git.cmd: Popen(['git', 'fetch', '-v', '--', 'origin'], cwd=/home/russell/.cache/briefcase/templates/briefcase-linux-system-template, stdin=None, shell=False, universal_newlines=False)

Could not find a template branch for Briefcase v0.3.18.
```

I don't have a recommended implementation but this [GitHub blogpost](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/) looks insightful.

rmartin16:
A separate comment that may warrant independent effort.

While this strategy gets us away from depending on the `~/.cookiecutter` directory, the `~/.cookiecutter_replay` directory is still utilized. While we don't use replay functionality and I don't see a way to turn it off, we may want to consider using a custom location to avoid this noise in the user's home directory.

The only configuration I can find for this is via a file in the file system; for instance:
```
❯ cat ~/.cache/briefcase/templates/cookiecutter.yml 
replay_dir: "/home/russell/.cache/briefcase/templates/cookiecutter-replay"
```
This file can be passed to `cookiecutter()` via `config_file`.

src/briefcase/commands/base.py
freakboy3742: That's a good idea - I'll add it in.

freakboy3742:
> I like this idea but I'm not sure I understand how it's going to work currently.
> 
> If we do a shallow clone of the repo, then all Briefcase can know about the repo is the last commit of the default branch. Therefore, it will not be able to checkout the necessary branch for the version of Briefcase....or an arbitrary branch from the CLI.

Dang... I was afraid I might have missed something in my testing. Back to the drawing board, I guess...

mhsmith:
> If we do a shallow clone of the repo, then all Briefcase can know about the repo is the last commit of the default branch. Therefore, it will not be able to checkout the necessary branch for the version of Briefcase....or an arbitrary branch from the CLI.

In a shallow clone, I think that could be done with a `git fetch` command specifically asking for the version branch, and maybe a `--depth` argument to the fetch command as well.

But if we're about to remove the binaries from the templates anyway, it might not be worth the effort of getting a shallow clone working. When we remove the binaries, we can force-push the main branch to a commit which either has no history, or a rewritten history with the binaries omitted. After that, as long as we're careful not to add anything large to the templates, a full clone should be no problem.

mhsmith:
We also noted just now that `git clone` has `--branch` and `--single-branch` options which are independent of the shallow setting.

freakboy3742:
> In a shallow clone, I think that could be done with a `git fetch` command specifically asking for the version branch, and maybe a `--depth` argument to the fetch command as well.
 ...
> But if we're about to remove the binaries from the templates anyway, it might not be worth the effort of getting a shallow clone working. When we remove the binaries, we can force-push the main branch to a commit which either has no history, or a rewritten history with the binaries omitted. After that, as long as we're careful not to add anything large to the templates, a full clone should be no problem.

Even if we do this, the full repo history will include the blobs for all the historical branches - so the *repo* will still be ~500MB, even though only 1MB is actually checked out on any recent branches. 

Another option may be a blobless-clone - this would download all the hashes (so there's full branch history), but only download the blobs when they're needed. I'll investigate further.

src/briefcase/commands/base.py
freakboy3742: Actually - on second thought; I'm not sure deleting is a good idea. It might make sense to delete if we could "delete and retry"; but that's not going to be simple to do here (possible, but not trivial). In the case that we *do* have a corrupt repo, it might be helpful to investigate *why* it is corrupt, which we won't be able to do if it's deleted as soon as it's found to be corrupt.

freakboy3742:
Looks like replacing `depth=1` with `filter=[blob:none]` does the job without any additional handling. Using the current state of the repos, an initial macOS template clone is 59 MB (236MB without blob filtering). Switching to the v0.3.15 template increases the repo size to 61MB, with the extra blob content being downloaded when the branch is checked out. 

I've also explicitly disabled replay usage on the cookiecutter calls; that should disable the use of the replay folder, without the need to write a configuration file.

src/briefcase/commands/base.py
```diff
@@ -957,20 +966,22 @@ def _generate_template(self, template, branch, output_path, extra_context):
...
             # Unroll the template. Use a copy of the context to ensure that any
-            # mocked calls have an unmodified copy.
+            # mocked calls have an unmodified copy. Disable the use of replay.
             self.tools.cookiecutter(
                 str(cached_template),
                 no_input=True,
                 output_dir=str(output_path),
                 checkout=branch,
                 extra_context=extra_context.copy(),
+                replay=False,
```
rmartin16, line 984: I'll still need to look more closely at the repo mgmt....but I can say that this will not disable capturing replay data, unfortunately. The `replay` argument only has affect when truthy and that's to perform a "replay"....the replay data is still captured in `~/.cookiecutter_replay`, irrespective.

freakboy3742: Sigh... oh well, a full configuration file it is then.

freakboy3742: That definitely sounds preferable; better still would be passing in the parsed configuration as a dict (so we can pass in an explicit replay path without needing to "fake" a YAML file).

On that basis, I'll leave this as is for now (disabling replay is still an improvement), and update the release note to only talk about the `.cookiecutters` folder.

rmartin16:
> Looks like replacing `depth=1` with `filter=[blob:none]` does the job without any additional handling. Using the current state of the repos, an initial macOS template clone is 59 MB (236MB without blob filtering). Switching to the v0.3.15 template increases the repo size to 61MB, with the extra blob content being downloaded when the branch is checked out

So, since we're still requiring a non-trivial amount of data to be downloaded _and_ we've taken over the Git clone process from Cookiecutter, do we want to consider implementing a progress bar for GitPython? There's an example [here](https://stackoverflow.com/a/71285627/8549606). Before I read your comment fully that there's still 59MB to download....I was surprised when the process seemed to freeze for a minute in my VM.

I'll be able to do more robust testing of this strategy tomorrow.

mhsmith:
I'll leave this to @rmartin16 as he's more familiar with this area.

src/briefcase/commands/base.py
rmartin16: - https://github.com/beeware/briefcase/issues/1857

src/briefcase/commands/base.py
```diff
@@ -876,16 +871,39 @@ def update_cookiecutter_cache(self, template: str, branch="master"):
...
+            else:
+                # This is the first time seeing this template. Perform a shallow clone.
+                try:
+                    self.logger.info(f"Cloning template {template!r}...")
+                    cached_template.mkdir(exist_ok=True, parents=True)
+                    repo = self.tools.git.Repo.clone_from(
+                        url=template,
+                        to_path=cached_template,
+                        filter=["blob:none"],
+                    )
```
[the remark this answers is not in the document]
rmartin16: Of note, I think the fallback to "full clone" happens because I think that Git just sends the `filter=blob:none` to the Git server to see if it will support and respect filter....if the server rejects the filter for whatever reason, then I think the Git client just falls back the default behavior.

src/briefcase/commands/base.py
freakboy3742: That's a good point. In my experience, Git is reasonably resilient once the initial clone is complete, so we only need to deal with cleanup in the the "new repo" case. 

I'll add `KeyboardInterrupt` protection around the initial new repo clone, as well ensuring a clone failure on the *initial* failure does cleanup; if there's a failure past that point, there's definitely something wonky going on, and we probably want to investigate before cleaning up.

freakboy3742: > > Actually - on second thought; I'm not sure deleting is a good idea. It might make sense to delete if we could "delete and retry"; but that's not going to be simple to do here (possible, but not trivial).
> 
> I'm not sure I'm following....can't we just fall in to "first time clone" logic below if `GItPython` falls to detect the template as a repo and we've deleted whatever might be there? Is there something I'm missing?

Huh - I hadn't considered that - if I turn the "if/else" into an "if exists; if not exists", the re-clone can follow the failed update. It's obvious now I look at it. I'll add that change/

src/briefcase/commands/base.py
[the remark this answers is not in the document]
freakboy3742: If the fallback behavior is "full clone", and the version required is 6 years old, I'm comfortable opting into this behavior as the default.

freakboy3742:
> > It's 59MB right now - but once #1849 lands, that should drop dramatically (hopefully to < 4MB, like the Xcode template).
> 
> All right...although, if the plan is to rewrite the currently large repos' histories, then this seems like it will mostly be a service for large third-party repos since the benefits will be marginal for Briefcase defaults....which lends credence towards Malcolm's comment about "why do this at all?". However, we have already done it at this point and the change doesn't seem that risky to me.

To be clear - the current plan is to *not* rewrite the repo histories, as the blobless clone means users won't have any large binary artefacts in their repo as soon as #1849 (and the related template changes) land.

rmartin16 (changes requested):
Since we haven't spent a lot of time in this code, I spent a bit of time nitpicking....but I'm done digging in after these comments :) I expect to approve once you send it back tonight/in the morning.

src/briefcase/commands/base.py
```diff
@@ -887,16 +882,57 @@ def update_cookiecutter_cache(self, template: str, branch="master"):
         if is_repo_url(template):
             # The app template is a repository URL.
             #
-            # When in `no_input=True` mode, cookiecutter deletes and reclones
-            # a template directory, rather than updating the existing repo.
+            # When in `no_input=True` mode, cookiecutter deletes and reclones a template
+            # directory, rather than updating the existing repo.
```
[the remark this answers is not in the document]
freakboy3742: I have a *very* vague recollection that it might have been a *long* time ago, when we were doing a lot more reliance on Cookiecutter to manage repo clones... but it definitely isn't true now. I'll remove the comment.

src/briefcase/commands/base.py
```diff
@@ -907,10 +943,9 @@ def update_cookiecutter_cache(self, template: str, branch="master"):
                     # Attempt to update the repository
                     remote.fetch()
```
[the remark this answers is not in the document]
freakboy3742: Nice catch... but I think it's sufficiently niche that I'm comfortable ignoring it, especially if there's no observable error that we can inspect. AFAICT, we'd need to do the clone, then inspect the clone to make sure it's the branch and origin we expected, which is a lot of headache for an area where the user is already well off the "normal" use case; if they're getting unexpected behavior, I don't think it's unreasonable to read the logs to diagnose what might be happening.

freakboy3742:
I've cleaned up those comments and error messages; #1849 has also been merged, and with the new "stables" repo, a clean macOS app template is 1.3 MB. Ready for final(?) review

src/briefcase/commands/base.py
```diff
@@ -967,17 +1001,16 @@ def update_cookiecutter_cache(self, template: str, branch="master"):
...
                 raise BriefcaseCommandError(
-                    f"Git repository in a weird state, delete {cached_template} and try briefcase create again"
+                    "Unable to check out template branch.\n"
+                    "\n"
+                    "This may be because your computer is offline, or because the template repository\n"
+                    "is in a weird state. If you have a stable network connection, try deleting:\n"
+                    "\n"
+                    f"    {cached_template}\n"
+                    "\n"
+                    "retrying your command."
```
rmartin16, line 1013:
```suggestion
                    "Unable to check out template branch.\n"
                    "\n"
                    "This may be because your computer is offline, or because the template repository\n"
                    "is in a weird state. If you have a stable network connection, try deleting:\n"
                    "\n"
                    f"    {cached_template}\n"
                    "\n"
                    "and retrying your command."
```

rmartin16 (approved):
One small typo fix and we're good to go 👍🏼

PR #1856, opened 2024-06-05
Add signing and notarisation to PKG installers

freakboy3742:
Unsigned macOS Installers are a bit difficult to use, as the OS "unknown developer" handling prevents installation. 

This PR adds signing and notarisation options to ``.pkg`` builds.

This signing requires a different signing identity - you need a "Development ID Installer" identity, rather than a "Development ID Application". To facilitate this, a ``--installer-identity`` option has been added (plus a ``--no-sign-installer`` option to avoid signing entirely). 

Identity selection is a little complicated, because macOS doesn't appear to give a way to *just* list the installer certificates; so - this implementation lists all non-app certificates from the same Team ID as the application certificate.

## PR Checklist:


mhsmith:
I'll have to wait for my Apple developer account to reactivate before I can test this – I paid today, but it apparently takes [24-48 hours](https://forums.developer.apple.com/forums/thread/743261) to take effect.

mhsmith:
Signing works fine, but notarization has now been "in progress" for over an hour. This probably isn't Briefcase's fault, but I'll leave it overnight and see what happens.

freakboy3742:
Yeah - notarisation is known to be a bit flaky. Not sure what we can do to make it less flaky though; we're entirely dependent on Apple's servers returning a response.

mhsmith (approved):

mhsmith:
My first notarization attempt was still "in progress" after 10 hours. I interrupted it and tried again, and it succeeded after an additional 4 hours and 25 minutes.

PR #1826, opened 2024-05-21
Added Files using tenacity for retrying renaming a directory on a opened filepath.

spa51273:
* Issue #1780
* On Windows, renaming a directory that was just created can fail with PermissionError: [WinError 5] Access denied

## PR Checklist:

- [ x] All new features have been tested
- [ ] All new features have been documented
- [ x] I have read the **CONTRIBUTING.md** file
- [ x] I will abide by the code of conduct

PR #1871, opened 2024-06-09
Introduce the `File` tool and tweak stub binary implementation

rmartin16:
## Changes
- Introduce the `File` tool
  - Encapsulates complex file operations such as archive unpacking
  - Incorporates the (now removed) Download tool's functionality
- Some updates to the stub binary error handling

## Notes
- I was making the stub binary updates while rebasing #1714....but it seems more appropriate on its own

## PR Checklist:


src/briefcase/integrations/base.py
```diff
@@ -234,6 +234,22 @@ def system_encoding(self) -> str:
 
         return encoding.upper()
 
+    def unpack_archive_kwargs(self, archive_path: str | os.PathLike) -> dict[str, str]:
```
rmartin16, line 237: Putting this here is less than ideal...

Like we talked about at PyCon, we should create a `File` tool that would encompass #1826 and subsume the `Download` tool. So, if the `File` tool already existed, I would just put this there.

Furthermore, though, I think we should just wrap the whole archive unpacking process instead of calling `shutil.unpack_archive()` throughout Briefcase.

Given #1826 is still in limbo and I don't want to introduce the `File` tool right now....putting it here seems like a reasonable compromise...

src/briefcase/commands/create.py
```diff
@@ -420,19 +411,41 @@ def install_stub_binary(self, app: AppConfig):
...
+            supported_archive_extensions = {
+                ext for format in shutil.get_unpack_formats() for ext in format[1]
+            }
+            stub_path_exts = {
+                # captures extensions like .tar.gz, .tar.bz2, etc.
+                "".join(stub_binary_path.suffixes[-2:]),
+                # as well as .tar, .zip, etc.
+                stub_binary_path.suffix,
+            }
+            is_archive = not stub_path_exts.isdisjoint(supported_archive_extensions)
```
rmartin16, line 425: Similar to the `unpack_archive_kwargs` comment, this would ideally go in to a `File` tool....but in lieu of that existing just yet....it's here...

src/briefcase/integrations/base.py
rmartin16: Also, I gave up trying to add this `kwargs` to the `shutil.unpack_archive()` calls for Android SDK and Java....the tests were really disagreeing....and those archives are also ZIPs.

src/briefcase/exceptions.py
```diff
@@ -156,7 +156,13 @@ def __init__(self, platform):
 class InvalidSupportPackage(BriefcaseCommandError):
     def __init__(self, filename):
         self.filename = filename
-        super().__init__(f"Unable to unpack support package {filename!r}")
+        super().__init__(f"Unable to unpack support package '{filename}'.")
```
rmartin16, line 159: If you just call `repr()` here and `filename` is a `Path`, it looks like this:

```
Unable to unpack support package PosixPath("/asdf")
```
so, this strategy gets what we actually want.
```
Unable to unpack support package '/asdf'
```

rmartin16:
ive nerdsniped myself in to implementing a `File` tool....

src/briefcase/integrations/base.py
rmartin16: Moved in to `File` tool

src/briefcase/commands/create.py
rmartin16: Moved in to `File` tool

rmartin16:
The updates for the `File` tool were relatively extensive but also pretty banal. Reviewing the changes for each commit will help discern what I changed for the stub binary versus the `File` tool.

src/briefcase/exceptions.py
freakboy3742: Torn between whether we should use the explicit `'`, or `{str(filename)!r}` - that should ensure that any inline quotes are correctly escaped (I think...)

rmartin16: You mean like these differences?
```pycon
>>> file_path=Path("/pa't'h/to/file")
>>> 
>>> print(f"File is {file_path}")
File is /pa't'h/to/file
>>> 
>>> print(f"File is {str(file_path)!r}")
File is "/pa't'h/to/file"
>>> 
>>> print(f"File is '{file_path}'")
File is '/pa't'h/to/file'
```
Manually calling `str()` does seem to handle that edge case best.

src/briefcase/integrations/file.py
```diff
@@ -18,21 +20,87 @@
...
+        self.tools.shutil.unpack_archive(
+            filename=filename,
+            extract_dir=extract_dir,
+            **{
+                **self._unpack_archive_kwargs(filename),
+                **kwargs,
+            },
+        )
+
+    def _unpack_archive_kwargs(self, archive_path: str | os.PathLike) -> dict[str, str]:
```
freakboy3742, line 87: Possibly a little over-enthusiastic refactoring here; function calls aren't free, and in practice, this method can't be overwritten by a subclass. Given it's wrapping a single if clause that won't be needed in a couple of Python releases, I'm not sure the extra method is worth it.

freakboy3742 (changes requested):
A large PR, but largely straightforward, with some nice cleanups along the way. One minor suggestion about some possibly overenthusiastic modularity; weak preference would be to remove it, but if you feel particularly strongly to the contrary, I'm not going to fight you too hard over it.

src/briefcase/exceptions.py
freakboy3742: That's what I had in mind, yes.

src/briefcase/integrations/file.py
rmartin16: That makes sense. I also tweaked it a bit so (in theory, anyway) pyupgrade should have an easier time trimming it down later on.

rmartin16:
A new problem from this is a corrupted `build` directory if the stub can't be installed.

Wasn't there work being done to ensure the bundle was deleted if the `CreateCommand` goes bad?

freakboy3742 (approved):
👍 LGTM.

PR #1922, opened 2024-07-19
Added option to run ADB in no-streaming mode

xperroni:
ADB can install an app either directly from the host ("streamed install"), or by copying the APK to the device and then running the installation process. The former is the current default, but the latter is still needed for some devices. Added an option allowing streamed install to be disabled.

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1493,14 +1493,20 @@ class ADB:
                 raise InvalidDeviceError("device id", self.device) from e
             raise
 
-    def install_apk(self, apk_path: str | Path):
+    def install_apk(self, apk_path: str | Path, no_streaming=False):
         """Install an APK file on an Android device.
 
         :param apk_path: The path of the Android APK file to install.
+        :param no_streaming: Whether to push APK to device and invoke Package Manager as separate steps.
         :returns: `None` on success; raises an exception on failure.
         """
+        command = (
+            ["install", "-r"]
+            + (["--no-streaming"] if no_streaming else [])
+            + [apk_path]
+        )
         try:
-            self.run("install", "-r", apk_path)
+            self.run(*command)
         except subprocess.CalledProcessError as e:
             raise BriefcaseCommandError(
                 f"Unable to install APK {apk_path} on {self.device}"
--- a/src/briefcase/platforms/android/gradle.py
+++ b/src/briefcase/platforms/android/gradle.py
@@ -354,6 +354,12 @@ class GradleRunCommand(GradleMixin, RunCommand):
             required=False,
         )
         parser.add_argument(
+            "--no-streaming",
+            action="store_true",
+            help="Push APK to device and invoke Package Manager as separate steps",
+            required=False,
+        )
+        parser.add_argument(
             "--shutdown-on-exit",
             action="store_true",
             help="Shutdown the emulator on exit",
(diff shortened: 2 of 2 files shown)


mhsmith:
In my experience ADB always chooses the mode automatically based on the version of the device. Why would you need to override that?

xperroni:
> In my experience ADB always chooses the mode automatically based on the version of the device.

That's not always the case. For example, it's an [ongoing issue in LineageOS](https://gitlab.com/LineageOS/issues/android/-/issues/7400#note_2008487373). That's actually what motivated me to write this patch.

mhsmith:
OK, but since this is a rare issue, I don't think it justifies adding a specific Briefcase argument. Instead, we could create an `--Xadb-install` argument which follows the pattern of the existing `--Xemulator`, allowing anything to be passed through to the subprocess.

freakboy3742:
> OK, but since this is a rare issue, I don't think it justifies adding a specific Briefcase argument. Instead, we could create an `--Xadb-install` argument which follows the pattern of the existing `--Xemulator`, allowing anything to be passed through to the subprocess.

I agree. This seems like a very specific niche workaround for a specific Linux distro, not something that needs a generic capability added to Briefcase. However, `adb` does take a lot of options, and some of them may be useful under edge-case circumstances, so we should provide a generic mechanism to expose those options.

xperroni:
I replaced the `--no-streaming` option with the more general `--Xadb-install` as discussed, but now I'm getting a coverage error in the CI / unit test suite. I'm having trouble understanding what the exact issue is; I suppose this means I need to add more tests for the new option, but how do I know what the coverage system is looking for?

xperroni:
OK, I seem to have figured a way to placate the code coverage system, though to be honest I'm still not sure what was wrong with the previous version of the code.

freakboy3742:
> OK, I seem to have figured a way to placate the code coverage system, though to be honest I'm still not sure what was wrong with the previous version of the code.

FWIW: The coverage gap was reported as `1503->1506`. This is telling you about *branch* coverage - the branch that jumps from line 1503 to line 1506 (i.e., the implicit `else: pass` that was part of the original `if extra_args is None` statement).

If you just have line coverage, calling the method with `extra_args = None` will cover every line of code that is written. With branch coverage, you also need to cover the implicit `else: pass` case, to ensure that you've tested every way the code can be used.

This isn't picked up when you convert the line into an inline ternary statement, due to a bug/missing feature in how coverage works - but it should ideally still have a test case.

changes/1922.feature.rst
```diff
@@ -0,0 +1 @@
+Added to ``briefcase run android`` a ``--Xadb-install`` option for passing extra arguments to ``adb install``. This makes it possible to handle various corner cases, for example, setting the ``--no-streaming`` option to make installation work for devices running recent versions of LineageOS.
```
freakboy3742, line 1: These notes should be written as a description of the "exciting new feature in the release", not a rehash of the commit message. 
```suggestion
The ``--Xadb-install`` option can now be used to pass arguments to ``adb install`` when running apps on Android. 
```

freakboy3742 (changes requested):
Looks good! 

One suggestion about the release note; and there's a need for documentation of the new option in `docs/reference/platforms/android/gradle.rst`. You can use the documentation for `--Xemulator` as a template for this.

Also, as noted in my note about coverage, there's one more test needed around invoking `install_apk` with `extra_args=None`. That's probably a good opportunity to revert to the older `if extra_args is None` syntax - to me, it's a little clearer than the "all in one line, but avoids the coverage problem" approach.

PR #1933, opened 2024-07-29
Modifications to support Python 3.13 support packages

freakboy3742:
Python 3.13 will introduce official iOS and Android support; supporting this new format requires some small changes to how Briefcase deals with iOS apps.

As a side effect of these changes, we're also in a position to use a dynamically loaded macOS framework based on the official distributable, rather than compiling it ourselves. This is an interim measure until an official "embeddable" or XCFramework-based macOS distribution is available; but it also requires some minor changes to how Briefcase handles macOS support packages.

* Modifies the filtering pattern for iOS logs. When redirecting to NSLog, stdout appears to come from the `_ctypes` library. In the official 3.13 packaging arrangement, the binary no longer has a `.cpython-313-iphonesimulator.dylib` extension - it is just `_ctypes`.
* Frameworks often include symlinks, so there's a need to copy those symlinks correctly, protect against cleaning up symlinks, and not attempt to sign symlinks.
* The new framework-based macOS template doesn't include a standalone `python-stdlib`; instead, a framework can be copied from the XCFramework. The `briefcase.toml` can now include a `runtime_path` definition that declares what subfolder of the support package should be copied to the `support_path` location.

## PR Checklist:

### DIFF
--- a/src/briefcase/platforms/macOS/app.py
+++ b/src/briefcase/platforms/macOS/app.py
@@ -47,6 +47,12 @@ class macOSAppCreateCommand(macOSAppMixin, macOSCreateMixin, CreateCommand):
         else:
             return self.bundle_path(app) / "support"
 
+    def runtime_path(self, app: AppConfig) -> Path:
+        try:
+            return self.path_index(app, "runtime_path")
+        except KeyError:
+            return "python-stdlib"
+
     def install_app_support_package(self, app: AppConfig):
         """Install the application support package.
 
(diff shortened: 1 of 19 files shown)


src/briefcase/platforms/macOS/app.py
```diff
@@ -61,9 +67,15 @@ def install_app_support_package(self, app: AppConfig):
...
+            # actually needed at runtime. Only copy the runtime_path into
+            # the support location nominated by the template. Also ensure
+            # that symlinks are preserved.
+            runtime_path = self.runtime_path(app)
             self.tools.shutil.copytree(
-                self.support_path(app) / "python-stdlib",
-                runtime_support_path / "python-stdlib",
+                self.support_path(app) / runtime_path,
+                runtime_support_path / Path(runtime_path).name,
+                # symlinks=True,
```
mhsmith, line 78: Isn't this supposed to be uncommented? If so, the tests should also be improved so they would have caught this.

freakboy3742: Dangit... I commented this out specifically to verify that the test was working... and then the test passed, so I didn't add the extra test that was required.

mhsmith (approved):

PR #1934, opened 2024-07-30
Drop Python 3.8 support and bump macOS/iOS template epoch

freakboy3742:
Python 3.8 is EOL in October, so the PEP 730 patches for iOS support won't be back ported to that release. There's also a need for a template epoch change because of the change in binary format on both iOS and macOS.

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/file.py
+++ b/src/briefcase/integrations/file.py
@@ -22,7 +22,7 @@ from briefcase.exceptions import (
 from briefcase.integrations.base import Tool, ToolCache
 
 if TYPE_CHECKING:
-    from typing import Iterable, Sequence
+    from collections.abc import Iterable, Sequence
 
 
 class File(Tool):
--- a/src/briefcase/integrations/file.py
+++ b/src/briefcase/integrations/file.py
@@ -5,10 +5,10 @@ import os
 import shutil
 import sys
 import tempfile
+from collections.abc import Iterable, Sequence
 from contextlib import suppress
 from email.message import Message
 from pathlib import Path
-from typing import TYPE_CHECKING
 from urllib.parse import urlparse
 
 import requests.exceptions as requests_exceptions
@@ -20,9 +20,6 @@ from briefcase.exceptions import (
     NetworkFailure,
 )
 from briefcase.integrations.base import Tool, ToolCache
-
-if TYPE_CHECKING:
-    from collections.abc import Iterable, Sequence
 
 
 class File(Tool):
--- a/tests/integrations/base/test_tool_registry.py
+++ b/tests/integrations/base/test_tool_registry.py
@@ -26,8 +26,8 @@ def tools_for_module(tool_module_name: str) -> dict[str, type[Tool]]:
             sys.modules[f"briefcase.integrations.{tool_module_name}"],
             lambda klass: (
                 inspect.isclass(klass)
+                and not inspect.isabstract(klass)
                 and issubclass(klass, (Tool, ManagedTool))
-                and klass not in {Tool, ManagedTool}
             ),
         )
     )
--- a/tests/integrations/base/test_tool_registry.py
+++ b/tests/integrations/base/test_tool_registry.py
@@ -1,7 +1,6 @@
 import inspect
 import pkgutil
 import sys
-from typing import Dict, Set, Type
 
 import pytest
 
@@ -9,7 +8,7 @@ import briefcase.integrations
 from briefcase.integrations.base import ManagedTool, Tool, tool_registry
 
 
-def integrations_modules() -> Set[str]:
+def integrations_modules() -> set[str]:
     """All modules in ``briefcase.integrations`` irrespective of whether they are
     defined in ``briefcase.integrations.__all__``"""
     return {
@@ -19,7 +18,7 @@ def integrations_modules() -> Set[str]:
     }
 
 
-def tools_for_module(tool_module_name: str) -> Dict[str, Type[Tool]]:
+def tools_for_module(tool_module_name: str) -> dict[str, type[Tool]]:
     """Return classes that subclass Tool in a module in ``briefcase.integrations``, e.g.
     {"android_sdk": AndroidSDK}."""
     return dict(
@@ -35,7 +34,7 @@ def tools_for_module(tool_module_name: str) -> Dict[str, Type[Tool]]:
 
 
 @pytest.fixture
-def all_defined_tools() -> Set[Type[Tool]]:
+def all_defined_tools() -> set[type[Tool]]:
     """All classes under ``briefcase.integrations`` that subclass Tool."""
     return {
         tool
--- a/tox.ini
+++ b/tox.ini
@@ -16,18 +16,17 @@ extend-ignore =
     E203,
 
 [tox]
-envlist = towncrier-check,docs-lint,pre-commit,py{38,39,310,311,312,313}-cov,coverage
+envlist = towncrier-check,docs-lint,pre-commit,py{39,310,311,312,313}-cov,coverage
 labels =
-    test = py{38,39,310,311,312,313}-cov,coverage
-    test38 = py38-cov,coverage38
+    test = py{39,310,311,312,313}-cov,coverage
     test39 = py39-cov,coverage39
     test310 = py310-cov,coverage310
     test311 = py311-cov,coverage311
     test312 = py312-cov,coverage312
     test313 = py313-cov,coverage313
-    test-fast = py{38,39,310,311,312,313}-fast
-    test-platform = py{38,39,310,311,312,313}-cov,coverage-platform
-    ci = towncrier-check,docs-lint,pre-commit,py{38,39,310,311,312,313}-cov,coverage-platform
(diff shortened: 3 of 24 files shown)


tests/integrations/base/test_tool_registry.py
```diff
@@ -19,23 +18,23 @@ def integrations_modules() -> Set[str]:
...
+def tools_for_module(tool_module_name: str) -> dict[str, type[Tool]]:
     """Return classes that subclass Tool in a module in ``briefcase.integrations``, e.g.
     {"android_sdk": AndroidSDK}."""
     return dict(
         inspect.getmembers(
             sys.modules[f"briefcase.integrations.{tool_module_name}"],
             lambda klass: (
                 inspect.isclass(klass)
+                and not inspect.isabstract(klass)
                 and issubclass(klass, (Tool, ManagedTool))
```
freakboy3742, line 30: This is the only substantive logic change required by the version upgrade. 

With the move of Sequence from `typing.Sequence` to `collections.abc.Sequence`, `SubprocessArgsT` now returns True from `inspect.isclass()`, but raises an error if you try to use it in `issubclass`. Adding an additional check for `inspect.isabstract()` avoids the error, and also allows removing the explicit reference to `Tool` and `ManagedTool` classes.

src/briefcase/integrations/file.py
```diff
@@ -22,7 +22,7 @@
 from briefcase.integrations.base import Tool, ToolCache
 
 if TYPE_CHECKING:
-    from typing import Iterable, Sequence
+    from collections.abc import Iterable, Sequence
```
rmartin16, line 25: I'm a bit curious...why gate this with `TYPE_CHECKING`?

tox.ini
rmartin16, line 1: just in case it wasn't intentionally omitted, you can make these changes as well if you'd like:
```yaml
[docs]
docs_dir = {tox_root}{/}docs
build_dir = {[docs]docs_dir}{/}_build
# replace when Sphinx>=7.3 and Python 3.8 is dropped:
#  -T => --show-traceback
#  -W => --fail-on-warning
#  -b => --builder
#  -v => --verbose
#  -a => --write-all
#  -E => --fresh-env
sphinx_args = -T -W --keep-going --jobs auto
```

src/briefcase/integrations/file.py
freakboy3742: Most of these changes are what pyupgrade generated automatically; if it was in `TYPE_CHECKING` before, it is after as well. There's a *really* minor import speed gain... but no functional reason that I can think of.

tox.ini
freakboy3742: Good catch - will do that.

freakboy3742: Just to confirm - the update here is to replace the short form arguments with the long forms? Or is there some other update that is on the table?

rmartin16: Just to use long-form of the arguments....at least, for the ones where it isn't self-evident at all what they do.

freakboy3742:
Regarding review/merge strategy, the sequence is:

[ ] Tag and publish updated support packages using the new frameworks
[ ] Re-run CI on the macOS Xcode and iOS templates
[ ] Merge the macOS Xcode and iOS templates
[ ] Tag and publish updated stub binaries from the macOS Xcode template
[ ] Re-run CI on the macOS app PR
[ ] Merge the macOS app template
[ ] Re-run CI on this PR
[ ] Merge this PR
[ ] ... 
[ ] profit.

rmartin16:
Extracted from https://github.com/beeware/briefcase/pull/1943#issuecomment-2276958967:

> As a broader topic; it's worth keeping in mind how this sort of change interacts with [this sort of landing sequence](https://github.com/beeware/briefcase/pull/1934#issuecomment-2272584424). In the case of that PR, we're bumping the support revision on macOS templates, and bumping the template epoch, which has complicated the landing sequence. I don't think the changes you've suggested here would change anything - but if we're able to make changes that would simplify things, that's definitely something we should consider.

Yeah; I've definitely encountered this - even as recently as https://github.com/beeware/briefcase/pull/1929. The solution I've been thinking about is basically the same approach we resolved this in other similar situations.

That is, when an app template repo needs to test against a specific version of Briefcase, you can specify `BRIEFCASE_REPO` and `BRIEFCASE_REF` in the text of the PR Body to change what CI uses for testing. Further, we added support for any CI running `app-build-verify` to use `BRIEFCASE_TEMPLATE_REPO` and `BRIEFCASE_TEMPLATE_REF`.

So, we could just extend this pattern to the build format templates for any use of `app-build-verify`.

Maybe something like:
- `BRIEFCASE[_-]<platform>[_-]<format>[_-]TEMPLATE[_-]REPO`
- `BRIEFCASE[_-]<platform>[_-]<format>[_-]TEMPLATE[_-]REF`

So, for example:
- `Briefcase-macOS-app-Template-Repo`/`Briefcase-macOS-app-Template-Ref`,
- `Briefcase-macOS-Xcode-Template-Repo`/`Briefcase-macOS-Xcode-Template-Ref`,

This would also probably be a good time to refactor all the regex logic on PR Body in to its own GitHub Action instead of continuing to duplicate it.

freakboy3742:
> > As a broader topic; it's worth keeping in mind how this sort of change interacts with [this sort of landing sequence](https://github.com/beeware/briefcase/pull/1934#issuecomment-2272584424). 
...
> Yeah; I've definitely encountered this - even as recently as #1929. The solution I've been thinking about is basically the same approach we resolved this in other similar situations.

The change you're suggesting would address the issue of running CI for a PR on Briefcase prior to landing the PRs for the template - but I'm not sure that's actually a problem that is causing us the problem here. We've got a reasonably well established pattern for landing template changes - rely on CI in the template (pointing at a branch of Briefcase) to validate the template works; merge the template PR; then re-run CI on the Briefcase PR; then merge the Briefcase PR. I'm not sure we gain a lot by further complicating our CI tooling just so that we can have a clean CI run on the Briefcase side prior to merging the template.

The problem in the case of this specific PR is that the branch redirection only applies *on the PR* - and so when merged to main, the template's merge CI run will fail. That's *usually* short lived because you can (almost) immediately merge the Briefcase PR, but in this case we can't because we need to retag the stub binary on the Xcode template before we can land the app template, which means the Xcode template's *mainline* needs to use a branch of Briefcase.

I guess one approach in this case would be to tag the new binary *on the PR branch*. That breaks the usual convention of tagging stub binaries on the main branch... but I guess the tagged commit will become part of the mainline as soon as the template is merged to main, so maybe that doesn't matter.

Even then, that wouldn't address the issue with the TODOs on the beeware/Python-Apple-support#191 - again, because the changes need to exist on the main branch (or, the 3.X branches in that case).

So - while this change would fix *a* problem, I'm not sure it's a problem we actually have in practice, and the complication to the CI workflows is such that I'm not sure it's worth it.

rmartin16:
I see. Out of curiosity, is there another example of this happening? At least where the CI for `main` needs configuration to pull from other repos for other than their `main` branch? Been racking my brain but this level of dependency seems rare.

Another thought is GitHub supports variables that can be configured outside of the CI workflows themselves. So, for these temporary periods where CI on `main` must behave differently, such configuration could control this.

freakboy3742:
> I see. Out of curiosity, is there another example of this happening? At least where the CI for `main` needs configuration to pull from other repos for other than their `main` branch? Been racking my brain but this level of dependency seems rare.

Yes, it's rare - hence why we've been willing to live with some janky landing sequences on the rare occasion it happens.

Other than template and epoch changes, the only example I can think of where it has come up recently is where a template adds a dependency on a new template filter. This currently requires adding (and landing) the filter in one PR, then updating the template to use that filter, then updating briefcase to do whatever change needed the new template. Again, we have a workaround; it *might* be nice to be able to temporarily tell the template to use a branch of Briefcase when running on main.
 
> Another thought is GitHub supports variables that can be configured outside of the CI workflows themselves. So, for these temporary periods where CI on `main` must behave differently, such configuration could control this.

I hadn't thought of using GitHub environment variables for this. The UX for setting this up is a bit messy, but I guess it would work, and it wouldn't be *that* much more complicated than the current "regex over PR body" approach. We'd probably also want a safety catch to make sure that we accidentally release Briefcase with the environment variable set on a template.

freakboy3742:
@mhsmith FYI - I had to revert the merge of the macOS Xcode template because having that template merged was breaking all the dependabot runs. When we do merge that template, we'll need to follow the remaining steps in rapid succession because until this PR is merged, any other PRs on Briefcase or a template will fail.

mhsmith:
OK, then I'll approve the new PRs but leave them unmerged.

freakboy3742:
@mhsmith I've tagged and published the updated stub binaries on the `framework-lib` branch, so I've been able to (1) run CI on the macOS-app template branch, and (2) revert the change in the binary publication script that temporarily used this Briefcase branch.

That means the macOS-app template and this PR are ready for review.

Assuming you're OK with both, this template PRs can be merged, the CI re-run on this PR, and then merge this PR. If you want to hold off on the merging, I'm happy to do that in the morning.

mhsmith (approved):

PR #1942, opened 2024-08-07
Fix docs environment to Python 3.12.

freakboy3742:
The changes in #1940 didn't have the desired effect on dependabot (see #1937).

This takes a different approach - since RTD *always* uses Python3.12 as the build environment, we can force tox to also use Python3.12 as a development environment. This radically simplifies dependency configuration (as we only
need to worry about a single Python version), and eliminates a potential source of bugs (as different Python versions can render details like type annotations differently).

## PR Checklist:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -98,16 +98,10 @@ commands =
 [docs]
 docs_dir = {tox_root}{/}docs
 build_dir = {[docs]docs_dir}{/}_build
-# replace when Sphinx>=7.3 and Python 3.8 is dropped:
-#  -T => --show-traceback
-#  -W => --fail-on-warning
-#  -b => --builder
-#  -v => --verbose
-#  -a => --write-all
-#  -E => --fresh-env
-sphinx_args = -T -W --keep-going --jobs auto
+sphinx_args = --show-traceback --fail-on-warning --keep-going --jobs auto
 
 [testenv:docs{,-lint,-all,-live}]
+basepython = python3.12
 package = wheel
 wheel_build_env = .pkg
 extras = docs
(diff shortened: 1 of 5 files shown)


rmartin16:
FWIW, while I philosophically support this change, I expect new contributors to not understand how to satisfy this requirement....and it's usually non-trivial to explain to someone how to side-by-side Python installs 🫤

tox.ini
```diff
@@ -98,16 +98,10 @@ commands =
...
-#  -W => --fail-on-warning
-#  -b => --builder
-#  -v => --verbose
-#  -a => --write-all
-#  -E => --fresh-env
-sphinx_args = -T -W --keep-going --jobs auto
+sphinx_args = --show-traceback --fail-on-warning --keep-going --jobs auto
 
 [testenv:docs{,-lint,-all,-live}]
+basepython = python3.12
```
rmartin16, line 104: For consistency :) or perhaps the coverage section could be updated to match this

```suggestion
base_python = py312
```

freakboy3742:
I agree this is a concern. However, the Sphinx ecosystem is making supporting "all valid Python versions" exceedingly difficult (deprecating Python 3.9 a full year before it goes EOL), and there are some advantages (such as being consistent with RTD rendering) that I think it's worth the pain. 

If it turns out that it's more trouble than wrangling version dependencies, we can easily revert this and find another way to keep dependabot happy.

tox.ini
freakboy3742: WHY ARE THERE TWO VALID SYNTAXES FOR THIS!!>!!KJHKJH!OI&!*^*&!

mhsmith:
We now have 4 places referencing the Python version number. There's a very good chance that we'll forget one of them next time we have to update this.

Three of them are unavoidable: contribute-docs.rst, tox.ini and readthedocs.yaml. But they should each have a comment referencing the other two.

In pyproject.toml we can simply remove the expicit Python version number and change it to something like "Docs are always built on a specific Python version; see ...".

mhsmith (approved):

PR #2027, opened 2024-10-14
Correct handling of arguments that take values.

### LINKED ISSUE #2026: Handling of options with values broken by Python 3.12.7
### Describe the bug

Any command other than `convert`, `new`, `dev`, or `upgrade` that uses an option that takes a value (e.g., `-C <config value>`, `-d <device ID>`) breaks when using Python 3.12.7.

### Steps to reproduce

1. Generate an empty Briefcase project using Toga
2. Try to run `briefcase run -d "iPhone 15 Pro"`
3. See error:
```
Invalid format 'iPhone 15 Pro'; (choose from: Xcode)
```
 

### Expected behavior

Options specified after flags should be accepted.

### Screenshots

_No response_

### Environment

- Operating System: All
- Python version: 3.12.7
- Software versions:
  - Briefcase: 0.3.19+

### Logs

_No response_

### Additional context

The problem appear to be function of the complex sequence of multiple *partial* parsings of the command line arguments performed by Briefcase.

In the first pass, Briefcase's command line parser only knows about the core arguments `-V` and `<command>` - the values that are valid to a bare `briefcase` invocation. These are partially parsed, allowing Briefcase to determine which subcommand will be invoked (or if `-V` or `--help` should be handled.

In the second pass, the parser adds handling for the *optional* `<platform>` and `<output_format>` placeholders, then reparses the arguments. This allows Briefcase to validate the platform and output format, choosing defaults if they aren't provided.

However, there's an inherent ambiguity at this point. In `briefcase run -d 'iPhone'` - is `'iPhone'` a value for the `-d` argument, or the value of the platform to pass to the run command? At this point, the parser doesn't have enough information to determine, as `-d` hasn't been configured in the parser.

Prior to python/cpython#124345 (released in Python 3.12.7; will be in 3.13.1 and 3.13.0a1) the value was considered to be a value for `-d`. As of 3.12.7, it's considered a value for the platform (i.e., as if `-d` took no value). 

It's difficult to argue this is a bug in Python, because `-d` hasn't been defined, so there's no reason to believe that `"iPhone"` *will* be a value for the `-d` flag.

This is currently breaking CI, both because of test cases that validate `-C foo=bar` handling for the `run` command, plus commands deploying iOS that pass in the device identifier for the test configuration.

The immediate workaround is to not fall back on any platform/output format defaults when using an argument that requires values - i.e., use `briefcase run iOS Xcode -d "iPhone 15 Pro"`, rather than `briefcase run iOS -d "iPhone 15 Pro"`. This ensures that the platform and output format are fully parsed before attempting to handle arguments with values.

freakboy3742:
Python 3.12.7 modified the handling of command line arguments in a way that broke Briefcase's handling of arguments that take values (e.g., `-C <config=value>`, `-d <device>`, etc).

Briefcase's command line handling is based on pushing argparse to its limits, performing partial parsing of arguments before all options have been registered. This is needed to allow commands to register their own arguments, and for the help for commands to be responsive to the target platform and output format.

The parser *also* has 2 optional positional arguments for most commands (`platform` and `output_format`). Because they're optional, there's an ambiguity between these positional arguments, and arguments with flags that take values that haven't been registered yet. 

This PR takes a quick-and-dirty approach, removing any argument that looks like it might be passing a value from consideration from the partial parse. It's not pretty, but it works.

Fixes #2026.

## PR Checklist:

### DIFF
--- a/changes/2026.bugfix.rst
+++ b/changes/2026.bugfix.rst
@@ -1,0 +1,1 @@
+Python 3.12.7 introduced an incompatibility with the handling of ``-C``, ``-d`` and other flags that accept values. This incompatibility has been corrected.
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -110,9 +110,23 @@ def parse_cmdline(args, console: Console | None = None):
         return {n.lower(): n for n in platforms.keys()}.get(name.lower(), name)
 
     # argparse handles `--` specially, so make the passthrough args bypass the parser.
+    # Also remove from partial processing any flag that is followed by a non-flag
+    # argument. This is required because we are partially processing arguments *before*
+    # declaring those arguments; that means we need to be able to parse `-X foobar` as a
+    # 2-part argument before actually defining the existence of `-X`. See
+    # beeware/briefcase#2026.
     def parse_known_args(args):
         args, passthrough = split_passthrough(args)
-        options, extra = parser.parse_known_args(args)
+        simple_args = []
+        complex_args = []
+        while args:
+            arg = args.pop(0)
+            if arg.startswith("-") and args and not args[0].startswith("-"):
+                complex_args.extend([arg, args.pop(0)])
+            else:
+                simple_args.append(arg)
+        options, extra = parser.parse_known_args(simple_args)
+        extra += complex_args
         if passthrough:
             extra += ["--"] + passthrough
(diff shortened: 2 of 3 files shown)


freakboy3742:
Superseded by #2028, which fixes the underlying problem with parsing rather than trying to work around it.

Issue #2016, opened 2024-10-06
Honor the PEP 621  `requires-python` setting


freakboy3742:
### What is the problem or limitation you are having?

Although Briefcase is able to target a wide range of Python versions, the app being published may not be equally flexible. As an example, the Toga testbed *requires* the use of Python 3.10 as a result of the specific dependencies that have been pinned. 

### Describe the solution you'd like

Briefcase should honor the PEP 612 [`requires-python`](https://packaging.python.org/en/latest/specifications/pyproject-toml/#requires-python) configuration item. If a project defines `requires-python` setting, and the currently active Python doesn't meet the requirement defined by that setting, Briefcase should raise an error.

### Describe alternatives you've considered

Do nothing, and rely on project documentation to define acceptable deployment platforms.

### Additional context

The `BaseCommand.verify_app()` method would be a reasonable place for this check.

sarayourfriend:
Hi @freakboy3742, I'd like to work on this issue. I hope to have a PR up for it later today, but if you see this before I manage to get that up, can you confirm that the best place for me to add the check would be in `merge_pep621_config`, in other words, when the PEP621 properties are parsed in the first place? If I've followed the code correctly, [the config is parsed every time a Briefcase command runs](https://github.com/beeware/briefcase/blob/c11a391c574c08bda8cc4bd5a700dddfeda27d88/src/briefcase/__main__.py#L25-L28), so that would prevent running any command with a Python version that does not satisfy `requires-python` (when defined).

freakboy3742:
@sarayourfriend No - `merge_pep621_config` isn't the right place, essentially for the reason you've identified.

There's two parts to the process of reading a configuration. The first is reading the configuration file; this involves a series of merges so that all the various parts of a configuration are pulled together into a single "flat" configuration dictionary for each app, no matter how complex your app's configuration is. This is where the `merge_pep621_config` comes in - it pulls known PEP 621 configuration items into the app's configuration. This *is* doing the parsing, which will sometimes involve some checks that the values are reasonable (e.g, that you've actually provided a string for a value that needs to be a a string); but it doesn't do anything to verify that the app configuration itself makes sense in the context of the app.

Then, when you run a command, there's [the `verify_app(app)` method](https://github.com/beeware/briefcase/blob/main/src/briefcase/commands/base.py#L619). This is defined on the base command; but other commands extend this implementation to do other checks, once more context about what has been requested is available. *That* is where any validations should be performed.

[...]

PR #2029, opened 2024-10-14
Honor PEP 621 `requires-python` setting.

sarayourfriend:
Fixes https://github.com/beeware/briefcase/issues/2016

While I believe this changeset addresses the issue as described, I have some lingering questions about the approach I took. Namely, the Linux system command mixins already include some Python interpreter version validation to ensure the running interpreter is compatible with Briefcase, and I haven't found a good way to integrate that into this change. I wondered for a bit whether a basic definition of `LinuxSystemMostlyPassiveMixin::verify_python` should be hoisted to the `BaseCommand`, and then `super()` called from those Linux system commands instead of creating an entirely new method on the `BaseCommand`. I chose against trying to do that because it seemed the concerns were getting mixed a bit, and I wanted to see what the maintainers thought before making any bigger changes.

If I were going to fully integrate this new bit of interpreter version validation with the existing bits, I think I'd like to rename `verify_python` to `verify_interpreter_briefcase_compatibility` (or something along those lines). Then name the `BaseCommand` method just `verify_python` and call to `verify_interpreter_briefcase_compatibility` from an override of that base command method in the mixin.

Speaking generally, it would probably be a good idea to rename `verify_python` in that mixin anyway, to at least nominally clarify the difference between its responsibility and that of the base command's own Python interpreter verification. But, like I said, I wasn't too confident in any of the approaches I could come up with, and wanted to avoid creating a bigger changeset than absolutely necessary. That being said, I'm happy to keep working on this to create a more cohesive solution, just let me know :slightly_smiling_face:.

Also: thank you for the excellent contributor documentation. It was easy to get started contributing to BeeWare! And a pleasant codebase to navigate around too.

## PR Checklist:


freakboy3742 (changes requested):
Initial quick review looks really good - will do a full review when CI is passing.

From the look of it, the issue is that `config.py` hasn't yet acquired the `from __future__ import annotations` that enables PEP 563 deferred annotation handling. This has been added to *most* files in Briefcase;l but for some reason, `config.py` hasn't been updated yet.

sarayourfriend:
Aha! Thanks, I previously pushed a commit to just remove that annotation, but have replaced it with the future import :+1:

tests/commands/base/test_verify_requires_python.py
```diff
@@ -0,0 +1,42 @@
...
+    """Validation passes if requires-python specifies a version lower than the running interpreter."""
+
+    major, minor, micro, *_ = sys.version_info
+    spec = f">{major}.{minor - 1}.{micro}"
+    base_command.global_config = _get_global_config(requires_python=spec)
+    base_command.verify_required_python(my_app)
+
+
+def test_requires_python_unmet(base_command, my_app):
+    """Validation fails if requires-python specifies a version lower than the running interpreter."""
```
sarayourfriend, line 35:
```suggestion
    """Validation fails if requires-python specifies a version greater than the running interpreter."""
```

sarayourfriend:
Okay, seems like the CI is good aside from the problem fixed in https://github.com/beeware/briefcase/pull/2028 :tada:

tests/commands/base/test_verify_requires_python.py
```diff
@@ -0,0 +1,51 @@
...
+    """If requires-python isn't set, no verification is necessary."""
+
+    base_command.global_config = _get_global_config(requires_python=None)
+    base_command.verify_required_python(my_app)
+
+
+def test_requires_python_met(base_command, my_app):
+    """Validation passes if requires-python specifies a version lower than the running interpreter."""
+
+    major, minor, micro, *_ = sys.version_info
```
freakboy3742, line 28: This is a good test; but it would be better if it were extended with `pytest.mark.parameterize` to verify a range of specifiers that meet requirements (>, >=, <, <=, ==), rather than a single "lower than" test.

tests/commands/base/test_verify_requires_python.py
```diff
@@ -0,0 +1,51 @@
...
+    spec = f">{major}.{minor - 1}.{micro}"
+    base_command.global_config = _get_global_config(requires_python=spec)
+    base_command.verify_required_python(my_app)
+
+
+def test_requires_python_unmet(base_command, my_app):
+    """Validation fails if requires-python specifies a version greater than the running interpreter."""
+
+    major, minor, micro, *_ = sys.version_info
+    spec = f">{major}.{minor + 1}.{micro}"
```
freakboy3742, line 38: As with the "met" test, this would be better if parameterised to cover more possible values.

freakboy3742 (changes requested):
This looks really good! A couple of small suggestions inline; we'll also need to wait until #2028 lands to confirm we're good to merge.

Regarding your comments about the Linux version checks - those are slightly different, because they are quite often verifying the version of Python in a Docker container that Briefcase is managing, rather than Briefcase itself. To that end, they're a slightly different verification to what is being validated here.

However, I think your instinct about renaming those validations from `verify_python` to something else is sound; `verify_interpreter_briefcase_compatibility` is a little more  verbose than I'd prefer. Something like `verify_system_python` would be a little more palatable.

freakboy3742:
FYI - #2028 has now landed, so if you merge with main, CI should now pass.

src/briefcase/commands/base.py
```diff
@@ -662,6 +665,27 @@ def verify_app_template(self, app: AppConfig):
...
+            return
+
+        try:
+            spec = Specifier(requires_python)
+        except InvalidSpecifier as e:
+            raise BriefcaseConfigError(
+                f"Invalid requires-python in pyproject.toml: {e}"
+            ) from e
+
+        running_version = f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.minor}"
```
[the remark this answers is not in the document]
sarayourfriend: TIL, thanks!

tests/commands/base/test_verify_requires_python.py
sarayourfriend: Good idea, done in https://github.com/beeware/briefcase/pull/2029/commits/dde2dafcc01646aceab4475d4e4718d9c539fa76

tests/commands/base/test_verify_requires_python.py
sarayourfriend: :+1: Done in https://github.com/beeware/briefcase/pull/2029/commits/dde2dafcc01646aceab4475d4e4718d9c539fa76

src/briefcase/commands/base.py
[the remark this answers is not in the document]
sarayourfriend: Done in https://github.com/beeware/briefcase/pull/2029/commits/4aa29d535fc7a4a2c51810879704c7e9d612c465

sarayourfriend:
You may be partial to `verify_system_python` because it already exists :sweat_smile: 

https://github.com/beeware/briefcase/blob/63f726094058d43d7e61545ed91c8ac02b1293b4/src/briefcase/platforms/linux/system.py#L447

After looking at the usage of the existing `verify_python` method a bit, it seems to _only_ be used for validating the Docker container's interpreter. The method's docstring claims this, and it is called only in this branch guarded by `if self.use_docker`:

https://github.com/beeware/briefcase/blob/63f726094058d43d7e61545ed91c8ac02b1293b4/src/briefcase/platforms/linux/system.py#L578-L594

I've renamed it to `verify_docker_python`, which matches the verbosity of the other methods, and (hopefully) accurately reflects the method's scope. Let me know what you think, happy to use any name you'd like or leave this change out of this PR altogether.

freakboy3742 (approved):
I've made a small tweak to the parameterized cases to (hopefully) make them a little more clear, and be a little less reliant on the specific Python version - but otherwise, this looks great - Thanks for the contribution! 

As luck has it, you've wrapped this up just before I was about to push a release, so this feature should be in the wild in a couple of hours as part of Briefcase 0.3.20.

(On an unrelated note - I notice from your bio that you're (a) looking for work, and (b) based in naarm; if you're not already planning to be at [PyCon AU 2024](https://2024.pycon.org.au), that should be some companies looking to hire there - and, for bonus points, you'll be able to claim your [BeeWare challenge coin](https://beeware.org/contributing/challenge-coins/), because I'll be there :-))

sarayourfriend:
> (On an unrelated note - I notice from your bio that you're (a) looking for work, and (b) based in naarm; if you're not already planning to be at [PyCon AU 2024](https://2024.pycon.org.au/), that should be some companies looking to hire there - and, for bonus points, you'll be able to claim your [BeeWare challenge coin](https://beeware.org/contributing/challenge-coins/), because I'll be there :-))

Thanks! The timing of this contribution may or may not be related to my presence at PyCon AU next month and having seen you speak at KiwiPycon last month :slightly_smiling_face:. I very much enjoyed your talk in Wellington, and suddenly have a lot of time on my hands to explore contributions to projects other than Openverse, so thought I'd take your advice from the Q&A at that talk and look at BeeWare.

I'll see you at PyCon, and around GitHub :grin:

PR #2032, opened 2024-10-15
Remove use of macOS-12; add explicit use of system python

freakboy3742:
A collection of CI updates reflecting platform changes:

1. macOS-12 runners will be turned off in just over a month, so we should stop using them. macOS-13 is the last x86-64 runner that is available; macOS-latest will be ARM64 from now on.
2. beeware/.github#172 adds the ability to explicitly reference a "system" python, rather than pinning the Ubuntu release and ensuring the `python-version` matches the Ubuntu release.

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ "macos-13", "macos-latest", "windows-latest", "ubuntu-latest" ]
-        python-version: [ "3.9", "3.13", "3.14" ]
+        python-version: [ "3.9", "3.13" ]
         include:
           - experimental: false
           # Ensure the Python versions between min and max are tested
@@ -78,9 +78,9 @@ jobs:
           - platform: "ubuntu-latest"
             python-version: "3.12"
             experimental: false
-          # Allow dev Python to fail without failing entire job
-          - python-version: "3.14"
-            experimental: true
+          # # Allow dev Python to fail without failing entire job
+          # - python-version: "3.14"
+          #   experimental: true
     steps:
     - name: Checkout
       uses: actions/checkout@v4.2.1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,4 +210,14 @@ jobs:
 
           # We need to use the system Python on Ubuntu
           - runner-os: ubuntu-latest
+            framework: toga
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -64,22 +64,23 @@ jobs:
...
             experimental: false
-          # Allow dev Python to fail without failing entire job
-          - python-version: "3.13-dev"
-            experimental: true
+          - platform: "ubuntu-latest"
+            python-version: "3.12"
+            experimental: false
+          # # Allow dev Python to fail without failing entire job
+          # - python-version: "3.14"
+          #   experimental: true
```
rmartin16, line 83: I was thinking...is it possible to default `experimental` to `false` so it only has to be set for dev Python?

For instance, maybe:
```yaml
continue-on-error: ${{ matrix.experimental || false }}
```
or perhaps:
```yaml
continue-on-error: ${{ matrix.experimental && true || false }}
```

I think originally `continue-on-error` did not support arbitrary expressions so everyone defaulted a value in their matrix config.

.github/workflows/ci.yml
```diff
@@ -185,23 +187,27 @@ jobs:
...
         framework: [ "toga", "pyside6", "pygame", "console" ]
-        runner-os: [ "macos-12", "macos-14", "ubuntu-22.04", "windows-latest" ]
+        runner-os: [ "macos-13", "macos-latest", "ubuntu-latest", "windows-latest" ]
 
   verify-apps:
     name: Build app
     needs: [ package, package-automation, unit-tests ]
-    uses: beeware/.github/.github/workflows/app-build-verify.yml@main
+    # TODO: Revert to main when .github#172 is merged
+    uses: beeware/.github/.github/workflows/app-build-verify.yml@599f2fbb61c17aa94a962f26647f31a879a31052
```
rmartin16, line 196: Do you want to confirm there won't be any surprises with `app-create-verify.yml@599f2fbb61c17aa94a962f26647f31a879a31052` as well?

rmartin16 (commented):
2 comments

rmartin16 (approved):
👍🏼 

Looks good. On a sidenote, I realized we're not actually using the logic in `app-create-verify` to skip `actions/setup-python` anywhere since we never pass `system` for `python-version` to it. I don't think that's a problem, though.

PR #2033, opened 2024-10-15
Forward git error message when template repo clone fails

### LINKED ISSUE #1118: Improve Error Handling for Invalid Briefcase Template

The error state for using an invalid remote repo can be improved. While the "invalid repo" example is completely manufactured, I encountered the non-existent repo one with a typo.

A non-existent repo

```
❯ briefcase new -t https://github.com/beeware/beeware-template.git --template-branch linux-system --no-input

Generating a new application 'Hello World'
Using app template: https://github.com/beeware/beeware-template.git, branch linux-system

Log saved to /home/russell/tmp/beeware/logs/briefcase.2023_03_05-16_19_56.new.log

  File "/home/russell/github/beeware/briefcase/venv-3.10-briefcase/bin/briefcase", line 8, in <module>
    sys.exit(main())
  File "/home/russell/github/beeware/briefcase/src/briefcase/__main__.py", line 20, in main
    command(**options)
  File "/home/russell/github/beeware/briefcase/src/briefcase/commands/new.py", line 528, in __call__
    return self.new_app(
  File "/home/russell/github/beeware/briefcase/src/briefcase/commands/new.py", line 477, in new_app
    self.generate_template(
  File "/home/russell/github/beeware/briefcase/src/briefcase/commands/base.py", line 814, in generate_template
    cached_template = self.update_cookiecutter_cache(
  File "/home/russell/github/beeware/briefcase/src/briefcase/commands/base.py", line 758, in update_cookiecutter_cache
    remote = repo.remote(name="origin")
  File "/home/russell/github/beeware/briefcase/venv-3.10-briefcase/lib/python3.10/site-packages/git/repo/base.py", line 414, in remote
    raise ValueError("Remote named '%s' didn't exist" % name)
ValueError: Remote named 'origin' didn't exist
```

An invalid repo

```
❯ briefcase new -t https://example.com --template-branch linux-system --no-input

Generating a new application 'Hello World'
Using app template: https://example.com, branch linux-system

Log saved to /home/russell/tmp/beeware/logs/briefcase.2023_03_05-16_20_46.new.log

  File "/home/russell/github/beeware/briefcase/venv-3.10-briefcase/bin/briefcase", line 8, in <module>
    sys.exit(main())
  File "/home/russell/github/beeware/briefcase/src/briefcase/__main__.py", line 20, in main
    command(**options)
  File "/home/russell/github/beeware/briefcase/src/briefcase/commands/new.py", line 528, in __call__
    return self.new_app(
  File "/home/russell/github/beeware/briefcase/src/briefcase/commands/new.py", line 477, in new_app
    self.generate_template(
  File "/home/russell/github/beeware/briefcase/src/briefcase/commands/base.py", line 820, in generate_template
    self.tools.cookiecutter(
  File "/home/russell/github/beeware/briefcase/venv-3.10-briefcase/lib/python3.10/site-packages/cookiecutter/main.py", line 69, in cookiecutter
    repo_dir, cleanup = determine_repo_dir(
  File "/home/russell/github/beeware/briefcase/venv-3.10-briefcase/lib/python3.10/site-packages/cookiecutter/repository.py", line 106, in determine_repo_dir
    cloned_repo = clone(
  File "/home/russell/github/beeware/briefcase/venv-3.10-briefcase/lib/python3.10/site-packages/cookiecutter/vcs.py", line 72, in clone
    repo_type, repo_url = identify_repo(repo_url)
  File "/home/russell/github/beeware/briefcase/venv-3.10-briefcase/lib/python3.10/site-packages/cookiecutter/vcs.py", line 45, in identify_repo
    raise UnknownRepoType
cookiecutter.exceptions.UnknownRepoType
```

Catch these errors and inform the user to verify the template.

Interpret the errors myself.

### Additional context

[briefcase.2023_03_05-16_19_56.new.log](https://github.com/beeware/briefcase/files/10892997/briefcase.2023_03_05-16_19_56.new.log)
[briefcase.2023_03_05-16_20_46.new.log](https://github.com/beeware/briefcase/files/10892998/briefcase.2023_03_05-16_20_46.new.log)

sarayourfriend:
Fixes #1118.

When cloning goes wrong, git usually provides a useful message. Examples of these are listed in the linked issue, and as far as I can tell, they always include a `fatal:` before the message details. Git sometimes also outputs information _before_ the `fatal:` sigil, but the exact content of that message varies, and appears to do so based on the server responding. From what I've seen, doesn't offer more information than Briefcase's "Unable to clone ..." message. For example, GitHub and Sourcehut say almost, but not quite the same things when you try to clone a repository that doesn't exist.

- GitHub: `ERROR: Repository not found.` with only a single trailing newline
- Sourcehut: `Repository not found.` with two trailing newlines and no `ERROR:` preamble

That message is then followed by identical `fatal: ...` messages. Git already introduces enough variability in the formatting of its `fatal` messages (see the normalisation part of this PR), so trying to account for the variations introduced by any given git server would be futile, and with dubious value. So far as I can tell, git's `fatal` message provides the context that would be useful to a user, so that's what I've chosen to surface to the Briefcase user.

The messages used in the test cases are based on the handful of real messages discussed in the issue, but have manufactured aspects documented in the `pytest.param` `id`, meant to test the particular cases of the hint normalisation. I have not tried to find every error git can emit when cloning, and have not verified that any of them don't include the `fatal:` sigil. It could be that the fallback hint will never be used in real life.

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -981,15 +981,12 @@ Did you run Briefcase in a project directory that contains {filename.name!r}?"""
                     git_fatal_message = re.findall(r"(?<=fatal: ).*?$", e.stderr, re.S)
                     if git_fatal_message:
                         # GitError captures stderr with single quotes. Because the regex above
-                        # takes everything after git's "fatal" message, we need to strip that final single quote
-                        hint = git_fatal_message[0].rstrip("'")
+                        # takes everything after git's "fatal" message, we need to strip that final single quote.
+                        hint = git_fatal_message[0].rstrip("'").strip()
 
                         # git is inconsistent with capitalisation of the first word of the message
-                        # and about periods at the end of the message. Normalise those here to ensure
-                        # a clean and uniform presentation
-                        # Also strip the string of any trailing whitespace before applying the period fix.
-                        hint = f"{hint[0].upper()}{hint[1:]}".strip()
-                        hint += "." if hint[-1] != "." else ""
+                        # and about periods at the end of the message.
+                        hint = f"{hint[0].upper()}{hint[1:]}{'' if hint[-1] == '.' else '.'}"
                     else:
                         hint = (
                             "This may be because your computer is offline, or "
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -6,6 +6,7 @@ import importlib.metadata
 import inspect
 import os
 import platform
+import re
 import subprocess
 import sys
 from abc import ABC, abstractmethod
@@ -977,11 +978,26 @@ Did you run Briefcase in a project directory that contains {filename.name!r}?"""
                     # any partial remnants of this initial clone.
                     # If we're getting a GitError, we know the directory must exist.
                     self.tools.shutil.rmtree(cached_template)
+                    git_fatal_message = re.findall(r"(?<=fatal: ).*?$", e.stderr, re.S)
+                    if git_fatal_message:
+                        # GitError captures stderr with single quotes. Because the regex above
+                        # takes everything after git's "fatal" message, we need to strip that final single quote
+                        hint = git_fatal_message[0].rstrip("'")
+
+                        # git is inconsistent with capitalisation of the first word of the message
+                        # and about periods at the end of the message. Normalise those here to ensure
+                        # a clean and uniform presentation
+                        # Also strip the string of any trailing whitespace before applying the period fix.
+                        hint = f"{hint[0].upper()}{hint[1:]}".strip()
+                        hint += "." if hint[-1] != "." else ""
+                    else:
+                        hint = (
+                            "This may be because your computer is offline, or "
+                            "because the repository URL is incorrect."
+                        )
+
                     raise BriefcaseCommandError(
-                        f"Unable to clone repository {template!r}.\n"
-                        "\n"
-                        "This may be because your computer is offline, or "
-                        "because the repository URL is incorrect."
+                        f"Unable to clone repository {template!r}.\n\n{hint}"
                     ) from e
 
             try:
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -32,10 +32,10 @@ from briefcase.console import MAX_TEXT_WIDTH, Console, Log
 from briefcase.exceptions import (
     BriefcaseCommandError,
     BriefcaseConfigError,
+    InvalidTemplateBranch,
     InvalidTemplateRepository,
     MissingAppMetadata,
     NetworkFailure,
-    TemplateUnsupportedVersion,
     UnsupportedHostError,
     UnsupportedPythonVersion,
 )
@@ -1037,7 +1037,7 @@ Did you run Briefcase in a project directory that contains {filename.name!r}?"""
                     head.checkout()
                 except IndexError as e:
                     # No branch exists for the requested version.
-                    raise TemplateUnsupportedVersion(branch) from e
+                    raise InvalidTemplateBranch(template, branch) from e
             except (ValueError, self.tools.git.exc.GitError) as e:
                 raise BriefcaseCommandError(
                     "Unable to check out template branch.\n"
@@ -1094,7 +1094,7 @@ Did you run Briefcase in a project directory that contains {filename.name!r}?"""
             raise InvalidTemplateRepository(template) from e
         except cookiecutter_exceptions.RepositoryCloneFailed as e:
             # Branch does not exist.
-            raise TemplateUnsupportedVersion(branch) from e
+            raise InvalidTemplateBranch(template, branch) from e
         except cookiecutter_exceptions.UndefinedVariableInTemplate as e:
             raise BriefcaseConfigError(e.message) from e
 
@@ -1135,7 +1135,7 @@ Did you run Briefcase in a project directory that contains {filename.name!r}?"""
                 output_path=output_path,
                 extra_context=extra_context,
             )
-        except TemplateUnsupportedVersion:
+        except InvalidTemplateBranch:
             # Only use the main template if we're on a development branch of briefcase
             # and the user didn't explicitly specify which branch to use.
             if version.dev is None or branch is not None:
--- a/src/briefcase/exceptions.py
+++ b/src/briefcase/exceptions.py
@@ -137,11 +137,10 @@ class UpgradeToolError(BriefcaseCommandError):
         super().__init__(msg=error_msg, skip_logfile=True)
 
 
-class TemplateUnsupportedVersion(BriefcaseCommandError):
-    def __init__(self, briefcase_version):
-        self.briefcase_version = briefcase_version
+class InvalidTemplateBranch(BriefcaseCommandError):
+    def __init__(self, template_repo, branch):
         super().__init__(
-            f"Could not find a template branch for Briefcase {briefcase_version}."
+            f"{branch} did not match any branches in template repository {template_repo}."
         )
 
 
--- a/src/briefcase/exceptions.py
+++ b/src/briefcase/exceptions.py
@@ -139,8 +139,10 @@ class UpgradeToolError(BriefcaseCommandError):
 
 class InvalidTemplateBranch(BriefcaseCommandError):
     def __init__(self, template_repo, branch):
+        self.template_repo = template_repo
+        self.branch = branch
         super().__init__(
-            f"{branch} did not match any branches in template repository {template_repo}."
+            f"Could not find a branch named {branch!r} in template repository {template_repo!r}."
         )
 
 
--- a/tests/commands/base/test_update_cookiecutter_cache.py
+++ b/tests/commands/base/test_update_cookiecutter_cache.py
@@ -185,13 +185,6 @@ def test_repo_clone_error(stderr_string, error_message, base_command, mock_git):
             branch="briefcase-template",
         )
 
-    base_command.tools.git.Repo.clone_from.assert_called_once_with(
-        url="https://example.com",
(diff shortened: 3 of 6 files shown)


src/briefcase/commands/base.py
```diff
@@ -977,11 +978,26 @@ def update_cookiecutter_cache(self, template: str, branch="master"):
...
+                        # GitError captures stderr with single quotes. Because the regex above
+                        # takes everything after git's "fatal" message, we need to strip that final single quote
+                        hint = git_fatal_message[0].rstrip("'")
+
+                        # git is inconsistent with capitalisation of the first word of the message
+                        # and about periods at the end of the message. Normalise those here to ensure
+                        # a clean and uniform presentation
+                        # Also strip the string of any trailing whitespace before applying the period fix.
+                        hint = f"{hint[0].upper()}{hint[1:]}".strip()
+                        hint += "." if hint[-1] != "." else ""
```
sarayourfriend, line 992: Please share any ideas of cleaner ways to do this cleanup, particularly the last two normalisation lines. I wasn't sure how deeply to comment this, so I went with as maximal approach as I could think of without just providing the particular examples. Suggestions for better comments are welcome as well :pray:

tests/commands/base/test_update_cookiecutter_cache.py
```diff
@@ -126,6 +127,72 @@ def test_new_repo_invalid_template_url(base_command, mock_git):
...
+            template=repository,
+            branch="briefcase-template",
+        )
+
+    base_command.tools.git.Repo.clone_from.assert_called_once_with(
+        url="https://example.com",
+        to_path=base_command.data_path / "templates" / "example.com",
+        filter=["blob:none"],
+        no_checkout=True,
+    )
```
sarayourfriend, line 193: I'm not confident this assertion is necessary in this test. The fact that this call happened is evident in the fact of the particular error strings being emitted, and isn't directly relevant to the question of the error messages themselves :thinking: 

```suggestion
```

src/briefcase/commands/base.py
freakboy3742: I think `str.capitalize()` is what you're looking for here. 
```suggestion
                        # GitError captures stderr with single quotes. Because the regex above
                        # takes everything after git's "fatal" message, we need to strip that final single quote.
                        hint = git_fatal_message[0].rstrip("'").strip()

                        # git is inconsistent with capitalisation of the first word of the message
                        # and about periods at the end of the message. 
                        hint = f"{hint.capitalize()}{'' if hint[-1] == '.' else '.'}"
```

sarayourfriend: Nice, thanks! Suggestion applied :+1:

tests/commands/base/test_update_cookiecutter_cache.py
```diff
@@ -126,6 +127,65 @@ def test_new_repo_invalid_template_url(base_command, mock_git):
     assert not (base_command.data_path / "templates" / "special-template").exists()
 
 
+@pytest.mark.parametrize(
+    ("stderr_string", "error_message"),
+    (
+        pytest.param(
+            "\n    stderr: '\nfatal: could not clone repository 'https://example.com' \n'",
+            "Could not clone repository 'https://example.com'.",
+            id="tailing-whitespace-no-caps-no-period",
```
freakboy3742, line 136: Nice - TIL about using pytest.param to specify a test ID

freakboy3742 (changes requested):
Looks great... except for the failing test cases :-)

sarayourfriend:
Ahh, `str::capitalize` won't work, because it lowercases the rest of the string, and ignores sentences:

```
>>> "Hello, World. Briefcase is Cool.".capitalize()
'Hello, world. briefcase is cool.'
```

I'll switch back to `upper` on the first character, plus the other improvements you made :+1: Should have time to do that in the next couple of hours.

freakboy3742:
> Ahh, `str::capitalize` won't work, because it lowercases the rest of the string, and ignores sentences:
> 
> ```
> >>> "Hello, World. Briefcase is Cool.".capitalize()
> 'Hello, world. briefcase is cool.'
> ```

... well that sucks. `'.'.join(t.capitalize() for t in text.split("."))`, anyone? 🤣 

(seriously - `upper()` on the first character is definitely preferable).

freakboy3742 (changes requested):
This looks great!

At the risk of scope creep, manual testing has revealed one additional change that is required before we can close out #1118 - the error handling around bad template branches. If I provide a valid template URL, but a bad branch name, this is the current output:
```
$ briefcase new -t https://github.com/beeware/briefcase-template.git --template-branch linux-system --no-input

[helloworld] Generating a new application 'Hello World'
Using app template: https://github.com/beeware/briefcase-template.git, branch linux-system
Cloning template 'https://github.com/beeware/briefcase-template.git'...

Could not find a template branch for Briefcase linux-system.
```
This error is currently being generated by the `TemplateUnsupportedVersion` exception - this makes sense when the template branch is a version number (which it will be in many cases), but doesn't make any sense when a specific template branch has been provided.

I suspect all is required here is to modify the exception to a `InvalidTemplateBranch` exception that accepts both the template URL and the branch name, and outputs an error message that reflects that context.

sarayourfriend:
Easy! I'll work on that tomorrow :slightly_smiling_face:

sarayourfriend:
@freakboy3742 I'm a bit stuck after revisiting this today.

Looking at the code, I think `TemplateUnsupportedVersion` is only used when the branch does not exist. Is that what you're getting at with respect to changing it to a new `InvalidTemplateBranch` exception? Basically: to clarify that it is not an "unsupported" branch, insofar as it does not even exist—it's just an invalid input?

Would the change then be to rename the exception to `InvalidTemplateBranch`, and update the message to something like `{branch} did not match any branches in the template repository {repo}.`?

---

A slight complication... the branch input is treated as any generic git ref, rather than explicitly a branch. It could be a tag or commit too. Does the "branch" terminology make sense, in that case? The CLI argument uses "branch", but could that name be reconsidered? With respect to the error message, I'm not sure how accurately such an error message needs to reflect the real usage of the argument, at least not if the real usage differs in scope from the name. It would be confusing for a user to pass a commit SHA and then be told it didn't match any branches. The user might interpret that as briefcase requiring specifically a _branch_ and only a branch, rather than any git ref[^git-output]. On the other hand, I might just be over-thinking this bit.

[^git-output]: And, for what it's worth, git's own message is laughably obtuse in this situation, so not much help as guidance. `git checkout not-a-branch` results in `error: pathspec 'not-a-branch' did not match any file(s) known to git`.

(This bit, whether to call it a branch or not, also feels definitely outside the scope of this particular issue, even if it potentially muddies the communication still; clarifying it would probably be better suited for at least a separate PR, if not an entirely separate issue?)

src/briefcase/exceptions.py
```diff
@@ -137,11 +137,10 @@ def __init__(self, error_msg):
         super().__init__(msg=error_msg, skip_logfile=True)
 
 
-class TemplateUnsupportedVersion(BriefcaseCommandError):
-    def __init__(self, briefcase_version):
-        self.briefcase_version = briefcase_version
+class InvalidTemplateBranch(BriefcaseCommandError):
+    def __init__(self, template_repo, branch):
```
freakboy3742, line 141: It's helpful for exceptions to keep their properties, so that they can be interrogated programmatically.
```suggestion
    def __init__(self, template_repo, branch):
        self.template_repo = template_repo
        self.branch = branch
```

src/briefcase/exceptions.py
```diff
@@ -137,11 +137,10 @@ def __init__(self, error_msg):
...
         super().__init__(
-            f"Could not find a template branch for Briefcase {briefcase_version}."
+            f"{branch} did not match any branches in template repository {template_repo}."
```
freakboy3742, line 143: The `!r` annotation is helpful for differentiating which part of text is user specified; it wasn't required previously because it was assumed the branch was always in the form of a Briefcase version number.
```suggestion
            f"Could not find a branch named {branch!r} in template repository {template_repo!r}."
```

freakboy3742 (approved):
I've made two minor tweaks to the cosmetics of the exception; but otherwise, this is great - thanks for another awesome contribution!

freakboy3742:
> Looking at the code, I think `TemplateUnsupportedVersion` is only used when the branch does not exist. Is that what you're getting at with respect to changing it to a new `InvalidTemplateBranch` exception? Basically: to clarify that it is not an "unsupported" branch, insofar as it does not even exist—it's just an invalid input?
> 
> Would the change then be to rename the exception to `InvalidTemplateBranch`, and update the message to something like `{branch} did not match any branches in the template repository {repo}.`?

Didn't see this comment until after reviewing the code - but yes - that was exactly what I had in mind. Your analysis is dead on - the primary usage case for branches is supporting different Briefcase versions, which is why almost all the usage is tied to "template version". However, you *can* specify any branch name you want, which is where the current exception text falls apart.
 
> A slight complication... the branch input is treated as any generic git ref, rather than explicitly a branch. It could be a tag or commit too. Does the "branch" terminology make sense, in that case? The CLI argument uses "branch", but could that name be reconsidered? ... On the other hand, I might just be over-thinking this bit.

I think that's probably overthinking it 😄 You're correct of course - but the set of people who work out you can use a tag or hash in place of a branch name aren't going to get confused if an error message refers to "branch <tag name>". 

> 1. And, for what it's worth, git's own message is laughably obtuse in this situation, so not much help as guidance. 

Yeah - "don't follow git's UX" is good advice in almost any scenario... 🤣

src/briefcase/exceptions.py
sarayourfriend: TIL, thanks!

PR #2040, opened 2024-10-18
Capture network errors when downloading files

sarayourfriend:
Fixes #1960.

I'm not very happy with the error message I've come up with, it's a bit of riffing on pip's "likely not a problem with pip" error. I think the same situation applies here, because either the server didn't send a chunked response, or the response it sent was unintelligible or had an SSL cert issue. With the server not sending a chunked response, perhaps Briefcase could try requesting an un-chunked response, but that's not going to work well for downloads the size of the Android SDK or Visual Studio, I don't think.

Regardless, the logs will contain more detailed information, including the handled requests exception.

Suggestions welcome, of course.

## PR Checklist:

- [x] All new features have been tested
- [x] All new features have been documented
- [x] I have read the **CONTRIBUTING.md** file
- [x] I will abide by the code of conduct

PR #2041, opened 2024-10-18
Replace `requests` with `httpx`

### LINKED ISSUE #1960: Handle Requests' `ChunkedEncodingError` exceptions when downloading files
### Describe the bug

A download failure mode for Requests is a `ChunkedEncodingError` and Briefcase does not handle this when downloading arbitrary files.

For instance:
```
2024-08-14T08:50:59.1255448Z [verifyapp] Installing support package...
2024-08-14T08:50:59.1358324Z Using custom support package https://github.com/rmartin16/python-standalone-releases/releases/download/continuous/cpython-3.9-aarch64-unknown-linux-gnu-noopt.tar.gz
2024-08-14T08:50:59.3866734Z Downloading cpython-3.9-aarch64-unknown-linux-gnu-noopt.tar.gz...
2024-08-14T08:51:00.2657989Z    \ ━━━━━━━━━━━━━━━━━━━━━━━╸                           47.9% • 00:01
2024-08-14T08:51:00.2659826Z 
2024-08-14T08:51:00.2902173Z 
2024-08-14T08:51:06.1506679Z Log saved to /home/runner/work/briefcase-arch-test/briefcase-arch-test/verifyapp/logs/briefcase.2024_08_14-08_51_00.create.log
2024-08-14T08:51:06.1572024Z 
2024-08-14T08:51:06.1636586Z 
2024-08-14T08:51:06.1643694Z 
2024-08-14T08:51:06.1659243Z Traceback (most recent call last):
2024-08-14T08:51:06.1664565Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/urllib3/response.py", line 748, in _error_catcher
2024-08-14T08:51:06.1687342Z     yield
2024-08-14T08:51:06.1689610Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/urllib3/response.py", line 894, in _raw_read
2024-08-14T08:51:06.1699939Z     raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
2024-08-14T08:51:06.1704736Z urllib3.exceptions.IncompleteRead: IncompleteRead(20971520 bytes read, 22835863 more expected)
2024-08-14T08:51:06.1708533Z 
2024-08-14T08:51:06.1709127Z The above exception was the direct cause of the following exception:
2024-08-14T08:51:06.1709865Z 
2024-08-14T08:51:06.1710088Z Traceback (most recent call last):
2024-08-14T08:51:06.1713666Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/requests/models.py", line 820, in generate
2024-08-14T08:51:06.1722126Z     yield from self.raw.stream(chunk_size, decode_content=True)
2024-08-14T08:51:06.1723614Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/urllib3/response.py", line 1060, in stream
2024-08-14T08:51:06.1737314Z     data = self.read(amt=amt, decode_content=decode_content)
2024-08-14T08:51:06.1738885Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/urllib3/response.py", line 949, in read
2024-08-14T08:51:06.1750958Z     data = self._raw_read(amt)
2024-08-14T08:51:06.1752209Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/urllib3/response.py", line 902, in _raw_read
2024-08-14T08:51:06.1762120Z     self._fp.close()
2024-08-14T08:51:06.1762900Z   File "/usr/lib/python3.9/contextlib.py", line 135, in __exit__
2024-08-14T08:51:06.1767196Z     self.gen.throw(type, value, traceback)
2024-08-14T08:51:06.1768320Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/urllib3/response.py", line 772, in _error_catcher
2024-08-14T08:51:06.1778412Z     raise ProtocolError(arg, e) from e
2024-08-14T08:51:06.1780954Z urllib3.exceptions.ProtocolError: ('Connection broken: IncompleteRead(20971520 bytes read, 22835863 more expected)', IncompleteRead(20971520 bytes read, 22835863 more expected))
2024-08-14T08:51:06.1781773Z 
2024-08-14T08:51:06.1782017Z During handling of the above exception, another exception occurred:
2024-08-14T08:51:06.1782547Z 
2024-08-14T08:51:06.1782744Z Traceback (most recent call last):
2024-08-14T08:51:06.1784508Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/bin/briefcase", line 8, in <module>
2024-08-14T08:51:06.1785451Z     sys.exit(main())
2024-08-14T08:51:06.1786402Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/briefcase/__main__.py", line 29, in main
2024-08-14T08:51:06.1789169Z     command(**options)
2024-08-14T08:51:06.1790447Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/briefcase/commands/create.py", line 949, in __call__
2024-08-14T08:51:06.1801804Z     state = self.create_app(app, **full_options(state, options))
2024-08-14T08:51:06.1803740Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/briefcase/commands/create.py", line 888, in create_app
2024-08-14T08:51:06.1814779Z     self.install_app_support_package(app=app)
2024-08-14T08:51:06.1816771Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/briefcase/commands/create.py", line 313, in install_app_support_package
2024-08-14T08:51:06.1823926Z     support_file_path = self._download_support_package(app)
2024-08-14T08:51:06.1825991Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/briefcase/commands/create.py", line 374, in _download_support_package
2024-08-14T08:51:06.1831846Z     return self.tools.file.download(
2024-08-14T08:51:06.1833614Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/briefcase/integrations/file.py", line 214, in download
2024-08-14T08:51:06.1840003Z     self._fetch_and_write_content(response, filename)
2024-08-14T08:51:06.1841931Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/briefcase/integrations/file.py", line 250, in _fetch_and_write_content
2024-08-14T08:51:06.1845802Z     for data in response.iter_content(chunk_size=1024 * 1024):
2024-08-14T08:51:06.1847381Z   File "/runner/work/briefcase-arch-test/briefcase-arch-test/venv/lib/python3.9/site-packages/requests/models.py", line 822, in generate
2024-08-14T08:51:06.1857449Z     raise ChunkedEncodingError(e)
2024-08-14T08:51:06.1859009Z requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(20971520 bytes read, 22835863 more expected)', IncompleteRead(20971520 bytes read, 22835863 more expected))
```

### Steps to reproduce

Cannot be reproduced at will.

### Expected behavior

Catch `ChunkedEncodingError` for file downloads. Or maybe just catch any exception that Requests can raise since they should all mean the same thing: the download was unsuccessful.

### Screenshots

_No response_

### Environment

- Operating System: virtualized arm64 Bullseye on GitHub Linux runner 2.319.0
- Python version: 3.9
- Software versions: 
  - Briefcase: c080702

### Logs

_No response_

### Additional context

_No response_

### LINKED ISSUE #2039: Replace `requests` with `httpx`

Briefcase currently uses the `requests` library to manage downloads etc. Requests is a well known quantity in the Python ecosystem, but it contains some architectural dead ends that prevent it from adapting to modern web and Python standards.

We should replace the use of `requests` in the Briefcase codebase with [`httpx`](https://www.python-httpx.org). The two share very similar APIs; but httpx has features and benefits that requests doesn't have (and, in some cases, *can't* have without a major rewrite), including:

* Support for http/2
* Support for asyncio
* More flexible choices in streaming downloads
* Performance (httpx generally beats requests in transfer performance; however, usually only in workloads that don't reflect Briefcase's use case) 

Do nothing. Requests definitely works as-is. We could safely do nothing and Briefcase will keep working.

The other big `requests` alternatives library is `aiohttp`; however, that's more of an 'async only' solution.

### Additional context

This isn't an urgent update. It's more of a "future-proofing" thing. There's also the benefit of using BeeWare's position to send a social signal that despite `requests` being a well-known and entrenched solution for HTTP management in Python, it's not the *only* solution, and in many cases, it's *not* the best solution.

sarayourfriend:
Fixes #2039

This was surprisingly more straightforward than I anticipated, but there are a few quirks worth explaining in the PR, which I'll leave as inline comments.

This PR definitely supersedes #2040, so I will close that PR in favour of this one. In which case, this also fixes #1960.

## PR Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,7 @@ dependencies = [
     "platformdirs >= 2.6, < 5.0",
     "psutil >= 5.9, < 7.0",
     "python-dateutil >= 2.9.0.post0",  # transitive dependency (beeware/briefcase#1428)
-    "requests >= 2.28, < 3.0",
+    "httpx >= 0.20, < 1.0",
     "rich >= 12.6, < 14.0",
     "tomli >= 2.0, < 3.0; python_version <= '3.10'",
     "tomli_w >= 1.0, < 2.0",
--- a/src/briefcase/exceptions.py
+++ b/src/briefcase/exceptions.py
@@ -90,9 +90,12 @@ class BriefcaseCommandError(BriefcaseError):
 
 
 class NetworkFailure(BriefcaseCommandError):
-    def __init__(self, action):
+    DEFAULT_HINT = "is your computer offline?"
+
+    def __init__(self, action, hint=None):
         self.action = action
-        super().__init__(msg=f"Unable to {action}; is your computer offline?")
+        self.hint = hint if hint else self.DEFAULT_HINT
+        super().__init__(msg=f"Unable to {action}; {self.hint}")
 
 
 class MissingNetworkResourceError(BriefcaseCommandError):
--- a/tests/commands/create/test_install_stub_binary.py
+++ b/tests/commands/create/test_install_stub_binary.py
@@ -3,8 +3,8 @@ import shutil
 import sys
 from unittest import mock
 
+import httpx
 import pytest
-from requests import exceptions as requests_exceptions
 
 from briefcase.exceptions import (
     InvalidStubBinary,
@@ -412,8 +412,9 @@ def test_offline_install(
     stub_binary_revision_path_index,
 ):
     """If the computer is offline, an error is raised."""
-    create_command.tools.requests.get = mock.MagicMock(
-        side_effect=requests_exceptions.ConnectionError
+    stream_mock = create_command.tools.httpx.stream = mock.MagicMock()
+    stream_mock.return_value.__enter__.side_effect = httpx.TransportError(
+        "Unstable connection"
     )
 
     # Installing while offline raises an error
--- a/tests/integrations/file/test_File__download.py
+++ b/tests/integrations/file/test_File__download.py
@@ -2,12 +2,12 @@ import os
 import platform
 import shutil
 import stat
+from collections.abc import Iterable, Iterator
 from pathlib import Path
 from unittest import mock
 
 import httpx
 import pytest
-from urllib3._collections import HTTPHeaderDict
 
 from briefcase.exceptions import (
     BadNetworkResourceError,
@@ -25,6 +25,56 @@ def mock_tools(mock_tools) -> ToolCache:
     # Restore move so the temporary file can be moved after downloaded
     mock_tools.shutil.move = mock.MagicMock(wraps=shutil.move)
     return mock_tools
+
+
+class _IteratorByteSteam(httpx.SyncByteStream):
+    """Shim that satisfies ``httpx.Response`` ``stream`` parameter type.
+
+    Cannot be replaced by any ``Iterable[bytes]`` because the base class requires
+    an explicit finalization method ``close``."""
+
+    def __init__(self, iterable: Iterable[bytes]) -> None:
+        self.iterable = iterable
+
+    def __iter__(self) -> Iterator[bytes]:
+        return iter(self.iterable)
+
+
+def _make_httpx_response(
+    *,
+    url: str,
+    status_code: int,
+    stream: list[bytes],
+    method: str = "GET",
+    headers: dict = {},
+) -> httpx.Response:
+    """Create a real ``httpx.Response`` with key methods wrapped by ``mock.Mock`` for spying.
+
+    Wrapped methods:
+        response.read
+        response.iter_bytes
+        response.headers.get
+    """
+    response = httpx.Response(
+        request=httpx.Request(
+            method=method,
+            url=httpx.URL(url),
+        ),
+        status_code=status_code,
+        headers=httpx.Headers(headers),
+        # Always use ``stream`` rather than content because it's more flexible
+        # even if the request is made non-streaming or the response is read with
+        # ``response.read()``, httpx will still consume the ``stream`` response
+        # content internally. This allows testing both the non-streaming and
+        # streaming download paths without needing to complicate the response params
+        stream=_IteratorByteSteam(stream),
+    )
+
+    response.read = mock.Mock(wraps=response.read)
+    response.iter_bytes = mock.Mock(wraps=response.iter_bytes)
+    response.headers.get = mock.Mock(wraps=response.headers.get)
+
+    return response
 
 
 @pytest.fixture
@@ -80,21 +130,21 @@ def file_perms() -> int:
     ],
 )
 def test_new_download_oneshot(mock_tools, file_perms, url, content_disposition):
-    response = httpx.Response(
+    """If no content-length is provided, ``File`` downloadds the file all at once."""
+    response = _make_httpx_response(
+        method="GET",
+        url=url,
         status_code=200,
-        request=httpx.Request(method="GET", url=httpx.URL(url)),
-        headers=httpx.Headers(
+        headers=(
             {
                 "content-disposition": content_disposition,
             }
             if content_disposition is not None
             else {}
         ),
-        stream=httpx.ByteStream(b"all content"),
+        stream=[b"all content"],
     )
     mock_tools.httpx.stream.return_value.__enter__.return_value = response
-    response.read = mock.Mock(wraps=response.read)
-    response.headers.get = mock.Mock(wraps=response.headers.get)
 
     # Download the file
     filename = mock_tools.file.download(
@@ -134,16 +184,17 @@ def test_new_download_oneshot(mock_tools, file_perms, url, content_disposition):
 
 
 def test_new_download_chunked(mock_tools, file_perms):
-    response = mock.MagicMock()
-    response.url = httpx.URL("https://example.com/path/to/something.zip")
-    response.status_code = 200
-    response.headers.get.return_value = "24"
-    response.iter_bytes.return_value = iter(
-        [
+    """If a content-length is provided, ``File`` streams the response rather than downloading it all at once."""
+    response = _make_httpx_response(
+        method="GET",
+        url="https://example.com/path/to/something.zip",
+        status_code=200,
+        headers={"content-length": "24"},
+        stream=[
             b"chunk-1;",
             b"chunk-2;",
             b"chunk-3;",
-        ]
+        ],
     )
     mock_tools.httpx.stream.return_value.__enter__.return_value = response
 
@@ -186,27 +237,40 @@ def test_new_download_chunked(mock_tools, file_perms):
 
 
 def test_already_downloaded(mock_tools):
+    """If the file already exists on disk, it isn't re-downloaded.
+
+    The request is still made to derive the filename, but the content
+    is never streamed."""
+
     # Create an existing file
     existing_file = mock_tools.base_path / "something.zip"
     with existing_file.open("w", encoding="utf-8") as f:
         f.write("existing content")
 
-    response = mock.MagicMock()
-    response.headers.get.return_value = ""
-    response.url = httpx.URL("https://example.com/path/to/something.zip")
-    response.status_code = 200
+    url = "https://example.com/path/to/something.zip"
+
+    response = _make_httpx_response(
+        status_code=200,
+        url=url,
+        # Use content and a content-encoding that would cause a DecodeError
+        # if ``File`` tried to read the response content.
+        # Because the file already exists, ``File`` shouldn't try to read
+        # the response, and the ``DecodingError`` won't occur.
+        headers={"content-length": "100", "content-encoding": "gzip"},
+        stream=[b"definitely not gzip content"],
+    )
     mock_tools.httpx.stream.return_value.__enter__.return_value = response
 
     # Download the file
     filename = mock_tools.file.download(
-        url="https://example.com/support?useful=Yes",
+        url=url,
         download_path=mock_tools.base_path,
     )
 
     # The GET request will have been made
     mock_tools.httpx.stream.assert_called_with(
         "GET",
-        "https://example.com/support?useful=Yes",
+        url,
         follow_redirects=True,
     )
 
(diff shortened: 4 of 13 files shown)


pyproject.toml
```diff
@@ -85,7 +85,7 @@ dependencies = [
     "platformdirs >= 2.6, < 5.0",
     "psutil >= 5.9, < 7.0",
     "python-dateutil >= 2.9.0.post0",  # transitive dependency (beeware/briefcase#1428)
-    "requests >= 2.28, < 3.0",
+    "httpx >= 0.20, < 1.0",
```
sarayourfriend, line 88: Really wasn't sure what version range to select here. I chose 0.20 by searching httpx's changelog and finding the last time they made any noted change to streaming. Other notable API usages are `response.url` as `httpx.URL`, which at least pre-dates 0.20 based on changelog notes that mention it receiving bug fixes in 0.18.0, as does the introduction of retries to `httpx.HTTPTransport` (0.15.0).

Aside from that, 0.20.0 is kind of arbitrary. It might be possible to go back even further like to 0.15.0, but 0.20.0 is 3 years old at this point, and I don't really know how to validate that any older version is actually supported (to be fair, I haven't done anything to validate that 0.20.0 is either, other than check feature availability).

src/briefcase/exceptions.py
```diff
@@ -90,9 +90,12 @@ def __str__(self):
 
 
 class NetworkFailure(BriefcaseCommandError):
-    def __init__(self, action):
+    DEFAULT_HINT = "is your computer offline?"
```
sarayourfriend, line 93: The default hint is meant to prevent needing to modify other code paths that already relied on `NetworkFailure`. Doing the change this way prevents those areas from needing to change or duplicate the generic hint.

tests/commands/create/test_install_stub_binary.py
```diff
@@ -412,8 +412,9 @@ def test_offline_install(
     stub_binary_revision_path_index,
 ):
     """If the computer is offline, an error is raised."""
-    create_command.tools.requests.get = mock.MagicMock(
-        side_effect=requests_exceptions.ConnectionError
+    stream_mock = create_command.tools.httpx.stream = mock.MagicMock()
+    stream_mock.return_value.__enter__.side_effect = httpx.TransportError(
```
sarayourfriend, line 416: `httpx.stream` returns a context manager, so it's necessary to mock `__enter__` with either the mocked response or exception.

tests/utils.py
```diff
@@ -330,9 +328,17 @@ def file_content(path: Path) -> str | None:
 
 def assert_url_resolvable(url: str):
     """Tests whether a URL is resolvable with retries; raises for failure."""
-    with requests.session() as sess:
-        adapter = HTTPAdapter(max_retries=Retry(status_forcelist={500, 502, 504}))
-        sess.mount("http://", adapter)
-        sess.mount("https://", adapter)
-
-        sess.head(url).raise_for_status()
+    transport = httpx.HTTPTransport(
```
sarayourfriend, line 331: httpx has no equivalent to urllib3's Retry and the requests adapter pattern.

Generally speaking, httpx's transports replace the adapter pattern, and in this case, they handle retrying in the case of network connectivity issues.

There is nothing like `Retry` and its `status_forcelist`, so the while loop reimplements that aspect.

I chose to add 3 retries somewhat randomly. `Retry` uses 10 retries by default, so I could increase `retries` and `bad_response_retries` to 5 to match, but the even split between networking errors and status code errors is arbitrary. requests does not operate that way, and there's currently no way to implement a combined retry count in httpx without overriding private methods deep in httpcore's connection handling.

sarayourfriend:
Hmmm, I'm not understanding where the coverage issue is coming from. When I run the test locally I can reproduce the same issue with 223->230... but if I put a breakpoint, the `TooManyRedirects` branch is definitely exercised in the test that mocks that exception (`test_get_connection_error`).

Maybe I'm misunderstanding what 223->230 means in this failure :thinking:

freakboy3742 (changes requested):
Wow - they weren't kidding when they said that the httpx API is mostly requests-compatible... I'm surprised it's *this* compatible. 

Regarding the coverage failure - the case it's reporting is the "bucket else" - when the exception type isn't any of the three cases that have been explicitly reported. I'm not sure if it's actually *possible* to have an exception other than those three types; but regardless, there's no way for coverage to do that reasoning because exceptions aren't type described by Python code. 

My usual approach to getting coverage in a case like this is to make the "default" case (i.e., TransportError in this case) the else case. If you can't do that because the order of the `isinstance` checks is significant (e.g., because of the specific class hierarchy), then you can put an `else: pass` and annotate it with `#pragma: no cover` (with a comment describing why that branch can't happen).

Once the coverage issue is resolved, this looks pretty much good to go. Nice work!

sarayourfriend:
> Wow - they weren't kidding when they said that the httpx API is mostly requests-compatible... I'm surprised it's this compatible.

It's pretty nice!

> Regarding the coverage failure - the case it's reporting is the "bucket else" - when the exception type isn't any of the three cases that have been explicitly reported. I'm not sure if it's actually possible to have an exception other than those three types; but regardless, there's no way for coverage to do that reasoning because exceptions aren't type described by Python code.

Aha! Thank you for the explanation.

The `isinstance` check order does not matter in this case, so I've pushed a change to match your suggestion with `TransportError` as the default `else` case.

---

As an aside, if httpx introduces a fourth kind of `RequestError`, that `else` branch starts to handle more than just the `TransportError`. Ideally, the check would be exhaustive, and report when the technically possible but practically (currently) impossible case occurs. Is the better thing to do to add an `else` that re-raises the unknown error type instead of swallowing it into the logs, or transforms into a Briefcase error that explicitly indicates an unknown error occurred and to file a bug report with logs attached? Not sure how defensive you prefer to be in these kinds of situations. The stakes in this particular case are pretty low, as are the chances it ever matters, and the error is generic enough that a user might check the logs anyway, so truly exhaustive checks is maybe going overboard.

sarayourfriend:
Looks like the failing test found an unhandled edge case: https://github.com/beeware/briefcase/actions/runs/11411470307/job/31756731325?pr=2041#step:27:511

It should be fixed with this patch, I'll push that, but only after revisiting the tests to figure out how to accurately exercise that case:

```diff
diff --git a/src/briefcase/integrations/file.py b/src/briefcase/integrations/file.py
index 81922290..002bd4f2 100644
--- a/src/briefcase/integrations/file.py
+++ b/src/briefcase/integrations/file.py
@@ -256,6 +256,7 @@ class File(Tool):
             with temp_file:
                 total = response.headers.get("content-length")
                 if total is None:
+                    response.read()
                     temp_file.write(response.content)
                 else:
                     progress_bar = self.tools.input.progress_bar()
```

The current tests don't mock `httpx.Response` correctly, so never catch this case. They mock `response.content` directly with bytes, but it's a `property` in the real `httpx.Response` class. This gets to a limitation of mocking httpx in this way that I was a little worried about when refactoring, and which `pytest-httpx` or `pook` might be better suited to handle, they let the testing code get away from mocking the request/response cycle and act more like integration tests (but eliminating the outbound network requests).

However, a less invasive change would be to use a real response object and only mock (with `wraps`) the particular methods that need asserting on. I'll try to make that work.

sarayourfriend:
Okay, https://github.com/beeware/briefcase/pull/2041/commits/5ffe0f13c087892633b5e763e1b5af021e9195a9 refactors one of the file__download tests to reduce the extent of mocking and expose the bug. The following commit, https://github.com/beeware/briefcase/pull/2041/commits/568644d7371ad0024e73db00e7eeefe6228649d3, addresses the bug.

I'm going to convert this to a draft because I'd like to refactor the rest of the tests in `test_File__download.py` to match the more thinly mocked approach of the test I refactored in 5ffe0f13c087892633b5e763e1b5af021e9195a9, and make sure to look for other tests that would benefit from the same treatment with respect to HTTP client mocking.

To clarify what I said before about `pook` and `pytest-httpx`, they significantly improve the ergonomics of writing tests like this. Compare the setup of the refactored test in 5ffe0f13c087892633b5e763e1b5af021e9195a9 to the `pook` version, (which is very similar to the `pytest-httpx` but slightly more comprehensive in its requirements by default):

```py
# one shot
pook.get("https://example.com/support?useful=Yes").reply(200).body(b"all content")

# chunked
pook.get("https://example.com/support?useful=Yes").reply(200).body([b"chunk 1;", b"chunk 2;", b"chunk 3;"], chunked=True)
```

Briefcase's usage is so minimal, that it doesn't feel worth pulling in a test dependency for just this `test_File__download.py` suite, but something to keep in mind if the need to mock HTTP request/response cycles grows.

Anyway, I'll work on the rest of this later this weekend or early next week :slightly_smiling_face:

freakboy3742:
> Not sure how defensive you prefer to be in these kinds of situations. The stakes in this particular case are pretty low, as are the chances it ever matters, and the error is generic enough that a user might check the logs anyway, so truly exhaustive checks is maybe going overboard.

In this case, the error message in the default case is already framed in a "maybe...?", so if the eventuality you've flagged ever happens, the error won't be *wrong*, just not as helpful as it could be under that specific failure mode, and we can easily adapt. If the error message pointed at a *specific* cause and mitigation, I'd probably prefer explicit and distinct bucket handling.

> Briefcase's usage is so minimal, that it doesn't feel worth pulling in a test dependency for just this `test_File__download.py` suite, but something to keep in mind if the need to mock HTTP request/response cycles grows.

Good call. I'm not opposed to introducing a dependency if it's going to radically improve testing; but given we've already got 98% of the testing of downloads that we need without the dependency, and we've pretty much hit the limit of the downloading that Briefcase will need to do, if we can live without the extra dependency, that would be my preference for now.

tests/integrations/file/test_File__download.py
```diff
@@ -22,12 +21,62 @@
...
+            url=httpx.URL(url),
+        ),
+        status_code=status_code,
+        headers=httpx.Headers(headers),
+        # Always use ``stream`` rather than content because it's more flexible
+        # even if the request is made non-streaming or the response is read with
+        # ``response.read()``, httpx will still consume the ``stream`` response
+        # content internally. This allows testing both the non-streaming and
+        # streaming download paths without needing to complicate the response params
+        stream=_IteratorByteSteam(stream),
```
freakboy3742, line 70: It took me a second to work out what was going on here - but once I did get it, it's a very nice simplification.

freakboy3742 (approved):
Another awesome contribution - thanks!

Issue #2006, opened 2024-09-25
Provide a more extensive default bootstrap


freakboy3742:
### What is the problem or limitation you are having?

If a user selects "No GUI framework", no bootstrap content is selected. As a result, the generated app configuration is *really* bare bones. This puts the onus on the developer to discover all the important configuration keys that they might want to use.

### Describe the solution you'd like

A new "Empty" bootstrap that contains all the same configuration *keys* as the Toga backend, but with no *values* specified. This would be used in the New and Convert wizards.

For example, the content for each operating system section (e.g., macOS, Linux, ...) would be something like:
```
# Delete the next line to support macOS
supported = false
requires = [
    # Python requirements for macOS
]
```

The Debian configuration section might be:
```
system_requires = [
    # System packages needed to run the app on Debian
]

system_runtime_requires = [
    # System packages needed to run the app on Debian
]
```

There would be no *app* content; this is purely to ensure that there are empty stub configurations for each platform that the user can "fill in", rather than needing to read the docs (and potentially hit typos etc).

### Describe alternatives you've considered

Do nothing. The status quo works, and the empty template case is an edge case.

### Additional context

Bootstrap definitions are in `src/briefcase/bootstraps`; they're registered in `pyproject.toml`, and installed in `NewCommand._gui_bootstrap_choices()` and `ConvertCommand.build_gui_context()`.

QB3-11:
I was wondering if the user will add the requirements, or should we define them?

freakboy3742:
@QB3-11 The user will be defining the requirements - we're not in a position to know what those requirements will be. This proposal is purely about ensuring that the "empty" stub has some better basic structure so it's obvious the options that might need to be filled in.

PR #2045, opened 2024-10-23
Provide a more extensive default bootstrap #2006

Xzenergy:
## PR Checklist:

- [x ] All new features have been tested
- [x ] All new features have been documented
- [x ] I have read the **CONTRIBUTING.md** file
- [x ] I will abide by the code of conduct

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -230,6 +230,9 @@ filterwarnings = [
     "error",
 ]
 
+[tool.briefcase]
+bootstraps = ["toga", "empty"]  # Add "empty" to the list of bootstraps
+
 # need to ensure build directories aren't excluded from recursion
 norecursedirs = []
 
--- a/src/briefcase/bootstraps/empty/empty.toml
+++ b/src/briefcase/bootstraps/empty/empty.toml
@@ -1,0 +1,22 @@
+# Configuration for an empty bootstrap
+
+# macOS configuration
+[macOS]
+# Delete the next line to support macOS
+supported = false
+requires = [
+    # Python requirements for macOS
+]
+
+# Linux configuration
+[Linux]
+system_requires = [
+    # System packages needed to run the app on Linux
+]
+system_runtime_requires = [
+    # System runtime packages needed to run the app on Linux
+]
+
+# Windows configuration
+[Windows]
+# Add any platform-specific configuration here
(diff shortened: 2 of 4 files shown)


freakboy3742 (changes requested):
Thanks for the PR, but there's a couple of issues here.

Firstly, [our contribution guide](https://briefcase.readthedocs.io/en/stable/how-to/contribute-code.html) details how to set up a pull request so it has the best chance of passing CI. At present, your PR is failing some basic checks.

More importantly, the implementation isn't really solving the problem. You've added an "empty.toml"... which doesn't reflect the way any of the other bootstraps are defined. You've also made a lot of changes to the new and convert commands that only appears to mention the Toga and Empty templates... missing the other pre-existing templates. Lastly, you've added *something* to briefcase.toml... but the code you've added won't do anything.

A better approach for this would be to take an *existing* bootstrap, search for everywhere in the code that it is mentioned, and duplicate those references - i.e., create a "toga2" bootstrap and then rename that bootstrap "empty" - then use *that* bootstrap when "no GUI template" is specified (the current "None") option.

Xzenergy:
> Thanks for the PR, but there's a couple of issues here.
> 
> Firstly, [our contribution guide](https://briefcase.readthedocs.io/en/stable/how-to/contribute-code.html) details how to set up a pull request so it has the best chance of passing CI. At present, your PR is failing some basic checks.
> 
> More importantly, the implementation isn't really solving the problem. You've added an "empty.toml"... which doesn't reflect the way any of the other bootstraps are defined. You've also made a lot of changes to the new and convert commands that only appears to mention the Toga and Empty templates... missing the other pre-existing templates. Lastly, you've added _something_ to briefcase.toml... but the code you've added won't do anything.
> 
> A better approach for this would be to take an _existing_ bootstrap, search for everywhere in the code that it is mentioned, and duplicate those references - i.e., create a "toga2" bootstrap and then rename that bootstrap "empty" - then use _that_ bootstrap when "no GUI template" is specified (the current "None") option.

Thanks so much for the feedback, this is my first attempt at contributing anything to anywhere! Thanks for the contributions guide, I hadn't visited that yet. I'll do some reading and then try to re-tackle the problem. Thanks again!

freakboy3742:
Resolved by #2119

Issue #1270, opened 2023-05-11
Install a python requirement with an --extra-index-url argument


mocaccano:
### What is the problem or limitation you are having?

Hi, I have a project which requires PyTorch CPU version, not the default GPU version. Normally, it's possible to install this in one of two ways using pip:
```
pip install torch torchvision --extra-index-url https://download.pytorch.org/whl/cpu
```
or
```
pip install -r requirements.txt
````
where the requirements look like:
```
--extra-index-url https://download.pytorch.org/whl/cpu
torch==2.0.1
torchvision==0.15.2
```

### Describe the solution you'd like

How do I translate this to the Briefcase environment? This is a question rather than a solution, but I'm new to Beeware and don't see any obvious way of including `--extra-index-url` in the `requires` section of the `pyproject.toml` file.

### Describe alternatives you've considered

Thanks!

**Comment by freakboy3742** [comment]
Agreed that this is something we should formally support. 

In addition to the PIP_EXTRA_INDEX_URL workaround, you could download the wheel manually, and add a direct reference to the wheel in your requires list.

Another workaround - although one that is a lot more fragile, and I wouldn't really want to rely on long term - would be to add [`"--extra-index-url", "https:///...."` to your `requires` list. By a quirk of implementation, it should work on macOS, iOS and Cocoa. On Android and Linux, adding `"--extra-index-url https:///...."` might work. This is obviously very fragile, but as a workaround, it might do the job.

**Comment by mocaccano** [comment]
Thanks for your responses. I tried adding  `"--extra-index-url https:///...."` to `requires` but it didn't work. I am on Ubuntu.
On the other hand `PIP_EXTRA_INDEX_URL` works very well, so I'm using this for now. Manual wheel could work, but haven't tried it so far.

**Comment by rmartin16** [comment]
> I tried adding `"--extra-index-url https://...."` to `requires` but it didn't work. I am on Ubuntu.

To add to the fragility of this workaround, you can't have the space separating the option and its value...an equal sign must be used. This should be true for any `pip` option you want to use:
```toml
 requires = [
     "--extra-index-url=https://download.pytorch.org/whl/cpu",
     "torch==2.0.1",
     "toga-gtk~=0.3.1",
 ]
```

wyrd-0:
> Agreed that this is something we should formally support.
> 
> In addition to the PIP_EXTRA_INDEX_URL workaround, you could download the wheel manually, and add a direct reference to the wheel in your requires list.
> 
[...]

PR #2057, opened 2024-11-05
Add mechamism for explicit `requires` package repository references

sarayourfriend:
Part of #1270.

Does not close out #1270 because @freakboy3742 also requested adding support for pip's `--find-links`. I chose to leave that out of this PR because this PR already has enough edge cases to consider (see the issue for discussion about them), and supporting `--find-links` requires thinking about how to generically describe it.

While this PR is feature complete for `--index-url` and `--extra-index-url`, which were the original concerns of the linked issue, I'm leaving it as a draft in order to propose a very different approach, which might simplify description of the feature, and better anticipate future changes like support/hooks for different package installation backends.

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/create.py
+++ b/src/briefcase/commands/create.py
@@ -513,12 +513,19 @@ class CreateCommand(BaseCommand):
             written.
         """
 
+        extra_pip_args = self._extra_pip_args(app)
+
         with self.input.wait_bar("Writing requirements file..."):
             with requirements_path.open("w", encoding="utf-8") as f:
-                if requires:
+                if requires or extra_pip_args:
                     # Add timestamp so build systems (such as Gradle) detect a change
                     # in the file and perform a re-installation of all requirements.
                     f.write(f"# Generated {datetime.now()}\n")
+
+                for arg in extra_pip_args:
+                    f.write(f"{arg}\n")
+
+                if requires:
                     for requirement in requires:
                         # If the requirement is a local path, convert it to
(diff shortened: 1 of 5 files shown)


src/briefcase/commands/create.py
```diff
@@ -513,12 +513,19 @@ def _write_requirements_file(
             written.
         """
 
+        extra_pip_args = self._extra_pip_args(app)
+
         with self.input.wait_bar("Writing requirements file..."):
             with requirements_path.open("w", encoding="utf-8") as f:
-                if requires:
+                if requires or extra_pip_args:
```
freakboy3742, line 520: My inclination is that this is the right idea, not taken far enough. If we're generating a new file, we should *always* be writing the timestamp - worst case, we end up with a file that only contains the timestamp, rather than a literally empty file.

sarayourfriend: Totally, makes sense to do away with the check altogether.

Also noting that I'd originally moved the check to the function root and done an early return when nothing would be written, but if requirements are being removed, it still needs to write the file to reflect requirements changed. Removing the conditional on the timestamp write makes that aspect a bit clearer as well (maybe with a small comment to help explain).

sarayourfriend:
Closing this based on the discussion in the linked issue, and [in particular this comment from Russell confirming the other approach](https://github.com/beeware/briefcase/issues/1270#issuecomment-2458522177).

I'll work on a changeset for that this week and hope to have a PR up in the next couple of days :slightly_smiling_face:

PR #2059, opened 2024-11-07
Add per-app pip install argument configuration

### LINKED ISSUE #1270: Install a python requirement with an --extra-index-url argument

sarayourfriend:
Fixes #1270 

This PR adds a new app config option `requirement_installer_args`, which is a list of strings, as discussed in the issue.

The goal is to allow users to configure `pip install` with any options they want. As discussed in the issue, arguments that appear to be local paths in format and resolve to an existing local path relative to the command's base path are transformed to absolute paths when calling `pip install`.

For most build targets, the arguments are added directly to the `pip install` invocation, alongside the requirements. Users have full control over the options they choose to pass.

For some build targets (Flatpak and Android), Briefcase hands off `pip install` invocation to a specialised build tool (Flatpak builder and Chaquo respectively). In these cases, Briefcase must supply the user-defined installer arguments in a way the build tool can reference. For both these cases, I've updated the templates (https://github.com/beeware/briefcase-linux-flatpak-template/pull/53, https://github.com/beeware/briefcase-android-gradle-template/pull/95) to process a `pip-options.txt` file defined in the path index as `app_requirement_installer_args_path`. When this path is defined, Briefcase will write a new-line separated list of the requirement installer args. It is up to the template to decide how to utilise this file. If no such path is defined, Briefcase will not write the installer args, but will still write the `requirements.txt` as it does today. This ensures seamless backwards compatibility with templates that do not define the new key in the path index.

The PR also changes the generated `requirements.txt` file to always write the timestamp comment, even if the file would end up empty otherwise.

## PR Checklist:


src/briefcase/platforms/linux/flatpak.py
```diff
@@ -149,6 +152,38 @@ def permissions_context(self, app: AppConfig, x_permissions: dict[str, str]):
...
+        self, app: AppConfig, requires: list[str], requirements_path: Path
+    ):
+        super()._write_requirements_file(app, requires, requirements_path)
+
+        # Flatpak runs ``pip install`` using an ``install_requirements.sh`` which Briefcase uses
+        # to indicate user-configured arguments to the command
+
+        pip_install_command = " ".join(
+            [
+                "/app/bin/python3",
```
rmartin16, line 165: Don't want to interject myself too much here....but potentially something to consider.

I've tried to be leery of coupling Briefcase too much to its first-party templates. For instance, here, we're assuming where Python will be available in the Flatpak; this will limit where a custom template can put Python without also re-implementing this packaging format for Briefcase. So (without having considered any potential complexity for this...), perhaps Python could be added to `PATH` in the Flatpak (or a similar mechanism) to prevent this coupling. freakboy3742, I'll ofc defer to you for any prioritization of this concern :)

p.s. - definitely appreciating your contributions :)

freakboy3742: Agreed that we should be avoiding coupling Briefcase to an "official" template wherever possible; but in this case, I think we're reasonable safe. `/app/bin` is where binaries will almost certainly be installed for virtual all practical Flatpak purposes; if that turns out to *not* be the case, we could add an extension point to the template that defines `flatpak_python_exe = "/app/bin/python3"` (similar to how we configure the location of the support package etc). However, that seems like a fairly niche requirement; and it's a fairly simple thing to add if and when the requirement emerges.

docs/reference/configuration.rst
```diff
@@ -309,6 +309,53 @@ multiple paragraphs, if necessary. The long description *must not* be a copy of
...
+The following examples will have the relative path transformed to an absolute one when Briefcase
+runs the requirement installation command if the path ``wheels`` exists relative to the configuration file:
+
+.. code-block:: TOML
+
+    requirement_installer_args = ["--find-links", "./wheels"]
+
+    requirement_installer_args = ["-f", "../wheels"]
+
+On the other hand, the next two examples avoid it because the string starts with ``-``, does not start with
```
sarayourfriend, line 336:
```suggestion
On the other hand, the next examples avoid it because the string starts with ``-``, does not start with
```

src/briefcase/platforms/linux/flatpak.py
mhsmith: > I've tried to be leery of coupling Briefcase too much to its first-party templates. For instance, here, we're assuming where Python will be available in the Flatpak

I suppose by this argument, we should also allow the template to control where the pip options are written, rather than assuming that install_requirements.sh is in the same directory as requirements.txt.

In fact, I have an idea which would solve both this and the Python executable path issue. In #1270 I suggested that the Android mechanism would be:

> make Briefcase write the pip options to a simple text file, one option per line, and put some code in the template's build.gradle to read the options from the file and pass them to the `pip { options }` method

What if Flatpak worked in the same way? 

* Briefcase would use the same code for both Flatpak and Android.
* This code would write only the options file, not a full shell script.
* The location and name of the file would be controlled by a variable in the template itself, just like [the requirements.txt file](https://github.com/beeware/briefcase-linux-flatpak-template/blob/v0.3.20/%7B%7B%20cookiecutter.format%20%7D%7D/briefcase.toml#L7).
* The details of how the file is used would be contained entirely within the template.

docs/reference/configuration.rst
```diff
@@ -309,6 +309,53 @@ multiple paragraphs, if necessary. The long description *must not* be a copy of
...
+
+    requirement_installer_args = ["-f", "wheels"]
+
+    requirement_installer_args = ["-f", "./this/path/does/not/exist"]
+
+.. admonition:: Supported arguments
+
+    The arguments supported in ``requirement_installer_args`` depend on the requirement installer backend.
+
+    The only currently supported requirement installer is ``pip``. As such, the list should only contain valid
```
mhsmith, line 353: We prefer wrapping both code and documentation at 88 columns. This also makes PR diffs easier to read on GitHub on laptop-size screens with a directory tree sidebar.

src/briefcase/platforms/android/gradle.py
```diff
@@ -282,6 +282,24 @@ def permissions_context(self, app: AppConfig, x_permissions: dict[str, str]):
             "features": features,
         }
 
+    def _write_requirements_file(
+        self, app: AppConfig, requires: list[str], requirements_path: Path
+    ):
+        super()._write_requirements_file(app, requires, requirements_path)
+
+        # Flatpak runs ``pip install`` using an ``install_requirements.sh`` which Briefcase uses
```
mhsmith, line 290: See comment above about line length.

src/briefcase/platforms/android/gradle.py
```diff
@@ -282,6 +282,24 @@ def permissions_context(self, app: AppConfig, x_permissions: dict[str, str]):
...
+        # to indicate user-configured arguments to the command
+        pip_options = "\n".join(
+            [f"# Generated {datetime.datetime.now()}"] + self._extra_pip_args(app)
+        )
```
mhsmith, line 294: Unlike the requirements file, the options file doesn't need a timestamp comment to trigger a re-run of pip when it changes, because it's read by Gradle on every build. And if we use this file format in Flatpak as I suggested [above](https://github.com/beeware/briefcase/pull/2059#discussion_r1844710348), then eliminating the comment would make it easier to process with a shell script.

src/briefcase/platforms/linux/flatpak.py
sarayourfriend: @mhsmith so would the file to write have a generic name like the configuration option? I suppose I can see a version of this feature where the arguments are always written in the same format regardless of target (and therefore, it can be part of the base `CreateCommand` without private method overrides like I've done in this PR)... and then it's up to the template whether or how it uses those arguments.

On the other hand, keeping in mind that arbitrary requirement installer backends could be supported aside from pip (like conda, uv), writing the pip options to a file makes the requirement installer backend a template detail, rather than a Briefcase detail... but only for Flatpak and Android which run `pip install` themselves rather than other targets where Briefcase executes it.

Implementing another installer backend then gets split. For most targets, it would just be a change to Briefcase to support it. For targets that use the `pip-options` file, they would need to have logic in the template to use the correct installer backend, adding complexity to the template.

Of course for Flatpak more would need to be done to support some other installer backend in this case (i.e., providing conda/uv/whatever in the Flatpak environment) but that could be put into the `install-requirements.sh` script. Android seems like potentially the most restrictive because Chaquopy exposes a custom pip interface rather than allowing any old package installer command running? I'd worry that taking the approach necessary for Android for everything else potentially limits the possibilities for targets that don't have the same limitations.

docs/reference/configuration.rst
sarayourfriend: Thanks for letting me know. flake8 is configured for max line length of 119. I suppose if 88 is the expectation, it's worth changing the linter to catch this so it doesn't need to come up in reviews.

src/briefcase/platforms/android/gradle.py
sarayourfriend: Fixed in https://github.com/beeware/briefcase/pull/2059/commits/2807664b3406436f505a181f8367761b9d10543e

docs/reference/configuration.rst
sarayourfriend: Fixed in https://github.com/beeware/briefcase/pull/2059/commits/2807664b3406436f505a181f8367761b9d10543e

src/briefcase/platforms/linux/flatpak.py
freakboy3742: Agreed that it would be desirable to use the same format for both Flatpak and Android (if only because the code can be shared); and that the other benefits you describe would also be useful.

My only hesitation is whether there's an easy way to manage this in the Flatpak template itself. I'm not certain that subshell invocations (like `$(cat somefile.txt)` ) are legal in a Flatpak manifest. Definitely happy to be proven wrong on this front, though.

src/briefcase/platforms/android/gradle.py
freakboy3742: It's not needed to trigger a re-run, but I'd argue it's helpful as a diagnostic tool. However, if it interferes with easy command line processing, then it's a "nice-to-have" we can definitely lose.

src/briefcase/platforms/linux/flatpak.py
sarayourfriend: It's kind of ugly, but this works as a build command in Flatpak manifest to read the arguments from a newline delimited `pip-options.txt`:

```yaml
      - readarray -t EXTRA_PIP_ARGS < pip-options.txt; CC="gcc -pthread" /app/bin/python3 -m pip install --no-cache-dir -r requirements.txt --target /app/briefcase/app_packages "${EXTRA_PIP_ARGS[@]}"
```

I don't really know if this is portable. `$SHELL` is reported as `/bin/sh` but on my Fedora installation that's just bash. I'm concerned this won't work if the build host doesn't use bash for `sh` and adding a dependency on it doesn't seem right.

I suppose this argument reading and command preparation could be written as a Python script executed with `os.execv` (something along those lines), removing the build host's shell environment as a factor altogether.

freakboy3742: A python script sounds like a reasonable approach to me - we definitely don't have any guarantees that the end-user will be using bash.

sarayourfriend: I've opened a PR with a potential approach for using this options-file approach for Flatpak: https://github.com/beeware/briefcase-linux-flatpak-template/pull/53

src/briefcase/commands/create.py
```diff
@@ -652,8 +673,21 @@ def install_app_requirements(self, app: AppConfig, test_mode: bool):
...
+                app
+            )
+        except KeyError:
+            requirement_installer_args_path = None
+
+        if requirements_path:
+            self._write_requirements_file(
+                app, requires, requirements_path, requirement_installer_args_path
+            )
+        else:
```
sarayourfriend, line 690: This change in logic is to isolate the indication of whether the requirements files should be written to only if `app_requirements_path` is defined. To ensure backwards compatibility, `app_requirement_installer_args_path` needs to be optional. Existing template versions won't define it but should continue to write the requirements path.

src/briefcase/platforms/linux/flatpak.py
mhsmith: > Implementing another installer backend then gets split. For most targets, it would just be a change to Briefcase to support it. For targets that use the `pip-options` file, they would need to have logic in the template to use the correct installer backend, adding complexity to the template.

On Flatpak, I think the main reason we run pip inside a container rather than directly from Briefcase is to support building binary packages from source. But that wouldn't be a factor with Conda packages, which are always binary. 

So, although I haven't thought through the details, I suspect that when we add Conda support to Briefcase, we'll be running Conda directly from Briefcase on all desktop platforms including Flatpak. This will inevitably complicate the template, but I don't think anything in this PR will make it any worse.

Mobile platforms aren't currently supported by Conda at all, so we don't need to worry about that until much later, and maybe never.

mhsmith:
For future reference, force pushing is fine if it's to clean up something that other people probably haven't seen yet. But please don't use a force push to alter commits which have already been commented on, as it makes the conversation difficult to follow.

sarayourfriend:
Sure, no problem, I will avoid it in the future. I chose to force push here because I noticed PRs are squashed, and I wanted to avoid introducing commits on `main` that didn't reflect the desired implementation. Apologies. Maybe this could have been better approached as a separate PR, considering the degree of change in the approach (though it's also why I've left this as a draft for now).

In any case, I'll avoid it in this repository from now on.

sarayourfriend:
I've updated the PR description to reflect the current implementation. Looking forward to reviews when y'all have a chance to revisit this :slightly_smiling_face:.

src/briefcase/commands/create.py
```diff
@@ -544,7 +555,17 @@ def _extra_pip_args(self, app: AppConfig):
...
-        return []
+        args: list[str] = []
+        for argument in app.requirement_installer_args:
+            if relative_path_matcher.match(argument) and _is_local_path(argument):
+                abs_path = os.path.abspath(self.base_path / argument)
+                if Path(abs_path).exists():
+                    args.append(abs_path)
+                    continue
+
+            args.append(argument)
```
freakboy3742, line 566: A minor code style thing - I try to avoid `continue` and `break` as much as possible, as it's easy to misread the intention of the statement (for a long read example of this happening, [in 1990 almost the entire US phone system was taken down by a stray `break` statement](https://www.mit.edu/hacker/part1.html)).

In a case such as this, I'd lean towards defining a `final_arg` value that is `abs_path` in the continued case here, but set to `argument` otherwise, and then unconditionally appending `final_arg`.

src/briefcase/commands/create.py
```diff
@@ -618,7 +639,7 @@ def _install_app_requirements(
             self.tools.os.mkdir(app_packages_path)
 
         # Install requirements
-        if requires:
+        if requires or app.requirement_installer_args:
```
freakboy3742, line 642: I'm not sure I understand this one - why invoke `pip install`  if there are pip arguments, but aren't any requirements to install?

tests/commands/create/test_install_app_requirements.py
```diff
@@ -491,6 +674,130 @@ def test_app_requirements_requires(
...
+
+    # requirements.txt exists either
+    assert app_requirements_path.exists()
+    with app_requirements_path.open(encoding="utf-8") as f:
+        assert f.read() == f"{GENERATED_DATETIME}\nmy-favourite-package\n"
+
+    assert app_requirement_installer_args_path.exists()
+    assert (
+        app_requirement_installer_args_path.read_text(encoding="utf-8")
+        == "--no-cache\n-f\nwheels with space"
```
freakboy3742, line 729: A mostly cosmetic thing - but should this have a trailing newline at the end?

freakboy3742 (changes requested):
This looks really good. I've flagged a couple of minor code style things inline, and I need to take it for a manual walk through on each platform - but other than that, this looks really solid.

src/briefcase/commands/create.py
sarayourfriend: A user could define a requirement in the new option, for example, to install it as editable. Maybe the "right" way to do that is `--editable=../package` in `requires`. Regardless, it's an edge case that (probably) only applies in a development environment, and using `requires` for it would still work. There may be other such examples, I'm not sure. I don't have a strong feeling about this, happy to revert it.

tests/commands/create/test_install_app_requirements.py
sarayourfriend: I'd left out the trailing newline because I was worried about needing to filter empty lines in the templates, but the templates already do anyway. I can add it for consistency.

src/briefcase/commands/create.py
sarayourfriend: Fair enough, thanks for the link to the story!

Done in https://github.com/beeware/briefcase/pull/2059/commits/ca791f43fd7caf445c6387c4f3b1b842204a0809

tests/commands/create/test_install_app_requirements.py
sarayourfriend: Done in https://github.com/beeware/briefcase/pull/2059/commits/7666f2f208469d63df78c6a6bb52aac56d187aca

src/briefcase/commands/create.py
mhsmith: I think `--editable` doesn't really make sense in a Briefcase context, much like `--user` and other options which change the location of the installed code. Briefcase's equivalent to `--editable` is `briefcase dev`.

sarayourfriend: Wouldn't you want to use editable in, say, a monorepo situation, so you don't have to re-`pip install` (via `briefcase dev -r`)? I've done this myself a bit when working on Toga GTK features that I wanted to try out in an actual application, it makes it much nicer than needing to wait for `pip install` each time I restart `briefcase dev`.

Maybe I've missed some part of the dev workflow there, though.

freakboy3742: You're correct that editable installs can be useful for `briefcase dev` purposes - I do the same thing as you've described for Toga development. Editable installs *might* even work in some scenarios for a `briefcase run` purposes - but they definitely won't on anything deployment target that is sandboxed (e.g., Flatpak, or anything mobile) because the `.pth` file that manifests the editable install won't resolve at runtime. 

In terms of this specific feature, I could probably go either way. An empty `requires` is definitely a niche edge case; I think we can leave it as right now, and if it turns out to be necessary or useful, it's an easy addition later on.

sarayourfriend: Sounds good! Thanks for the input, all. I'll resolve this thread as addressed in https://github.com/beeware/briefcase/pull/2059/commits/b67eed533818c10ee48da41b3e7b8de712dabfe8.

freakboy3742 (changes requested):
This looks great; 2 issues emerged in testing.

The first is dev requirements - In `briefcase dev -r`, pip arguments aren't passed into pip install for the dev environment. This should be a fairly straightforward fix - it's a separate pip call that needs to follow the same patterns as the one in `create`.

The second is more complex - Docker builds for Linux (e.g., `briefcase create linux --target ubuntu:24.04`). `create` converts the file system references into absolute references, but those absolute references are then passed verbatim to the Docker container, where they aren't readable because they don't exist inside the container.

Briefcase's Docker tooling has mechanisms in place to convert any "external" filesystem location that matches a filesystem mount into an "internal" reference when a command is passed to Docker for execution; but the mount needs to exist in the first place. I suspect this will mean a different path through the `requirement_installer_args` handling, keeping track of (and *also* adding mounts for) each filesystem location mentioned in the requirement args. 

That said - I could also live with the Docker piece being future work. Docker builds are always the more complex case of everything, so I don't think it's the end of the world if `requirement_installer_args` has a known limitation for Docker paths, as long as it's documented as such, and we open a ticket for someone else to look into at some point.

sarayourfriend:
Thanks, I've pushed a commit to fix `briefcase dev`. Good catch!

Regarding Docker...

> That said - I could also live with the Docker piece being future work. Docker builds are always the more complex case of everything, so I don't think it's the end of the world if requirement_installer_args has a known limitation for Docker paths, as long as it's documented as such, and we open a ticket for someone else to look into at some point.

I think this is the way to go. If you'll accept the PR in its current state (provided I push another commit with documentation to outline this limitation for Docker), I would be happy to work on a follow-up PR to explore the Docker side of things for this. I'd prefer it in a separate PR as this one's already pretty big, and having briefly looked at the Docker path stuff when working on earlier implementations to support #1270, I suspect sorting out the Docker side will be a significant PR on its own.

If that works for you, I'll push the documentation change. Otherwise, we can discuss next steps :slightly_smiling_face:

freakboy3742:
> If that works for you, I'll push the documentation change. 

Deferring the Docker part is completely OK with me.

freakboy3742 (changes requested):
`dev` fixes work great; with the extra "quirks" docs, this looks ready to go!

sarayourfriend:
Sounds good, thanks again for the review. I will work on the docs update tomorrow.

sarayourfriend:
New docs added in https://github.com/beeware/briefcase/pull/2059/commits/741366f4781d11dc790705b1dda3ce052dfa0d6e, though I'm not confident in the phrasing I've used.

freakboy3742 (approved):
Thanks - those docs are close enough that I can prod them the rest of the way.

Another awesome contribution in the bag!

Issue #1876, opened 2024-06-16
Minor syntax bug when using quotes in the app's name


Malasaur:
### Describe the bug

I've just noticed that if you create an app and include a quote in the name (`'`), `briefcase` does not escape it inside the `settings.gradle` file. Nothing very concerning, just a bit misleading for people who may not know about this kinda stuff

### Steps to reproduce

1. Create a new app with `briefcase`
2. Include `'` inside the name, e.g. "John Doe's amazing application"
3. Build it for Android
Why does writing this feel like greentext?

### Expected behavior

I expected it to compile without errors. lol

### Environment

Like, are these really needed?
- Operating System: Endeavour OS
- Python version: 3.12.3
- Software versions: 
  - Briefcase: 0.3.19
  - Toga: 0.4.5

freakboy3742:
Thanks for the report. This is a class of problem we've been slowly squashing over time; See #905, #1746, and some others.

If you're interested in trying your hand at fixing this, the history and linked issues for #1746 are a good point of comparison.

The fix for this comes in 2 parts. 

The first part is that we need to modify the [briefcase-android-gradle-template](https://github.com/beeware/briefcase-android-gradle-template) so that anywhere user-generated content is inserted, it is appropriately escaped.

The second part is that we need an appropriate escape filter. We have escaping filters for XML, TOML and PLIST; we clearly need one for gradle as well. This will likely be structurally similar to the `TOML` filter, except that it's the `'` character that needs escaping. 

It would also be worth auditing all the other places that user-generated content is being inserted. Based on a quick inspection, it looks like an app with a formal name or app name that contains `<` or `>` will likely cause issues, as they need to be escaped in `strings.xml`. Anywhere that a `{{ cookiecutter.something }}` value is inserted in a template, we should be ensuring that the right escaping is occurring. Unless the value is intended as a template insertion (e.g., `android_manifest_activity_attrs_extra_content`), all user-provided content should be escaped.

qilyn:
Keen to grab this one.

freakboy3742:
Please do - I'm around all tomorrow if you want to discuss any details!

PR #2068, opened 2024-11-25
Draft: Create GradleEscape extension for #1876

qilyn:
- [x] Create GradleEscape extension
- [ ] Address https://github.com/beeware/briefcase-android-gradle-template

- User-entered data should be appropriately escaped, rather than break the resulting template
-

## PR Checklist:


src/briefcase/integrations/cookiecutter.py
```diff
@@ -77,6 +77,28 @@ def escape_non_ascii(obj):
...
+class GradleEscape(Extension):
+    """Jinja2 extension to escape strings for Gradle as well."""
+
+    def  __init__(self, environment):
+        """Initialize the extension with the given environment."""
+        super().__init__(environment)
+
+        def escape_toml(obj):
+            """Escapes single quotes."""
+            return obj.replace("'", '\\"')
```
freakboy3742, line 89: Is this escaping a single quote, or replacing a single quote with an escaped double quote?

src/briefcase/integrations/cookiecutter.py
```diff
@@ -77,6 +77,28 @@ def escape_non_ascii(obj):
...
 
 
+class GradleEscape(Extension):
+    """Jinja2 extension to escape strings for Gradle as well."""
+
+    def  __init__(self, environment):
+        """Initialize the extension with the given environment."""
+        super().__init__(environment)
+
+        def escape_toml(obj):
```
freakboy3742, line 87: This method name should match the name it is declared as (i.e., `escape_gradle`)

src/briefcase/integrations/cookiecutter.py
```diff
@@ -77,6 +77,28 @@ def escape_non_ascii(obj):
         environment.filters["escape_non_ascii"] = escape_non_ascii
 
 
+class GradleEscape(Extension):
+    """Jinja2 extension to escape strings for Gradle as well."""
+
+    def  __init__(self, environment):
```
freakboy3742, line 83: This has been picked up by the pre-commit handler; there's an extra space here:
```suggestion
    def __init__(self, environment):
```

src/briefcase/integrations/cookiecutter.py
```diff
@@ -77,6 +77,28 @@ def escape_non_ascii(obj):
...
+            """Escapes single quotes."""
+            return obj.replace("'", '\\"')
+
+        # def escape_non_ascii(obj):
+        #     """Quotes obj if non ascii characters are present."""
+        #     if obj.isascii():
+        #         return obj
+        #     else:
+        #         return '"' + obj + '"'
+
```
freakboy3742, line 97: If Gradle doesn't require non-ascii escaping, we can delete this block rather than commenting it out.

freakboy3742 (changes requested):
A couple of notes inline, but this look like it's on the right track - I'm assuming the other side of this will be a template update to use this filter. 

The other issue that has been picked up by CI is the need for a changenote - we require every pull request to include the text that will become part of the release notes of the next release. The [contributor guide has some details on what is needed here](https://briefcase.readthedocs.io/en/stable/how-to/contribute-code.html#add-change-information-for-release-notes); or if you're at the PyCon AU sprints, come me in person and I'll walk you through what is needed.

qilyn:
Wasn't expecting such fast feedback on a draft MR, thanks Russell! :)

Working on it, and the template library. @pengufrost is doing great support :)

freakboy3742 (approved):
This looks great - thanks for the fix!

PR #2069, opened 2024-11-26
[1876] fix(cookiecutter): Adds XML quote and backslashes escape support

Sharealikelicence:
Adds XML quote and backslashes escape support

[#1876](https://github.com/beeware/briefcase/issues/1876)

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/cookiecutter.py
+++ b/src/briefcase/integrations/cookiecutter.py
@@ -1,6 +1,7 @@
 """Jinja2 extensions."""
 
 import uuid
+import html
 
 from jinja2.ext import Extension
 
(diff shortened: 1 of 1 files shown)


src/briefcase/integrations/cookiecutter.py
```diff
@@ -139,6 +139,11 @@ def bool_attr(obj):
             """Render value in XML format appropriate for an attribute."""
             return "true" if obj else "false"
 
+        def escape_xml(obj):
+            """Escapes single quotes and backslashes."""
+            return obj.replace("\\", "\\\\").replace("'", "\\'")
```
kesara, line 144: May be best way to escape `'` and `"` is using `"'"` or `'"'`.
>To allow attribute values to contain both single and double quotes, the apostrophe or single-quote character (') may be represented as " &apos; ", and the double-quote character (") as " &quot; ".
See https://www.w3.org/TR/REC-xml/#syntax

kesara: So best thing to do here might be replacing `'` with `&apos;`.

kesara: Also https://docs.python.org/3/library/xml.sax.utils.html#xml.sax.saxutils.quoteattr might be of help.

Sharealikelicence:
Unneeded change

PR #2082, opened 2024-12-12
WIP: Mount relative local paths in docker command args

### LINKED ISSUE #2081: Support path references external to `sources` for Docker builds in `requires` and `requirement_installer_args`

Once https://github.com/beeware/briefcase/pull/2059 is merged and released, the `requirement_installer_args` option will be available. Local, relative paths used in the option (e.g., as an argument to `--find-links`) are automatically transformed into absolute paths, which enables seamless use for all build targets **that build on the host**. Docker, however, does not build on the host, so these path references are meaningless to it.

This also affects usages of local paths in `requires`, for which Briefcase handles local paths in a manner similar to `requirement_installer_args`.

Add support for these by mounting referenced local paths into the container. Any solution for this will require tracking any local references found in `requires` or `requirement_installer_args` and exposing them in such a way that the Docker integration is able to include those mounts in calls to commands in the build container.

I can see two ways of actually mounting them:
1. Mount the host paths to a location like `/opt/{mount-name}` in the container, and re-write any usage of those paths in the command arguments to reference the location in the container. Re-writing the arguments could be risky, but Briefcase already does this when it converts relative paths to absolute paths anyway, so it doesn't seem like this would really be too unique.
2. Mount the host paths to the same location in the container, which avoids needing to re-write the actual command that is executed. It will appear as though the host's path is explicitly available to the container. However, this might also cause a subtle confusion in error messages. For example, if a command ends with an error about not being able to find a particular referenced path, and that path appears to be a host path rather than one obviously in the container's context, the path could present as a red herring, when the underlying issue is actually that the path doesn't exist anywhere at all.

The first option seems better to me, even though it's yet another level of argument re-writing. If the re-writing is implemented in `DockerAppContext._dockerize_args` itself, then it could work automatically for any path referenced on the host. In other words, it may be possible to implement this in such a way that it solves it for any command running on the host that happens to use a local path reference, rather than only for `requires` and `requirement_installer_args`. On the other hand, if more care is desired, a solution scoped specifically to just those two arguments should be explored. This is perhaps best decided in PR review, particularly if an answer isn't immediately obvious.

See above.

### Additional context

_No response_

sarayourfriend:
Fixes #2081.

Very much still WIP and ideation, and almost completely unverified, but wanted to push the branch into a draft PR to get things moving.

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/create.py
+++ b/src/briefcase/commands/create.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import hashlib
 import os
 import platform
-import re
 import shutil
 import subprocess
 import sys
@@ -25,6 +24,7 @@ from briefcase.exceptions import (
 )
 from briefcase.integrations.git import Git
 from briefcase.integrations.subprocess import NativeAppContext
+from briefcase.utils import is_local_path, is_relative_local_path
 
 from .base import BaseCommand, full_options
 
@@ -529,7 +529,7 @@ class CreateCommand(BaseCommand):
                         # If the requirement is a local path, convert it to
                         # absolute, because Flatpak moves the requirements file
                         # to a different place before using it.
-                        if _is_local_path(requirement):
+                        if is_local_path(requirement):
                             # We use os.path.abspath() rather than Path.resolve()
                             # because we *don't* want Path's symlink resolving behavior.
                             requirement = os.path.abspath(self.base_path / requirement)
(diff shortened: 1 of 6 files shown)


sarayourfriend:
I have a feeling the particular approach in https://github.com/beeware/briefcase/pull/2082/commits/50feda09a6535dde16ae76a404823be54a7c2808 is non-viable. It is not necessarily a safe assumption that _any_ local relative path in command arguments to execute inside a Docker container which matches a real local relative path on the host should be transformed, especially if, for example, `--workdir` is being changed and the path is meant to refer to a known path in the container which "just so happens" to match one on the host.

I think the solution must constrain itself to the specific inputs known to potentially contain local relative paths, `requires` and `requirement_installer_args`.

PR #2096, opened 2024-12-31
Replace use of build-essential with constituent parts.

freakboy3742:
When building a system package on a Debian derivative, Briefcase checks for the `build-essential` meta-package as a pre-requisite. This ensures that gcc, g++, libc6-dev, make, and dpkg-dev are installed. 

However, it is also possible to satisfy these requirements by installing those packages directly. This was done recently on GitHub Actions configurations, causing build failures (see beeware/toga#3051). 

The immediate workaround is "just install the meta-package"; but installing a no-op package isn't a great option. This PR modifies the runtime check to look for the actual constituent packages.

The Dockerfiles will continue to install `build-essential`; but they will pass the system check because installing that meta-package will cause these constituent packages to be installed. However, we'll be able to verify an environment that doesn't have these packages installed; and the help message presented to the user will reference the meta-package, not the constituent packages.

Fixes beeware/toga#3051.

## PR Checklist:

### DIFF
--- a/changes/2096.bugfix.rst
+++ b/changes/2096.bugfix.rst
@@ -1,0 +1,1 @@
+A Debian-based system that does *not* have ``build-essential`` installed, but *does* have the constituent packages of ``build-essential`` installed, can now build Briefcase system packages.
--- a/src/briefcase/platforms/linux/system.py
+++ b/src/briefcase/platforms/linux/system.py
@@ -479,11 +479,11 @@ class LinuxSystemMostlyPassiveMixin(LinuxSystemPassiveMixin):
             base_system_packages = [
                 "python3-dev",
                 # The consitutent parts of build-essential
-                "dpkg-dev",
-                "g++",
-                "gcc",
-                "libc6-dev",
-                "make",
+                ("dpkg-dev", "build-essential"),
+                ("g++", "build-essential"),
+                ("gcc", "build-essential"),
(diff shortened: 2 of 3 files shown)


rmartin16 (approved):
I found it isn't actually practical to test this. The dependencies for Toga ultimately require `build-essential` to be installed. So, one can't easily use `apt` to get a Debian system in to a state where Toga works but `build-essential` is not installed. Nonetheless, this works for me on jammy, noble, oracular, and even plucky.

freakboy3742:
> I found it isn't actually practical to test this. The dependencies for Toga ultimately require `build-essential` to be installed. So, one can't easily use `apt` to get a Debian system in to a state where Toga works but `build-essential` is not installed. 

Agreed that it's difficult to test - However, I'm not sure what you mean by this comment. Isn't that what Github Actions is doing (installing gcc et al, but not build-essential)?

rmartin16:
Yeah; thinking about it more, I was able to contrive such an environment. Notably, install `toga-textual` instead, remove `system_requires`, and use `apt install --no-install-recommends dpkg-dev g++ gcc libc6-dev make` to get the environment set up since `build-essential` is a recommended package of one them.

The GitHub runner image appears rather novel...seems like they are definitely cherry-picking packages to install.

freakboy3742:
Ok - in which case, I'll call this working and merge. Thanks for the extra manual validation.

PR #2101, opened 2025-01-06
Modify pip handling to use platform tag for iOS.

freakboy3742:
With the release of pip 24.3, we're now in a position to use explicit `--platform` tags when installing wheels for iOS.

This removes the need for the platform-site shim to "fake" pip installation; we can instead rely on default pip/Python platform handling. 

As part of this change, the use of wheels is now *required*, through use of the `--only-binary=:all:`. This is required as part of using `--platform`, as pip can't install from source when installing onto a foreign target platform, even if the wheel being installed is pure Python. There is a small minority of packages that are only packaged as source tarballs; but the broader Python ecosystem is firmly promoting the use of wheels, so this seems like a good opportunity to force the issue.

## PR Checklist:

### DIFF
--- a/changes/2101.feature.rst
+++ b/changes/2101.feature.rst
@@ -1,0 +1,1 @@
+Briefcase now uses native pip handling for cross-platform installs, rather than the site-based shim.
--- a/docs/reference/platforms/iOS/xcode.rst
+++ b/docs/reference/platforms/iOS/xcode.rst
@@ -133,13 +133,12 @@ PyPI to provide those wheels. Briefcase uses a `secondary repository
 
 This repository is maintained by the BeeWare project, and as a result, it does not have
 binary wheels for *every* package that is available on PyPI, or even every *version* of
-every package that is on PyPI. If you see any of the following messages when building an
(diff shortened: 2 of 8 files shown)


mhsmith (approved):

mhsmith:
I patched Chaquopy's copy of pip to deal with this, though at that time `sys.platform` returned "linux" on Android:

* https://github.com/chaquo/chaquopy/blob/16.0.0/product/gradle-plugin/src/main/python/pip/_internal/cli/cmdoptions.py#L557
* https://github.com/chaquo/chaquopy/blob/16.0.0/product/gradle-plugin/src/main/python/pip/_vendor/packaging/markers.py#L291

freakboy3742:
That's a good point - looks like we can't abandon the platform site package after all. `--platform` will guarantee a matching binary wheel - but we still need a cross-platform-enabled `pip` to do the install. 

Looking at the bigger picture - I wonder if it would be possible to use `PYTHON_HOST_SYSCONFIG` as a configuration point, store the host overrides in the platform-specific sysconfig, and have them incorporated into pip (or anything else that needs it) as required...

PR #2110, opened 2025-01-15
Improve macOS notarization tools

### LINKED ISSUE #1218: maybe briefcase might hypothetically generate an archive invalid for notarization sometimes when using a zip archive for mac app distribution, and should probably use the recommended command to generate them
### Describe the bug

I do not technically know that this bug affects `shutil.make_archive`, but it does affect the operating system's built-in `zip` command, so this *might* not affect briefcase, but reproducing it is weird and difficult, and I *did* hit it with some builds of an app with Encrust, so I am leaving this here regardless.

Anyway, the magic password is:

```
ditto -c -k --sequesterRsrc --keepParent MyApp.app MyApp.app.zip
```

The bug is described here:

https://blog.frostwire.com/2019/08/27/apple-notarization-the-signature-of-the-binary-is-invalid-one-other-reason-not-explained-in-apple-developer-documentation/

The explanation — that the paths themselves are being re-encoded with a different Unicode normalization — does not make sense for my application, because I do not have any non-ASCII in any of the paths in my app. It seems to have something to do with the code signature itself, or some other thing in a resource fork somewhere.  Switching to `ditto` did reliably fix it, though.

### Steps to reproduce

I dunno, maybe try to notarize a zip file who has an embedded signature that doesn't cleanly decode as UTF-8?

### Expected behavior

N/A

### Screenshots

_No response_

### Environment

N/A

### Logs

N/A

### Additional context

This seems heavily dependent on the content of the app being signed and the time at which it is signed, so I have no idea how to submit a reproducer.

### LINKED ISSUE #1472: Print notarization log on failure

freakboy3742:
Makes a number of improvements to Briefcase's wrapper around app notarization for macOS.

1. Replaces the use of `shutil.make_archive` with `ditto`. There is an edge case of app notarization where UTF-8 encoding of some file system metadata can result in notarization failures; this is because the default zip tools don't guarantee that these file system metadata are preserved correctly in the archive. Apple provides a tool called "ditto" that produces zip archives, but guarantees the integrity of file system metadata. The use of ditto is strongly recommended by Apple anyway, so this is broadly a good idea.
2. The time required to notarize an app is highly variable - it can take *hours*, depending on the load at Apple's end. The current notarization tooling requires users to leave a terminal open, and in a position to continuously contact Apple's servers, which is less than ideal. This PR submits the app for notarization, then outputs a submission ID, before going into a polling loop checking for notarization approval. If this loop is interrupted, the user can provide the submission ID as a command line argument to resume the previous notarization pass.
3. If notarisation fails, Briefcase will now output a summary of the notarization problems that have been reported.
4. The filename used when submitting a bare app for notarization now uses the app's formal name. This is mostly for the benefit of debugging when notarization history is inspected - the current name of "archive.zip" doesn't tell you which app has been archived.
5. Some of the helper messages related to selecting identities have been clarified, ensuring that they accurately reflect the context in which they were invoked
6. If verbosity is set to level 3 (i.e., `-vvv`), any multi-threaded tasks are reduced to a single thread. This makes it easier to identify failures when they occur in the multi-threaded section of code (such as app signing).

Fixes #1218 
Fixes #1472

## PR Checklist:


src/briefcase/platforms/macOS/__init__.py
```diff
@@ -755,28 +772,129 @@ def verify_app(self, app):
...
+        :param filename: The file to notarize.
+        :param identity: The code signing used to notarize the app.
+        :param installer_identity: The signing identity to use when signing the
+            installer. Optional unless the packaging format is ``pkg``.
+        """
+        # Determine the arguments that would be needed to reproduce this notarization
+        if installer_identity:
+            identity_args = (
+                f"--identity {identity.id} "
+                f"--installer_identity {installer_identity.id}"
```
mhsmith, line 846:
```suggestion
                f"--installer-identity {installer_identity.id}"
```

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -840,19 +957,21 @@ def notarize(self, filename, identity: SigningIdentity):
...
-                            "submit",
-                            archive_filename,
-                            "--keychain-profile",
-                            profile,
-                            "--wait",
-                        ],
-                        check=True,
-                    )
-                    submitted = True
+                    with self.input.wait_bar("Submitting app for notariztion..."):
```
mhsmith, line 960:
```suggestion
                    with self.input.wait_bar("Submitting app for notarization..."):
```

freakboy3742:
> I tried notarizing all 3 formats, but they all failed in different ways.
> As discussed in person, this was fixed by upgrading dmgbuild from 1.6.2 to 1.6.4. Notarization then succeeded, both with and without an interruption.

Corrected by bumping the minimum version requirement from `dmgbuild`.

> .pkg returned some signature errors:

I swear I tested this, but clearly I didn't - I must have checked that it started the notarization wait process, then killed the check and not waited for completion, assuming that if it works for `dmg`, it musts work for `pkg` as well.

It turns out the individual signatures were fine - but when we build the `pkg` file, we duplicate the app into a new location using `shutil.copytree()` ... which unrolls symlinks by default. As a result, when the symlinks are unrolled, the *overall* signature for frameworks (and then the binary, and then the app) fail because the file manifest has changed since it was signed.

This wasn't an issue prior to Briefcase 0.3.20, as the support package didn't contain any symlinks. The most recent support packages use dynamic linking and a repackaged version of the official macOS distribution, so there are symlinks in the Framework package.

> .zip completed successfully when not interrupted, but failed to resume after an interruption:

This was an edge case that didn't get caught because I was doing lots of repeated testing, so I end up in a situation where the file it was looking for *did* exist as a result of a previous testing run. However, this pointed out a bunch of edge cases caused by the fact that .zip distribution artefacts aren't created until *after* notarisation is complete (since it's the .app that is stapled); I've reworked things to address this.

mhsmith (approved):

Issue #1898, opened 2024-07-07
Do not override existing `pyproject.toml` keys during conversion


rmartin16:
### What is the problem or limitation you are having?

When an existing project is converted, keys should not be added to `[tool.briefcase]` when they already exist in `[project]`. For instance, if `version` is already specified in `[project]`, do not add `version = "0.0.1"` to `[tool.briefcase]`. This should be considered for all PEP-621 `pyproject.toml` keys.

### Describe the solution you'd like

Omit `[tool.briefcase]` keys from `pyproject.toml` that are already represented in `[project]`.

### Describe alternatives you've considered

Manually remove overriding keys from `[tool.briefcase]` post-conversion.

freakboy3742:
Conceptually a good idea; but could be difficult to implement; the mechanism we're using to append `pyproject.toml` is (AFAIR) concatenation, rather than parsing and processing, so removing duplicate keys requires some pre-processing or a formatting-sensitive parser.

It may be better to consider this part of a broader migration to PEP621 format, so that a `version` field in `[tool.briefcase]` is no longer required or preferred.

CarliJoy:
I think this is solved already by https://github.com/beeware/briefcase/pull/2772

freakboy3742:
@CarliJoy Not really. This issue is about the `convert` command, which takes an existing project with a `pyproject.toml` configuration and adds a Briefcase configuration. In the process, it duplicates some of the keys (like `version`) which *can* be picked up from the `[project]` section, but can't easily be removed from the templated `[tool.briefcase]` section that the `convert` command adds to the existing configuration.

CarliJoy:
🤦🏻  Sorry my bad.

PR #2119, opened 2025-01-18
Add a post-generate extension point for bootstraps.

### LINKED ISSUE #1899: Include all Briefcase tables in `pyproject.toml` during conversion

When an existing project is converted to use Briefcase, only the `[tool.briefcase]` and `[tool.briefcase.app.<app name>]` tables are populated. This makes it more difficult to take advantage of Briefcase's packaging controls because all of the those `pyproject.toml` tables must be created manually.

Populate `pyproject.toml` with the platform-specific tables when converting an existing project. At the least, I see benefit from creating the empty tables; however, there are likely at least some keys that are worth exposing for all projects.

Manually create all platform tables in `pyproject.toml`.

### Additional context

_No response_

### LINKED ISSUE #2006: Provide a more extensive default bootstrap

freakboy3742:
In the process of working on #1288, I discovered another extension point that is required on bootstraps - a "post generation" step that is invoked after the template has been generated. This is required because the Briefcase template has a fixed set of files; if you need to write an additional file (for example, to write an `index.html` for an embedded web server), there's not currently any way to do that with the bootstrap.

This requires a slight reorganisation to how the bootstrap is created, so that the bootstrap instance can be retained until after the template is generated. However, by doing this, it becomes helpful for there to *always* be a bootstrap... which also provides an opportunity to address #2006.

Makes 2 other small changes to the Console bootstrap:
1. Adds some additional documentation to the code generated by the Console bootstrap to match that generated by new Empty bootstrap
2. Renames the override variable for specifying a console app in the convert command. This hasn't been in an official release, so there's no backwards compatibility concerns; but specifying "-Q console_app=GUI` doesn't make any sense from a UX point of view.

Refs #1288
Fixes #1899
Fixes #2006

## PR Checklist:


src/briefcase/bootstraps/base.py
```diff
@@ -120,3 +121,12 @@ def pyproject_table_web(self) -> str | None:
 
     def pyproject_extra_content(self) -> str | None:
         """Additional TOML to add to the bottom of pyproject.toml."""
+
+    def post_generate(self, base_path: Path) -> None:
+        """Code to run after the template has been generated.
```
rmartin16, line 126:
```suggestion
        """Runs after the template has been generated.
```

src/briefcase/commands/convert.py
```diff
@@ -326,10 +327,10 @@ def input_url(self, app_name, override_value: str | None) -> str:
 
         return url
 
-    def input_console_app(self, override_value: str | None) -> bool:
+    def input_app_type(self, override_value: str | None) -> bool:
```
rmartin16, line 330: I was a bit caught off-guard when this didn't work:
```
briefcase convert -Q app_type=console
```
```
-- Interface Style -----------------------------------------------------------

Using override value 'console'

Invalid value; console
Invalid override value 'console' for Interface Style, using user-provided value.

Is this a GUI application or a console application?

  1) GUI
  2) Console

Interface Style [1]:
```
Maybe better to make these comparisons case-insensitive....

rmartin16 (approved):
lgtm w/ some minor comments.

This also addresses https://github.com/beeware/briefcase/issues/1899. And it may make https://github.com/beeware/briefcase/issues/1898 even more difficult to address....although, as you said, a more general approach with PEP-621 will likely reshapre that whole conversation.

freakboy3742:
> This also addresses #1899.

Good catch - I've added a changenote and a  "Fixes" annotation in the PR body. 

>  And it may make #1898 even more difficult to address....although, as you said, a more general approach with PEP-621 will likely reshapre that whole conversation.

Yeah - I think the real fix for #1898 is to stop using keys that are already covered by PEP 621, which is a much bigger project. I'm not sure this makes #1898 substantially *worse* though - the only project-wide keys it's added are ones that are cumulative; and the platform-specific keys are all either cumulative, or not covered by PEP 621.

PR #2124, opened 2025-01-20
Unify Printer, Logger and Console classes

freakboy3742:
Following the discussion in #2115 - unifies the `Printer`, `Logger` and `Console` classes into a single Console class. 

This allows a single representation of "interaction with the user", capturing both input and output, which also writes to a log file as required. It also cleans up some odd API usage (such as needing to invoke `input.wait_bar()`, and the `__call__` interface on input (it can now be replaced with an `input()` method.

~This PR builds on #2115 to avoid collisions with that work. You could either review, approve and merge this PR by itself; or review and merge #2115, then rebase, review and merge this one.~ UPDATE: PR has now been rebased onto main.

It is, unfortunately, a very large PR... but most of the changes are renames.

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/new.py
+++ b/src/briefcase/commands/new.py
@@ -528,7 +528,7 @@ class NewCommand(BaseCommand):
         bootstrap = self.create_bootstrap(context, project_overrides)
         context.update(self.build_gui_context(bootstrap, project_overrides))
 
-        self.console.debug()  # close the prompting section of output
+        self.console.divider()  # close the prompting section of output
 
         self.warn_unused_overrides(project_overrides)
 
--- a/src/briefcase/commands/new.py
+++ b/src/briefcase/commands/new.py
@@ -127,7 +127,7 @@ class NewCommand(BaseCommand):
         """
         if not make_class_name(candidate):  # Check whether a class name may be derived
             raise ValueError(
-                self.input.textwrap(
+                self.console.textwrap(
                     f"{candidate!r} is not a valid formal name.\n"
                     "\n"
                     "Formal names must include at least one valid Python identifier character."
@@ -143,7 +143,7 @@ class NewCommand(BaseCommand):
         """
         if (self.base_path / candidate).exists():
             raise ValueError(
-                self.input.textwrap(
+                self.console.textwrap(
                     f"A {candidate!r} directory already exists.\n"
                     f"\n"
                     f"Select a different name, move to a different parent directory, or "
@@ -160,7 +160,7 @@ class NewCommand(BaseCommand):
         """
         if not is_valid_app_name(candidate):
             raise ValueError(
-                self.input.textwrap(
+                self.console.textwrap(
                     f"{candidate!r} is not a valid app name.\n"
                     "\n"
                     "App names must not be reserved keywords such as 'and', 'for' and "
@@ -192,7 +192,7 @@ class NewCommand(BaseCommand):
         """
         if not is_valid_bundle_identifier(candidate):
             raise ValueError(
-                self.input.textwrap(
+                self.console.textwrap(
                     f"{candidate!r} is not a valid bundle identifier.\n"
                     "\n"
                     "The bundle should be a reversed domain name. It must contain at "
@@ -253,7 +253,7 @@ class NewCommand(BaseCommand):
         return f"https://{self.make_domain(bundle)}/{app_name}"
 
     def input_project_name(self, formal_name, override_value):
-        return self.input.text_question(
+        return self.console.text_question(
             intro=(
                 "Briefcase can manage projects that contain multiple applications, so "
                 "we need a Project name.\n"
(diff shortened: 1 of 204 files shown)


freakboy3742:
Apologies in advance for this monster :-)

rmartin16 (commented):
The important bits in `console.py` look good. Since we're delineating the groups of related methods with "hashtag boxes", do you want to add one above `textwrap()` for "user input and output formatting"....or something like that :)?

Otherwise, I'm hoping to get through this tomorrow.

src/briefcase/console.py
rmartin16, line 1: So, I started to look more closely at what this input API is looking like:

**Input API**
- Primitives
  - `input()`
  - `prompt()`
  - `divider()`
  - `textwrap()`
- Basic Solicitation
  - `_selection()`
  - `input_text()`
  - `input_boolean()`
- Advanced Solicitation
  - `text_question()`
  - `selection_question()`

Should we round this out?

For example, we provide the primitives to recreate framing and prompting for arbitrary text; but we aren't exposing the basic selection prompt logic in `_selection()`. Should we convert that to `input_selection()`?

Going the other way, we are providing APIs to fully frame and ask text & selection questions but not boolean. Should we create a `boolean_question()` API as well?

src/briefcase/console.py
```diff
@@ -920,20 +935,21 @@ def selection_question(
...
-            input_value = self.input(escaped_prompt, markup=True)
+            # Use underlying Rich input() to read from user
+            input_value = self._console_impl.input(escaped_prompt, markup=True)
         except EOFError:
             raise KeyboardInterrupt
 
-        self.print.to_log(prompt)
-        self.print.to_log(f"User input: {input_value}")
+        self.to_log(prompt)
+        self.to_log(f"User input: {input_value}")
```
rmartin16, line 953: One thing I haven't been able to try is seeing if this works in a bootstrap.

If you run `python -c 'import briefcase.console; briefcase.console.Console().input("is anybody out there?")'`, for instance, this'll error.

rmartin16 (commented):
A few comments while I'm making my way.

src/briefcase/commands/new.py
```diff
@@ -525,18 +521,18 @@ def new_app(
...
-        self.input.prompt("Let's build a new Briefcase app!")
+        self.console.prompt()
+        self.console.prompt("Let's build a new Briefcase app!")
 
         context = self.build_app_context(project_overrides)
         bootstrap = self.create_bootstrap(context, project_overrides)
         context.update(self.build_gui_context(bootstrap, project_overrides))
 
-        self.input.divider()  # close the prompting section of output
+        self.console.debug()  # close the prompting section of output
```
rmartin16, line 531:
```suggestion
        self.console.divider()  # close the prompting section of output
```

src/briefcase/console.py
freakboy3742: I thought about `_selection`->`input_selection()`; I ended up making it an internal method because I couldn't see a use case for it as written... but I guess the same could also be said for `input_text()`. At the very least, using `input_selection()` would be consistent from a naming perspective, so I'll make that change. 

As for `boolean_question()` - I agree that it would make sense from the perspective of symmetry... but we don't actually have a use case for it. I'm inclined to call that "future work" and maybe log it as a first timer ticket.

src/briefcase/commands/new.py
freakboy3742: The same error exists in convert; I've fixed it there as well.

src/briefcase/console.py
rmartin16: I was able to get that bootstrap to work with some massaging of things.

It does look like `Console.input()` will not crash for foreign bootstraps since the frame stack is deep enough...but the frame reference is wrong unfortunately. This would probably be annoying to resolve, though....and doesn't really provide that much useful information.

Added a prompt to the static site bootstrap:
```python
     def extra_context(self, project_overrides):
         self.console.input_text("How are you?")
```
Got this in the log:
```
[20:26:54] How are you?                                                                                                                                                   new.py:467
           User input: asdf                                                                                                                                               new.py:467
```

src/briefcase/console.py
rmartin16: Created https://github.com/beeware/briefcase/issues/2128.

rmartin16:
Looks like `tests/console/Console/test_selection.py` needs to be updated for `console._selection()` but I think I'm good with this. I didn't review all of the changes in `tests` like I did `src`....but if anything isn't causing the tests to fail, we'll have to fix it later :)

rmartin16 (approved):

freakboy3742:
> Looks like `tests/console/Console/test_selection.py` needs to be updated for `console._selection()` but I think I'm good with this. I didn't review all of the changes in `tests` like I did `src`....but if anything isn't causing the tests to fail, we'll have to fix it later :)

/me grumbles... I thought I ran those tests...

PR #2126, opened 2025-01-21
Corrections to bootstrap handling

freakboy3742:
A bootstrap post_generate step was added in #2119, but there was not explicit test for the step. As a result, it missed the fact that in the context of `new`, `base_path` is the path where the *parent* of the path that is usually the base_path (i.e., the path that contains the `pyproject.toml`), because the `new` command is responsible for *creating* the folder that contains pyproject.toml.

Since you'll always be wanting to post_generate content in the project folder, this adds the extra path element - and adds a test that the path is correct.

This was revealed when I re-tested my prototype Positron bootstrap against the merged version of the code. 

## PR Checklist:

### DIFF
--- a/changes/2126.misc.rst
+++ b/changes/2126.misc.rst
@@ -1,0 +1,1 @@
+The base path provided to post_generate was corrected to include the app name.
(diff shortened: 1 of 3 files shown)


rmartin16 (approved):
obvious in retrospect...but I don't think I ever considered `base_path` is  different for the `new` command....probably even more important for `convert` command.

PR #2149, opened 2025-02-06
Correct pip usage and notarization logging.

freakboy3742:
Fixes 2 small issues:

1. Removes the use of the `--no-python-version-warning`. pip 25.0 displays the following warning if it is used:
```
DEPRECATION: --no-python-version-warning is deprecated. pip 25.1 will enforce this behaviour change. A possible replacement is to remove the flag as it's a no-op. Discussion can be found at https://github.com/pypa/pip/issues/13154
```
2. Adds `quiet=1` to two uses of `parse_output` in the notarization code. When we introduced #2125, we added `quiet=1` to all uses of `check_output()` where a error code was normal expected operation, and added a call to `output_error()` if the error received wasn't an expected value. We audited the code for all `check_output()` calls... but forgot that `parse_output()` builds on `check_output()`. The notarisation code includes 2 such uses; there are no others that I can find.

Given the 25.1 release could happen at any time, it may be advisable to push out a new release after merging this PR. The quiet=1 fix isn't a major issue, but it would be nice to clean up at the same time.

## PR Checklist:


mhsmith (approved):
> pip 25.1 will enforce this behaviour change

I've [asked them](https://github.com/pypa/pip/pull/13180) about the fact that this appears to be earlier than their deprecation policy allows.

> Given the 25.1 release could happen at any time

It looks like they still have a [long list in the milestone](https://github.com/pypa/pip/milestone/78), so it probably isn't imminent.

freakboy3742:
> > Given the 25.1 release could happen at any time
> 
> It looks like they still have a [long list in the milestone](https://github.com/pypa/pip/milestone/78), so it probably isn't imminent.

True - but given we're prone to months between releases, it would be easy to be bitten by pip releasing before our own informal release process decides it's time for 0.3.21.

PR #2187, opened 2025-03-09
Include girepository-2.0 as a system depdendency for Toga on Debian

freakboy3742:
Adds libgirepository-2.0-dev as a system dependency for Toga apps on Debian-based releases. This is required by 
PyGObject 3.52.1. 

This library isn't available on Debian 12/Ubuntu 22.04. To accomodate those releases, the workaround is included in the template, commented out.

Refs beeware/toga#3143

## PR Checklist:

### DIFF
--- a/changes/2187.bugfix.rst
+++ b/changes/2187.bugfix.rst
@@ -1,1 +1,1 @@
-New Toga projects on Debian-derived distributions now include ``libgirepository-2.0`` and ``libgirepository-2.0-dev`` as a system requirement. This is needed for support for PyGobject 3.52.1 and later.
+New Toga projects on Debian-derived distributions now include ``libgirepository-2.0-0`` and ``libgirepository-2.0-dev`` as a system requirement. This is needed for support for PyGObject 3.52.1 and later.
--- a/changes/2187.bugfix.rst
+++ b/changes/2187.bugfix.rst
@@ -1,1 +1,1 @@
-New Toga projects now enforce an upper version bound of ``pygobject < 3.52.1``. PyGObject 3.52.1 introduced a system requirement of ``libgirepository-2.0-dev``; that library isn't available on Debian 12 or Ubuntu 22.04 LTS. If you have a project that isn't targeting those platforms, you can likely remove this pin.
+New Toga projects on Debian-derived distributions now include ``libgirepository-2.0`` and ``libgirepository-2.0-dev`` as a system requirement. This is needed for support for PyGobject 3.52.1 and later.
--- a/changes/2187.bugfix.rst
+++ b/changes/2187.bugfix.rst
@@ -1,0 +1,1 @@
+New Toga projects now enforce an upper version bound of ``pygobject < 3.52.1``. PyGObject 3.52.1 introduced a system requirement of ``libgirepository-2.0-dev``; that library isn't available on Debian 12 or Ubuntu 22.04 LTS. If you have a project that isn't targeting those platforms, you can likely remove this pin.
--- a/src/briefcase/bootstraps/toga.py
+++ b/src/briefcase/bootstraps/toga.py
@@ -60,18 +60,15 @@ requires = [
         return """\
 requires = [
     "toga-gtk~=0.4.7",
-]
-"""
-
-    def pyproject_table_linux_system_debian(self):
-        return """\
-requires = [
     # PyGObject 3.52.1 enforces a requirement on libgirepository-2.0-dev. This library
     # isn't available on Debian 12/Ubuntu 22.04. If you need to support those (or older)
     # releases, uncomment this version pin. See beeware/toga#3143.
     # "pygobject < 3.52.1",
 ]
+"""
 
+    def pyproject_table_linux_system_debian(self):
+        return """\
 system_requires = [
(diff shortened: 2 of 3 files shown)


src/briefcase/bootstraps/toga.py
```diff
@@ -60,6 +60,11 @@ def pyproject_table_linux(self):
         return """\
 requires = [
     "toga-gtk~=0.4.7",
+    # PyGObject 3.52.1 enforces a requirement on libgirepository-2.0-dev. This library
+    # isn't available on Debian 12/Ubuntu 22.04. Enforce an upper pin that prevents that
+    # release (or newer) from being selected. If you don't need to support older
+    # Debian/Ubuntu releases, you can likely remove this pin. See beeware/toga#3143.
+    "pygobject < 3.52.1",
```
rmartin16, line 67: FWIW, the _only_ PyGObject release less than `3.52.1` that works with `toga >= 0.4.7` is `3.50.0`. Prior versions of pygobject require gbulb and later versions have this new libgirepository-2.0 dependency.

freakboy3742: Sure - but that will get picked by the version bound math. `pygobject >= 3.50.0, < 3.52.1` will always resolve to 3.50.0.

changes/2187.bugfix.rst
```diff
@@ -0,0 +1 @@
+New Toga projects on Debian-derived distributions now include ``libgirepository-2.0`` and ``libgirepository-2.0-dev`` as a system requirement. This is needed for support for PyGobject 3.52.1 and later.
```
rmartin16, line 1:
```suggestion
New Toga projects on Debian-derived distributions now include ``libgirepository-2.0-0`` and ``libgirepository-2.0-dev`` as a system requirement. This is needed for support for PyGObject 3.52.1 and later.
```

src/briefcase/bootstraps/toga.py
```diff
@@ -65,17 +65,33 @@ def pyproject_table_linux(self):
 
     def pyproject_table_linux_system_debian(self):
         return """\
+requires = [
+    # PyGObject 3.52.1 enforces a requirement on libgirepository-2.0-dev. This library
+    # isn't available on Debian 13/Ubuntu 22.04. If you need to support those (or older)
```
rmartin16, line 70:
```suggestion
    # isn't available on Debian 12/Ubuntu 22.04. If you need to support those (or older)
```

src/briefcase/bootstraps/toga.py
```diff
@@ -65,17 +65,33 @@ def pyproject_table_linux(self):
...
+    # "pygobject < 3.52.1",
+]
+
 system_requires = [
     # Needed to compile pycairo wheel
     "libcairo2-dev",
-    # Needed to compile PyGObject wheel
-    "libgirepository1.0-dev",
+    # One of the following two packages are needed to compile PyGObject wheel.
+    # Debian 13/Ubuntu 22.04 only has the older 1.0 version of the library.
```
rmartin16, line 79:
```suggestion
    # Debian 12/Ubuntu 22.04 only has the older 1.0 version of the library.
```

src/briefcase/bootstraps/toga.py
```diff
@@ -65,17 +65,33 @@ def pyproject_table_linux(self):
...
-    "libgirepository-1.0-1",
+    # One of the following two packages are needed to use PyGObject at runtime.
+    # Debian 13/Ubuntu 22.04 only has the older 1.0 version of the library.
+    # If you use the older version, you'll also need to apply the pygobject pin
+    # (see the linux requires table). See beeware/toga#3143.
+    "libgirepository-2.0-0",
+    # "libgirepository-1.0-1",
     # Dependencies that GTK looks for at runtime
     "libcanberra-gtk3-module",
     # Needed to provide WebKit2 at runtime
```
rmartin16, line 97: We can remove the note below about Debian 11 and 20.04; they aren't supported anymore anyway.

rmartin16 (changes requested):
couple minor notes

src/briefcase/bootstraps/toga.py
```diff
@@ -65,17 +65,33 @@ def pyproject_table_linux(self):
...
+    "libgirepository-2.0-dev",
+    # "libgirepository1.0-dev",
 ]
 
 system_runtime_requires = [
     # Needed to provide GTK and its GI bindings
     "gir1.2-gtk-3.0",
-    "libgirepository-1.0-1",
+    # One of the following two packages are needed to use PyGObject at runtime.
+    # Debian 13/Ubuntu 22.04 only has the older 1.0 version of the library.
```
freakboy3742, line 90:
```suggestion
    # Debian 12/Ubuntu 22.04 only has the older 1.0 version of the library.
```

src/briefcase/bootstraps/toga.py
rmartin16: Oh, I guess it's just Ubuntu 20.04 that could be removed; looks like we still support Debian 11. Ubuntu 22.04 doesn't work anymore, though, since it uses Python 3.8.

We can worry about this a different time, though; more pressing matters right now.

Issue #2128, opened 2025-01-22
Add `boolean_question()` API to `Console`


rmartin16:
### What is the problem or limitation you are having?

in #2124, we added a set of APIs for soliciting user input. We added `text_question()` and `selection_question()` but didn't include `boolean_question()` at the time....but its ultimate inclusion is worthy.

### Describe the solution you'd like

Along with `text_question()` and `selection_question()`, we should add `boolean_question()`; it would have the same semantics but only accept boolean responses.

### Describe alternatives you've considered

Boolean question framing and prompting can be done manually.

dkarvan8:
Hi, I want to work on this issue. Can you assign it to me?

mhsmith:
We don't use the GitHub assignment feature – posting a comment is enough.

You've posted this comment on 5 issues today, but I recommend you only work on one at a time until you're more familiar with the project.

dkarvan8:
Apologies for commenting on multiple issues earlier—I was just eager to contribute as I’m new to open-source contributions. I appreciate the guidance, and **I’ll focus on this specific issue first**. Looking forward to working on it

mEp3ii2:
This enhancement has been implemented now
https://github.com/beeware/briefcase/pull/2198 
Can we close the issue, I didn't seem to link them properly unfortunately

freakboy3742:
Yup - we can close. The PR needs to include the magic text `Fixes #XXX` to auto-close the issue (or `Refs #xxx` to include an issue reference other than a "mention" in the timeline). Thanks staying on top of the housekeeping!

PR #2198, opened 2025-03-17
Add boolean question

mEp3ii2:
Added a boolean_question API to the console, allowing users to be prompted with a Yes/No Question

Previously, there was no dedicated method for handling boolean questions in Console.
This PR, provides a structured way to prompt users following the example from text_Question and selection_Question.

Add boolean_question() API to Console #2128
https://github.com/beeware/briefcase/issues/2128 

## PR Checklist:


src/briefcase/console.py
freakboy3742, line 1: As a heads up - this is rendering as a change to every line of the file. I suspect this may be because you're editing on Windows, and your IDE and/or git has modified every line to use Windows-style CRLF line endings rather Unix-style CR line endings. Take a look at [this article](https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings) about configuring line endings on Windows.

mEp3ii2: I've made modifications to the console.py to amend this and I've also done the same for the test_boolean_question.py file as well to avoid any future issues

src/briefcase/console.py
```diff
@@ -960,3 +960,45 @@ def selection_question(
...
+            )
+
+        self.divider(title=description)
+
+        if override_value is not None:
+            self.print()
+            self.print(f"Using override value {override_value!r}")
+            if isinstance(override_value, bool):
+                return override_value
+            else:
```
freakboy3742, line 995: The error handling here is a good catch, but I think it might be better to take a different approach because of how the API will be used in practice.

The role of `override_value` is to allow the user to provide the answers for the wizard at the command line. We added this feature mostly for our own testing purposes - so we can "generate a project" in a test case without needing to intercept/fake stdin. This means you can run something like `briefcase new -Q formal_name=Foobar`, and generate a new project that has a formal name of "Foobar". 

Essentially we're shortcutting the value that will be read from stdin - which will *always* be a string.

So - when we get to boolean questions, the format for specifying the override would be `-Q should_explode=False`... but in this case the value will be the *string* `"False"`, not a boolean. 

This *could* be handled by the caller - that is, requiring the user calling `boolean_question` to convert the literal string value into a boolean value when calling the method - but that then requires every user of `boolean_question` to remember that this parsing is required, and to invent their own parsing mechanism. From a developer experience perspective, it's better to accept the input in the form it's likely to be provided, and incorporate the parsing piece into this method.

That then opens the question of what value should be parsed? If the user is providing input, they're asked to type `y/n`... so `-Q should_explode=N` should probably be expected - but given it's a boolean, I can't see any reason we shouldn't be as liberal as possible here, allowing `True`, `true`, `TRUE`, `TrUe` `Yes`, `YES`, `Y`, `1`, or `on` as a True values, and similarly for "False", "FALSE" etc.

Python will technically allow "" as a False value, and any non-empty string as a True value; however, I think it would be advisable to diverge from Python's behavior in this case. Given this is a user input, having an *explicitly* True/False value will avoid errors like `-Q should_explode=$EXPLODE` being interpreted as False because the shell didn't define `EXPLODE`.

src/briefcase/console.py
```diff
@@ -960,3 +960,45 @@ def selection_question(
...
+        :param default: The default option for empty user input.
+        :param override_value: A pre-selected answer for the question. This can be used
+            to shortcut asking the question, such as when a command line option provides
+            a value. If provided and valid, the header bar will be displayed, but the
+            intro paragraph and option list will not.
+        :returns: The user's chosen answer or none if closed without input
+        """
+        if default not in {True, False, None}:
+            raise ValueError(
+                f"'{default}' is not a valid default value for {description}? Must be True, False, or None."
```
freakboy3742, line 985: Adding this handling is a good instinct, but not required in this case. The type declaration forces this to True/False/None; plus, at runtime, Python will duck type the value. Anything that quacks like a True value will behave like "True".

changes/2128.feature.rst
```diff
@@ -0,0 +1 @@
+Add `boolean_question()` API to Console.
```
freakboy3742, line 1: As I mentioned during our call this week, the release note should be in the voice of "advertising/marketing copy", not a commit message:
```suggestion
It is now possible to ask boolean questions using the Console interface.
```

tests/console/test_boolean_question.py
```diff
@@ -0,0 +1,119 @@
...
+    """Test that boolean_question returns False when user selects No."""
+    console = DummyConsole("n")
+
+    result = console.boolean_question(
+        description="Confirm?",
+        intro="Are you sure?",
+        default=None,
+    )
+
+    assert console.prompts == ["Confirm? y/n? "]
```
freakboy3742, line 32: This is an interesting discrepancy - it's not an error of your making, but since we're in the area, we should ensure that that the "no default" case is consistent in using `[]` around the `y/n`:
```suggestion
    assert console.prompts == ["Confirm? [y/n]? "]
```

freakboy3742 (changes requested):
Looking good - you've got the core functionality nailed, and the testing regimen is solid. A couple of comments inline, mostly about context that may not be obvious looking at this function in isolation.

tests/console/test_boolean_question.py
mEp3ii2: I've made changes to address the discrepancy in the input_boolean method so that all prompts will have the [] wrapping and changed my test case to follow this

src/briefcase/console.py
mEp3ii2: I've added a method to convert a string value into a bool using some standard cases as you suggested an use that for checking and converting the passed override value
Is this on the right track?

src/briefcase/console.py
```diff
@@ -719,6 +719,23 @@ def input(self, prompt: str, *, markup: bool = False):
 
         return input_value
 
+    def parse_boolean(value: str) -> str:
```
freakboy3742, line 722: Return type declaration should be bool; and if it's a member function, the first argument should be `self`:
```suggestion
    def parse_boolean(self, value: str) -> bool:
```

However - there's an argument to be made that it shouldn't be a member function of the Console class, but a general utility functionin `briefcase.utils` - after all, there's nothing "console" specific about the parsing that is being done.

freakboy3742 (changes requested):
Definitely on the right track; the test suite and coverage report is telling you what is missing - tests for the `parse_boolean()` function, and a case exercising when that parsing goes wrong. It looks like there is a test that *claims* to be a test of the override being an invalid value, but evidently that test isn't exercising the branch you think it is. That may be due to an earlier error preventing the exception from being raised.

src/briefcase/console.py
mEp3ii2: That was a silly oversight on my part with the return value, Was thinking that there would be a better place for this function as well since its a more utility method but wasn't sure to put it, I had a look and couldn't see the briefcase utils except for some files present in the tests directory and one in the macOS directory, is this a separate repo you're referring to or am I just missing it ?

freakboy3742: My apologies - the reference to a `utils` module was a red herring. I forgot there's only a `utils` module in the `tests` folder.

If we were going to move it to a different module, it would likely be the `config` module on the basis that it's parsing a configuration value... but that's a bit of a stretch. 

However, I think it would be worth pulling it out as a standalone method, rather than being a member of Console. as noted previously, there's nothing "Console" specific about the method; plus, as currently defined, it's part of the public API of the Console class (because it's a non-underscore method). While it's a useful and valuable method, it's not something that needs to be part of Briefcase's public API.

mEp3ii2: I've moved it into the config module as a solution for now, do we think their is a need to create a utils modules is there other more general utility functions present that would benefit from this ?

freakboy3742: We've survived so far without a utils module; I don't think this is enough of a trigger to add one now. 

The other option would be to include it in the `console` module itself - that would avoid the need for the import, but it's much of a muchness which one we actually use.

src/briefcase/console.py
```diff
@@ -29,7 +29,7 @@
 )
 from rich.traceback import Trace, Traceback
 
-from briefcase import __version__
+from briefcase import __version__, config
```
freakboy3742, line 32: This works; but stylistically, we generally pull specific methods or data structures rather than an entire module, unless the module itself is "an API" (like `os.path`). In this case, we're definitely just using the `parse_boolean()` function, so we should just import that.

mEp3ii2: I've made the modifications so that we are only importing the specific method, this has been done in the test file as well

freakboy3742 (approved):
Awesome - nice work!

Issue #1520, opened 2023-11-01
Catch exceptions when deleting JDK while it's running on Windows


rmartin16:
### Describe the bug

If you run `briefcase upgrade` while java is running, Briefcase crashes out.

```
❯ briefcase upgrade -vv
Using JDK at C:\Users\user\AppData\Local\BeeWare\briefcase\Cache\tools\java17
Using Android SDK at C:\Users\user\AppData\Local\BeeWare\briefcase\Cache\tools\android_sdk

[upgrade] Briefcase will upgrade the following tools:
 - Android SDK (android_sdk)
 - Java JDK (java)
 - RCEdit (rcedit)

[android_sdk] Upgrading Android SDK...

>>> Running Command:
>>>     'C:\Users\user\AppData\Local\BeeWare\briefcase\Cache\tools\android_sdk\cmdline-tools\9.0\bin\sdkmanager.bat' --update
>>> Working Directory:
>>>     C:\Users\user\github\beeware\briefcase\tmp\helloworld
>>> Environment Overrides:
>>>     ANDROID_HOME=C:\Users\user\AppData\Local\BeeWare\briefcase\Cache\tools\android_sdk
>>>     ANDROID_SDK_ROOT=C:\Users\user\AppData\Local\BeeWare\briefcase\Cache\tools\android_sdk
>>>     JAVA_HOME=C:\Users\user\AppData\Local\BeeWare\briefcase\Cache\tools\java17
[===                                    ] 10% Computing updates...
No updates available
[=======================================] 100% Computing updates...

>>> Return code: 0

[java] Upgrading Java JDK...
Removing old JDK install...

Log saved to C:\Users\user\github\beeware\briefcase\tmp\helloworld\logs\briefcase.2023_11_01-14_06_04.upgrade.log

Traceback (most recent call last):
  File "C:\Users\user\.pyenv\pyenv-win\versions\3.10.11\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\user\.pyenv\pyenv-win\versions\3.10.11\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Users\user\github\beeware\briefcase\venv-3.10-briefcase\Scripts\briefcase.exe\__main__.py", line 7, in <module>
  File "C:\Users\user\github\beeware\briefcase\src\briefcase\__main__.py", line 25, in main
    command(**options)
[...]

PR #2199, opened 2025-03-17
Adding catch exception when deleting jdk

vt37:
Adding error handling in the `uninstall()` method to catch `PermissionError` exceptions that happens if Java processes are still active or if file permissions prevent deletion. Provided a clear error message advising users to ensure no Java processes are running before retrying uninstallation.

Previously, when uninstalling the JDK while Java was running, the tool would crash with no meaningful error message.

Catch exceptions when deleting JDK while it's running on Windows #1520

## PR Checklist:


src/briefcase/integrations/java.py
```diff
@@ -310,8 +310,17 @@ def install(self):
...
+                if self.tools.host_os == "Darwin":
+                    self.tools.shutil.rmtree(self.java_home.parent.parent)
+                else:
+                    self.tools.shutil.rmtree(self.java_home)
+        except PermissionError as e:
+            raise BriefcaseCommandError(
+                f"""\
+Permission denied when trying to remove {self.java_home}.
+
+Ensure no java processes are running and try again.
```
freakboy3742, line 324: This is a user-facing error message, so it's important that we preserve appropriate case and adherence to branding guidelines.
```suggestion
Ensure no Java processes are running and try again.
```

src/briefcase/integrations/java.py
```diff
@@ -310,8 +310,17 @@ def install(self):
...
+        try:
+            with self.tools.console.wait_bar("Removing old JDK install..."):
+                if self.tools.host_os == "Darwin":
+                    self.tools.shutil.rmtree(self.java_home.parent.parent)
+                else:
+                    self.tools.shutil.rmtree(self.java_home)
+        except PermissionError as e:
+            raise BriefcaseCommandError(
+                f"""\
+Permission denied when trying to remove {self.java_home}.
```
freakboy3742, line 322: This message will be incorrect (or at least a little misleading) if you're on macOS (i.e., Darwin).

changes/1520.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added an error message when Java is used during a JDK upgrade.
```
freakboy3742, line 1: These messages will be compiled into release notes - so we're looking for more of a "marketing voice" than a commit message:
```suggestion
Error handling during JDK upgrades has been improved.
```

freakboy3742 (changes requested):
A good first pass! The approach you've taken is exactly what is needed; I've flagged some cosmetic issues with the actual message that is raised.

However, you may have noticed that CI is all red - that's because you've added functionality, but no *test* of the new functionality. We need to add a test that the behavior of the `uninstall` method when `rmtree` raises a PermissionError.

src/briefcase/integrations/java.py
```diff
@@ -310,8 +310,17 @@ def install(self):
...
+        try:
+            with self.tools.console.wait_bar("Removing old JDK install..."):
+                if self.tools.host_os == "Darwin":
+                    self.tools.shutil.rmtree(self.java_home.parent.parent)
+                else:
+                    self.tools.shutil.rmtree(self.java_home)
+        except PermissionError as e:
+            raise BriefcaseCommandError(
+                """\
+Permission denied when trying to remove Java.
```
freakboy3742, line 322: So - this is *a* fix for the problem I reported; but not the one I intended. That's my fault for not being explicit. 

It *is* desirable to give a specific location for the JDK that can't be removed - that way the user has the relevant detail to diagnose any possible problems. However, that location (and thus the message) is platform dependent. This means the location that you put in this error message needs to be conditional on the `self.tools.host_os` check a couple of lines above. Instead of just `rmtree(<some location>)` retain that location as a variable, `rmtree` that variable, and then include the variable in the error message.

tests/integrations/java/test_JDK_uninstall.py
freakboy3742, line 1: A minor detail - for consistency, this filename should be `test_JDK__uninstall.py`. The double underscore identifies that they're testing methods on the JDK class, not a method called `JDK_uninstall`, as a method name would never use a double underscore.

tests/integrations/java/test_JDK_uninstall.py
```diff
@@ -0,0 +1,58 @@
...
+)
+from briefcase.integrations.base import ToolCache
+from briefcase.integrations.java import JDK
+
+
+@pytest.fixture
+def mock_tools(mock_tools) -> ToolCache:
+    mock_tools.host_os = "Windows"
+    mock_tools.host_arch = "x86_64"
+    return mock_tools
```
freakboy3742, line 16: This `mock_tools` isn't needed - there's nothing platform or architecture specific about the test, so there's no need to force the test to be a test of Windows behavior specifically.

When you modify the logging message to be location dependent, there *will* be platform-specific behavior - however, there's no need to have a complete `mock_tools`. By default, it will inherit a `mock_tools` implementation from the `integrations` context, so you can *reference* the `mock_tools` fixture in a test without declaring it *in this file*. 

So, your `test_java_running` test can be parameterised for three different `host_os` values, with a different value being asserted for the test message; see `test_JDK__verify.py::test_invalid_jdk_version` for an example of how to do this.

tests/integrations/java/test_JDK_uninstall.py
```diff
@@ -0,0 +1,58 @@
...
+        "File is being used by other processes"
+    )
+
+    # Uninstalling JDK
+    with pytest.raises(BriefcaseCommandError) as exc_info:
+        jdk.uninstall()
+
+    # Exception raised
+    assert "Permission denied when trying to remove Java." in str(exc_info.value)
+    assert "Ensure no Java processes are running and try again." in str(exc_info.value)
```
freakboy3742, line 39: This is the right idea, but there's a better approach: the `pytest.raises` context manager can accept a `match` argument that allows you to define a regular expression that will match the raised exception text.

freakboy3742 (changes requested):
This is on the right track; I've flagged a couple of issues inline about the content of the error message and how that error is being tested.

vt37:
Thanks for the helpful feedback! I've changed the code based on your suggestions and made sure to adjust the platform-specific behavior accordingly in my code. Let me know if there’s anything else I can improve.

src/briefcase/integrations/java.py
```diff
@@ -310,8 +310,19 @@ def install(self):
...
+                if self.tools.host_os == "Darwin":
+                    jdk_location = self.java_home.parent.parent
+                else:
+                    jdk_location = self.java_home
+                self.tools.shutil.rmtree(jdk_location)
+
+        except PermissionError as e:
+            raise BriefcaseCommandError(
+                f"""\
+Permission denied when trying to remove {jdk_location}.
```
freakboy3742, line 324: A really minor UX thing - but a common usage pattern if this problem occurs will be to copy the path out of the terminal output and look there; the trailing `.` can be confusing as to whether it's a part of the path or not. In cases like this, it's better to either use "bad English" and omit the period, or use quotes around the string.
```suggestion
Permission denied when trying to remove {jdk_location!r}.
```

tests/integrations/java/test_JDK_uninstall.py
freakboy3742: You've added the new file; so this one can now be deleted.

freakboy3742 (changes requested):
Almost there - two really minor issues left to clean up, and we'll be done!

freakboy3742 (approved):
Nice work - thanks for the PR!

PR #2201, opened 2025-03-18
Add --app option to briefcase build/package (#2148)

kavi2du:
This pull request introduces the `-a / --app` option for `briefcase build` and `briefcase package`, allowing users to specify which app(s) to build/package instead of processing all apps in a project.

Allow specifying app name(s) to briefcase build and briefcase package #2148

## PR Checklist:

### DIFF
--- a/changes/2148.feature.rst
+++ b/changes/2148.feature.rst
@@ -1,0 +1,1 @@
+Added the `-a / --app` option to `briefcase build` and `briefcase package` to allow building/packaging a specific app instead of all apps.
--- a/changes/2148.feature.rst
+++ b/changes/2148.feature.rst
@@ -1,1 +1,1 @@
-The `build` and `package` commands can now be run on a single or multiple app(s) in a project by using the `-a / --app` option
+The `build` and `package` commands can now be run on a single app within a multi-app project by using the `-a / --app` option.
--- a/changes/2148.feature.rst
+++ b/changes/2148.feature.rst
@@ -1,1 +1,1 @@
-Added the `-a / --app` option to `briefcase build` and `briefcase package` to allow building/packaging a specific app instead of all apps.
+The `build` and `package` commands can now be run on a single or multiple app(s) in a project by using the `-a / --app` option
--- a/src/briefcase/commands/build.py
+++ b/src/briefcase/commands/build.py
@@ -14,6 +14,13 @@ class BuildCommand(BaseCommand):
         self._add_update_options(parser, context_label=" before building")
         self._add_test_options(parser, context_label="Build")
 
+        parser.add_argument(
+            "-a",
+            "--app",
+            dest="app_name",
+            help="Name of the app to build (if multiple apps exist in the project)",
+        )
+
     def build_app(self, app: AppConfig, **options):
         """Build an application.
 
@@ -86,6 +93,7 @@ class BuildCommand(BaseCommand):
     def __call__(
         self,
         app: AppConfig | None = None,
+        app_name: str | None = None,  # New argument for filtering
         update: bool = False,
         update_requirements: bool = False,
         update_resources: bool = False,
@@ -123,19 +131,24 @@ class BuildCommand(BaseCommand):
         # and that the app configuration is finalized.
         self.finalize(app)
 
-        if app:
-            state = self._build_app(
-                app,
-                update=update,
-                update_requirements=update_requirements,
-                update_resources=update_resources,
-                update_support=update_support,
-                update_stub=update_stub,
-                no_update=no_update,
-                test_mode=test_mode,
-                **options,
-            )
+        if app_name:
+            # Build only the specified app
+            if app_name in self.apps:
+                state = self._build_app(
+                    self.apps[app_name],
+                    update=update,
+                    update_requirements=update_requirements,
+                    update_resources=update_resources,
+                    update_support=update_support,
+                    update_stub=update_stub,
+                    no_update=no_update,
+                    test_mode=test_mode,
+                    **options,
+                )
+            else:
+                raise BriefcaseCommandError(f"App '{app_name}' not found in project.")
         else:
+            # Default: Build all apps
             state = None
             for app_name, app in sorted(self.apps.items()):
                 state = self._build_app(
--- a/src/briefcase/commands/build.py
+++ b/src/briefcase/commands/build.py
@@ -17,9 +17,8 @@ class BuildCommand(BaseCommand):
         parser.add_argument(
             "-a",
             "--app",
-            dest="apps",
-            action="append",
-            help="Name of the app(s) to build (if multiple apps exist in the project)",
+            dest="app_name",
+            help="Name of the app to build (if multiple apps exist in the project)",
         )
 
     def build_app(self, app: AppConfig, **options):
@@ -94,7 +93,7 @@ class BuildCommand(BaseCommand):
     def __call__(
         self,
         app: AppConfig | None = None,
-        apps: list[str] | None = None,
+        app_name: str | None = None,
         update: bool = False,
         update_requirements: bool = False,
         update_resources: bool = False,
@@ -132,15 +131,12 @@ class BuildCommand(BaseCommand):
         # and that the app configuration is finalized.
         self.finalize(app)
 
-        if apps:
-            selected_apps = {}
-            for name in apps:
-                if name not in self.apps:
-                    raise BriefcaseCommandError(
-                        f"App '{name}' does not exist in this project."
-                    )
-                selected_apps[name] = self.apps[name]
-            apps_to_build = selected_apps
+        if app_name:
+            if app_name not in self.apps:
+                raise BriefcaseCommandError(
+                    f"App '{app_name}' does not exist in this project."
+                )
+            apps_to_build = {app_name: self.apps[app_name]}
         elif app:
             apps_to_build = {app.app_name: app}
         else:
--- a/src/briefcase/commands/build.py
+++ b/src/briefcase/commands/build.py
@@ -17,8 +17,9 @@ class BuildCommand(BaseCommand):
         parser.add_argument(
             "-a",
             "--app",
-            dest="app_name",
-            help="Name of the app to build (if multiple apps exist in the project)",
+            dest="apps",
+            action="append",
+            help="Name of the app(s) to build (if multiple apps exist in the project)",
         )
 
     def build_app(self, app: AppConfig, **options):
@@ -93,7 +94,7 @@ class BuildCommand(BaseCommand):
     def __call__(
         self,
         app: AppConfig | None = None,
-        app_name: str | None = None,  # New argument for filtering
+        apps: list[str] | None = None,
         update: bool = False,
         update_requirements: bool = False,
         update_resources: bool = False,
@@ -131,36 +132,32 @@ class BuildCommand(BaseCommand):
         # and that the app configuration is finalized.
         self.finalize(app)
 
-        if app_name:
-            # Build only the specified app
-            if app_name in self.apps:
-                state = self._build_app(
-                    self.apps[app_name],
-                    update=update,
-                    update_requirements=update_requirements,
-                    update_resources=update_resources,
-                    update_support=update_support,
-                    update_stub=update_stub,
-                    no_update=no_update,
-                    test_mode=test_mode,
-                    **options,
-                )
-            else:
-                raise BriefcaseCommandError(f"App '{app_name}' not found in project.")
+        if apps:
+            selected_apps = {}
+            for name in apps:
+                if name not in self.apps:
+                    raise BriefcaseCommandError(
+                        f"App '{name}' does not exist in this project."
+                    )
+                selected_apps[name] = self.apps[name]
+            apps_to_build = selected_apps
+        elif app:
+            apps_to_build = {app.app_name: app}
         else:
-            # Default: Build all apps
-            state = None
-            for app_name, app in sorted(self.apps.items()):
-                state = self._build_app(
-                    app,
-                    update=update,
-                    update_requirements=update_requirements,
-                    update_resources=update_resources,
-                    update_support=update_support,
-                    update_stub=update_stub,
-                    no_update=no_update,
-                    test_mode=test_mode,
-                    **full_options(state, options),
-                )
+            apps_to_build = self.apps
+
+        state = None
+        for app_name, app_obj in sorted(apps_to_build.items()):
+            state = self._build_app(
+                app_obj,
+                update=update,
+                update_requirements=update_requirements,
+                update_resources=update_resources,
+                update_support=update_support,
+                update_stub=update_stub,
+                no_update=no_update,
+                test_mode=test_mode,
+                **full_options(state, options),
+            )
 
         return state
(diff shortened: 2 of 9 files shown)


changes/2148.feature.rst
```diff
@@ -0,0 +1 @@
+Added the `-a / --app` option to `briefcase build` and `briefcase package` to allow building/packaging a specific app instead of all apps.
```
freakboy3742, line 1: The changenote should be phrased in the way that you'd see a release note describing the "cool features in the new release!!1!11!". You're looking for a voice that's a lot more like "marketing copy" than a commit message:
```suggestion
The `build` and `package` commands can now be run on a single app in a project by using the `-a / --app` option.
```

src/briefcase/commands/build.py
```diff
@@ -86,6 +93,7 @@ def _build_app(
     def __call__(
         self,
         app: AppConfig | None = None,
+        app_name: str | None = None,  # New argument for filtering
```
freakboy3742, line 96: There's no need for a comment like this one. It may be a new argument *in this PR*, but as soon as it's merged to main, it won't be a "new" argument any more.

src/briefcase/commands/build.py
```diff
@@ -123,19 +131,24 @@ def __call__(
...
-                update_stub=update_stub,
-                no_update=no_update,
-                test_mode=test_mode,
-                **options,
-            )
+        if app_name:
+            # Build only the specified app
+            if app_name in self.apps:
+                state = self._build_app(
+                    self.apps[app_name],
```
freakboy3742, line 138: This logic doesn't seem right.

1. If you specify `app_name`, it will build *every* app, with the specified app's configuration - so the same app will be built N times.
2. If you *don't* specify `app_name` the fallback is to the default "build all apps", which ignores the case where `app` is passed in as an argument - for example, when you call `run -u`.

src/briefcase/commands/package.py
```diff
@@ -151,20 +159,32 @@ def add_options(self, parser):
...
         **options,
     ) -> dict | None:
         # Confirm host compatibility, that all required tools are available,
         # and that the app configuration is finalized.
         self.finalize(app)
 
-        if app:
-            state = self._package_app(app, update=update, **options)
+        if app_name:
+            # Package only the specified app
```
freakboy3742, line 171: This has the same set of problems as the `build` implementation.

tests/commands/build/test_call.py
```diff
@@ -11,11 +11,11 @@ def test_specific_app(build_command, first_app, second_app):
         "second": second_app,
     }
 
-    # Configure no command line options
-    options, _ = build_command.parse_options([])
+    # Configure command line options for building a specific app
+    options, _ = build_command.parse_options(["--app", "first"])
```
freakboy3742, line 15: The existing test is verifying something specific. If you're adding behavior, you need to add a *new* test for the new test configuration, not just modify an existing one.

freakboy3742 (changes requested):
A good first pass - I've flagged a few issues inline; and the test suite is flagging some test failures and coverage issues that will need to be addressed.

In particular, make sure you're paying attention to how you could be modifying existing behavior - if you're adding new functionality, you need to be clear on how existing behavior will be impacted by this change. 

The good news is that Briefcase has 100% branch coverage - so all existing use cases and behaviours *should* be tested - as long as you don't modify existing tests to change what the tests are validating.

One last detail - I noticed that you submitted this as #2200 before closing that PR and opening this one. There's no need to close and reopen a pull request for the same issue - once you've got an open PR, you can push updates to that PR. Each time you do, the test suite will run again. When we merge the PR, we squash all the individual commits into a single commit, so you don't have to worry about a "dirty" commit history.

src/briefcase/commands/build.py
```diff
@@ -17,8 +17,9 @@ def add_options(self, parser):
         parser.add_argument(
             "-a",
             "--app",
-            dest="app_name",
-            help="Name of the app to build (if multiple apps exist in the project)",
+            dest="apps",
+            action="append",
```
freakboy3742, line 21: A couple of notes here:

Firstly, `--app` is an existing flag that can be used on the `run` command. In that context, the `--app` argument on `run` *has* to be non-append, because you can only run one app at a time. 

I'm guessing the argument you're making here is that for the `build` and `package` context, it's possible to build/package a subset of N apps - and yes, that's true - but there's also consistency to consider. If there's a flag on one command, and it's appropriate for that flag to exist on a second command, then the flag should generally speaking behave the same. That sort of consistency is a subtle form of documentation, because once you've learned a skill in one part of the project, that skill is transferrable to other situations.

Yes, dropping `append` will mean that you won't be able to build 2 of 3 projects. However, projects with multiple apps are a niche to start with; having a use case for building N out of M (n<m) projects is even more niche. I'm entirely comfortable saying "build all" or "build 1" are the only options here.

Secondly - variable naming matters. In the usage here, `app_name` implies a string, `app` implies an AppConfig object. Briefcase's codebase is fairly consistent about this usage. Based on the naming of the `apps` argument, I'd assume that it is a "list of AppConfig objects"... which it won't be. `app_names` would be a more appropriate name for the "append" version of the command; but since we won't be using `append`, ` here, `app_name` is the more appropriate name.

tests/commands/package/test_call.py
```diff
@@ -758,3 +762,174 @@ def test_already_packaged(package_command, first_app, tmp_path):
...
+
+
+def test_package_with_update(package_command, first_app):
+    """If --update is passed, app is updated before packaging."""
+    package_command.apps = {"first": first_app}
+    options, _ = package_command.parse_options(["--app", "first", "--update"])
+    package_command(**options)
+
+    # Find the 'update' action for the app
+    update_action = next(
```
freakboy3742, line 926: It may be long-winded, but it's important to test the *entire* action sequence here, not just the one action that we think is important. The purpose of this test is to verify that Briefcase will perform the right sequence of actions - we can't leave anything up to assumed behavior here. We need to be certain that an updated package has everything updated (with the expected arguments), then built, then packaged.

freakboy3742 (changes requested):
Definitely getting closer. Some comments inline about the change in semantics of the `--app` argument; the testing regimen looks on the right track (and will be simplified by the requested change to how `--app` is interpreted). 

The one last detail that is missing: documentation. This adds a command line option to `update` and `package`; so the documentation for those commands should be updated to include `--app` as an available option. The text from the `run` command can be duplicated for this (noting also that there's a minor typo in markup of the `--app` option in the `run` docs that you might as well fix while you're in the area).

The other thing that occurs to me - although the original ticket doesn't mention them, `create`, `update` should probably also support the `--app` argument. You don't have to include that change in this PR unless you're feeling particularly enthusiastic; but if you are, the general pattern of the fix for those commands will be identical.

src/briefcase/commands/build.py
kavi2du: Hi @freakboy3742, appriciate the detailed and prompt feedback. I just have couple of questions that I would like to clarify.
1. Where can I find the documentation for the commands such as build and package? I only found the changes documentation folder in the briefcase repo (briefcase/changes).
2. I understand the need for the --app flag consistency accross the commands. I will address that issue. That said, I had a quick follow up idea, would it be worth considering a different flag (maybe something like -m/--multiple) specifically for commands such as build and package, so they will have the functionality of selecting multiple apps?

Moreover, I will look into adding this functionality to create and update as well. I probably won't be able to finalise it within this sprint but I will surely work on it soon.

freakboy3742: > 1. Where can I find the documentation for the commands such as build and package? I could only find the changes documentation folder in the briefcase repo (briefcase/changes).

The docs for Briefcase are all contained in the `docs` folder of the Briefcase repository. The layout of that part of the repository pretty closely follows the navigation hierarchy of the live docs site. The pages are also in `.rst` format.

There's also a [documentation contribution guide](https://briefcase.readthedocs.io/en/stable/how-to/contribute-docs.html) for details on how.

> 2. I understand the need for the --app flag consistency accross the commands. I will address that issue. That said, I had a quick follow up idea, would it be worth considering a different flag (maybe something like -m/--multiple) specifically for commands such as build and package, so they will have the functionality of selecting multiple apps?

If we were going to go down this path, then modifying `-a/--app` to be "append mode" would be the approach to take, rather than adding a second argument to do almost exactly the same thing. However, as I flagged earlier, this is a sufficiently niche use case that I'm not convinced we need to support it.

> Moreover, I will look into adding this functionality to create and update as well. I probably won't be able to finalise it within this sprint but I will surely work on it soon.

It's fine if the `create`/`update` additions are  a follow up PR. If nothing else, it would be good to get this PR to a finished state, and then use that solution as a model for the next update.

src/briefcase/commands/build.py
```diff
@@ -132,15 +131,12 @@ def __call__(
...
-                        f"App '{name}' does not exist in this project."
-                    )
-                selected_apps[name] = self.apps[name]
-            apps_to_build = selected_apps
+        if app_name:
+            if app_name not in self.apps:
+                raise BriefcaseCommandError(
+                    f"App '{app_name}' does not exist in this project."
+                )
+            apps_to_build = {app_name: self.apps[app_name]}
```
freakboy3742, line 139: This works - but "good Python" code works on the principle of "it's better to ask forgiveness than permission" - that is - assume the operation will succeed, and catch the error:
```suggestion
            try
                apps_to_build = {app_name: self.apps[app_name]}     
            except KeyError:
                raise BriefcaseCommandError(
                    f"App '{app_name}' does not exist in this project."
                )
```

This ends up being slightly faster, because the "normal" case does the operation, and doesn't need the "pre-check" that the app name is valid.

src/briefcase/commands/package.py
```diff
@@ -169,15 +168,12 @@ def __call__(
...
-                        f"App '{name}' does not exist in this project."
-                    )
-                selected_apps[name] = self.apps[name]
-            apps_to_package = selected_apps
+        if app_name:
+            if app_name not in self.apps:
+                raise BriefcaseCommandError(
+                    f"App '{app_name}' does not exist in this project."
+                )
+            apps_to_package = {app_name: self.apps[app_name]}
```
freakboy3742, line 176: As above - ask forgiveness, not permission.

tests/commands/build/test_call.py
```diff
@@ -1251,16 +1251,67 @@ def test_build_app_single(build_command, first_app, second_app):
...
+
+    # Run the build command
+    result = build_command(**options)
+
+    # Nothing is built
+    assert result is None
+    assert build_command.actions == [("verify-host",), ("verify-tools",)]
+
+
+def test_build_app_no_args(build_command):
```
freakboy3742, line 1289: How is this different to the previous test?

tests/commands/build/test_call.py
```diff
@@ -1251,16 +1251,67 @@ def test_build_app_single(build_command, first_app, second_app):
...
+
+    # Run the build command
+    result = build_command(**options)
+
+    # Nothing is built
+    assert result is None
+    assert build_command.actions == [("verify-host",), ("verify-tools",)]
+
+
+def test_build_with_update(build_command, first_app, second_app):
```
freakboy3742, line 1305: You need to be a little careful about *over* testing. 

Technically, yes: a full black-box test should have every possible combination. However - every possible combination is an O(n^2) proposition; and each test you add takes time to write, time to run, and will need to be maintained over time. So - this is a case where it's OK to take strategic shortcuts in what is tested, informed by the known complexity and implementational risk factors of the thing we're testing.

So - I suspect most of the these "combination of flag" tests can be dropped. Testing each individual flag, and combination of flags, and error conditions for combinations of flags... that's starting to get excessive. For example - we already have a test that `--update` and `--no-update` are mutually exclusive; there's nothing about adding `-a` to the mix that makes that *more* mutually exclusive. 

We definitely need an explicit test of `--app first` and `-a first`, and a test of `app not-an-app`, as that's literally the feature we're adding. It might be worth throwing in a "kitchen sink" test of "--app first -u -r --update-resources" to make sure that all the other flags are actually plumbed in as expected. But past that, I'm not sure we need many more tests.

tests/commands/package/test_call.py
```diff
@@ -758,3 +762,141 @@ def test_already_packaged(package_command, first_app, tmp_path):
...
+    assert result is None
+    assert package_command.actions == [
+        # Host OS is verified
+        ("verify-host",),
+        # Tools are verified
+        ("verify-tools",),
+    ]
+
+
+def test_package_with_update(package_command, first_app, tmp_path):
```
freakboy3742, line 844: This is lot closer to the level of testing needed here - but we should also be testing the `-a first` version of the flag.

freakboy3742 (changes requested):
Implementation is looking good, with one minor style tweak; tests need a little more work (or, rather... a little less work :-) - details inline)

tests/commands/build/test_call.py
```diff
@@ -1217,3 +1217,123 @@ def test_test_app_unbuilt(build_command, first_app_unbuilt, second_app):
...
+def test_build_app_single(build_command, first_app, second_app):
+    """If the --app flag is used, only the selected app is built."""
+    # Add two apps
+    build_command.apps = {
+        "first": first_app,
+        "second": second_app,
+    }
+
+    # Configure command line options
+    options, _ = build_command.parse_options(["--app", "first"])
```
freakboy3742, line 1231: This test should be parameterised to confirm that both `--app` and `-a` work. (You could also duplicate the test - but given the test is otherwise identical, parameterization makes more sense.)

tests/commands/package/test_call.py
```diff
@@ -758,3 +762,149 @@ def test_already_packaged(package_command, first_app, tmp_path):
...
     # of this file shows that the old one has been deleted.
     assert not artefact_path.exists()
+
+
+def test_package_app_single(package_command, first_app):
+    """If the --app flag is used, only the selected app is packaged."""
+    # Add a single app
+    package_command.apps = {
+        "first": first_app,
+    }
```
freakboy3742, line 772: The test you're running here is testing the right thing - but it's only testing a single app. The setup should include multiple apps to confirm that `--app` only picks up a single

tests/commands/package/test_call.py
```diff
@@ -758,3 +762,149 @@ def test_already_packaged(package_command, first_app, tmp_path):
...
+
+    # Configure the --app option
+    options, _ = package_command.parse_options(["--app", "first"])
```
freakboy3742, line 775: As with the build test - this test should be parameterised to confirm both `--app` and `-a` work as arguments.

freakboy3742 (approved):
So very close! A couple of final test cases to add/fix and we'll be there.

freakboy3742 (changes requested):
Sorry - selected the wrong button - there's still some minor changes required here.

freakboy3742 (approved):
Awesome - thanks for the PR!

kavi2du:
No worries, really appreciate your help throughout the process!

kavi2du:
> Awesome - thanks for the PR!
No worries, really appreciate your help throughout the process!

kavi2du:
> Awesome - thanks for the PR!

No worries, really appreciate your help throughout the process!

Issue #2103, opened 2025-01-08
Briefcase fails to escape ampersands in company name when generating [appname].wxs when packaging for Windows


miss-programgamer:
### Describe the bug

Exactly as the title says, ampersands need to be escaped when generating .wxs files, but briefcase doesn't perform such escaping, leading to confusing errors while attempting to package an app and having to manually edit [appname].wxs files in the build folder to force it to work.

### Steps to reproduce

1. Create a briefcase project with an ampersand in the company name (or even the app name? Haven't tested that case)
2. Run `briefcase package -p msi --adhoc-sign` (after doing any boilerplate initialization I may have forgotten about)

### Expected behavior

I expect a working windows installer for my app, instead I get an error and no installer for my app. Sadge :(

### Environment

- Operating System: Windows 11
- Python version: 3.13
- Briefcase version: 0.3.20

### Logs

```
error CNDL0104 : Not a valid source file; detail: An error occurred while parsing EntityName. Line 10, position 43.
```

freakboy3742:
Thanks for the report. Escaping content correctly is a class of bug we've been slowly squashing over time. The model for fixing them is well understood, and not that difficult to implement - see #1876 and #1746 for two recent examples.

In this case, we need to add an XML content escaping filter to Briefcase (there's already an XML extension with a boolean value filter; it needs a new entity-escaping filter using Python's built-in escaping tools), and then use that filter in the windows [app](https://github.com/beeware/briefcase-windows-app-template) and [Visual Studio](https://github.com/beeware/briefcase-windows-visualstudio-template) templates. 

There's likely a bunch of other places where escaping *could* fail in the WiX template, so it would be worth auditing all the cookiecutter inclusions to make sure we've got them all.

caydnn:
happy to take a look at this one

caydnn:
looking into this

freakboy3742:
Accidental close due to ticket content; this still requires an app template update.

PR #2203, opened 2025-03-19
Add XML content escaping filter to cookiecutter,py

caydnn:
Added an XML content escaping filter to cookiecutter.py.
By using this filter in the windows app and visual studio templates, special characters  & < and > will be escaped. refs #2103 

## PR Checklist:

### DIFF
--- a/changes/2103.bugfix.rst
+++ b/changes/2103.bugfix.rst
@@ -1,0 +1,1 @@
+Cookiecutter has been improved to escape special characters & < and >.
--- a/changes/2103.bugfix.rst
+++ b/changes/2103.bugfix.rst
@@ -1,1 +1,0 @@
-Cookiecutter has been improved to escape special characters & < and >.
--- a/changes/2103.feature.rst
+++ b/changes/2103.feature.rst
@@ -1,0 +1,2 @@
+A Cookiecutter filter for escaping XML content has been added.
+A Cookiecutter filter for quoting XML attributes appropriately.
--- a/changes/2103.feature.rst
+++ b/changes/2103.feature.rst
@@ -1,2 +1,1 @@
-A Cookiecutter filter for escaping XML content has been added.
-A Cookiecutter filter for quoting XML attributes appropriately.
+Cookiecutter filters for escaping XML content and attributes have been added.
--- a/tests/integrations/cookiecutter/test_XMLExtension.py
+++ b/tests/integrations/cookiecutter/test_XMLExtension.py
@@ -53,10 +53,8 @@ def test_xml_escape(value, expected):
         ("Hello ' World", '"Hello \' World"'),
         # A double quote wrapped in single quotes
         ('Hello " World', "'Hello \" World'"),
-        # A double quote and a single quote wrapped in single quotes
+        # A double quote and a single quote in the same string value
         ("Hello \" And ' World", '"Hello &quot; And \' World"'),
-        # A double quote and a single quote wrapped in double quotes
-        ("Hello \" Or ' World", '"Hello &quot; Or \' World"'),
     ],
 )
 def test_xml_attr(value, expected):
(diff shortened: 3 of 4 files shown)


changes/2103.bugfix.rst
```diff
@@ -0,0 +1 @@
+Cookiecutter has been improved to escape special characters & < and >.
```
freakboy3742, line 1: Cookiecutter is a standalone project, not part of Briefcase; we're adding a specific filter, not fixing Cookiecutter itself.
```suggestion
A Cookiecutter filter for escaping XML content has been added.
```

Also - although this change in service of a fixing a bug, this particular change is adding a feature - a new cookiecutter filter for quoting XML content. As such, the file should be `2103.feature.rst`.

freakboy3742 (changes requested):
I've flagged a couple of minor tweaks to the release notes, but otherwise, this looks great - nice work!

The only other suggestion is whether we want to add an `xml_attr` filter while we're at it. I haven't done a deep audit of the WiX file, but I'd be surprised if there isn't string content somewhere in the file that is defined as an XML attribute that *could* include a quote (i.e., `<actor name="Dwayne &quot;The Rock&quot; Johnson">`. And even if we don't need it *now*, it's such a predictable potential future need that we might as well save ourselves the future effort while we're making changes in the area.

changes/2103.feature.rst
```diff
@@ -0,0 +1,2 @@
+A Cookiecutter filter for escaping XML content has been added.
+A Cookiecutter filter for quoting XML attributes appropriately.
```
mhsmith, line 2: A few issues here:

* The second line is incomplete.
* Towncrier doesn't format multi-line change notes properly, so they should be split into two files called `2103.feature.1.rst` and `2103.feature.2.rst`.
* The `.bugfix.rst` file still exists.

freakboy3742: Alternatively, this single feature could be written to encompass both:

```suggestion
Cookiecutter filters for escaping XML content and attributes have been added.
```

Regardless - the `.bugfix` version needs to be deleted as well.

tests/integrations/cookiecutter/test_XMLExtension.py
```diff
@@ -24,3 +24,43 @@ def test_bool_attr(value, expected):
...
+    "value, expected",
+    [
+        # A single quote wrapped in double quotes
+        ("Hello ' World", '"Hello \' World"'),
+        # A double quote wrapped in single quotes
+        ('Hello " World', "'Hello \" World'"),
+        # A double quote and a single quote wrapped in single quotes
+        ("Hello \" And ' World", '"Hello &quot; And \' World"'),
+        # A double quote and a single quote wrapped in double quotes
+        ("Hello \" Or ' World", '"Hello &quot; Or \' World"'),
```
freakboy3742, line 59: So... Black (our code formatter) is going to fight you here :-) Black will normalise the content of strings to be double quoted if there's any path do that unambiguously. As a result, these last two tests end up being identical.

The good news is that I don't think you actually need to test both. We need *a* legal string that contains both a `"` and a `'` - but as long as we've got that, we don't care how it was escaped or quoted for input. However the value has been passed in (be it TOML or Python), it will need to be correctly escaped to pass the parser that read it - so additional checks beyond that are more a test of TOML and Python parsers than they are of your new XML escaping filter.
```suggestion
        # A double quote and a single quote in the same string value
        ("Hello \" And ' World", '"Hello &quot; And \' World"'),
```

freakboy3742 (changes requested):
This is really close to being done - one last issue to clean up in the release note, and one test case that has been "helpfully fixed" by the pre-commit tooling.

caydnn:
My apologies for my mistakes here, I have now deleted 2103.bugfix.rst and applied the suggestions :) Hopefully this is good to go now
I'll get working on the changes to the templates this afternoon.

freakboy3742 (approved):
Looks good - Thanks for the updates!

Issue #2151, opened 2025-02-09
Error finding stub app on macOS when formal name includes a `.`


freakboy3742:
### Describe the bug

If the formal name of an app contains a period, Briefcase will fail to build the app, as the `unbuilt_executable_path` will contain the "suffix" suggested by the period.

### Steps to reproduce

1. `briefcase new --no-input -Q "formal_name=Hello 2.0"`
2. `briefcase run`
3. See error:
```

Traceback (most recent call last):
  File "/Users/rkm/beeware/briefcase/venv3.13/bin/briefcase", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/rkm/beeware/briefcase/src/briefcase/__main__.py", line 27, in main
    command(**options)
    ~~~~~~~^^^^^^^^^^^
  File "/Users/rkm/beeware/briefcase/src/briefcase/commands/run.py", line 309, in __call__
    state = self.build_command(
        app,
    ...<7 lines>...
        **options,
    )
  File "/Users/rkm/beeware/briefcase/src/briefcase/commands/build.py", line 127, in __call__
    state = self._build_app(
        app,
    ...<7 lines>...
        **options,
    )
  File "/Users/rkm/beeware/briefcase/src/briefcase/commands/build.py", line 52, in _build_app
    state = self.create_command(app, test_mode=test_mode, **options)
  File "/Users/rkm/beeware/briefcase/src/briefcase/commands/create.py", line 984, in __call__
    state = self.create_app(app, **options)
  File "/Users/rkm/beeware/briefcase/src/briefcase/commands/create.py", line 938, in create_app
    self.install_stub_binary(app=app)
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/rkm/beeware/briefcase/src/briefcase/commands/create.py", line 433, in install_stub_binary
    self.tools.os.chmod(unbuilt_executable_path, 0o755)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/Users/rkm/beeware/briefcase/local/tmp/hello20/build/hello20/macos/app/Hello 2.0.app/Contents/MacOS/Stub.0'
(
```

### Expected behavior

App should build and run successfully.

### Environment

- Operating System: macOS
- Python version: any (tested on 3.13, 3.12)
- Software versions:
[...]

PR #2213, opened 2025-03-26
fix: finding stub app on macOS with `.` in formal name

Mr-Sunglasses:
fix: #2151 

- fix issue in finding stub app on macOS with `.` in formal name

## PR Checklist:

- [] All new features have been tested

### DIFF
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -401,9 +401,12 @@ a custom location for Briefcase's tools.
 
         :param app: The app config
         """
-        return self.binary_executable_path(app).parent / (
-            "Stub" + self.binary_executable_path(app).suffix
-        )
+        if sys.platform == "win32":
+            return self.binary_executable_path(app).parent / (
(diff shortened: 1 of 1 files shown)


freakboy3742 (changes requested):
Thanks for the PR!

You've got the general form of the solution correct - but there's a couple of housekeeping issues you need to take care of to get CI to pass. The most immediate problem is the lack of a changenote - you need to add a file that describes the change as it will appear in the release notes. Once you've fixed that, I suspect you'll hit two other problems:
* a code style issue picked up by `black`; and 
* a test coverage failure. We have 100% branch coverage in our test suite; you're adding a fix for a bug here, so it should be accompanied by a test for that bug.

The [Briefcase contribution guide](https://briefcase.readthedocs.io/en/stable/how-to/contribute-code.html) describes the things you need to do as part of a PR, and how to set up your local environment to perform those checks.

JohananOppongAmoateng:
@freakboy3742 I think we can close this too. The referenced issue has been fixed

PR #2229, opened 2025-03-31
Accept other changelog name and extension #2116

### LINKED ISSUE #2116: Please accept CHANGELOG.md name
### Describe the bug

Please accept CHANGELOG.md name.
The .md extensions allows a nice rendering on GitHub and similar websites.

It is also a recommended name in [keepachangelog.com spec](https://keepachangelog.com/en/1.1.0/#frequently-asked-questions).

### Steps to reproduce

On Linux run `briefcase build` in a project which has `CHANGELOG.md` (instead of `CHANGELOG`) in the project root directory.
You'll get the following error:

```
Installing license... done
Installing changelog... errored

Your project does not contain a CHANGELOG file.

Create a file named `CHANGELOG` in the same directory as your `pyproject.toml`
with details about the release.
```

and the build will stop.

### Expected behavior

The build should find `CHANGELOG.md` and complete successfully.

My current workaround is copying/symlinking the expected name, but I'd rather avoid it.

Thanks for considering this.

### Screenshots

_No response_

### Environment

- Operating System: Linux (Fedora 41)
- Python version: 3.13.1
- Software versions:
  - Briefcase: 0.3.20

### Logs

_No response_

### Additional context

_No response_

JMah007:
This pull request fixes BUILD and PACKAGE in linux to accept changelog name and extension in any combination of CHANGELOG, HISTORY, NEWS and RELEASES with an extension of like .md, .rst, .txt or no extension.

Fixes #2116

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/convert.py
+++ b/src/briefcase/commands/convert.py
@@ -686,10 +686,10 @@ class ConvertCommand(NewCommand):
 
         if matched_changelog_name is False:
             self.console.warning(
-                f"\nChangelog file not found in '{self.base_path}'. You should either "
-                f"create a new '{self.base_path / 'changelog_file'}' file, or rename an "
-                f"existing changelog file with the following as its name "
-                f"(CHANGELOG, HISTORY, NEWS or RELEASES) with extensions (.md, .rst, .txt or no extension)"
+                f"create a new changelog file in {self.base_path}, or rename an "
+                "existing file to a known changelog file name (one of 'CHANGELOG', "
+                "'HISTORY', 'NEWS' or 'RELEASES'; the file may have an extension of "
+                "'.md', '.rst', '.txt', or have no extension)"
             )
 
         # Copy tests or test entry script
--- a/src/briefcase/commands/convert.py
+++ b/src/briefcase/commands/convert.py
@@ -674,7 +674,8 @@ class ConvertCommand(NewCommand):
 
         if changelog is None:
             self.console.warning(
-                f"Create a new changelog file in {self.base_path}, or rename an "
+                f"\nChangelog file not found in {self.base_path!r}. You should either "
+                f"create a new changelog file in {self.base_path!r}, or rename an "
                 "existing file to a known changelog file name (one of 'CHANGELOG', "
                 "'HISTORY', 'NEWS' or 'RELEASES'; the file may have an extension of "
                 "'.md', '.rst', '.txt', or have no extension)"
--- a/src/briefcase/commands/convert.py
+++ b/src/briefcase/commands/convert.py
@@ -799,8 +799,12 @@ To run your application, type:
 
 
 def find_changelog_filename(base_path):
-    """Search for a valid changelog file in the specified directory
-    and return its format name if it exists, otherwise return None"""
+    """Find a valid changelog file in a given directory.
+    
+    :param base_path: The directory to search
+    :returns: The filename of the changelog that was found; None if a changelog
+        with a known name could not be found.
+    """
     changelog_formats = [
         f"{name}{extension}"
         for name in ["CHANGELOG", "HISTORY", "NEWS", "RELEASES"]
--- a/src/briefcase/commands/convert.py
+++ b/src/briefcase/commands/convert.py
@@ -670,21 +670,9 @@ class ConvertCommand(NewCommand):
             copy2(project_dir / "LICENSE", self.base_path / "LICENSE")
 
         # See if changefile exists
-        matched_changelog_name = False
-        changelog_formats = [
-            f"{name}{extension}"
-            for name in ["CHANGELOG", "HISTORY", "NEWS", "RELEASES"]
-            for extension in ["", ".md", ".rst", ".txt"]
-        ]
+        changelog = find_changelog_filename(self.base_path)
 
-        # Stops checking once a match is found
-        for format in changelog_formats:
-            changelog_source = self.base_path / format
-            if changelog_source.is_file():
-                matched_changelog_name = True
-                break
-
-        if matched_changelog_name is False:
+        if changelog is None:
             self.console.warning(
                 f"create a new changelog file in {self.base_path}, or rename an "
                 "existing file to a known changelog file name (one of 'CHANGELOG', "
@@ -808,3 +796,19 @@ To run your application, type:
                 project_overrides=parse_project_overrides(project_overrides),
                 **options,
             )
+
+
+def find_changelog_filename(base_path):
+
+    changelog_formats = [
+        f"{name}{extension}"
+        for name in ["CHANGELOG", "HISTORY", "NEWS", "RELEASES"]
+        for extension in ["", ".md", ".rst", ".txt"]
+    ]
+
+    # Stops checking once a match is found
+    for format in changelog_formats:
+        changelog = base_path / format
+        if changelog.is_file():
+            return format
+    return None
--- a/src/briefcase/commands/convert.py
+++ b/src/briefcase/commands/convert.py
@@ -799,7 +799,8 @@ To run your application, type:
 
 
 def find_changelog_filename(base_path):
-
+    """Search for a valid changelog file in the specified directory
+    and return its format name if it exists, otherwise return None"""
     changelog_formats = [
         f"{name}{extension}"
         for name in ["CHANGELOG", "HISTORY", "NEWS", "RELEASES"]
--- a/src/briefcase/commands/convert.py
+++ b/src/briefcase/commands/convert.py
@@ -670,8 +670,33 @@ class ConvertCommand(NewCommand):
             copy2(project_dir / "LICENSE", self.base_path / "LICENSE")
 
         # Copy changelog file
-        changelog_file = self.base_path / "CHANGELOG"
-        if not changelog_file.is_file():
+        changelog_name_formats = [
+            "CHANGELOG",
+            "HISTORY",
+            "NEWS",
+            "RELEASES",
+            "CHANGELOG.md",
+            "CHANGELOG.rst",
+            "CHANGELOG.txt",
+            "HISTORY.md",
+            "HISTORY.rst",
+            "HISTORY.txt",
+            "NEWS.md",
+            "NEWS.rst",
+            "NEWS.txt",
+            "RELEASES.md",
+            "RELEASES.rst",
+            "RELEASES.txt",
+        ]
+        matched_changelog_name = None
+
+        for name in changelog_name_formats:
+            changelog = self.base_path / name
+            if changelog.is_file():
+                matched_changelog_name = name
+                break
+
+        if matched_changelog_name is None:
             self.console.warning(
                 f"\nChangelog file not found in '{self.base_path}'. You should either "
                 f"create a new '{self.base_path / 'CHANGELOG'}' file, or rename an "
--- a/src/briefcase/commands/convert.py
+++ b/src/briefcase/commands/convert.py
@@ -674,7 +674,7 @@ class ConvertCommand(NewCommand):
 
         if changelog is None:
             self.console.warning(
-                f"create a new changelog file in {self.base_path}, or rename an "
+                f"Create a new changelog file in {self.base_path}, or rename an "
                 "existing file to a known changelog file name (one of 'CHANGELOG', "
                 "'HISTORY', 'NEWS' or 'RELEASES'; the file may have an extension of "
                 "'.md', '.rst', '.txt', or have no extension)"
--- a/src/briefcase/commands/convert.py
+++ b/src/briefcase/commands/convert.py
@@ -801,7 +801,7 @@ To run your application, type:
 
 def find_changelog_filename(base_path):
     """Find a valid changelog file in a given directory.
-    
+
     :param base_path: The directory to search
     :returns: The filename of the changelog that was found; None if a changelog
         with a known name could not be found.
--- a/src/briefcase/commands/convert.py
+++ b/src/briefcase/commands/convert.py
@@ -669,38 +669,27 @@ class ConvertCommand(NewCommand):
             )
             copy2(project_dir / "LICENSE", self.base_path / "LICENSE")
 
-        # Copy changelog file
-        changelog_name_formats = [
-            "CHANGELOG",
-            "HISTORY",
-            "NEWS",
-            "RELEASES",
-            "CHANGELOG.md",
-            "CHANGELOG.rst",
-            "CHANGELOG.txt",
-            "HISTORY.md",
-            "HISTORY.rst",
-            "HISTORY.txt",
-            "NEWS.md",
-            "NEWS.rst",
-            "NEWS.txt",
-            "RELEASES.md",
-            "RELEASES.rst",
-            "RELEASES.txt",
+        # See if changefile exists
+        matched_changelog_name = False
+        changelog_formats = [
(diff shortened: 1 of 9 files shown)


src/briefcase/platforms/linux/system.py
```diff
@@ -760,15 +760,39 @@ def build_app(self, app: AppConfig, **kwargs):
...
+                "HISTORY.md",
+                "HISTORY.rst",
+                "HISTORY.txt",
+                "NEWS.md",
+                "NEWS.rst",
+                "NEWS.txt",
+                "RELEASES.md",
+                "RELEASES.rst",
+                "RELEASES.txt",
+            ]
```
freakboy3742, line 780: Two notes: 

1. When you're enumerating business logic that includes a complex list like this for the second time, you should be considering whether it should be pulled out as a utility method (or at the very least, a constant). A `find_changelog_filename()` method that returns the match (or None/exception if doesn't exist) would seem appropriate here.
2. You can generate this list in a much less error prone way with a list comprehension: `[f"{name}{ext}" for name in ["CHANGELOG", "HISTORY"] for extension in ["", ".md"]]`

src/briefcase/platforms/linux/system.py
```diff
@@ -1139,18 +1163,43 @@ def _package_rpm(self, app: AppConfig, **kwargs):  # pragma: no-cover-if-is-wind
...
+
+                if matched_changelog_name is None:
                     raise BriefcaseCommandError(
                         """\
-Your project does not contain a CHANGELOG file.
-
-Create a file named `CHANGELOG` in the same directory as your `pyproject.toml`
-with details about the release.
-"""
+        Your project does not contain a CHANGELOG file.
```
freakboy3742, line 1197: The indentation here means the error message will start with 8 spaces when printed on the console.

Also - it references a single filename (CHANGELOG), but the code now accepts multiple valid names. The error message should reflect the fact that there are multiple valid inputs.

freakboy3742 (changes requested):
You've got the core of the logic sorted out. Some comments about code re-use inline. 

The two other major problems with this PR: 

1. There are no tests. It's passing all the existing tests; but you're adding new functionality, but no new tests. This is *always* a red flag - if you're adding functionality, how do you have *any* confidence that your new functionality works?

The fact that it's passing coverage tests is mostly accidental - the existing paths through the code matching "has CHANGELOG" and "doesn't have CHANGELOG" is enough to give you branch coverage. 

At a minimum, you need to be validating that:
* It's generating the expected list of valid changelog filenames.
* CHANGELOG isn't the only filename that works
* No matter what the input changelog filename, the RPM package will compile that into `changelog.gz`
* No matter what the input changeling filename the PKG package will include that file.
* If more than one valid changelog file exists, only one is found (and is found in a predictable way). If more than one is found, it might be also advisable to warn the user - and *that* needs to be validated.

These tests should be making good use of test parameterisation. Any test that uses "CHANGELOG" can be parameterised to accept "NEWS.rst" (or similar) - that 

2. There are no documentation changes. You're modifying behavior - there *must* be a change in documentation somewhere. Otherwise, how will users know what a valid changelog file is?

Lastly, on an administrative note: if this PR will close a ticket, it needs to use the magic words `Fixes #XXX` for the ticket number that it will resolve.

JMah007:
The only concern I have is regarding the issue (Changelog formats) I was working on, during my research I could not find solid information even in the documentation for the packagers specifying what changelog names and formats are "legally" allowed. For rpm the most I could find was that a common practise would be to use CHANGELOG or CHANGELOG.md. Apart from that it seems rpm and pkg dont have an enforced format and hence in my PR I allow all the formats.

freakboy3742:
> For rpm the most I could find was that a common practise would be to use CHANGELOG or CHANGELOG.md. Apart from that it seems rpm and pkg dont have an enforced format and hence in my PR I allow all the formats.

Yeah - you’re not missing anything. The Linux packaging tools are… badly documented. Unfortunately, such is the way with Linux tooling generally; and the closer you get to the core of the system, the worse it gets.

These tools do, however, have very good internal validation - so if there’s a requirement that is a MUST have, they’ll usually raise errors if that requirement isn’t met. 

So, the acid test here is “does it work?” If you’re able to generate an RPM/PKG package *at all* without raising warnings or errors, and any “package linting” tools don’t raise an eyebrow, I happy to call that a success, and handle any “but you’re doing it wrong!!1!!” comments from RPM/PKG users as bug reports that can be corrected as necessary. 

To that end, preserving the user’s choice of filename is the preferred option; if that isn’t valid, then we’re looking for the most consistent/explainable set of modifications that are possible.

src/briefcase/platforms/linux/system.py
```diff
@@ -760,24 +760,25 @@ def build_app(self, app: AppConfig, **kwargs):
...
+            if changelog is None:
                 raise BriefcaseCommandError(
                     """\
-Your project does not contain a CHANGELOG file.
-
-Create a file named `CHANGELOG` in the same directory as your `pyproject.toml`
+Your project does not contain a changelog file with a valid file name.
+Create a changelog file with the following as its name (CHANGELOG, HISTORY, NEWS or RELEASES)
+with extensions (.md, .rst, .txt or no extension) in the same directory as your `pyproject.toml`
 with details about the release.
```
freakboy3742, line 771:
```suggestion
Your project does not contain a changelog file with a known file name. You 
must provided a changelog file in the same directory as your `pyproject.toml`,
with a known changelog file name (one of 'CHANGELOG', 'HISTORY', 'NEWS' or 
'RELEASES'; the file may have an extension of '.md', '.rst', or '.txt', or have 
no extension).
```

src/briefcase/platforms/linux/system.py
```diff
@@ -1141,16 +1142,20 @@ def _package_rpm(self, app: AppConfig, **kwargs):  # pragma: no-cover-if-is-wind
...
+                if changelog is None:
                     raise BriefcaseCommandError(
                         """\
-Your project does not contain a CHANGELOG file.
-
-Create a file named `CHANGELOG` in the same directory as your `pyproject.toml`
+Your project does not contain a changelog file with a valid file name.
+Create a changelog file with the following as its name (CHANGELOG, HISTORY, NEWS or RELEASES)
+with extensions (.md, .rst, .txt or no extension) in the same directory as your `pyproject.toml`
 with details about the release.
```
freakboy3742, line 1153:
```suggestion
Your project does not contain a changelog file with a known file name. You 
must provided a changelog file in the same directory as your `pyproject.toml`,
with a known changelog file name (one of 'CHANGELOG', 'HISTORY', 'NEWS' or 
'RELEASES'; the file may have an extension of '.md', '.rst', or '.txt', or have 
no extension).
```

src/briefcase/platforms/linux/system.py
```diff
@@ -1295,6 +1304,22 @@ class LinuxSystemPublishCommand(LinuxSystemMixin, PublishCommand):
     description = "Publish a Linux system project."
 
 
+def find_changelog_filename(base_path):
```
freakboy3742, line 1307: Two comments:
1. This method should have a docstring, describing what it does
2. Changelogs are a cross-platform feature, so this discovery mechanism should be in a cross-platform location. Since it's could be used by the `convert` command, I'd suggest that as a better location for defining this method.

tests/commands/convert/test_migrate_necessary_files.py
```diff
@@ -190,17 +190,27 @@ def test_warning_without_changelog_file(
...
     )
 
 
+@pytest.mark.parametrize(
+    "changelog_filename",
+    [
+        f"{name}{extension}"
+        for name in ["CHANGELOG", "HISTORY", "NEWS", "RELEASES"]
+        for extension in ["", ".md", ".rst", ".txt"]
+    ],
```
freakboy3742, line 205: While this is exhaustive, each test does take *some* time to run, so it would be worth    reducing this to maybe 2 explicit representative cases (say, `CHANGELOG` and `NEWS.txt`), and only do the exhaustive test in the actual `find_changelog_file` method.

tests/platforms/linux/system/test__find_changelog_filename.py
```diff
@@ -0,0 +1,48 @@
+import pytest
+
+from briefcase.platforms.linux.system import find_changelog_filename
+
+from ....utils import create_file
+
+
+def test_no_changelog(tmp_path):
```
freakboy3742, line 8: Tests should all have a docstring that explain what is being tested, however, obvious the title might be.

tests/platforms/linux/system/test__find_changelog_filename.py
```diff
@@ -0,0 +1,48 @@
...
+
+def test_multiple_changefile(tmp_path):
+    """
+    More than one changelog exists but only first one according
+    the below search order is accepted:
+
+    CHANGELOG, CHANGELOG.md, CHANGELOG.rst, CHANGELOG.txt,
+    HISTORY, HISTORY.md, HISTORY.rst, HISTORY.txt,
+    NEWS, NEWS.md, NEWS.rst, NEWS.txt,
+    RELEASES, RELEASES.md, RELEASES.rst, RELEASES.txt"""
```
freakboy3742, line 42: There's no need to enumerate the full logic here. 
```suggestion
    """If there's more than one changelog, only one if found."""
```

tests/platforms/linux/system/test_build.py
```diff
@@ -33,9 +33,25 @@ def build_command(tmp_path, first_app):
...
 
 
+@pytest.mark.parametrize(
+    "changelog_filename",
+    [
+        f"{name}{extension}"
+        for name in ["CHANGELOG", "HISTORY", "NEWS", "RELEASES"]
+        for extension in ["", ".md", ".rst", ".txt"]
+    ],
+)
```
freakboy3742, line 43: As previously, there's no need to be exhaustive here; we can lean on the tests of the underlying method to verify *all* the combinations.

tests/platforms/linux/system/test_package__pkg.py
```diff
@@ -146,16 +146,31 @@ def test_verify_docker(package_command, first_app_pkg, monkeypatch):
...
 
 
+@pytest.mark.parametrize(
+    "changelog_filename",
+    [
+        f"{name}{extension}"
+        for name in ["CHANGELOG", "HISTORY", "NEWS", "RELEASES"]
+        for extension in ["", ".md", ".rst", ".txt"]
+    ],
+)
```
freakboy3742, line 156: Again, no need to be exhaustive.

tests/platforms/linux/system/test_package__rpm.py
```diff
@@ -146,11 +146,26 @@ def test_verify_docker(package_command, first_app_rpm, monkeypatch):
...
 
 
+@pytest.mark.parametrize(
+    "changelog_filename",
+    [
+        f"{name}{extension}"
+        for name in ["CHANGELOG", "HISTORY", "NEWS", "RELEASES"]
+        for extension in ["", ".md", ".rst", ".txt"]
+    ],
+)
```
freakboy3742, line 156: More exhaustion here :-)

freakboy3742 (changes requested):
This is getting pretty close. A couple of cosmetic changes to error text and docstrings, and one more opportunity for code re-use by relocating where the `find` method is defined, plus a couple of places where the tests can be simplified. 

Also - keep in mind that when you move the `find` method, you should also move the test file for the find method to reflect its new home.

tests/platforms/linux/system/test_build.py
JMah007: I must have misunderstood the concept of 100 percent coverage then as I thought all branches and possibilities had to be covered. But I do completely understand what you mean by leaning on the tests of the underlying method to fulfill 100 percent coverage as well

src/briefcase/commands/convert.py
```diff
@@ -808,3 +796,20 @@ def __call__(
                 project_overrides=parse_project_overrides(project_overrides),
                 **options,
             )
+
+
+def find_changelog_filename(base_path):
+    """Search for a valid changelog file in the specified directory
+    and return its format name if it exists, otherwise return None"""
```
freakboy3742, line 803: The formatting of docstrings is significant:
- the first line should be a brief description; ideally a single line.
- Parameters are annotated with `:param:`
- Return values are annotated with `:returns:`
```suggestion
    """Find a valid changelog file in a given directory.
    
    :param base_path: The directory to search
    :returns: The filename of the changelog that was found; None if a changelog
        with a known name could not be found.
    """
```

tests/commands/convert/test__find_changelog_filename.py
```diff
@@ -32,15 +33,7 @@ def test_has_changelog(tmp_path, changelog_filename):
...
-    """
-    More than one changelog exists but only first one according
-    the below search order is accepted:
-
-    CHANGELOG, CHANGELOG.md, CHANGELOG.rst, CHANGELOG.txt,
-    HISTORY, HISTORY.md, HISTORY.rst, HISTORY.txt,
-    NEWS, NEWS.md, NEWS.rst, NEWS.txt,
-    RELEASES, RELEASES.md, RELEASES.rst, RELEASES.txt"""
-
+    """If there's more than one changelog, only one if found."""
```
freakboy3742, line 36:
```suggestion
    """If there's more than one changelog, only one is found."""

```

src/briefcase/commands/convert.py
```diff
@@ -669,13 +669,15 @@ def migrate_necessary_files(self, project_dir, test_source_dir, module_name):
...
-        if not changelog_file.is_file():
+        # See if changefile exists
+        changelog = find_changelog_filename(self.base_path)
+
+        if changelog is None:
             self.console.warning(
-                f"\nChangelog file not found in '{self.base_path}'. You should either "
-                f"create a new '{self.base_path / 'CHANGELOG'}' file, or rename an "
-                "already existing changelog file to 'CHANGELOG'."
+                f"Create a new changelog file in {self.base_path}, or rename an "
```
freakboy3742, line 677: You've reproduced the text that I provided in my suggestion; but you haven't applied that change in the way the diff suggested. You've dropped the "entry" text - `f"\nChangelong file not found in..."`.

It's worth noting that when I'm providing suggestions in this form, you can commit them *directly* in Github (including batching a group of them together). Not only does it guarantee that the change is applied *exactly* as intended - it saves you a bunch of effort in checking out the code, making the change, committing, etc:
```suggestion
                f"\nChangelog file not found in {self.base_path!r}. You should either "
                f"create a new changelog file in {self.base_path!r}, or rename an "
```

tests/commands/convert/test_migrate_necessary_files.py
freakboy3742: You haven't responded to this feedback.

freakboy3742 (changes requested):
Getting really close now. A couple more comments - most of them about paying attention to small details, especially in user-facing text.

The one additional comment that isn't inline: you've removed *all* the additional tests for the package command, and haven't made *any* changes to the `migrate_necessary_files` test. In my previous review, I *didn't* say remove *all* the tests. I said cut down the list of test cases from "exhaustive" to "maybe 2 explicit representative cases".

freakboy3742 (approved):
A couple of simplifications to how you've expressed the non-exhaustive test cases - but otherwise, I think that's done it!

Issue #1696, opened 2024-03-14
Normalize contribution docs with Toga


freakboy3742:
### What is the problem or limitation you are having?

Briefcase has a contribution guide for docs and code; so does Toga. However, there are some elements of the Toga docs that would be useful to include in the Briefcase docs. 

### Describe the solution you'd like

Replicate the sections of the Toga contribution guides that aren't Toga-specific in the Briefcase docs. For example, we don't need the discussion about core and backends; but we *do* need the section about not submitting from your main branch, and waiting for feedback.

### Describe alternatives you've considered

Consolidate these docs upstream (on the beeware.org website) so that the content isn't duplicated unnecessarily. However, developing a good narrative flow with an external site like that may be complex.

### Additional context

See also beeware/toga#2453, beeware/rubicon-objc#427

kattni:
A sync between the Toga and Briefcase contribution guides is in progress.

PR #2236, opened 2025-04-14
Normalise contribution docs with Toga

kavi2du:
This pull request updates the Briefcase contribution guides to include the missing sections, closing the gaps with Toga's guides and reducing duplication.

Refs : #1696 

## PR Checklist:


docs/how-to/contribute-code.rst
```diff
@@ -610,6 +610,58 @@ coverage tox environment names, for instance:
...
+if you *don't* do this. If you submit a pull request that is *almost* right, the
+core team member who reviews your pull request may be able to make the necessary
+changes, rather than giving feedback asking for a minor change. However, if you
+submit your pull request from your ``main`` branch, reviewers are prevented from
+making modifications.
+
+Instead, you should make your changes on a *feature branch*. A feature branch
+has a simple name to identify the change that you've made. For example, if
+you've found a bug in Briefcase's layout algorithm, you might create a feature branch
+``fix-windows-signing``. If your bug relates to a specific issue that has been
```
freakboy3742, line 637: The example looks like it's been partially updated here.

freakboy3742 (changes requested):
This has definitely captured a couple of sections from the Toga docs that would benefit from being in Briefcase's docs. One minor note flagged inline.

Also flagging a potential clash with #2237 - whichever one lands first will require a merge to resolve the fact that a file is being altered in one PR and moved in another.

However, the bigger issue I'm seeing is that there are still inconsistencies between the Toga and Briefcase forms of this documentation. There are definitely details in the Toga docs that don't need to be reproduced here (like details on how to run the testbed); and details in the Briefcase docs that should probably be reproduced in the Toga docs (like the "tl;dr" section at the start). 

With the one minor issue and the potential merge conflict resolved, this would be a fine update to Briefcase's docs - but it doesn't really resolve #1696, because there's still lots of inconsistencies, even at the level of section headings and the markup being used at various levels (e.g., top level headings in Briefase are using `-`, but in Toga, they're using `=`). Ideally, we'd almost be able to use a diff tool to keep the two in sync; but as it stands, there's just enough differences that they haven't converged. There's almost a need to 2 parallel PRs that, when both are merged, results in the "best of both worlds" for everything that is common.

So - my question is whether you've got an appetite for the deeper dive, or if this was a fly-by PR that represents a definite improvement, but not the final "best case" scenario? Both answers are fine - I just need to work out whether I'm going to merge this with a couple of minor fixes, or whether this is going to become a more comprehensive set of changes.

kavi2du:
Hi @freakboy3742, thanks for the comments! Please correct me if I’m wrong, but from what I understand, the goal here is for Toga and Briefcase to have one-to-one identical docs, in terms of structure, content, and formatting (except for the parts that are specific to each project). If that’s the case, I’m happy to dig a bit deeper and work on aligning the docs to follow the same structure, same content, and same consistent formatting.

freakboy3742:
> Hi @freakboy3742, thanks for the comments! Please correct me if I’m wrong, but from what I understand, the goal here is for Toga and Briefcase to have one-to-one identical docs, in terms of structure, content (except for the parts that are specific to each project), and formatting. 

Not *completely*. There are - and always will be - *some* differences between Briefcase and Toga. For example, Toga has a testbed app; Briefcase does not. Briefcase's docs won't *ever* have a section about how to run the testbed app. 

However, the broad structure - you need to create an environment, you need to run a test suite, etc - these are common to both. We need everything that *is*  common to be reflected in both locations; and everything that is different to be clearly delineated.

So - there's two goals here:
1. Ensure that any details that are common are reflected in both sources. Right now, Briefcase doesn't mention pre-commit - that's a gap, and it's a gap that this PR fixes. But Toga's docs don't have a tl;dr section - and they could do.
2. Ensure that it's as easy as possible to maintain similarities between the two documents, while preserving the notable differences. It's not just a matter of copying content from Toga to Briefcase; it's ensuring that in future, if we make a change to Briefcase's docs, it's easy to port that change to Toga as well. 

The simple approach to (2) is to ensure that the documents are as similar as possible, so that a straightforward diff comparison tool (I use Kaleidoscope on macOS, but there are many others) can be used to reconcile differences. Right now, Toga and Briefcase docs use different markup for their headings - so a simple merge isn't possible.

The "deeper" fix is to factor out any repeated content so that there's a single source of truth. However, that's more of a "conceptual goal" - I don't have any good ideas on how to implement this in practice. 

So - there's effectively 3 outcomes for this PR:
1.  Finalise the last couple of typos, minor improvements (and merge conflicts with main, now that #2237 has landed), and merge. It's a solid improvement on the status quo, so we can call that a win and move on. This is the easiest option by far - you can probably wrap this up in a hour.
2. Open a second PR on Toga, doing the reverse set of changes - everything that is in the Briefcase docs that isn't in the Toga docs, but is "shared" content. This will also involve normalising markup so that a visual diff *will* show differences, but *only* where the content actually needs to be different. And that will be almost literally how I review it - pull up both documents in a visual diff, and check where the differences are. This is more ambitious - it's definitely a worthwhile contribution, but it's probably going to take a week or two for the pair of PRs to get to a final state. 
3. Address the deeper content duplication issue. I don't have any good ideas on how to implement this, though. It's almost a capstone project in itself :-)

My inclination is that (1) is probably the best option right now. This is an important topic, but it's not the topic of your formal capstone project, so it's somewhat of a distraction to let this turn into a multi-week code dive. I won't discourage you if you want to pursue (2) - but I'd consider it a "additional thing I'm doing in my spare time", rather than part of your capstone work.

freakboy3742 (commented):
From the look of it, CI is failing because of a linting issue (`tox -e docs-lint`) identifying an inconsistency in how headings are being used. This PR updates the "style" of underlines; Sphinx uses the style of underlines to determine the "type" of heading. However, it's not a hard "`---` maps to `<h2>`" - the order is determined by how the headings are used  in the document being processed. So - you need to be consistent across the entire document in how headings are used - if the first usage of a second level heading is an underline, then *every* second level heading needs to be an underline.

freakboy3742 (approved):
Thanks for those changes. I've added some more normalisations between the Toga and Briefcase docs - it's still not completely aligned, but a few more high profile sections have been pulled over.

Issue #1706, opened 2024-03-23
Add file associations to Windows and macOS  installers


davidfokkema:
### What is the problem or limitation you are having?

My application can save files containing the project you're working on. On macOS, I can double-click such a file to start my application and open it. On Windows, this works but I'll first have to manually associate my application with that type of file.

### Describe the solution you'd like

Just as on macOS, I want Windows installers to install and uninstall file associations.

### Describe alternatives you've considered

Briefcase uses WiX but I looked into other tools. These do not create native .msi installers so I dropped that. WiX is terrible, but it seems to be the only open-source tool to generate msi installers.

### Additional context

I've managed to get it working by manually editing a few files. To fix that in Briefcase is not _that_ easy since it touches the create, build and package stages. This is, more or less, what is needed (with my application as an example. The package step creates a file manifest using the WiX harvest tool (heat) which contains _all_ files. The executable stub file must be removed from that manifest, along with its component _and_ the reference to that component in the fragment at the end of the manifest. However, the `File` element must be included in the wxs file created in the create step. It seems to contain some kind of hash and I still need to figure out if and when that hash changes. If it does not, maybe we can choose our own hash and then it's not hard to include the line during the create step. If the hash changes based on the contents of the file created during the build step, we then need to update the .wxs file during the package step. In either case, the wxs file should include a new component for the file associations:
```xml
<Component Id="FileAssociation" Directory="tailor_ROOTDIR" Guid="*">
    <File Id="ProductIcon" Source="icon.ico" />
    <File Id="fil5E92122F5D3D3D9A85322DF825CEEBB9" KeyPath="yes" Source="$(var.SourceDir)\Tailor.exe" />
    <ProgId Id="MyApp.document" Description="Tailor Project" Icon="ProductIcon">
        <Extension Id="tlr" ContentType="application/tailor">
            <Verb Id="open" Command="Open" TargetFile="fil5E92122F5D3D3D9A85322DF825CEEBB9" Argument='"%1"' />
        </Extension>
    </ProgId>
</Component>
```
and that component must be included in the default feature:
```xml
<Feature Id="DefaultFeature" Level="1">
    <ComponentGroupRef Id="tailor_COMPONENTS" />
    <ComponentRef Id="ApplicationShortcuts"/>
    <ComponentRef Id="FileAssociation"/>
</Feature>
```
So, I see basically two options:
[...]

PR #2284, opened 2025-05-13
File associations

davidfokkema:
This PR adds file associations to macOS apps. The support in briefcase.toml is already there and the feature is documented. It wasn't yet generalised for non-package file types.

Refs https://github.com/beeware/briefcase/issues/1706 (the issue)
Refs https://github.com/beeware/briefcase-macOS-app-template/pull/72 (the PR for the macOS app template)
Refs https://github.com/beeware/briefcase-macOS-Xcode-template/pull/82 (the PR for the macOS Xcode template)

This PR completes the work in the briefcase CLI tool.

I tested with https://github.com/davidfokkema/helloworld. This repository may be removed when the PR is merged.

## PR Checklist:


davidfokkema:
I have two code coverage issues of which I'm not sure how to proceed. The aforementioned exception if the plist file is not in the correct location works fine in manual testing, but is not a part of the tests since the location is hardcoded in the file. So no 100% coverage. Should I change the path to a global variable, alter that in testing to check for the exception? Or should I skip the coverage for that branch?

Second issue is the if-statement in document validation where it checks the platform. If it is _not_ macOS, it should _not_ do the mime type to UTI conversion. I never test on macOS that it is not macOS, so coverage.py complains.

freakboy3742:
> There's currently one failed unit test because of a timeout which I don't really understand since it seems that the failing test _mocks_ a download, so... should _never_ time out? Not sure. Restarting the GitHub action will probably solve this for now.

Hrm... I don't think it's doing a download, but it *may* be probing a live URL, which would make it sensitive to a temporary network outage. I haven't seen that failure before, though - I've kicked off the test again, and it looks like it's passed this time.

davidfokkema:
A new issue has arisen: coverage. Coverage.py keeps complaining one branch of an if-statement is never reached in the tests, but I know for a fact that it _is_ reached. If I include an `else: 1 / 0` then the code crashes, if I include an `else: foo = 1` then the code does not crash _and_ coverage.py is happy, but apart from the ugly hack the pre-commit step fails because `foo` is never used. I could fix _that_ with another line `foo += 1` or something like that but that is even uglier. I removed it, and maybe someone else can figure out why coverage.py complains. I believe passing a string to `LSItemContentTypes` takes the 'missing' branch, and that is tested in the `test_document_type_macOS_config_with_list_of_single_content_type` test in `tests/config/test_document_type_config.py`.

mhsmith:
The coverage error is `src/briefcase/config.py 168->exit`. This means that the `if` statement on line 168 has been executed, but it always evaluates to true. So either the `if` is redundant and can be removed, or you need to add a test for the case where the `if` evaluates to false.

davidfokkema:
> The coverage error is `src/briefcase/config.py 168->exit`. This means that the `if` statement on line 168 has been executed, but it always evaluates to true. So either the `if` is redundant and can be removed, or you need to add a test for the case where the `if` evaluates to false.

I don't have much experience with coverage.py, so thanks for taking a look! The if-statement checks whether the supplied content type is a list, and the 'else' means 'no list'. I believe I test that here: https://github.com/beeware/briefcase/blob/5b28dedaa630f15edf4e48999e58dee313485ed6/tests/config/test_document_type_config.py#L236. If I _add_ an else: 1 / 0 statement that test crashes. So either I'm doing something wrong, I'm missing something obvious, or coverage.py is a bit confused.

mhsmith:
Are you sure you're adding the `else` at the correct indentation level? There are several nested `if`s here.

src/briefcase/config.py
```diff
@@ -146,6 +146,43 @@ def validate_document_type_config(document_type_id, document_type):
...
+multiple values in a LSItemContentTypes key is only valid when multiple document
+types are manually grouped together in the Info.plist file. For Briefcase apps,
+document types are always separately declared in the configuration file, so only
+a single value should be provided.
+                """
+                )
+            else:
+                macOS["LSItemContentTypes"] = content_types[0]
+        else:
+            True is True
```
davidfokkema, line 182: These two lines seem to fix the coverage check, but obviously they don't do anything.

davidfokkema: @mhsmith, but this fails the ruff pre-commit check with https://docs.astral.sh/ruff/rules/useless-comparison/. Of course. I don't know what the best course of action is.

mhsmith: We had a similar mysterious coverage failure in https://github.com/beeware/briefcase/pull/2308, but I don't think we ever worked out the cause.

davidfokkema: Ah, I see the commit, thanks. I've now pushed a no-op which hopefully satisfies both the coverage checker _and_ the pre-commit check and will leave that in for now. If it's a bug in coverage.py, this will hopefully be fixed in the future and this can go out. Not sure if I have enough time and inclination to create an MRE from this to file as an issue on coverage.py.

changes/2284.feature.rst
```diff
@@ -0,0 +1 @@
+Added improved document types (file associations) for Windows and macOS platforms
```
freakboy3742, line 1: Since 0.3.23 has been released, this feature is only macOS updates now.
```suggestion
Improved document types and file associations for macOS platforms.
```

src/briefcase/config.py
```diff
@@ -146,6 +146,44 @@ def validate_document_type_config(document_type_id, document_type):
...
+                    f"""
+Document type {document_type_id!r} has multiple content types. Specifying
+multiple values in a LSItemContentTypes key is only valid when multiple document
+types are manually grouped together in the Info.plist file. For Briefcase apps,
+document types are always separately declared in the configuration file, so only
+a single value should be provided.
+                """
+                )
+            else:
+                macOS["LSItemContentTypes"] = content_types[0]
```
freakboy3742, line 180: While I agree that there can only be one LSItemContentType entry, and we should be validating for this case,  I'd prefer to keep the pluralisation of LSItemContentTypes aligned with the storage - that is, normalise *to* a list, and then only use the `[0]` element in the template.

src/briefcase/config.py
```diff
@@ -146,6 +146,44 @@ def validate_document_type_config(document_type_id, document_type):
...
 
+    if sys.platform == "darwin":  # pragma: no-cover-if-not-macos
+        from briefcase.platforms.macOS.utils import mime_type_to_UTI
+
+        macOS = document_type.setdefault("macOS", {})
+        mime_type = document_type.get("mime_type", None)
+        if (uti := mime_type_to_UTI(mime_type)) is not None:
+            macOS.setdefault("is_core_type", True)
+            macOS.setdefault("LSItemContentTypes", uti)
+            macOS.setdefault("LSHandlerRank", "Alternate")
```
freakboy3742, line 157: Just confirming there's no need for a UTTypeConformsTo here?

tests/config/test_document_type_config.py
```diff
@@ -160,3 +163,80 @@ def test_validate_document_invalid_extension(invalid_extension, valid_document):
...
+@pytest.mark.skipif(sys.platform != "darwin", reason="Test runs only on macOS")
+def test_document_type_macOS_config_with_mimetype_single(valid_document):
+    """Valid document types don't raise an exception when validated.
+
+    application/pdf is the only valid MIME type for PDF files.
+    """
+    valid_document["mime_type"] = "application/pdf"
+    validate_document_type_config("ext", valid_document)
+    assert "LSItemContentTypes" in valid_document["macOS"].keys()
+    assert valid_document["macOS"]["LSItemContentTypes"] == "com.adobe.pdf"
```
freakboy3742, line 177: We should be checking `is_core_type`, `LSHandlerRank`, and `UTTypeConformsTo` in this and subsequent tests, just to confirm we're getting all the right keys.

tests/config/test_document_type_config.py
```diff
@@ -160,3 +163,80 @@ def test_validate_document_invalid_extension(invalid_extension, valid_document):
...
+    valid_document["mime_type"] = "custom/mytype"
+    validate_document_type_config("ext", valid_document)
+    assert "LSItemContentTypes" not in valid_document["macOS"].keys()
+
+
+@pytest.mark.skipif(sys.platform != "darwin", reason="Test runs only on macOS")
+def test_mime_type_to_uti_with_nonexisting_coretypes_file(monkeypatch):
+    """Test that mime_type_to_UTI returns None if the coretypes file doesn't exist."""
+    monkeypatch.setattr(utils, "CORETYPES_PATH", "/does/not/exist")
+    assert utils.mime_type_to_UTI("application/pdf") is None
```
freakboy3742, line 210: For consistency, this should either (a) be test at the `validate_document_type_config()` level (i.e., a reproduction of the actual `application/pdf` case, but with the "couldn't find UTI data" response, or a separate test in the macOS module that does a couple of superficial direct checks of `mime_type_to_UTI()`.

tests/config/test_document_type_config.py
```diff
@@ -160,3 +163,80 @@ def test_validate_document_invalid_extension(invalid_extension, valid_document):
...
+
+@pytest.mark.skipif(sys.platform != "darwin", reason="Test runs only on macOS")
+def test_document_type_macOS_config_with_list_of_single_content_type(valid_document):
+    """Single content type is allowed, even if a list is provided.
+
+    If a document type has a single content type, it is converted to a string.
+    """
+    valid_document.setdefault("macOS", {})["LSItemContentTypes"] = "com.adobe.pdf"
+    validate_document_type_config("ext", valid_document)
+    assert valid_document["macOS"]["LSItemContentTypes"] == "com.adobe.pdf"
```
freakboy3742, line 238: This should be split into 2 separate tests - or possibly a parameterised test with 2 inputs, since the output is the same for both.

freakboy3742 (changes requested):
This is getting pretty close; a couple of minor tweaks and suggestions inline.

src/briefcase/config.py
davidfokkema: Done!

tests/config/test_document_type_config.py
davidfokkema: Done

tests/config/test_document_type_config.py
davidfokkema: Done

tests/config/test_document_type_config.py
freakboy3742: All the other tests in this file are based on invoking `validate_document_type_config()`; but this test is invoking `utils.mime_type_to_UTI`. The test suite is broadly organised by test function; a test of `mime_type_to_UTI` should either be in a test file for that method; or the entry point for for this specific test should be modified to use `validate_document_type_config()` (i.e., testing the same thing, but doing so by setting up test conditions that will result in performing a mime type lookup).

To that end - this test file should also be called `test_validate_document_type_config`.

davidfokkema:
Added some documentation to describe the document types, especially the new macOS-specific support. Probably a work in progress, but it is a first draft.

docs/reference/configuration.rst
```diff
@@ -655,11 +657,43 @@ round and square icons, in sizes ranging from 48px to 192px; Briefcase will
...
+<https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types>`__.
+A full list is available at `IANA
+<https://www.iana.org/assignments/media-types/media-types.xhtml>`__. Please
+specify the MIME type of existing document types since they can be used by the
+operating system to find more information about the document type.
+
+If you do not specify a MIME type, Briefcase will generate a default MIME type
+of the *unregistered* type ``application/x-<app name>-<document type id>``, e.g.
+``application/x-myapp-data``. This is not a valid MIME type, and should not be
+used for production applications. It is only used for testing purposes, to allow
```
freakboy3742: AFAIK, this is incorrect. The `application/x-` prefix is defined by [RFC 2046](https://www.rfc-editor.org/rfc/rfc2046.html) for "private and experimental" MIME types - it exists specifically for expansion by end users without needing to go through formal IANA registration - so it *is* valid.

freakboy3742:
FYI - I wanted to let you know that I'm not ignoring this PR; I'm currently buried in post-PyCon US backlog. I'm hoping to get to this in a day or two.

davidfokkema:
Sure, no worries!

docs/reference/platforms/macOS/index.rst
```diff
@@ -151,6 +151,138 @@ only be executable on the host platform on which it was built - i.e., if you bui
...
+==============
+
+General usage of document types is described in the :ref:`document-types`
+section. That may be all you need! However, macOS has some additional quirks
+that you may need to be aware of when defining document types for your app.
+
+Background
+~~~~~~~~~~
+
+First, macOS document types are defined in the ``Info.plist`` file, and the
```
freakboy3742: These docs are great from the perspective of macOS internals, but they're not as practical for a Briefcase user. Reading these docs, my immediate question was "ok... but what do *I* have to do?". 

And - for most users, the answer is "nothing". I've pushed an update to these docs that changes the focus to more of a "how does Briefcase use these settings", rather than trying to explain Apple's Info.plist scheme.

src/briefcase/config.py
```diff
@@ -146,6 +146,58 @@ def validate_document_type_config(document_type_id, document_type):
...
+a single value should be provided.
+                """
+                )
+            else:
+                # This is basically a no-op to satisfy coverage checkers
+                content_types = "is a list with a single value"
+        elif isinstance(content_types, str):
+            # If the content type is a string, convert it to a list
+            macOS["LSItemContentTypes"] = [content_types]
+        else:
```
freakboy3742: These no-op branches can be cleaned up if the content type handling is performed before the UTI handling - I'll push an update to do this.

tests/config/test_AppConfig.py
```diff
@@ -60,6 +62,7 @@ def test_extra_attrs():
                 "extension": "doc",
                 "description": "A document",
                 "url": "https://testurl.com",
+                "mime_type": "text/x.my-doc-type",
```
freakboy3742: This isn't a valid doctype - we should be using `application/x-my-doc-type`.

tests/config/test_is_uti_core_type.py
```diff
@@ -0,0 +1,22 @@
+import sys
+
+import pytest
+
+from briefcase.platforms.macOS import utils
+
+
+@pytest.mark.skipif(sys.platform != "darwin", reason="Test runs only on macOS")
+def test_is_uti_core_type():
```
freakboy3742: This test can be parameterised.

tests/config/test_mime_type_to_uti.py
```diff
@@ -0,0 +1,22 @@
+import sys
+
+import pytest
+
+from briefcase.platforms.macOS import utils
+
+
+@pytest.mark.skipif(sys.platform != "darwin", reason="Test runs only on macOS")
+def test_mime_type_to_uti():
```
freakboy3742: This test can be parameterized.

davidfokkema:
No need to apologise, I first started working on this back in March, 2024, but did not have time to finish this for many, many months. Thank you for your encouragement, guidance, and careful reviews throughout the process! I've looked through your latest changes; I like them. Thanks for merging!

PR #2292, opened 2025-05-19
Switch from black and flake8 to ruff

### LINKED ISSUE #2064: Flake8 line limit is inconsistent with Black

Briefcase's flake8 configuration (set in tox.ini because of... *reasons*) enforces a line length of 119. However, Black format enforces an 88 character limit. So - our code is being formatted to 88 character lines, but that won't format some comments and strings, which can lead to PRs that don't auto-flag a formatting issue.

Briefcase's flake8 configuration should set an 88 characters limit (to match Black). 

Existing code that exceeds the 88 character limit should be reformatted.

None

### Additional context

After changing the tox.ini configuration, `pre-commit run --all` will highlight all the problem lines.

tekktrik:
- Switches from using black and flake8 to ruff.
- Removes references to flake8 in 
- Reformats any files ruff now wants to.
- Fixes a bug in a test the ruff had linted

Improves repository infrastructure

Resolves #2064 

## PR Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,15 +8,6 @@ repos:
       - id: check-docstring-first
       - id: end-of-file-fixer
       - id: trailing-whitespace
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v3.19.1
-    hooks:
-      - id: pyupgrade
-        args: [--py39-plus]
-  - repo: https://github.com/PyCQA/isort
-    rev: 6.0.1
-    hooks:
-      - id: isort
   # Docformatter 1.7.5 isn't compatible with Pre-commit 4.0
   # - repo: https://github.com/PyCQA/docformatter
   #   rev: v1.7.5
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,15 +23,11 @@ repos:
   #   hooks:
   #     - id: docformatter
   #       args: [--in-place, --black]
-  - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 25.1.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.11.10
     hooks:
-      - id: black
-        language_version: python3
-  - repo: https://github.com/PyCQA/flake8
-    rev: 7.2.0
-    hooks:
-      - id: flake8
+      - id: ruff-format
+      - id: ruff-check
   - repo: https://github.com/codespell-project/codespell
     rev: v2.4.1
     hooks:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -230,16 +230,16 @@ norecursedirs = []
 [tool.ruff.lint]
 # In addition to the default rules, these additional rules will be used:
 extend-select = [
-    "E",        # pycodestyle
-    "W",        # pycodestyle
-    "F",        # pyflakes
-    "UP",       # pyupgrade
-    "B",        # flake8-bugbear
-    "SIM",      # flake8-simplify
-    "YTT",      # flake8-2020
-    "ASYNC",    # flake8-async
-    "C4",       # flake8-comprehensions
-    "I",        # isort
+    "E",      # pycodestyle
+    "W",      # pycodestyle
+    "F",      # pyflakes
+    "UP",     # pyupgrade
+    "B",      # flake8-bugbear
+    "SIM",    # flake8-simplify
+    "YTT",    # flake8-2020
+    "ASYNC",  # flake8-async
+    "C4",     # flake8-comprehensions
+    "I",      # isort
 ]
 
 # Below is a list of ignored rules.  Many (if not all) of these have been disabled
@@ -247,27 +247,27 @@ extend-select = [
 # instances of these errors were caught.  Many if not all of these are candidates
 # for future fixes throughout the project.
 ignore = [
-    "E501",     # line length
-    "B007",     # loop control variable not used
-    "B018",     # useless expression
-    "B019",     # memory leaks from functools.lru_cache / functools.cache
-    "B024",     # abstract base class without abstract methods / properties
-    "B027",     # empty method in abstract base class
-    "B904",     # use raise ... from ...
-    "SIM102",   # combine if statements
-    "SIM103",   # return condition directly
-    "SIM105",   # use contextlib.suppress() instead of try-except-pass
-    "SIM108",   # use ternary operators
-    "SIM115",   # use context manager for opening files
-    "SIM117",   # combine with statements
-    "SIM118",   # use "key in dict" instead of "key in dict.keys()"
-    "SIM201",   # use "x != y" instead of "not x == y"
-    "SIM300",   # Yoda condition (0 != x)
-    "SIM905",   # use list literal instead of str.split()
-    "C400",     # unnecessary generator
-    "C408",     # use dict literal instead of dict()
-    "C414",     # unnecessary list() use within sorted()
-    "C416",     # unnecessary list comprehension
+    "E501",    # line length
+    "B007",    # loop control variable not used
+    "B018",    # useless expression
+    "B019",    # memory leaks from functools.lru_cache / functools.cache
+    "B024",    # abstract base class without abstract methods / properties
+    "B027",    # empty method in abstract base class
+    "B904",    # use raise ... from ...
+    "SIM102",  # combine if statements
+    "SIM103",  # return condition directly
+    "SIM105",  # use contextlib.suppress() instead of try-except-pass
+    "SIM108",  # use ternary operators
+    "SIM115",  # use context manager for opening files
+    "SIM117",  # combine with statements
+    "SIM118",  # use "key in dict" instead of "key in dict.keys()"
+    "SIM201",  # use "x != y" instead of "not x == y"
+    "SIM300",  # Yoda condition (0 != x)
+    "SIM905",  # use list literal instead of str.split()
+    "C400",    # unnecessary generator
+    "C408",    # use dict literal instead of dict()
+    "C414",    # unnecessary list() use within sorted()
+    "C416",    # unnecessary list comprehension
 ]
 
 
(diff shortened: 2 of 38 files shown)


pyproject.toml
```diff
@@ -236,6 +227,9 @@ filterwarnings = [
 # need to ensure build directories aren't excluded from recursion
 norecursedirs = []
 
+[tool.ruff.lint]
+select = ["I", "UP"]
```
danyeaw, line 231:
```suggestion
select = ["E4", "E7", "E9", "F", "I", "UP"]
```
The select overrides the defaults, so you need to add them and the additional I and UP. :+1:

danyeaw (changes requested):
Nice, this is looking really good. I noticed that you had an unused import of `lru_cache`, so I think you need to reselect the default Ruff linters with the additional ones.

tekktrik:
Ah, good catch - thanks!

danyeaw (approved):

freakboy3742 (changes requested):
It looks like a lot of the modifications here are caused because of 80 vs 88 line length; it would be preferable to keep the 88 if possible.

pyproject.toml
```diff
@@ -236,6 +227,9 @@ filterwarnings = [
 # need to ensure build directories aren't excluded from recursion
 norecursedirs = []
 
+[tool.ruff.lint]
+select = ["E4", "E7", "E9", "F", "I", "UP"]
```
rmartin16, line 231: FWIW, these are the ruff rules I landed on after reviewing nearly all of them:
```
# E, W   pycodestyle
# F      Pyflakes
# UP     pyupgrade
# B      flake8-bugbear
# SIM    flake8-simplify
# YTT    flake8-2020
# ASYNC  flake8-async
# C4     flake8-comprehensions
# I      isort
--extend-select="E,W,F,UP,B,SIM,YTT,ASYNC,C4,I"
```

Unfortunately, I think that astral is still trying to get around to [creating sane default profiles](https://github.com/astral-sh/ruff/issues/11415) for their rules...so, until then, we're left to assemble our own...

I might also consider using `extend-select` so we inherit whatever defaults astral does eventually land on.

tekktrik:
Couldn't find anything to limit some of the longer line  like in `tests/commands/create/test_install_app_requirements.py` so I would love any pointers or I'm happy to keep looking shortly if that's a deal-breaker.

tests/integrations/docker/test_DockerAppContext__verify.py
```diff
@@ -100,7 +100,7 @@ def test_docker_verify_fail(mock_tools, first_app_config, verify_kwargs, monkeyp
 def test_docker_image_build_fail(mock_tools, first_app_config, verify_kwargs):
     """Failure if Docker image build fails."""
     mock_tools.subprocess._subprocess.Popen.side_effect = subprocess.CalledProcessError(
-        returncode=80, cmd=["docker" "build"]
+        returncode=80, cmd=["dockerbuild"]
```
freakboy3742, line 103: This should be 2 strings, not one concatenated string
```suggestion
        returncode=80, cmd=["docker", "build"]
```
I'm guessing it was split on 2 lines at some point, but the comma was omitted, and black autocleaned it and we didn't notice.

tests/platforms/linux/system/test_run.py
```diff
@@ -359,7 +359,7 @@ def test_run_console_app(run_command, first_app, tmp_path):
     run_command.run_app(first_app, test_mode=False, passthrough=[])
 
     # The process was started
-    run_command.tools.subprocess.run.mock_calls == [
+    assert run_command.tools.subprocess.run.mock_calls == [
```
freakboy3742, line 362: Wow... that's a pretty glaring error...

pyproject.toml
```diff
@@ -236,6 +227,63 @@ filterwarnings = [
...
+# E, W   pycodestyle
+# F      Pyflakes
+# UP     pyupgrade
+# B      flake8-bugbear
+# SIM    flake8-simplify
+# YTT    flake8-2020
+# ASYNC  flake8-async
+# C4     flake8-comprehensions
+# I      isort
+extend-select = ["E" ,"W", "F", "UP", "B", "SIM", "YTT", "ASYNC", "C4", "I"]
```
freakboy3742, line 240: Can these comments be inlined against the definition? e.g.,
```
extend-select = [
    "E",  # pycodestyle
    "W",  # pycodestyle
    "F",  # pyflakes
...
```
It will take up the same amount of space, but there's immediate correspondence between definition and explanation. Same for the exception cases below.

pyproject.toml
```diff
@@ -236,6 +227,63 @@ filterwarnings = [
...
+# SIM105    use contextlib.suppress() instead of try-except-pass
+# B007      loop control variable not used
+# SIM102    combine if statements
+# SIM115    use context manager for opening files
+# B027      empty method in abstract base class
+# B024      abstract base class without abstract methods / properties
+# SIM103    return condition directly
+# C414      unnecessary list() use within sorted()
+# SIM201    use "x != y" instead of "not x == y"
+# SIM118    use "key in dict" instead of "key in dict.keys()"
```
freakboy3742, line 261: It might be worth commenting on why this list of exceptions. If it's a candidate list for fixes that could be made in the future, it would be good to be explicit about them.

freakboy3742 (changes requested):
For the most part this looks pretty uncontroversial - and in some cases, has picked up some glaring errors.

.pre-commit-config.yaml
```diff
@@ -8,30 +8,17 @@ repos:
       - id: check-docstring-first
       - id: end-of-file-fixer
       - id: trailing-whitespace
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v3.19.1
-    hooks:
-      - id: pyupgrade
-        args: [--py39-plus]
```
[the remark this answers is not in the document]
tekktrik: It should be using the versioning in `pyproject.toml`: https://docs.astral.sh/ruff/settings/#target-version

src/briefcase/config.py
```diff
@@ -594,7 +594,7 @@ def parse_config(config_file, platform, output_format, console):
     The current configuration for {section_name} has a 'license' field
     that is specified as a string:
 
-        license = "{config['license']}"
+        license = "{config["license"]}"
```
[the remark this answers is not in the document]
tekktrik: Let me look into this.  I ran `rudd` locally but it still should have used the target version.

pyproject.toml
tekktrik: Definitely!

src/briefcase/config.py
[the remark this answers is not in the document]
tekktrik: Oh dang, this is a warning too, so an explicit noqa is ugly and bad anyway :/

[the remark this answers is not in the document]
tekktrik: Interestingly, `print()`ing this f-string (with some made-up data for `section_name` and `config`) with 3.9.22 does yield the expected result - I think it's because the f-string is encapsulated by `"""`.  However, changing the changed quotes to be `"""` yields a `SyntaxError`, which seems consistent with the changelog notes:  https://docs.python.org/3/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings

Sure enough, the `SyntaxError` goes away when I run this with 3.12.10.  So I think this is actually valid given the encapsulating triple quotes.

tests/platforms/linux/system/test_run.py
tekktrik: The best kinds of tests are the ones you can't fail! :laughing:

tests/platforms/linux/system/test_mixin__verify_python.py
```diff
@@ -32,10 +32,7 @@ def test_match(create_command, first_app_config, capsys):
         [
             "python3",
             "-c",
-            (
-                "import sys; "
-                "print(f'{sys.version_info.major}.{sys.version_info.minor}')"
-            ),
+            ("import sys; print(f'{sys.version_info.major}.{sys.version_info.minor}')"),
```
tekktrik, line 35: Just wanted to check if you're okay with this on a single line, or if you want me to try to break it up with a `\`.  I'll do this for all command-like lines that got joined back up by ruff.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1146,9 +1145,9 @@ def finalize_notarization(
                                 f"Notarization was rejected: {summary}\n"
                                 + "\n".join(
                                     f"""
-    * ({issue.get('severity', "?")}) {issue.get('path')} [{issue.get('architecture', "unknown architecture")}]
-      {issue.get('message')}
-      {issue.get('docUrl', '(No additional help available)')}"""
+    * ({issue.get("severity", "?")}) {issue.get("path")} [{issue.get("architecture", "unknown architecture")}]
+      {issue.get("message")}
+      {issue.get("docUrl", "(No additional help available)")}"""
```
tekktrik, line 1150: Oh here's another example of the triple-quote f-strings!

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -912,8 +912,7 @@ def notarize(
         # Determine the arguments that would be needed to reproduce this notarization
         if installer_identity:
             identity_args = (
-                f"--identity {identity.id} "
-                f"--installer-identity {installer_identity.id}"
+                f"--identity {identity.id} --installer-identity {installer_identity.id}"
```
tekktrik, line 915: This also got pushed into a single line - I can also disable concats like this one by disabling ruff momentarily if you'd prefer.

freakboy3742 (approved):
Two minor formatting things noted inline, but otherwise, this all looks good to me now.

Given this is a moderately significant procedural change for the project, tagging @mhsmith so he's aware and doesn't have any objections.

tekktrik:
I'm also happy to start working through the ignored list as well - I'll probably break it up into a few PRs so as to keep PR diffs reasonable each round.

mhsmith:
LGTM

Issue #2188, opened 2025-03-10
Distro-specific Linux settings cumulative settings don't accumulate


freakboy3742:
### Describe the bug

If you put a `requires` definition into a Linux distro-specific app configuration block, the requires definition *overrides* the project, app, and Linux definitions, rather than appending to them.

### Steps to reproduce

1. Generate a "hello world" Toga project
2. Modify the `[tool.briefcase.helloworld.linux.debian]` section to include `requires = []`
3. Build the project for a Debian-derived release: `briefcase build linux system --target debian:bookworm`
4. Try to run the app. It will fail saying that `toga` could not be found.

Alternatively, check the build log from step 3; observe that `toga-gtk` was not installed.

### Expected behavior

Adding `requires`, `sources`, `test_requires`, `test_sources` or `permission` definitions to the distro-specific section of an app definition should be cumulative, not overwrites.

### Environment

- Operating System: Linux (or building on macOS, targeting linux)
- Python version: any
- Software versions:
  - Briefcase: Observed in 0.3.21; exists in older releases.

### Logs

```

```

### Additional context

When [distro-specific app settings are processed (LinuxSystemPassiveMixin.finalize_app_config)](https://github.com/beeware/briefcase/blob/main/src/briefcase/platforms/linux/system.py#L200), the code is explicitly *copying* properties, not merging. It should be using the [merge_config](https://github.com/beeware/briefcase/blob/main/src/briefcase/config.py#L420) utility method, rather than iterating over properties and copying them. 

Using this method will also require adding an `update` method to BaseConfig.

A1pha-Z3r0:
Hey, I'm new to PR's and stuff, do you think this would be a good first contribution? If so could you help me out?

freakboy3742:
@A1pha-Z3r0 Yes - this should be a good first contribution - that's why we flagged this issue with the `good-first-issue` tag!

We're happy to help where we can, but we need to know what help you need :-)

Briefcase has a [contribution guide](https://briefcase.readthedocs.io/en/stable/how-to/contribute-code.html) to help you get a development environment set up. 

I'd start by trying to reproduce the problem as described - the ticket description details how to do this. Once you've reproduced the problem, you can try to fix it; again, the ticket description contains a broad description of the change that is required. Implement that change, and confirm you can no longer reproduce the problem; then add a test for what you've changed, and you should be ready to submit a PR.

If you need more pointers than that, let us know what you need and we'll do our best to help!

[...]

PR #2297, opened 2025-05-19
Accumulate Linux distribution-specific configs

ekhunter123:
Bugfix for issue #2188 
Linux distribution-specific config settings can now be set in the pyproject.toml file, and they will accumulate with the higher-level requirements. 

The solution suggested in the issue only included adding an `update()` function to the `BaseConfig` class. However for a more complete solution I opted to create a `setdefault()` function as well. This will allow all `BaseConfig` objects to use the full functionality of the `merge_config()` function.

There are tests for these new functions (which pass), and the tox report indicates 100% code coverage.

Open to any and all suggestions! Thank you.

## PR Checklist:

- [ x] All new features have been tested
- [ x] All new features have been documented
- [ x] I will abide by the code of conduct


freakboy3742 (changes requested):
Thanks for the PR! This looks like it successfully implements the fix suggested in the original issue.

The only issue with this PR: although there's test coverage, there's no test of the specific problem case that this fix corrects. You've added an explicit test of `update` and `set_default`, which is great to for unit testing purposes - but the "functional" test of "are we actually merging correctly in the right place" isn't being asserted. 

Essentially, this is a test that would *fail* if it were run on the current codebase, and *pass* once we apply this patch.

freakboy3742 (approved):
One minor suggestion for the release note; but otherwise, this is great! Thanks for the PR!

ekhunter123:
Thanks @freakboy3742 for the comments and suggestions!

Issue #2071, opened 2024-11-29
Project Name with Extra Double Quote Causes pyproject.toml Error


NADOOITChristophBa:
### Describe the bug

When generating a new application with Briefcase, if the project name contains an extra double quote, it results in a syntax error in the pyproject.toml file. This error prevents the application from running.

### Steps to reproduce

Run the briefcase new command.
Enter a project name with an extra double quote, e.g., NADOO Websurfer".
Complete the rest of the prompts to generate the application.
Attempt to run the application with briefcase dev.
See the error in the terminal output.

### Expected behavior

The briefcase new command should either stop with a warning or remove the symbol

### Environment

Operating System: 14.5 (23F79) (macOS)
Python version: 3.12.7
Software versions:
Briefcase: 0.3.16
Toga: 0.4.1

### Logs

```

```

freakboy3742:
Thanks for the report - I thought we had found all of these, but obviously not.

dev-adas:
I'm on it!

dev-adas:
Running `briefcase dev` does not reproduce the error, however `briefcase run` does.

Log error:
```python
Using app template: https://github.com/beeware/briefcase-macOS-app-template.git, branch v0.3.23
Cloning template 'https://github.com/beeware/briefcase-macOS-app-template.git'...
Using existing template (sha f63353430d7c94e2efd4164c9dcd526ba5d03b5e, updated Thu Apr 17 13:23:14 2025)
  File "/var/folders/_p/5_kmj5r5145_n0410jzn6_xh0000gn/T/tmpcb7a8dh2.py", line 12
    INFO_PATH = Path("badFormalName".app/Contents/Info.plist")
                                                         ^
SyntaxError: unterminated string literal (detected at line 12)
Stopping generation because post_gen_project hook script didn't exit successfully
```

One proposed solution was to make the "fix" with `{{ cookiecutter.formal_name|replace('"', '\\"') }}`  [on line 12 of the macOS post gen template script](https://github.com/beeware/briefcase-macOS-app-template/blob/9d8494a5a365c74f03b075aa8f76c48494516b0b/hooks/post_gen_project.py#L12)

However, fixing this caused cascading issues (invalid strings) with the generation of the `briefcase.toml` corresponding to the user's application build.

See example for formal_name as `badFormalName"`
```python
# Generated using Python 3.13.3
[briefcase]
# This is the start of the framework-based support package era.
target_version = "0.3.20"

[paths]
[...]

PR #2298, opened 2025-05-19
Bugfix-2071: Fixed bug where a double quote in the Formal Name would cause invalid strings.

dev-adas:
This addresses #2071 (details were also left on this comment https://github.com/beeware/briefcase/issues/2071#issuecomment-2891810167)

Formal Names with a double quote was causing invalid strings, this was causing the `post_gen_project` hook for MacOS to fail.

Log error:
```python
Using app template: https://github.com/beeware/briefcase-macOS-app-template.git, branch v0.3.23
Cloning template 'https://github.com/beeware/briefcase-macOS-app-template.git'...
Using existing template (sha f63353430d7c94e2efd4164c9dcd526ba5d03b5e, updated Thu Apr 17 13:23:14 2025)
  File "/var/folders/_p/5_kmj5r5145_n0410jzn6_xh0000gn/T/tmpcb7a8dh2.py", line 12
    INFO_PATH = Path("badFormalName".app/Contents/Info.plist")
                                                         ^
SyntaxError: unterminated string literal (detected at line 12)
Stopping generation because post_gen_project hook script didn't exit successfully
```

One proposed solution was to make the "fix" with `{{ cookiecutter.formal_name|replace('"', '\\"') }}`  [on line 12 of the macOS post gen template script](https://github.com/beeware/briefcase-macOS-app-template/blob/9d8494a5a365c74f03b075aa8f76c48494516b0b/hooks/post_gen_project.py#L12)

However, fixing this caused cascading issues (invalid strings) with the generation of the `briefcase.toml` corresponding to the user's application build.

See example for formal_name as `badFormalName"`
```python
# Generated using Python 3.13.3
[briefcase]
# This is the start of the framework-based support package era.
target_version = "0.3.20"

[paths]
app_path = "badFormalName".app/Contents/Resources/app"
app_packages_path = "badFormalName".app/Contents/Resources/app_packages"
info_plist_path = "badFormalName".app/Contents/Info.plist"
entitlements_path = "Entitlements.plist"
support_path = "badFormalName".app/Contents/Frameworks"
runtime_path = "Python.xcframework/macos-arm64_x86_64/Python.framework"
support_revision = 6
stub_binary_revision = 11
cleanup_paths = [
    "badFormalName".app/Contents/Frameworks/Python.framework/**/Headers",
    "badFormalName".app/Contents/Frameworks/Python.framework/**/include",
    "badFormalName".app/Contents/Frameworks/Python.framework/**/config-*-darwin",
    "badFormalName".app/Contents/Frameworks/Python.framework/**/pkg-config",
]
icon = "badFormalName".app/Contents/Resources/badFormalName.icns"
```

Solution:
Do not allow users to create an application with a formal name containing a double quote(s).

## PR Checklist:


freakboy3742 (changes requested):
As noted in person - we can't eliminate this at the level of the new wizard - in part because the new wizard is only one place for a formal name to be specified, but also because a quote *is* a valid formal name. 

What we need here is more complete escaping across the board; but we can't do that on the Briefcase side, because the escaping used depends on the context in which the values is used.

changes/2071.bugfix.rst
```diff
@@ -0,0 +1 @@
+Handle paths on MacOS with double quotes.
```
freakboy3742, line 1: Apps whose formal name includes a `"` are now correctly handled on macOS.

freakboy3742 (changes requested):
This looks good - but it requires a test case to verify against future regressions.

JohananOppongAmoateng:
@freakboy3742 I think we can close this PR the corresponding issue has been closed

PR #2303, opened 2025-05-20
Improving build error messages when build fails due to missing wheels

### LINKED ISSUE #2230: Improve error message when a wheel for one platform is missing

Briefcase builds for iOS and macOS are multi-architecture - they involve compiling for multiple CPU architectures, which potentially involves downloading multiple binary wheels. For example building for iOS means downloading an `iphoneos` and `iphonesimulator` wheel.

However, if *one* of these wheels is missing, unless you're paying close attention, the error message that is returned suggests that *neither* wheel is available.

When performing a build on iOS and macOS, if a wheel isn't available on *one* platform, the error message should make it clear (1) that this issue is a *single* wheel, and (2) which platform is the problem.

I hit this recently testing a new build of numpy wheels in the Python-support-tested project. I copied the iphonesimulator wheel into the `wheels` folder, and ran a build; it reported:
```
Collecting pyspamsum
  Using cached pyspamsum-1.0.7-cp313-cp313-ios_13_0_arm64_iphoneos.whl.metadata (3.9 kB)
WARNING: Cache entry deserialization failed, entry ignored
ERROR: Could not find a version that satisfies the requirement numpy (from versions: none)
ERROR: No matching distribution found for numpy
Installing app requirements for iPhone device... errored

Unable to install requirements. This may be because one of your
requirements is invalid, or because pip was unable to connect
to the PyPI server.
```
At the time, cause I was doing some testing with IPHONEOS_DEPLOYMENT_TARGET, so I initially thought the issue was that the min-version tag wasn't matching. I\it took me a couple of rounds of debugging before I noticed that the issue was the missing *iPhone device* wheel was completely missing.

It would be preferable if, in the case of iOS builds, the error was more explicit - for example:
```
ERROR: No matching distribution found for numpy
Installing app requirements for iPhone device... errored

Unable to install requirements. This may be because one of your
requirements is invalid, or because pip was unable to connect
to the PyPI server.

It may also indicate that an `iphonesimulator` wheel could be found, 
but an `iphoneos` wheel could not be found. 
```

This message will need to be modified depending on which platform install failed; or, in the case of macOS, which architecture was missing.

Do nothing. The error is *there*, you just have to read it. However, as my recent experience shows, it's easy to miss this detail.

### Additional context

The same problem also exists for Android; however, the installation process there is handled by Chaquopy, so it will be more difficult for Briefcase to enhance the error message.

pjfan:
# Description of changes #
I added more helpful error messages (via the `install_hint` parameter on `_pip_install`) when `build iOS` and `build macOS` fail. I also added a new `install_hint` parameter to `_install_app_requirements` per suggestion so that it could be passed in directly instead of having to get passed in via `pip_kwargs`.

# What problem is solved #
`build macOS` and `build iOS` commands can sometimes fail because of a single missing wheel for a particular platform's version of a library.

Ex. 1 - a project that relies on a version of `numpy` before 1.2.0 won't be able to build on macOS if `universal_build=true` because there's no arm64 wheel for numpy versions before 1.2.0.

Ex. 2 - accidentally deleting the `iphonesimulator` wheel will cause `build iOS` to fail even if `iphoneos` is present but the error message won't indicate this / will make it sound like both wheels are missing.

# What issue is this PR related to #
Fixes #2230

## PR Checklist:


src/briefcase/commands/create.py
```diff
@@ -633,6 +633,7 @@ def _install_app_requirements(
         app_packages_path: Path,
         progress_message: str = "Installing app requirements...",
         pip_args: list[str] | None = None,
+        install_hint: str = "",
         pip_kwargs: dict[str, str] | None = None,
```
freakboy3742, line 637: From an organisational perspective, it would be preferable to keep `pip_args` and `pip_kwargs` together.
```suggestion
        pip_args: list[str] | None = None,
        pip_kwargs: dict[str, str] | None = None,
        install_hint: str = "",
```

src/briefcase/commands/create.py
```diff
@@ -643,6 +644,8 @@ def _install_app_requirements(
             requirements should be installed.
         :param progress_message: The waitbar progress message to display to the user.
         :param pip_args: Any additional command line arguments to use when invoking pip.
+        :param install_hint: Additional hint information to provide in the exception
+            message if the pip install call fails.
         :param pip_kwargs: Any additional keyword arguments to pass to the subprocess
             when invoking pip.
```
freakboy3742, line 650: Reordering to match suggested argument order.
```suggestion
        :param pip_args: Any additional command line arguments to use when invoking pip.
        :param pip_kwargs: Any additional keyword arguments to pass to the subprocess
            when invoking pip.
        :param install_hint: Additional hint information to provide in the exception
            message if the pip install call fails.
```

src/briefcase/exceptions.py
```diff
@@ -241,7 +241,9 @@ def __init__(self, install_hint=""):
             f"""\
 Unable to install requirements. This may be because one of your
 requirements is invalid, or because pip was unable to connect
-to the PyPI server.{install_hint}
+to the PyPI server.
+
+{install_hint}
```
freakboy3742, line 246: It's wild to me that this part was pre-existing - we thought about the need for an install hint... and then never used it 🤣 

However, the original form is preferable here; if there's no install hint, we don't want extra blank lines after the exception message. If the install hint needs some extra lines (and I agree it does), those should be factored into the hint itself.
```suggestion
to the PyPI server.{install_hint}
```

src/briefcase/platforms/iOS/xcode.py
```diff
@@ -370,10 +370,12 @@ def _install_app_requirements(
             pip_args=[
                 f"--platform=ios_{ios_min_tag}_arm64_iphoneos",
             ],
+            install_hint="It may also indicate that an `iphonesimulator` wheel could be found, "
+            "but an `iphoneos` wheel could not be found.",
             pip_kwargs={
                 "env": {
                     "PYTHONPATH": str(device_platform_site),
-                },
+                }
```
freakboy3742, line 378: As above, ordering should be consistent (and keep pip args/kwargs together) - unfortunately I can't put this in a code suggestion because there's a deleted line...

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -149,6 +149,14 @@ def _install_app_requirements(
                     "--platform",
                     f"macosx_{macOS_min_tag}_{self.tools.host_arch}",
                 ],
+                install_hint=f"""
+If an {self.tools.host_arch} wheel has not been published for one or more of your requirements,
```
freakboy3742, line 153: Really lucky that both arm and x86 start with a voiced vowel, so we can use "an" in both cases... :-)

With the removal of the newline from the exception usage, this will need an additional newline.
```suggestion
                install_hint=f"""
                
If an {self.tools.host_arch} wheel has not been published for one or more of your requirements,
```

src/briefcase/platforms/iOS/xcode.py
```diff
@@ -387,6 +389,8 @@ def _install_app_requirements(
             pip_args=[
                 f"--platform=ios_{ios_min_tag}_{self.tools.host_arch}_iphonesimulator",
             ],
+            install_hint="It may also indicate that an `iphoneos` wheel could be found, "
+            "but an `iphonesimulator` wheel could not be found.",
```
freakboy3742, line 393: As above, ordering of arguments needs tweaking. There's also a need to add extra whitespace at the start:
```suggestion
            install_hint="""

It may also indicate that an `iphoneos` wheel could be found, but an
`iphonesimulator` wheel could not be found.
"""
```

src/briefcase/platforms/iOS/xcode.py
```diff
@@ -370,10 +370,12 @@ def _install_app_requirements(
             pip_args=[
                 f"--platform=ios_{ios_min_tag}_arm64_iphoneos",
             ],
+            install_hint="It may also indicate that an `iphonesimulator` wheel could be found, "
+            "but an `iphoneos` wheel could not be found.",
```
freakboy3742, line 374: As above, ordering of arguments needs tweaking. There's also a need to add extra whitespace at the start:
```suggestion
            install_hint="""

It may also indicate that an `iphonesimulator` wheel could be found, but an
`iphoneos` wheel could not be found.
"""
```

freakboy3742 (changes requested):
This looks great; the one missing piece at this point is some extra test code to verify that the helper messages are being included when required.

tests/platforms/macOS/app/test_create.py
```diff
@@ -1012,3 +1014,30 @@ def test_install_support_package(
 
     # The legacy content has been purged
     assert not (runtime_support_path / "python-stdlib/old-Python").exists()
+
+
+def test_install_app_requirements_error_adds_install_hint(
```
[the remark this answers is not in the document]
pjfan: updated tests to test for each failure mode

pjfan:
Updated tests to test for each error mode

freakboy3742 (approved):
Awesome - thanks for those updates! 

I've made some cosmetic tweaks to the error message text, but otherwise, this is good to land - Thanks for the PR!

PR #2173, opened 2025-03-01
Add `--debug` option to `briefcase run`

timrid:
This is an first attempt for adding (remote) debugging support for bundled apps. This is not ready yet but i would like to share this anyways.

It supports:
- pdb
- debugpy (for VSCode)

I have tested the following platforms:
- Windows
- Android (on Windows)
- macOS
- iOS Simulator (on macOS)

The following points are missing:
- Test this on an real iPhone (unfortunately I don't have one)
- Currently you have to add `import __briefcase_startup__` in your `__main__.py`. Instead of this it is possible to use .pth files. On windows it is working, but this creates at least problems on Android... I dont have tested iOS and macOS yet.
- As suggested in https://github.com/beeware/briefcase/issues/2147#issuecomment-2646812554 `pyproject.toml` is probably not in the right place for the configuration. @freakboy3742 suggested to add user-level configuration. But this is totally out of scope for this PR. So I used the `pyproject.toml` even if it is not ideal.
- Adding pycharm is probably also possible, but it makes some problems, so it can be added later (https://github.com/beeware/briefcase/issues/2152).
- Enhance the documentation.

Fixes #2147 

## PR Checklist:

- [ ] All new features have been tested
- [ ] All new features have been documented
- [ ] I have read the **CONTRIBUTING.md** file
- [ ] I will abide by the code of conduct

PR #2327, opened 2025-06-04
Change `test_mode` to an app property

timrid:
In this PR, the `test_mode` Flag becomes a property of the `AppConfig` as suggested in https://github.com/beeware/briefcase/pull/2173#pullrequestreview-2873000047. This simplifies the code, as `test_mode` is only required at a few points and no longer has to be passed through all functions.

This refactoring simplifies adding a new `debug_mode` in #2173.

## PR Checklist:

### DIFF
--- a/changes/2327.misc.rst
+++ b/changes/2327.misc.rst
@@ -1,0 +1,1 @@
+Changed `test_mode` to an app property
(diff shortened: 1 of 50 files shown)


freakboy3742 (approved):
It's a big change, but I think the shape of the changes that have been required in the test suite indicates it's the right move - it's mostly removing `test_mode` context from places where it isn't really relevant, so we can inject it explicitly in the handful of places where it matters.

Thanks for taking this on - most appreciated!

PR #2328, opened 2025-06-06
Add briefcase config Command for User-Level and Global Configuration

### LINKED ISSUE #2279: Add ability for user-level configuration options

At present, the only mode of Briefcase configuration is "project level" - if a setting isn't in the `pyproject.toml`, it must be specified manually at the command line of any given Briefcase command.

However, there are some values that will be consistent on a per-user basis for a project, or across *all* projects. There should be a way to configure project per-user and global per-user options for Briefcase.

For example, settings that you may want to define as defaults, but not persist in a `pyproject.toml` may include
* Your preferred iOS/Android simulator
* The credential identifiers to use when signing a project
* Cache or proxy configurations

A `briefcase config` option that allows a setting to be defined:
* `briefcase config iOS.device "iPhone 16e"`
* `briefcase config macOS.identity "...."`

These configuration items would be saved in a `.briefcase/config.toml` file in the project directory; `.briefcase` should be added to the `.gitignore` for a new repository.

`briefcase config --global` option that would save the same options on a *global* basis. Global configuration items would be saved in a `config.toml` stored in a platform-appropriate configuration location (i.e., `PlatformDirs("org.beeware.briefcase", "BeeWare").user_config_dir / "config.toml"`).

Global configuration options should take priority over user-level project options, and project level options should take priority over user-level project options. (i.e., if you configure an "iPhone 14" global option, "iPhone 15" as a user-level project option, "iPhone 15" would take priority; if you specify "iPhone 16" at the command line, "iPhone 16" would take priority).

The configuration structure for configuration options should match pyproject.toml - so there should be possible to configure options at a `global`, `macOS`, and `macOS.Xcode` level, with the same prioritization as they have in `pyproject.toml`. 

Any option that has a command line option and can be set as a default should also allow for a `?` value - so, even if I configure a default simulator of `iPhone 15`, if I run `briefcase run iOS --device ?`, the command should override the configured option and ask.

Do nothing, and continue to require all options to be explicit.

### Additional context

As a first pass, I'd suggest the only options that should be exposed are:

* Author name (used in the `new` wizard)
* Email (used in the `new` wizard)
* simulator choices 
* signing identities

MariosLiapis:
## Summary
This PR introduces a new briefcase `config` command to allow setting and persisting configuration values at both project-specific and global **per-user** levels. This enables users to define default preferences without modifying `pyproject.toml`.

Currently, Briefcase requires all configurations to be either hardcoded in `pyproject.toml` or specified at every command invocation. This PR addresses the following use case gaps:
- Define simulator/emulator preferences
- Set author details 
- Persist signing identities

#### Examples:
```bash
briefcase config iOS.device "iPhone 16"  # project-level default device
briefcase config --global author.name "Jane Doe"  # global author name
briefcase config macOS.identity "Developer ID Application: Jane Doe (ABC123)"  # project-level identity
briefcase config --global author.email "jane@example.com"  # global email
```
### Config Storage
- Project-specific: `.briefcase/config.toml`
- Global: `PlatformDirs("org.beeware.briefcase", "BeeWare").user_config_dir/config.toml`

### Precedence Order
1. CLI argument
2. Project .briefcase/config.toml
3. Global user config
4. pyproject.toml

### Device Selection 
This feature adds fallback logic for selecting a simulator device when using the `briefcase run` command only for mobile platforms like `iOS`, `Android` or `WearOS` (it is not used for desktop or web targets).

**Resolution Order**:
1. CLI flag (`--simulator-device`)
2. Project config from `.briefcase/config.toml`
3. Global user config
4. Interactive prompt

If the CLI flag is set to `?`, like:
```bash
briefcase run iOS --simulator-device ?
```
Briefcase will skip the configuration resolution and instead prompt the user interactively to select a device, even if a default is already set. This behavior is handled by the `Config.get()` method in `integrations/config.py` and is currently only used in the `run` command.

#### Notes: 
1. This functionality (`--simulator-device ?`) is implemented specifically for `run` command and is not applied to other commands like `build`, `create`, or `package`.
2. Device options are not dynamically refreshed during config operations - selection is handled at runtime inside the `run` command. Enhancing this in the future to allow dynamic device discovery and validation during `briefcase config` might improve usability.

Fixes #2279 

## PR Checklist:

##  Implementation Checklist
- [x] Implements briefcase config CLI command
- [x] Supports global and project-level config writing
- [x] Resolves config values with proper precedence
- [x] Integrated config resolution into RunCommand
- [x] .briefcase/ included in .gitignore

## Tests Added
- `test_write_project.py`, `test_write_global.py`: ensure correct writes
- `test_resolutions.py`, `test_fallbacks.py`: validate precedence
- `test_errors.py`: cover edge cases like invalid keys, permission issues
- `test_device_selection.py`: confirm device resolution logic in `RunCommand`

### DIFF
--- a/changes/2320.misc.rst
+++ b/changes/2320.misc.rst
@@ -1,0 +1,1 @@
+Some small cleanups that make the code more readable.
--- a/src/briefcase/commands/create.py
+++ b/src/briefcase/commands/create.py
@@ -725,9 +725,7 @@ class CreateCommand(BaseCommand):
             self.tools.shutil.rmtree(app_path)
         self.tools.os.mkdir(app_path)
 
-        sources = app.sources.copy() if app.sources else []
-        if app.test_mode and app.test_sources:
-            sources.extend(app.test_sources)
+        sources = app.all_sources()
 
         # Install app code.
         if sources:
--- a/src/briefcase/config.py
+++ b/src/briefcase/config.py
@@ -445,6 +445,16 @@ class AppConfig(BaseConfig):
                 paths.append(path)
         return paths
 
+    def all_sources(self) -> list[str]:
+        """Get all sources of the application that should be copied to the app.
+
+        :returns: The Path to the dist-info folder.
+        """
+        sources = self.sources.copy() if self.sources else []
+        if self.test_mode and self.test_sources:
+            sources.extend(self.test_sources)
+        return sources
+
     def main_module(self):
         """The path to the main module for the app.
 


changes/2279.feature.rst
```diff
@@ -0,0 +1 @@
+Added a `briefcase config` command for setting and retrieving per-user and per-project configuration options. This supports setting defaults for values like author details, preferred simulators, and signing identities outside of `pyproject.toml`.
```
freakboy3742, line 1: Changenotes are written in the form they'd appear in the release notes for the next release, rather than descriptions of an implementation.
```suggestion
Briefcase can now store per-project and per-user defaults for some options, such as author details, preferred simulators, and signing identities.
```

src/briefcase/commands/config.py
```diff
@@ -0,0 +1,136 @@
...
+        "  - author.name\n"
+        "  - author.email\n"
+        "  - iOS.device\n"
+        "  - android.device\n"
+        "  - macOS.identity\n"
+        "  - network.proxy\n"
+        "  - network.cache_path\n\n"
+        "Precedence: CLI > project config > global config."
+    )
+    help = "Configure per-project or global settings."
```
freakboy3742, line 36:
```suggestion
    description = "Set a configuration value that Briefcase will use as a default."
```
The `description` key is used for `briefcase --help` - it can't run to multiple lines; and there's no need for a separate `help` key.

src/briefcase/commands/config.py
```diff
@@ -0,0 +1,136 @@
+from pathlib import Path
+
+import tomli
```
freakboy3742, line 3: If you're on Python 3.11 or greater, `tomli` isn't installed; you can use `tomllib` from the standard library. See `base.py` for the import pattern that is used.

src/briefcase/integrations/config.py
freakboy3742, line 1: Why treat this as an integration, rather than as part of the `config` module that already exists?

src/briefcase/commands/run.py
```diff
@@ -214,6 +214,13 @@ def add_options(self, parser):
             dest="appname",
             help="The app to run",
         )
+        # Add a command-line option to specify a target simulator/emulator device
```
freakboy3742, line 217: Simulator device handling is already implemented; and the base `run` command can't include anything that is platform specific.

freakboy3742 (changes requested):
Thanks for the PR. There's a lot of good material here, but there's still a lot of work needed.

The most immediate problem is that the test suite is currently failing. I've flagged at least one reason (the `tomli` import); There also appears to be coverage gaps.

From an architectural perspective, there's also the question of why you've implemented this as an integration, rather than as part of the `config.py` that already exists. Integrations are how we handle wrappers around specific tools (like Xcode or VisualStudio); a configuration file is a file that can be "just read". There is a `file` integration largely because we need to use Docker for some uses, and that leads to situations where we need an abstraction around accessing files.

The way you've integrated the default simulator handling is fundamentally wrong. The individual iOS and Android backends should be interrogating the config for a default value if an option hasn't been provided, rather than adding a whole new layer of simulator device specification.

As far as I can make out, the device configuration is the only use of user config that is currently supported by this PR; but the PR description suggests that there's more implemented.

Lastly, there's also a need for documentation; this is a new feature, introducing a new command; the usage of that command needs to be documented, as do the list of configuration keys that are supported.

MariosLiapis:
Okay got it, I will try to make the necessary changes. To be honest this is my very first contribution to an open-source project, as a part of a university assignment, so I apologize for any misunderstandings, confusion, or errors in the process — I'm still learning and truly appreciate your time, patience, and any feedback you can provide.

freakboy3742:
> Okay got it, I will try to make the necessary changes. To be honest this is my very first contribution to an open-source project, as a part of a university assignment, so I apologize for any misunderstandings, confusion, or errors in the process — I'm still learning and truly appreciate your time, patience, and any feedback you can provide.

Completely understood - and, for what it's worth: for a first contribution as a university assignment, it's a very impressive effort! We're here to provide any feedback or assistance you may require; if you've got any questions, let us know and we're happy to help.

freakboy3742:
@MariosLiapis Just checking in - are you planning to continue working on this, or can we release the issue for someone else to work on it?

MariosLiapis:
@freakboy3742 Yes, sorry for the delay. I was occupied with my exams up until now, but I will definitely continue working on this issue.

freakboy3742:
No worries - just checking if we should consider this ticket available for EuroPython sprint participants.

PR #2336, opened 2025-06-12
Add a check for iCloud after rolling out the template.

### LINKED ISSUE #1808: Code signing fails when running on an iCloud mounted drive
### Describe the bug

While following the Beware tutorial in the Packaging for distribution section, briefcase build fails whit the following error:

    Unable to code sign /Users/arana/Documents/PycharmProjects/Personal/IntroToBeewareRetread/helloworld/build/helloworld/macos/app/Hello World.app.

The actual error being raised by codesign is 

    Resource fork, Finder information, or similar detritus not allowed

### Steps to reproduce

1. Create a helloworld project in a folder that is being mirrored on iCloud
2. Enter briefcase create; briefcase build
3. See error

### Expected behavior

Build complete successfully.

### Screenshots

_No response_

### Environment

- Operating System: MacOS 14.4.1
- Python version: 3.12.0
- Software versions:
  - Briefcase: 0.3.18
  - Toga:
  - ...

### Logs

[briefcase.2024_05_20-12_21_49.build.log](https://github.com/beeware/briefcase/files/15379610/briefcase.2024_05_20-12_21_49.build.log)

### Additional context

Build succeeded on M1 Mac with similar software config.

freakboy3742:
iCloud creates (and aggressively maintains) resource forks on `.app` folders created on iCloud-synchronised drives. These resource forks cannot be removed (or rather, they can - but iCloud will put them back seconds later), but they also prevent app signing from succeeding.

This PR does a check for the resource fork after generating the app template; and again before Build and Package commands (the two commands that trigger signing); if iCloud markers are detected, an error is raised.

The check is performed during `briefcase create` immediately after generating the template; if a problem is detected, the macOS bundle is deleted (since the metadata seems to persist if the `.app` folder is copied to a non-synchronised location. Deletion is not performed on the build/package versions due to the potential for destroying user data, but the error message warns that deletion may be necessary.

As a final safety, the internal error code returned during signing is also surfaced.

Fixes #1808.

## PR Checklist:


src/briefcase/platforms/macOS/__init__.py
```diff
@@ -82,10 +82,92 @@ class macOSMixin:
...
+
+        Optionally, this method will clean up the bundle if the verification fails.
+
+        :param app: The app to check.
+        :param cleanup: Should the app bundle be deleted if verification fails?
+        """
+        if self.is_icloud_synced(self.binary_path(app)):
+            msg = [
+                "Briefcase cannot be used on an iCloud-mounted drive "
+                "(such as your Documents folder)."
```
mhsmith: This message could be confusing if the user's Documents folder isn't shared on iCloud, but their Briefcase project is.

freakboy3742: Good point; I've updated the exception messages.

PR #2338, opened 2025-06-13
Simplify Toga bootstrap

HalfWhitt:
- Those Pack imports weren't being used. `Pack` itself shouldn't often be needed anymore at all (at least by most users), but I could see there being an argument to including some imports of constants just to show it as a starting point.
- Main windows now get the app's title by default.
- Window's content can be specified upon creation. (I think this was at one point not the case, right?)

## PR Checklist:

### DIFF
--- a/src/briefcase/bootstraps/toga.py
+++ b/src/briefcase/bootstraps/toga.py
@@ -5,8 +5,6 @@ class TogaGuiBootstrap(BaseGuiBootstrap):
     def app_source(self):
         return '''\
 import toga
-from toga.style import Pack
-from toga.style.pack import COLUMN, ROW
 
 
 class {{ cookiecutter.class_name }}(toga.App):
@@ -14,13 +12,11 @@ class {{ cookiecutter.class_name }}(toga.App):
         """Construct and show the Toga application.
 
         Usually, you would add your application to a main content box.
-        We then create a main window (with a name matching the app), and
-        show the main window.
+        We then create a main window and show it.
         """
         main_box = toga.Box()
 
-        self.main_window = toga.MainWindow(title=self.formal_name)
-        self.main_window.content = main_box
+        self.main_window = toga.MainWindow(content=main_box)
         self.main_window.show()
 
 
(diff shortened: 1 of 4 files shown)


src/briefcase/bootstraps/toga.py
```diff
@@ -5,22 +5,18 @@ class TogaGuiBootstrap(BaseGuiBootstrap):
     def app_source(self):
         return '''\
 import toga
-from toga.style import Pack
-from toga.style.pack import COLUMN, ROW
```
rmartin16, line 9: Now, I'm curious to know why these were actually here. When we had moved everything in to these bootstraps, I ran black over them and it ofc removed all this but I left them in ultimately; I figured they were for visibility since they are very common Toga imports.

HalfWhitt: That makes sense... If that's the case, maybe we keep `COLUMN` and `ROW` but drop `Pack`, now that it doesn't need to be used directly?

freakboy3742: Yes - they're there because they're not *required* to start the app, but they're 3 symbols that will be needed by almost every app in practice. From a tutorial point of view, it's easier to have them there, rather than ask users to add them.

freakboy3742 (commented):
So - these are all completely legal updates - but I'm not sure it's worth it.

As noted about the style imports - those are so common that they're going to be need by almost everyone, so it's worth including them to save every user that piece of typing. 

The title is *technically* redundant for the MainWindow... but I'd argue it's at least *partially* helpful to point out a common option that users are likely to want to set. Similarly, using `main_window.content = ` hints that "changing content is something you can do"... which is apparently non-obvious to some users.

Plus, if we make these changes, we'd need to change a bunch of tutorial code and examples elsewhere to match. 

So - my inclination is to leave the bootstrap as is for now.

src/briefcase/bootstraps/toga.py
HalfWhitt: I agree about `COLUMN` and `ROW`, but how often are people going to be using `Pack` now that properties can be declared directly on widgets?

HalfWhitt:
I do feel like, were we writing this from scratch now, it would probably make sense to omit `Pack`. But I get not wanting to change horses midstream.

src/briefcase/bootstraps/toga.py
freakboy3742: Yeah - that's fair - Pack probably isn't needed now. But it's not really worth the churn.

PR #2340, opened 2025-06-16
Add suppport for Slint as a toolkit to bootstrap GUI apps from

tronical:
## PR Checklist:


tronical:
Maybe as a piece of background: I work on Slint, we love Python, and I love the way briefcase takes care of all the packaging and templating difficulties (especially iOS!).

I'm happy to make changes, file a separate feature issue for discussion, etc. - would love to see Slint support in briefcase.

freakboy3742:
@tronical Thanks for the contribution - and glad you've found Briefcase useful!

The plugin itself looks great. However, at this point, our preferred approach for new GUI toolkits is for the maintainers of those tools (or interested members of their communities) to add them as third-party plugins, and publish them independently on PyPI. The bootstrap mechanism has been designed so that a Python package can register a `[project.entry-points."briefcase.bootstraps"]` entry point, and that bootstrap will be added to the new project wizard as long as the package is installed. 

Moving GUI library support into plugins lowers our CI overhead, and means we (BeeWare) don't need to stay on top of the changes in the GUI frameworks we support, but don't maintain (or, alternatively, we don't need to give commit access to Briefcase for a maintainer of every GUI framework).

By convention, we recommend naming this plugin with a `briefcasex-` prefix (e.g., `briefcasex-slint`) to indicate a non-official Briefcase plugin; or, wrapping the plugin into the GUI framework itself (so `pip install slint` in the development environment registers the Briefcase plugin).

tronical:
Thanks for the quick feedback :)

> However, at this point, our preferred approach for new GUI toolkits is for the maintainers of those tools (or interested members of their communities) to add them as third-party plugins, and publish them independently on PyPI.

That's a bummer. I understand the reasoning though.

One thing that goes missing with this approach is discovery: Most people will just start with briefcase and this way they're missing out. It would be cool if briefcase could still offer other frameworks but install them implicitly, for example. Or at least print some options:

```
-- GUI Framework -------------------------------------------------------------

What GUI toolkit do you want to use for this project?

Additional GUI bootstraps are available from the community.

Check them out at https://beeware.org/bee/briefcase-bootstraps

  1) Toga
  2) PySide6 (does not support iOS/Android/Web deployment)
  3) Pygame  (does not support iOS/Android/Web deployment)
  4) Console (does not support iOS/Android/Web deployment)
  5) None
  
Additional frameworks can be installed:

  1) Slint (pip install slint)
  2) ...

GUI Framework [1]:
```

tronical:
Closing this as #2342 kindly distills the conclusion.

freakboy3742:
> One thing that goes missing with this approach is discovery: Most people will just start with briefcase and this way they're missing out. It would be cool if briefcase could still offer other frameworks but install them implicitly, for example. Or at least print some options:

Agreed that discoverability is an issue for plugins. Toga itself offers a Briefcase plugin for some additional app bootstraps - but unless you know that option exists, it's almost entirely hidden. The current selection of plugins does link to documentation that lists additional plugins; but that's an external documentation source, and as we all know, nobody *ever* reads the documentation :-)

So - the general UX you've suggested here makes sense to me. It's essentially providing the same information as the documentation, but presenting it in a form that is interactive with the user at the point they want that information. The responsibility for Briefcase is no different - provide a list of "known additional plugins" that can be selected from the list, and maybe some install instructions.

I've opened #2342 to track this idea; I've made a slightly different suggestion for how it might manifest, driven mostly by how the implementation works internally. If you're interested in improving the visibility of a Slint plugin, we'd love to see a PR; or, even just a PR to add to the documentation that is already there.

On that basis, I'm going to close this PR; we can continue the discussion about the menu item (if you're interested) over on the other ticket.

PR #2349, opened 2025-06-23
Update `tox`, `tox test` behavior.

kattni:
- This updates `tox` to match the `ci` target, and removes the `ci` target.
- Also updates `tox test` to run on the current version of Python.

## PR Checklist:
 
### DIFF
--- a/changes/2349.misc.rst
+++ b/changes/2349.misc.rst
@@ -1,0 +1,1 @@
+``tox`` command now runs all tests, and ``tox test`` runs on the current version of Python.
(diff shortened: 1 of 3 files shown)


kattni:
Tests run successfully locally, so I don't think the failure is due to my change, but I could absolutely be wrong here.

kattni:
> This change looks good to me; but if we're removing the `ci` target, we should also update the contribution guide to no longer reference that target, and reflect that tox default behavior is "ci -alike".

In my mind, this was in preparation for the docs sync, and I wasn't thinking about updating the docs yet. Hopefully this is what you had in mind.

PR #2351, opened 2025-06-24
Support `--debug` option on Android

timrid:
This PR adds debugger support for Android.

This PR is based on #2173 to have a working example for it. At first #2173 should be merged, then this PR.

## PR Checklist:

- [] All new features have been documented


src/briefcase/platforms/android/gradle.py
```diff
@@ -216,15 +220,19 @@ def output_format_template_context(self, app: AppConfig):
                 "androidx.swiperefreshlayout:swiperefreshlayout:1.1.0",
             ]
 
+        # Extract test packages, to enable features like test discovery and assertion rewriting.
+        extract_sources = app.test_sources or []
+
+        # In debug mode extract all source packages so that the debugger can get the source code
+        # at runtime. This is necessary for setting breakpoints in VSCode or when using 'll' in pdb.
+        if app.debugger:
+            extract_sources.extend(app.sources)
```
timrid: @mhsmith This is currently a problematic step in this PR, because `extractPackages` in `build.gradle` is not refreshed when rebuilding via `briefcase build -ru`. On other platforms like macOS or Windows that is enough. The only solution that is working for me is to delete the `build/helloworld/android` folder manually.

Additionally adding only `app.sources` will only allow me to set breakpoints in my custom source code, but not into libraries like toga, because they are not extracted an thus the source code is not available for the debugger. Is there a possibility to extract *all* sources in chaquopy?

timrid: To fix these problems I have created https://github.com/beeware/briefcase-android-gradle-template/pull/106 and https://github.com/chaquo/chaquopy/pull/1424. I will update this PR later, to use these new features.

timrid:
This PR is waiting for the following two patches to get merged. With these patches this PR is fully working:
- https://github.com/beeware/briefcase-android-gradle-template/pull/106
- https://github.com/chaquo/chaquopy/pull/1424

src/briefcase/platforms/android/gradle.py
```diff
@@ -315,6 +319,23 @@ def update_app_metadata(self, app: AppConfig):
...
+        with (
+            self.console.wait_bar("Setting packages to extract..."),
+            self.metadata_extract_packages_path(app).open("w", encoding="utf-8") as f,
+        ):
+            if app.debugger:
+                # In debug mode include the .py files and extract all of them so that the debugger can get
+                # the source code at runtime. This is e.g. necessary for setting breakpoints in VS Code.
+                extract_packages = ["*"]
+            else:
+                # Extract test packages, to enable features like test discovery and assertion rewriting.
```
mhsmith: Please wrap at 88 columns.

timrid: Fixed.

mhsmith:
Does this PR need any documentation updates? I guess at least some of the 4 places that say "only supported on Windows, macOS and iOS" will need to be updated. 

@freakboy3742: I don't know if this was mentioned somewhere in the gigantic discussion of #2173, but at what point are we going to stop calling this "experimental"? If we can do that now, then I think we can remove those paragraphs from the `build` and `dev` pages, and let the `pdb` and `vscode` pages be the single source of truth about which platforms they support.

mhsmith:
One unit test is failing because of the template target version change, but the test comment says it creates an empty folder "to mock a created project", which suggests it wasn't intended to actually expand a real template. 

If I understand correctly, the the idea is that the unit tests shouldn't depend on the template repository at all, and the `briefcase-template-repo` and `ref`, which I've just added to the top comment of this PR, should allow it to pass the later "build app" jobs even before https://github.com/beeware/briefcase-android-gradle-template/pull/106 is merged.

timrid:
> Does this PR need any documentation updates? I guess at least some of the 4 places that say "only supported on Windows, macOS and iOS" will need to be updated.

I added Android to the list of supported platforms in the docs.

> One unit test is failing because of the template target version change, but the test comment says it creates an empty folder "to mock a created project", which suggests it wasn't intended to actually expand a real template.

I fixed this by updating the target version constant in the unit tests.

> and the briefcase-template-repo and ref, which I've just added to the top comment of this PR, should allow it to pass the later "build app" jobs even before beeware/briefcase-android-gradle-template#106 is merged.

Wow, I did not know about this feature with `briefcase-template-repo` and `briefcase-template-ref`. This seems completly undocumented. It took me a while to understand how this is working and to find [the code](https://github.com/beeware/.github/blob/0e47df64d02ee56c262c98726f655a745e663acc/.github/actions/app-create/action.yml#L35-L56) where the PR body is parsed. `briefcase-template-repo` will be passed to [`--template` in `briefcase new`](https://briefcase.beeware.org/en/latest/reference/commands/new/?h=new#-t-template-template-template). But that has to be a template for an app based on [`briefcase-template`](https://github.com/beeware/briefcase-template) and not an platform specific template like the `briefcase-android-gradle-template` repo. So I forked the `briefcase-template` and set the correct android template. See [here](https://github.com/timrid/briefcase-template/commit/f9a651d85ede0d209171cb4836ae1930eadbeb50). Now the CI is passing.

mhsmith:
Thanks, I'll look at this next week.

mhsmith (commented):
Works fine for me in both PDB and VS Code, thanks! I have a few issues with the documentation, but they're not specific to Android, so I'll put them in a separate issue (#2621).

PR #2354, opened 2025-06-25
Correct version pin for PyGObject.

freakboy3742:
After we added an upper version pin to PyGObject in #2190, PyGObject published a new version that is *less* than the pinned version, but *doesn't* meet our binary requirements (not depending on libgirepository 2.0).

This modifies the version pin to what it probably should have been to start with a hard <= pin on a known good version, rather than a < pin on a known bad version.

## PR Checklist:

### DIFF
--- a/tests/commands/new/test_build_gui_context.py
+++ b/tests/commands/new/test_build_gui_context.py
@@ -86,10 +86,11 @@ requires = [
         pyproject_table_linux="""\
 requires = [
     "toga-gtk~=0.5.0",
-    # PyGObject 3.52.1 enforces a requirement on libgirepository-2.0-dev. This library
+    # PyGObject 3.50.0 is the last version that is known to be dependent on
+    # libgirepository1.0-dev. Newer releases depend on libgirepository-2.0-dev, which
     # isn't available on Debian 12/Ubuntu 22.04. If you don't need to support those (or
     # older) releases, you can remove this version pin. See beeware/toga#3143.
-    "pygobject < 3.52.1",
+    "pygobject <= 3.50.0",
 ]
 """,
         pyproject_table_linux_system_debian="""\
--- a/tests/commands/new/test_build_gui_context.py
+++ b/tests/commands/new/test_build_gui_context.py
@@ -90,7 +90,7 @@ requires = [
     # libgirepository1.0-dev. Newer releases depend on libgirepository-2.0-dev, which
     # isn't available on Debian 12/Ubuntu 22.04. If you don't need to support those (or
     # older) releases, you can remove this version pin. See beeware/toga#3143.
-    "pygobject <= 3.50.0",
+    "pygobject == 3.50.0",
 ]
 """,
         pyproject_table_linux_system_debian="""\
--- a/changes/2354.bugfix.rst
+++ b/changes/2354.bugfix.rst
@@ -1,0 +1,1 @@
+The version pin for PyGObject has been modified to ensure that the installed version is guaranteed to be compatible with ``libgirepository1.0``.
--- a/src/briefcase/bootstraps/toga.py
+++ b/src/briefcase/bootstraps/toga.py
@@ -60,10 +60,11 @@ requires = [
         return """\
 requires = [
     "toga-gtk~=0.5.0",
-    # PyGObject 3.52.1 enforces a requirement on libgirepository-2.0-dev. This library
+    # PyGObject 3.50.0 is the last version that is known to be dependent on
+    # libgirepository1.0-dev. Newer releases depend on libgirepository-2.0-dev, which
     # isn't available on Debian 12/Ubuntu 22.04. If you don't need to support those (or
     # older) releases, you can remove this version pin. See beeware/toga#3143.
-    "pygobject < 3.52.1",
+    "pygobject <= 3.50.0",
 ]
 """
 
--- a/src/briefcase/bootstraps/toga.py
+++ b/src/briefcase/bootstraps/toga.py
@@ -64,7 +64,7 @@ requires = [
     # libgirepository1.0-dev. Newer releases depend on libgirepository-2.0-dev, which
     # isn't available on Debian 12/Ubuntu 22.04. If you don't need to support those (or
     # older) releases, you can remove this version pin. See beeware/toga#3143.
-    "pygobject <= 3.50.0",
+    "pygobject == 3.50.0",
 ]
 """
 


tests/commands/new/test_build_gui_context.py
```diff
@@ -86,10 +86,11 @@ def main():
         pyproject_table_linux="""\
 requires = [
     "toga-gtk~=0.5.0",
-    # PyGObject 3.52.1 enforces a requirement on libgirepository-2.0-dev. This library
+    # PyGObject 3.50.0 is the last version that is known to be dependent on
+    # libgirepository1.0-dev. Newer releases depend on libgirepository-2.0-dev, which
     # isn't available on Debian 12/Ubuntu 22.04. If you don't need to support those (or
     # older) releases, you can remove this version pin. See beeware/toga#3143.
-    "pygobject < 3.52.1",
+    "pygobject <= 3.50.0",
```
rmartin16: Going back to [this comment](https://github.com/beeware/briefcase/pull/2187#discussion_r1986461528), I think it may be better to pin this to only `3.50.0`. Versions less than `3.50.0` are not compatible with Toga.

freakboy3742: Fair enough. Toga's version pin will combine with this <= pin and only resolve to 3.50.0, so we might as well make the math easier for all concerned.

rmartin16:
Actually, are we sure this is the issue?

The [3.50.0..3.50.1](https://gitlab.gnome.org/GNOME/pygobject/-/compare/3.50.0...3.50.1?from_project_id=416) diff on GitLab is quite limited and I can't reproduce this locally.

<details>
<summary><code>pygobject==3.50.1</code> succeeds</summary>

```
❯ python -m pip install pygobject==3.50.1 --no-cache
Collecting pygobject==3.50.1
  Downloading pygobject-3.50.1.tar.gz (1.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.1/1.1 MB 28.1 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... done
Collecting pycairo>=1.16 (from pygobject==3.50.1)
  Downloading pycairo-1.28.0.tar.gz (662 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 662.5/662.5 kB 84.3 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: pygobject, pycairo
  Building wheel for pygobject (pyproject.toml) ... done
  Created wheel for pygobject: filename=pygobject-3.50.1-cp313-cp313-linux_x86_64.whl size=272554 sha256=fa0f061b33bdb70eee2603db9a69ca9cddf65d6b94cc6b005ca304c212fbe37d
  Stored in directory: /tmp/pip-ephem-wheel-cache-v6ww8ofm/wheels/7e/d1/36/6e250eb59aedf0e230f7010c19ce5c22d49a99ce1cd2f5299e
/home/russell/tmp/venv/lib/python3.13/site-packages/pip/_internal/metadata/importlib/_dists.py:77: DeprecationWarning: Unimplemented abstract methods {'locate_file'}
  return cls(files, info_location)
  Building wheel for pycairo (pyproject.toml) ... done
  Created wheel for pycairo: filename=pycairo-1.28.0-cp313-cp313-linux_x86_64.whl size=129699 sha256=2023d43e33532fdb93fc21729ade4725eec8e8a15e5a36163227dc1de21ba2c2
  Stored in directory: /tmp/pip-ephem-wheel-cache-v6ww8ofm/wheels/18/6d/97/29ac8b19b40caca7aa0fa0fff7cf4e43dec78c23ee31273d76
Successfully built pygobject pycairo
Installing collected packages: pycairo, pygobject
/home/russell/tmp/venv/lib/python3.13/site-packages/pip/_internal/metadata/importlib/_dists.py:77: DeprecationWarning: Unimplemented abstract methods {'locate_file'}
  return cls(files, info_location)
Successfully installed pycairo-1.28.0 pygobject-3.50.1
```
</details>

<details>
<summary><code>pygobject==3.52.3</code> fails</summary>

```
❯ python -m pip install pygobject==3.52.3 --no-cache
Collecting pygobject==3.52.3
  Downloading pygobject-3.52.3.tar.gz (1.2 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.2/1.2 MB 30.8 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [23 lines of output]
      + meson setup /tmp/pip-install-8g8m9dw8/pygobject_738495545b6644cabe434a801af803da /tmp/pip-install-8g8m9dw8/pygobject_738495545b6644cabe434a801af803da/.mesonpy-cf7_n_bq -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md -Dtests=false -Dwheel=true --wrap-mode=nofallback --native-file=/tmp/pip-install-8g8m9dw8/pygobject_738495545b6644cabe434a801af803da/.mesonpy-cf7_n_bq/meson-python-native-file.ini
      The Meson build system
      Version: 1.8.2
      Source dir: /tmp/pip-install-8g8m9dw8/pygobject_738495545b6644cabe434a801af803da
      Build dir: /tmp/pip-install-8g8m9dw8/pygobject_738495545b6644cabe434a801af803da/.mesonpy-cf7_n_bq
      Build type: native build
      Project name: pygobject
      Project version: 3.52.3
      C compiler for the host machine: cc (gcc 11.4.0 "cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0")
      C linker for the host machine: cc ld.bfd 2.38
      Host machine cpu family: x86_64
      Host machine cpu: x86_64
      Program python3 found: YES (/home/russell/tmp/venv/bin/python)
      Found pkg-config: YES (/usr/bin/pkg-config) 0.29.2
      Run-time dependency python found: YES 3.13
      Found CMake: /usr/bin/cmake (3.22.1)
      Run-time dependency girepository-2.0 found: NO (tried pkgconfig and cmake)
      Not looking for a fallback subproject for the dependency girepository-2.0 because:
      Use of fallback dependencies is disabled.
      
      ../meson.build:31:9: ERROR: Dependency 'girepository-2.0' is required but not found.
      
      A full log can be found at /tmp/pip-install-8g8m9dw8/pygobject_738495545b6644cabe434a801af803da/.mesonpy-cf7_n_bq/meson-logs/meson-log.txt
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.

[notice] A new release of pip is available: 24.3.1 -> 25.1.1
[notice] To update, run: pip install --upgrade pip
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```
</details>

freakboy3742:
Ok... now I'm confused. I assumed this would resolve the issue that was reported... but didn't actually test it. Now that I'm testing it, I'm seeing the same thing you are - 3.50.1 works...

So... where did the original reporter go wrong? Back to the ticket, I guess... 

Although - it probably wouldn't be the worst idea in the world to just apply this pin (or apply "<= 3.50.1" at least), because we can confirm "known good" releases, but we can't (EDIT: did say "can") proactively confirm all releases < 3.52.1 will be good.

rmartin16:
One thing that comes to mind is they were possibly using a version of Briefcase before PyGObject had its upper-bound. Maybe Briefcase was already installed and the step to create a virtual environment was skipped. A potentially logical explanation anyway...

mhsmith:
Since the original report turned out to be incorrect, I don't think there's any reason to narrow the version range. All signs point to PyGObject only planning to make the incompatible change in 3.52.1 and later, so the current range is safe. It's unlikely, but if there was some package with a dependency on >= 3.50.1, then this PR would unnecessarily prevent it from being added to an app.

freakboy3742:
True - but if a user's new project has a dependency that specifies 3.50.1, they're in a position to fix that (and to identify that the cause is the hard pin in pyproject.toml). Someone doing the tutorial is a lot more vulnerable, on the basis that they're still working out what pyproject.toml *does*, and we need to explain the fix in the tutorial text, and so on.

But - as you point out, it turns out this isn't actually a problem right now, and it seems *unlikely* to re-emerge unless PyGObject publishes a release with a version number less than 3.52.1 that requires girepository-2.0, so lets close this for now.

Issue #747, opened 2022-05-27
Replace `upgrade` command with more generic `tool` command


freakboy3742:
The `briefcase upgrade` command provides a convenient interface to force an update of the tools Briefcase is managing. However, there are other "tool related" features that would be helpful to expose:

* Versions of currently managed tools
* Versions of tools that exist, but Briefcase can't update (e.g., Xcode)
* Versions of internal components (e.g., the SDKs managed by the Android SDKmanager
* The ability to *execute* tools (e.g., sdkmanager)

We currently compute the version of tools as part of the verification process, but we don't generally surface that information except in debug mode.

The Android backend will dump the contents of the sdkmanger during debug, but doesn't expose a "non-debug" way to list the same information.

The only way to easily invoke the Android SDK manager at present is to use `-v` to see the command line and environment used to issue commands, and manually reproduce those commands. 

There isn't an easy way to start a Docker shell in a project's environment.

**Describe the solution you'd like**

One option would be to replace the `briefcase upgrade` command with a new `briefcase tool` command. This new command would allow for the following usage patterns:

* `briefcase tool --list` - list all tools used by Briefcase, including the version numbers. 
* `briefcase tool --upgrade` - upgrade all tools currently managed by Briefcase
* `briefcase tool --upgrade <toolname>` - upgrade all tools currently managed by Briefcase
* `briefcase tool --run <toolname> <args...>` - run a specific tool with the provided arguments. This should configure any 

**Describe alternatives you've considered**

It may be preferable retain the `upgrade` command as-is, extend it to include *more* of the managed tools (e.g., it doesn't currently display the Xcode version because there's no upgrade path), and include the version numbers as part of the `upgrade --list` output.

The execution of specific tools would then be a feature of a specific backend - probably the `run` command. For example `briefcase run android --sdkmanager <args>` would invoke the SDK manager. This would then allow each backend to provide very specific tool executions, rather than needing to expose a generic API for gathering tools into the upgrade command.

freakboy3742:
Addendum: the contents of `briefcase tools --list` should be included as part of the output when the user runs any command with `-v` - if we ask a user to provide debug detail, automatically getting a dump of all the tools and their versions as part of that dump would be very helpful.

mhsmith:
`briefcase upgrade` should be deprecated, because every single new user probably confuses it with `briefcase update`.

mhsmith:
A couple of cosmetic points about the command line format, arising from #2639:

[...]

PR #2369, opened 2025-07-03
Add `--forward-port` and `--reverse-port` via ADB on Android

timrid:
This PR adds the possibility to forward/reverse a port to an Android app via ADB.

I needed the forwarding/reversing for adding debug capabilitys to Android in https://github.com/beeware/briefcase/pull/2351. But now I also have another use case, where I want to connect the android app to a REST-Server running on my host system. So I though it would be generally a useful feature for briefcase.

## PR Checklist:

### DIFF
--- a/src/briefcase/platforms/android/gradle.py
+++ b/src/briefcase/platforms/android/gradle.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 import datetime
 import re
 import subprocess
@@ -18,7 +19,7 @@ from briefcase.commands import (
 from briefcase.config import AppConfig, parsed_version
 from briefcase.console import ANSI_ESC_SEQ_RE_DEF
 from briefcase.exceptions import BriefcaseCommandError
-from briefcase.integrations.android_sdk import AndroidSDK
+from briefcase.integrations.android_sdk import ADB, AndroidSDK
 from briefcase.integrations.subprocess import SubprocessArgT
 
 
@@ -358,6 +359,22 @@ class GradleRunCommand(GradleMixin, RunCommand):
             help="Shutdown the emulator on exit",
             required=False,
         )
+        parser.add_argument(
+            "--forward-port",
+            action="append",
+            dest="forward_ports",
+            type=int,
+            help="Forward the specified port from host to device.",
+            required=False,
+        )
+        parser.add_argument(
+            "--reverse-port",
+            action="append",
+            dest="reverse_ports",
+            type=int,
+            help="Reverse the specified port from device to host.",
+            required=False,
+        )
 
     def run_app(
         self,
@@ -366,6 +383,8 @@ class GradleRunCommand(GradleMixin, RunCommand):
         device_or_avd=None,
         extra_emulator_args=None,
         shutdown_on_exit=False,
+        forward_ports: list[int] | None = None,
+        reverse_ports: list[int] | None = None,
         **kwargs,
     ):
         """Start the application.
@@ -376,6 +395,8 @@ class GradleRunCommand(GradleMixin, RunCommand):
             be asked to re-run the command selecting a specific device.
         :param extra_emulator_args: Any additional arguments to pass to the emulator.
         :param shutdown_on_exit: Should the emulator be shut down on exit?
+        :param forward_ports: A list of ports to forward for the app.
+        :param reverse_ports: A list of ports to reversed for the app.
         """
         device, name, avd = self.tools.android_sdk.select_target_device(device_or_avd)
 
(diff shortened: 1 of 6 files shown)


src/briefcase/platforms/android/gradle.py
```diff
@@ -358,6 +359,22 @@ def add_options(self, parser):
             help="Shutdown the emulator on exit",
             required=False,
         )
+        parser.add_argument(
+            "--forward-port",
+            action="append",
+            dest="forward_ports",
+            type=int,
+            help="Forward the specified port from host to device.",
+            required=False,
```
mhsmith: I don't know why the existing code specifies required=False so much when that's the default, but there's no need to perpetuate that.

```suggestion
```

src/briefcase/platforms/android/gradle.py
```diff
@@ -358,6 +359,22 @@ def add_options(self, parser):
...
+        )
+        parser.add_argument(
+            "--reverse-port",
+            action="append",
+            dest="reverse_ports",
+            type=int,
+            help="Reverse the specified port from device to host.",
+            required=False,
```
mhsmith:
```suggestion
```

tests/integrations/android_sdk/ADB/test_forward_reverse.py
```diff
@@ -0,0 +1,117 @@
+import subprocess
+
+import pytest
+
+from briefcase.exceptions import BriefcaseCommandError
+
+
+def test_forward(mock_tools, adb):
+    """An port forwarding."""
```
mhsmith:
```suggestion
    """A port forwarding"""
```

tests/integrations/android_sdk/ADB/test_forward_reverse.py
```diff
@@ -0,0 +1,117 @@
...
+    # Mock out the run command on an adb instance
+    mock_tools.subprocess.check_output.side_effect = subprocess.CalledProcessError(
+        returncode=1, cmd=""
+    )
+    with pytest.raises(BriefcaseCommandError):
+        adb.forward(5555, 6666)
+
+
+def test_forward_remove(mock_tools, adb):
+    """An port forwarding removing."""
```
mhsmith:
```suggestion
    """A port forwarding removal."""
```

tests/integrations/android_sdk/ADB/test_forward_reverse.py
```diff
@@ -0,0 +1,117 @@
...
+    # Mock out the run command on an adb instance
+    mock_tools.subprocess.check_output.side_effect = subprocess.CalledProcessError(
+        returncode=1, cmd=""
+    )
+    with pytest.raises(BriefcaseCommandError):
+        adb.forward_remove(5555)
+
+
+def test_reverse(mock_tools, adb):
+    """An port reversing."""
```
mhsmith:
```suggestion
    """A port reversing."""
```

tests/integrations/android_sdk/ADB/test_forward_reverse.py
```diff
@@ -0,0 +1,117 @@
...
+    # Mock out the run command on an adb instance
+    mock_tools.subprocess.check_output.side_effect = subprocess.CalledProcessError(
+        returncode=1, cmd=""
+    )
+    with pytest.raises(BriefcaseCommandError):
+        adb.reverse(5555, 6666)
+
+
+def test_reverse_remove(mock_tools, adb):
+    """An port reversing removing."""
```
mhsmith:
```suggestion
    """A port reversing removal."""
```

tests/integrations/android_sdk/ADB/test_forward_reverse.py
```diff
@@ -0,0 +1,117 @@
...
+            "exampleDevice",
+            "forward",
+            "--remove",
+            "tcp:5555",
+        ],
+    )
+
+
+def test_forward_remove_failure(adb, mock_tools):
+    """If port forwarding removing fails, the error is caught."""
```
mhsmith:
```suggestion
    """If port forwarding removal fails, the error is caught."""
```

tests/integrations/android_sdk/ADB/test_forward_reverse.py
```diff
@@ -0,0 +1,117 @@
...
+            "exampleDevice",
+            "reverse",
+            "--remove",
+            "tcp:5555",
+        ],
+    )
+
+
+def test_reverse_remove_failure(adb, mock_tools):
+    """If port reversing removing fails, the error is caught."""
```
mhsmith:
```suggestion
    """If port reversing removal fails, the error is caught."""
```

tests/platforms/android/gradle/test_run.py
```diff
@@ -137,6 +141,58 @@ def test_shutdown_on_exit_option(run_command):
...
         "extra_emulator_args": None,
         "shutdown_on_exit": True,
+        "forward_ports": None,
+        "reverse_ports": None,
+    }
+    assert overrides == {}
+
+
+def test_forward_ports_option(run_command):
+    """The -d option can be parsed."""
```
mhsmith:
```suggestion
    """The --forward-port option can be parsed."""
```

tests/platforms/android/gradle/test_run.py
```diff
@@ -137,6 +141,58 @@ def test_shutdown_on_exit_option(run_command):
...
+        "extra_emulator_args": None,
+        "shutdown_on_exit": False,
+        "forward_ports": [80, 81],
+        "reverse_ports": None,
+    }
+    assert overrides == {}
+
+
+def test_reverse_ports_option(run_command):
+    """The -d option can be parsed."""
```
mhsmith:
```suggestion
    """The --reverse-port option can be parsed."""
```

tests/platforms/android/gradle/test_run.py
```diff
@@ -306,6 +362,94 @@ def mock_stream_output(app, stop_func, **kwargs):
     run_command.tools.mock_adb.kill.assert_not_called()
 
 
+def test_run_forward_reverse_ports(run_command, first_app_config):
+    """An app can be run with passthrough args."""
```
mhsmith:
```suggestion
    """An app can be run with port forwarding and reversing."""
```

tests/platforms/android/gradle/test_run.py
```diff
@@ -306,6 +362,94 @@ def mock_stream_output(app, stop_func, **kwargs):
...
+    # Set up device selection to return a running physical device.
+    run_command.tools.android_sdk.select_target_device = mock.MagicMock(
+        return_value=("exampleDevice", "ExampleDevice", None)
+    )
```
mhsmith, line 376: There's a lot of stuff here which has been copied from other tests despite not being relevant to port forwarding or reversing. Can we reduce it to just the necessary code?

There should also be a test that the forward / reverse / remove methods are NOT called when there are no forwards.

tests/platforms/android/gradle/test_run.py
```diff
@@ -306,6 +362,94 @@ def mock_stream_output(app, stop_func, **kwargs):
...
+    run_command.tools.mock_adb.install_apk.assert_called_once_with(
+        run_command.binary_path(first_app_config)
+    )
+    run_command.tools.mock_adb.force_stop_app.assert_called_once_with(
+        f"{first_app_config.package_name}.{first_app_config.module_name}",
+    )
+
+    assert run_command.tools.mock_adb.forward.call_count == 2
+    run_command.tools.mock_adb.forward.assert_any_call(80, 80)
+    run_command.tools.mock_adb.forward.assert_any_call(81, 81)
```
mhsmith: Each of these 3-line groups could be replaced by a single assertion on [`mock_calls`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.mock_calls), which would give a much more useful message if it fails.

mhsmith (commented):
Thanks, this is very good. Obviously it only works when the app is launched via Briefcase, but like `--shutdown-on-exit`, it could still be useful for all kinds of scripted situations.

timrid:
> Obviously it only works when the app is launched via Briefcase, but like --shutdown-on-exit, it could still be useful for all kinds of scripted situations.

Yes, that was the intention of these options.

If you need a permanent connection to the Android device, you would have to use ADB manually. For this it would also be practical to have access to the ADB downloaded by `briefcase`. E.g. via a new command `briefcase tool adb`. Then you would use `briefcase tool adb forward tcp:5555 tcp:6666` to have a permanent connection. And to close this connection you can use `briefcase tool adb forward tcp:5555 --remove`. This would give you complete freedom to do every configuration with ADB without searching for the path where ADB is downloaded or manually downloading ADB. But this would be a different feature.

mhsmith:
There are some similar ideas in #747.

tests/platforms/android/gradle/test_run.py
timrid: Good point. I'm not so familiar with the mocking functions yet.

tests/platforms/android/gradle/test_run.py
timrid: I removed unnessesary parts of the testcase and added asserts in another testcase to verify that forward/reverse/remove methods are not called.

tests/platforms/android/gradle/test_run.py
```diff
@@ -211,12 +267,18 @@ def mock_stream_output(app, stop_func, **kwargs):
         f"{first_app_config.package_name}.{first_app_config.module_name}",
     )
 
+    assert run_command.tools.mock_adb.forward.call_count == 0
+    assert run_command.tools.mock_adb.reverse.call_count == 0
```
mhsmith: `assert_not_called` is better, because when the assertion fails, it prints details of the calls that were made.

```suggestion
    run_command.tools.mock_adb.forward.assert_not_called()
    run_command.tools.mock_adb.reverse.assert_not_called()
```

tests/platforms/android/gradle/test_run.py
```diff
@@ -211,12 +267,18 @@ def mock_stream_output(app, stop_func, **kwargs):
...
+
     run_command.tools.mock_adb.start_app.assert_called_once_with(
         f"{first_app_config.package_name}.{first_app_config.module_name}",
         "org.beeware.android.MainActivity",
         [],
     )
 
+    assert run_command.tools.mock_adb.forward_remove.call_count == 0
+    assert run_command.tools.mock_adb.reverse_remove.call_count == 0
```
mhsmith:
```suggestion
    run_command.tools.mock_adb.forward_remove.assert_not_called()
    run_command.tools.mock_adb.reverse_remove.assert_not_called()
```

PR #2375, opened 2025-07-07
Add BeeWare theme to Briefcase documentation

kattni:
Adds BeeWare theme to Briefcase documentation.

## PR Checklist:
 
### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,6 +13,24 @@ import os
 import sys
 from importlib.metadata import version as metadata_version
 
+import beeware_theme
+
+# BeeWare theme override for Furo Sphinx theme to add BeeWare features.
+templates_path = []
+html_static_path = []
+html_css_files = []
+html_context = {}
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -196,7 +214,7 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ["_static"]
+# html_static_path = ["_static"]
```
freakboy3742: So - there *is* content in the `_static` directory... but it's only being referenced by the `html_logo` definition. To avoid future confusion, we should make sure `_static` is preserved on the static path.
```suggestion
html_static_path.append("_static")
```

freakboy3742 (approved):
One minor tweak that doesn't actually impact anything, but should avoid future confusion when static files aren't being served.

Issue #2381, opened 2025-07-09
Resolve `BXXX` series ignored Ruff rules


kattni:
### What is the problem or limitation you are having?

When Ruff was enabled, for review purposes, the identified errors were globally ignored, with the intention of resolving them in subsequent PRs.

### Describe the solution you'd like

This repo currently ignores the following rules from the `BXXX` series:
* `B007`
* `B018`
* `B019`
* `B024`
* `B027`
* `B904`

Work through the issues as follows:
1. Comment out each error number in the `ignore` list found [here](https://github.com/beeware/briefcase/blob/1ffd045752d3a3d920a68d920201d890e8637e49/pyproject.toml#L257-L262).
2. Run `pre-commit run -a` to run the checks.
3. Work through each of the errors in the result until they are all resolved.
4. `pre-commit run -a` should pass without the errors in the `ignore` list.
5. Remove the `BXXX` series of rules from the `ignore` list.
6. If this is the last of the errors in the `ignore` list, remove the [`ignore` list](https://github.com/beeware/briefcase/blob/1ffd045752d3a3d920a68d920201d890e8637e49/pyproject.toml#L251-L267) completely, including the comment above.
7. Submit the PR.

### Describe alternatives you've considered

None.

### Additional context

The `SIMXXX` checks are explicitly disabled in the configuration. Do not remove the commented out `SIM` check, and avoid resolving the `SIMXXX` errors.

JeffersGlass:
I'm having a little look at these, if that's ok!

PR #2389, opened 2025-07-11
Resolve B007, B018, B019, B024, B027 Ruff Rules

JeffersGlass:
Resolves the errors for most of the "bugbear" flags that were being ignored by Ruff: B007, B018, B019, B024, B027. Mostly 'useless' statements, unused loop variables, etc. 

I see there is very little use of `#noqa` in this codebase, which is wonderful. I included one for the warning on `Docker.cache_image()` with a comment explaining why, but there could be other ways of solving the root issue.

With regard to B024 and B027, Ruff is complaining that the ABCs `BaseGuiBootstrap` and `BaseCommand` don't have any abstract methods... but those classes are directly instantiated in the tests, so they cannot currently have abstract methods. I believe the most correct thing to do is not to make them ABCs, though I see what that's meant to communicate. Perhaps comments there indicating that they're intended to be subclassed would be good?

I opted not to include the B904 in this roundup and will submit a separate PR. That error is: `Within an except* clause, raise exceptions with raise ... from err or raise ... from None to distinguish them from errors in exception handling`. Since this has a visible effect on the feedback that users receive (and in some cases means they could receive more or less context that previously), I wanted to make space for feedback on that.

With those errors no longer present (from the transition from Flake8 to Ruff), those errors are no longer ignored in `pyproject.toml`

Refs #2381

## PR Checklist:
- [x] All tests past


src/briefcase/commands/base.py
```diff
@@ -400,7 +399,6 @@ def bundle_path(self, app) -> Path:
         """
         return self.build_path(app) / self.output_format.lower()
 
-    @abstractmethod
     def binary_path(self, app) -> Path:
```
freakboy3742: I'm not sure why this was flagged as an issue to be cleaned up - it *is* an abstract method (there's currently documentation, but no implementation), and if the method isn't implemented, there's going to be some interesting fireworks from all the methods that use `binary_path`.

freakboy3742 (changes requested):
This all looks fairly straightforward; the only raised eyebrow is the abstract method that I've flagged inline. I'm not sure what the right approach is for that one, but it *is* an abstract method, and should probably be retained as one.

freakboy3742:
... and now I see your comment in the PR body. Apologies for the confusion.

BaseCommand is definitely *used* in tests, but I can't see any examples where it's directly instantiated in tests - all the uses I'm seeing are where it's been defined as a base class. What usage are you referring to?

JeffersGlass:
Ah, I think I conflated `BaseGuiBootstrap`, which instantiated in tests in a few places like:
https://github.com/beeware/briefcase/blob/c605f1342ebf9849951864648e2a72cb0bf9c49e/tests/commands/new/test_new_app.py#L54-L56
And `BaseCommand`, which you're right isn't instantiated at all. I'll clean that up.

JeffersGlass:
I think I understand better now - `binary_path()` should surely be an abstract method, since it's required to be implemented in all subclasses. 

However, there are five other methods of `BaseCommand` that seemingly aren't required to be re-implemented in subclasses - and so cannot be `@abstractmethods`, since that will prevent those subclasses from being instantiated - but that have no body, which Ruff doesn't like.

Those empty methods are `verify_tools`, `finalize_app_config`, `verify_app_tools`, `clone_options`, and `add_options`, for what it's worth.

For example, `verify_tools` is an empty method in `BaseCommand`, but the following classes which inherit from `BaseCommand` do not implement it: , `DevCommand, macOSAppRunCommand, UpgradeCommand, macOSAppPublishCommand, OtherDummyCommand, DefaultPackageCommand, DummyUpgradeCommand, macOSAppBuildCommand, DummySigningCommand, DummyInstallCommand, StaticWebBuildCommand, StaticWebPackageCommand, StaticWebRunCommand, WindowsAppRunCommand, WindowsVisualStudioRunCommand`. These were fond by marking `verify_tools` as abstract and running the tests.

Options to fix could include:

1. Make the methods non-empty in BaseCommand(e.g. raising an `UnsupportedHostError` or `NotImplementedError` or similar as appropriate by method) so that they are not empty
2. Remove the empty method from BaseCommand and only have them in subclasses with implementation. (I imagine this is bad for Docs reasons?)
3. Mark the methods as abstract, and modify subclasses/add mixins so that all used subclasses have implementations
4. Adding a `#noqa` to these methods with a comment explaining the empty body

freakboy3742:
> I think I understand better now - `binary_path()` should surely be an abstract method, since it's required to be implemented in all subclasses.
> 
> However, there are five other methods of `BaseCommand` that seemingly aren't required to be re-implemented in subclasses - and so cannot be `@abstractmethods`, since that will prevent those subclasses from being instantiated - but that have no body, which Ruff doesn't like.
> 
> Those empty methods are `verify_tools`, `finalize_app_config`, `verify_app_tools`, `clone_options`, and `add_options`, for what it's worth.

Yes - those are all methods that have deliberately empty default implementations, as "no behavior" *is* a reasonable default behavior for those methods (i.e, if there are no tools to verify, "verify tools" can safely be a no-op). 

`binary_path` is different because there isn't a "reasonable default" that works on every platform, but the method *must* return a value to be meaningful elsewhere.

So - I don't think there's any response needed here, other than reverting the "removal of abstract" that ruff was suggesting.

JeffersGlass:
Just to be a little more concise - if `BaseCommand` is an ABC, Ruff will complain if _any_ of its methods are empty, even if it's on purpose. If you run `pre-commit -a` on the previous commit i just pushed (https://github.com/beeware/briefcase/pull/2389/commits/bd421bd538793af33c29d5ec7ad419bc9153202e) - you can see the error messages like:

```
src/briefcase/commands/base.py:663:5: B027 `BaseCommand.verify_tools` is an empty method in an abstract base class, but has no abstract decorator
    |
663 | /     def verify_tools(self):
664 | |         """Verify that the tools needed to run this Command exist.
665 | |
666 | |         Raises MissingToolException if a required system tool is missing.
667 | |         """
    | |___________^ B027
```

So the question, I think is: better to make the methods non-empty (some other kind of no-op), or tell Ruff that everything's actually ok?

freakboy3742:
> So the question, I think is: better to make the methods non-empty (some other kind of no-op), or tell Ruff that everything's actually ok?

Ah, got it - I missed that detail. Thanks for clarifying.

I guess the question is what sort of "no-op" will make Ruff happy? I'm happy to add a `pass` as an implementation if that will get the job done; historically, that was something that flake8 would complain about. If the "new normal" requires us to put `pass` statements back, then that's fine.

JeffersGlass:
Sadly, Ruff also complains about `pass`, and also  `...`,

It's ok with just a bare `return` though, if that's acceptable?

mhsmith:
Yes, that should be fine.

changes/948.bugfix.rst
freakboy3742, line 1: Not sure where this change note has come from - seems like it might be a merge error.

changes/2394.misc.rst
freakboy3742, line 1: Not sure where this change note has come from - seems like it might be a merge error.

freakboy3742 (approved):
Awesome - thanks for those cleanups!

PR #2402, opened 2025-07-19
fix: toga gui bootstrap imports

amfibio:
This removes the `from toga.style import Pack` import from the TogaGuiBootstrap as it is no longer needed for the minimal application. The relevant unit test was adjusted. 

The [relevant tutorial on what was generated](https://docs.beeware.org/en/latest/tutorial/tutorial-2.html#what-was-generated) already does not include the import as expected. At the same time, the tutorial mentions the `Pack` object, including the import as an advanced technique via a [note](https://docs.beeware.org/en/latest/tutorial/tutorial-2.html#adding-some-content-of-our-own).

There is no PR. 

## PR Checklist:

### DIFF
--- a/src/briefcase/bootstraps/toga.py
+++ b/src/briefcase/bootstraps/toga.py
@@ -5,7 +5,6 @@ class TogaGuiBootstrap(BaseGuiBootstrap):
     def app_source(self):
         return '''\
 import toga
-from toga.style import Pack
 from toga.style.pack import COLUMN, ROW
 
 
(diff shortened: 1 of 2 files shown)


changes/2402.removal.rst
freakboy3742, line 1: The change note is good; but it's a `misc`, not a `removal` - removals are for backwards incompatible changes.

freakboy3742 (approved):
👍 Thanks for the cleanup!

Issue #2334, opened 2025-06-11
Platform-Aware briefcase dev Command with Editable Installs for Web Development


JMah007:
### What is the problem or limitation you are having?

Currently, `briefcase dev` is only supported on desktop platforms. For web backend, contributors must rebuild and repackage the entire project into wheels (e.g. `briefcase run web -r -u) even for minor changes. This process introduces significant wait times and slows down the speed of iteration for those developing Toga's web backend or debugging UI issues in the browser.

### Describe the solution you'd like

We propose extending `briefcase dev` to support web development. The solution should:

### Core Features

- Extend `briefcase dev` to be platform-aware and support web applications
- Use editable installs and `.pth` files instead of rebuilding `.whl` packages for pure Python modules
- Enable immediate code changes to be reflected in the browser with just a page refresh

### Technical Implementation

1. **Dedicated Dev Environment:** Create a separate development virtual environment that contains only the pure Python modules needed for development
2. **Editable Installs:** Install the application and pure Python dependencies (like Toga modules) as editable pip installs in the dev venv
3. **Dynamic Source Mapping:** Use `.pth` files to reference actual source folders and serve the latest versions directly to the browser via symlinks
4. **Static Asset Serving:** Serve Toga Python modules as standard `.py` files to PyScript's in-browser virtual filesystem

### Expected Outcome
- Immediate feedback when making code changes (refresh browser instead of rebuild)
- Faster development iteration cycles
- Seamless integration with the existing Briefcase workflow

### Alternative 1: File Watching with Auto-Rebuild

- Implement a file watcher that automatically triggers rebuilds when source files change
- **Limitation:** Still requires the rebuild step, only automates it. Doesn't eliminate the wait time.

### Alternative 2: Hybrid Approach with Partial Rebuilds

- Only rebuild changed modules instead of the entire application
- **Limitation:** Still involves a build step and doesn't provide the immediacy of serving source files directly.

The proposed solution with editable installs and `.pth` files is preferred because it:

- Leverages existing Python packaging mechanisms
- Provides the fastest possible feedback loop
- Maintains compatibility with the existing Briefcase architecture
- Allows for easy switching between dev and production modes

### Technical Breakdown
The work can be broken down into two main components:

1. **Enable Platform-aware dev commands** - Modify the core `briefcase dev `command to detect and handle web platform differently
2. **Local Python file serving** - Implement the venv setup, editable installs, and `.pth` file reading to serve source files to the browser

[...]

PR #2419, opened 2025-07-29
Make Dev command platform aware

mEp3ii2:
This PR refactors the `briefcase dev` command to make it platform-aware by adding  platform-specific `DevCommand` subclasses, such as `StaticWebDevCommand`, to handle platform-dependent workflows and logic. Currently, development mode for the web platform is not implemented, so an `UnsupportedCommandError` is raised when `dev web` is invoked. Additional tests have been added to verify that these new pathways are covered, and the `dev` command has been validated by successfully running Toga Tutorial 1.

Refs #2334 

## PR Checklist:


docs/reference/commands/dev.rst
```diff
@@ -7,12 +7,24 @@ Run the application in developer mode.
 Usage
 =====
 
-To run the app, run:
+To run your application on the current platform's default output format:
```
freakboy3742:
```suggestion
To run your application in development mode on the current platform's default output format:
```

src/briefcase/platforms/linux/system.py
```diff
@@ -892,6 +893,10 @@ def run_app(
                 )
 
 
+class LinuxSystemDevCommand(LinuxSystemMixin, DevCommand):
+    description = "Run a linux system app in development mode"
```
```suggestion
    description = "Run a Linux system app in development mode"
```

src/briefcase/platforms/windows/__init__.py
```diff
@@ -180,6 +180,10 @@ def run_app(
             )
 
 
+class WindowsDevCommand(DevCommand):
```
freakboy3742: Since we're not anticipating any difference between Windows app an Windows VisualStudio development modes, having this intermediate class is probably overkill. The macOS analog shows how you can avoid this intermediate class.

src/briefcase/platforms/web/static.py
```diff
@@ -447,6 +452,19 @@ class StaticWebPublishCommand(StaticWebMixin, PublishCommand):
...
 
 
+class StaticWebDevCommand(StaticWebMixin, DevCommand):
+    description = "Run a static web project in development mode. (Work in progress)"
+
+    def run_dev_app(self, app: AppConfig, env, passthrough=None, **kwargs):
+        raise UnsupportedCommandError(
+            platform="",
+            output_format="Web",
+            command="Dev",
```
freakboy3742: Web is the platform; the output format is static (we could, for example, deploy to docker).
```suggestion
            platform="web",
            output_format="static",
            command="dev",
```

src/briefcase/platforms/linux/system.py
```diff
@@ -892,6 +893,10 @@ def run_app(
                 )
 
 
+class LinuxSystemDevCommand(LinuxSystemMixin, DevCommand):
```
freakboy3742: In testing, I found an interesting failure mode of this specific inheritance heirarchy: If you run `briefcase dev linux` (or `briefcase dev linux system`) on a macOS machine, the error message says:
```
Linux system projects can only be built on Linux, or on macOS using Docker.
```

This is true... for full packages. However, in the context of dev mode, it's a little misleading. We aren't going to be able to add Docker support for Linux dev mode without a *lot* more work (which is *way* outside scope for what you're trying to do here). 

I suspect the base command here needs to be `LinuxMixin`, so that the Docker behaviors aren't picked up. This may also require defining `supported_host_os`, and a few other keys (see the definition of LinuxSystemPassiveMixin for details).

changes/2334.1.feature.rst
```diff
@@ -0,0 +1 @@
+Briefcase Dev command can now accept platforms arguments
```
freakboy3742: Right now, this should probably be a `misc` rather than feature, as there's no "user facing" manifestation of this command. When we add an implementation for `briefcase dev web`, *that* will be a feature.
```suggestion
Development mode behavior can now be platform-specific.
```

tests/test_cmdline.py
```diff
@@ -227,7 +227,7 @@ def test_dev_command(
 
     assert isinstance(cmd, DevCommand)
     assert cmd.platform == "macOS"
-    assert cmd.output_format is None
+    assert cmd.output_format == "app"
```
freakboy3742: This change is definitely required; but we should also check some other command line manifestations - `briefcase dev`, `briefcase dev macOS`, and `briefcase dev macOS Xcode` all need to work.

freakboy3742 (changes requested):
This looks pretty good; a couple of minor notes inline. 

The only other thing that I noticed is that `briefcase dev iOS` and `briefcase dev Android` are both untested. By good ~~fortune~~planning, they raise an error - but we should add a test to confirm this. Since there isn't an actual dev command to instantiate, these tests will need to be in the `test_cmdline` test.

src/briefcase/platforms/web/static.py
```diff
@@ -447,6 +452,19 @@ class StaticWebPublishCommand(StaticWebMixin, PublishCommand):
...
 
 
+class StaticWebDevCommand(StaticWebMixin, DevCommand):
+    description = "Run a static web project in development mode. (Work in progress)"
+
+    def run_dev_app(self, app: AppConfig, env, passthrough=None, **kwargs):
+        raise UnsupportedCommandError(
+            platform="web",
+            output_format="static",
+            command="Dev",
```
```suggestion
            command="dev",
```

tests/platforms/web/static/test_run_dev.py
```diff
@@ -0,0 +1,26 @@
...
+        base_path=tmp_path / "base_path",
+        data_path=tmp_path / "briefcase",
+    )
+
+
+def test_run_dev_app_unsupported(dev_command, first_app_built):
+    with pytest.raises(
+        UnsupportedCommandError,
+        match=re.escape(
+            "The Dev command for the web static format has not been implemented (yet!)."
```
freakboy3742:
```suggestion
            "The dev command for the web static format has not been implemented (yet!)."
```

freakboy3742 (approved):
Awesome! One minor typo, and this is good to land.

PR #2422, opened 2025-08-01
Fix some Liskov Substitution Principle crimes

ambv:
Makes mypy happier. The gist of it is that it should be possible to use instances of subclasses in any place an instance of a base class would be accepted. That means you:
- can't narrow field types in subclasses (this is why `platform_target_version` needs to still specify it's either a string or None);
- can't remove arguments from methods in subclasses or change their return type (that's why `run_app` and `install_app_requirements` were adjusted);
- assign `None` to fields in subclasses that declared a non-optional type in a base class (hence the `output_format` change to an empty string, which ends up being cleaner than doing the `str | None` forever like we did in `platform_target_version`).

## PR Checklist:

- [ ] All new features have been documented (this is effectively a refactor)

### DIFF
--- a/changes/2422.misc.rst
+++ b/changes/2422.misc.rst
@@ -1,0 +1,1 @@
+Some Liskov substitution issues with typing were resolved.
--- a/src/briefcase/commands/convert.py
+++ b/src/briefcase/commands/convert.py
@@ -27,7 +27,7 @@ class ConvertCommand(NewCommand):
     cmd_line = "briefcase convert"
     command = "convert"
     platform = "all"
-    output_format = None
+    output_format = ""
     description = "Set up an existing project for Briefcase."
 
     @cached_property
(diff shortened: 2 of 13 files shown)


ambv:
Towncrier's failing but I don't think this is a notable change. More sensible ones are coming.

mhsmith:
If the change isn't user-visible, you can create a "misc" category change note to pass the CI check.

freakboy3742:
> Towncrier's failing but I don't think this is a notable change. More sensible ones are coming.

To clarify - are saying you have some additional fixes for Liskov issues in the pipeline, or is the PR (modulo the towncrier issue) ready for review?

ambv:
The PR is ready for review, I just don't think there's anything that interesting here as to warrant a changelog entry.

mhsmith:
We use the presence of a "misc" change note file as a way of saying "I've thought about it, and it doesn't need a changelog entry". The text won't appear in the changelog, so you can just use the PR title.

It's a bit less convenient than Python's "skip-news" label, but that's how it's set up at the moment.

freakboy3742 (approved):
I've added a changenote for you, and corrected a couple of minor issues - but otherwise, this looks great - thanks for the cleanups.

PR #2423, opened 2025-08-01
docs: Direct people at cibuildwheel and PyPI uploads

ambv:
I mostly wanted to remove the no longer true assertion that PyPI doesn't support iOS wheels. I also mention cibuildwheel in one go as I've heard mobile-forge is winding down.

## PR Checklist:


freakboy3742 (approved):
Thanks for the suggestion - those are both additions we should definitely make.

I've expanded those notes out to Android as well; tagging @mhsmith for a review to make sure I haven't misrepresented the position of Chaquopy.

PR #2431, opened 2025-08-07
Drop support for Python 3.9, add Python 3.14

freakboy3742:
Drops support for Python 3.9, and adds support for Python 3.14.

Also modifies the Windows support package URL handling so that it can reference pre-releases.

## PR Checklist:

- [x]  I will abide by the code of conduct

### DIFF
--- a/src/briefcase/platforms/macOS/__init__.py
+++ b/src/briefcase/platforms/macOS/__init__.py
@@ -550,8 +550,8 @@ class macOSRunMixin:
             raise BriefcaseCommandError(f"Unable to start app {app.app_name}.")
         finally:
             # Ensure the App also terminates when exiting. The ordering here is a little
-            # odd; the if *could* (should?) be outside the context manager, but coverage
-            # has issues with that arrangement on some Python versions (3.10, 3.14)
+            # odd; the if could/should be outside the context manager, but coverage has
+            # issues with that arrangement on some Python versions (3.10, 3.14)
             with suppress(ProcessLookupError):
                 if app_pid:
                     self.tools.os.kill(app_pid, SIGTERM)
--- a/tests/console/test_Log.py
+++ b/tests/console/test_Log.py
@@ -356,7 +356,11 @@ def test_save_log_to_file_extra_interrupted(mock_now, console, command, tmp_path
 
 
 def test_save_log_to_file_missing_pyproject(
-    mock_now, console, command, tmp_path, monkeypatch
+    mock_now,
+    console,
+    command,
+    tmp_path,
+    monkeypatch,
 ):
     """Log file contains pyproject read exception if it's missing."""
     console.save_log = True


rmartin16:
I've been looking at this for a minute now....and it doesn't make sense.

I confirmed that the "unclosed file" is `/dev/null` opened in `Console`....and the `ResourceWarning` is being raised from pytest's call to `gc.collect()`. Of course, this should call `Console.__del__()` to close `_dev_null`....so the fact it's raising `ResourceWarning` suggests something is broken.

That said....adding this fixture "fixes" the problem....:
```python
@pytest.fixture(autouse=True)
def gc_for_each_test():
    gc.collect()
```

There is some strange interplay happening between pytest, `Console`, and garbage collection.  idk.

freakboy3742:
Can confirm that the extra GC calls the problem for me locally... at the cost of *massively* slowing down the tests. I'm seeing slowdowns from 50s to 170+s - sometimes more. That's a bit of a concern, even as a workaround.

rmartin16:
So, this is interesting....the tests succeed when you use the free-threading version of 3.14.0rc1... 🤔

freakboy3742:
> So, this is interesting....the tests succeed when you use the free-threading version of 3.14.0rc1... 🤔

That checks out - if the problem is related to exact shutdown sequences and garbage collection, FT builds are going to have different performance characteristics.

freakboy3742:
After a bit more investigation, I've found the source of the problem.

This is all caused because of a change in Python 3.14 - the introduction of an [incremental garbage collector](https://github.com/python/cpython/issues/108362). This alters the behavior of garbage collection so that it is called less frequently.

We have historically used the `__del__` handler in Console to ensure that `/dev/null` is closed at the end of a test pass. However, with this new garbage collector, unless garbage collection is invoked explicitly, there's no guarantee that the `__del__` method will be invoked at the end of a test.

Based on discussions like [this one](https://github.com/pytest-dev/pytest/issues/8153), it looks like this is "working as intended" by the pytest maintainers. They consider CPython implementation detail that shouldn't be used as the basis for reliably behavior.
Pytest maintainers consider relying on `__del__` for resource freeing to be depending on an implementation detail - which... it is. We've been relying on the garbage collector doing this cleanup at the end of each test; we can no longer do so.

The three options I can see are:
1. Force a garbage collection pass at the end of every test, which guarantees that `__del__` is invoked on every test. This is what @rmartin16 found as a solution; it works because it forces the GC cleanup pass, but at the cost of a 3-5x slower test suite.
2. Downgrade `ResourceWarning` so that it is an ignored warning. This is "ignoring the problem" - it's not a *completely* unreasonable 
3. Audit the test suite so that whenever we create instance of Console, and modify the fixture creating that instance to ensure the Console instance is explicitly shut down.

Although (3) seems fairly invasive, it's not as bad as you might think. Most of the usages are in test commands instantiating a default DummyConsole(); the only complex cases are in the tests for console itself. The patch is definitely *large*, but it's all fairly straightforward.

I'm about 80% of the way thorough at this point; I'll push an update when I'm done.

rmartin16:
well....if this is because of gc optimizations....here's a fun fix:
```diff
diff --git a/src/briefcase/console.py b/src/briefcase/console.py
index 5486c9a2..69aceb49 100644
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -43,6 +43,8 @@ SENSITIVE_SETTING_RE = re.compile(r"API|TOKEN|KEY|SECRET|PASS|SIGNATURE", flags=
 ANSI_ESC_SEQ_RE_DEF = r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])"
 ANSI_ESCAPE_RE = re.compile(ANSI_ESC_SEQ_RE_DEF)
 
+dev_null_handles = {}
+
 
 def sanitize_text(text: str) -> str:
     """Remove control codes and ANSI escape sequences from a line of text.
@@ -158,6 +160,7 @@ class Console:
         # writing to /dev/null allows Rich to do so without needing to print the
         # logs in the console or save them to file before it is known a file is wanted.
         self._dev_null = open(os.devnull, "w", encoding="utf-8", errors="ignore")
+        dev_null_handles[self._dev_null] = None
         self._log_impl = RichConsole(
             file=self._dev_null,
             record=True,
@@ -202,6 +205,7 @@ class Console:
         self.is_console_controlled = False
 
     def __del__(self):
+        dev_null_handles.pop(self._dev_null)
         self._dev_null.close() 
```

🙃 😉 

[edit]
this does work to stop the errors during test. but it's mostly a facetious proposal.

my assumption is gc is detecting circular refs between the `/dev/null` handle and the `Console` instance....and then freeing them out of order.

freakboy3742:
... 136 files later...

There's a lot of churn here - but it's almost entirely "use a fixture that cleans up after itself, rather than instantiating a Console() instance on the fly". The only place this is complicated is in the Console tests themselves, where we need to differentiate between low level tests of the internals of Console, and tests of services built on top of the base console.

The rest of the PR is fairly standard version update stuff - configuration updates, purges of old syntax and `if sys.version < ` branches, and a handful of miscellaneous tweaks to keep coverage happy.

mhsmith:
> my assumption is gc is detecting circular refs between the `/dev/null` handle and the `Console` instance....and then freeing them out of order.

@rmartin16: How could the /dev/null handle have a reference to the `Console` instance?

rmartin16:
> @rmartin16: How could the /dev/null handle have a reference to the `Console` instance?

Sorry; that's probably not well-worded.

What I mean is...once the `Console` instance has no more references, it is (obviously) available for GC. The only reference to the `/dev/null` handle is within the `Console` instance; therefore, the `/dev/null` handle is also available for GC because its only reference (i.e. the `Console` instance) is available for GC.

In this way, I am theorizing that GC is optimizing this situation and removing the reference from the `Console` instance to the `/dev/null` handle. At that point, this optimization could lead to GC freeing the `/dev/null` handle prior to the `Console` instance (where `__del__()` would run and close the `/dev/null` handle).

I believe my diff reinforces this because when I preserve a reference to the `/dev/null` handle until `Console.__del__()` runs, the error does not happen.

mhsmith:
For future reference, here's the failure message from 21c75ae:
```
    | Traceback (most recent call last):
    |   File "/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/unittest/mock.py", line 2252, in __init__
    |     def __init__(self, name, parent):
    |     
    | ResourceWarning: unclosed file <_io.TextIOWrapper name='/dev/null' mode='w' encoding='utf-8'>
    | 
    | The above exception was the direct cause of the following exception:
    | 
    | Traceback (most recent call last):
    |   File "/Users/runner/work/briefcase/briefcase/.tox/py-cov/lib/python3.14/site-packages/_pytest/unraisableexception.py", line 67, in collect_unraisable
    |     warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
    |     ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | pytest.PytestUnraisableExceptionWarning: Exception ignored while finalizing file <_io.FileIO name='/dev/null' mode='wb' closefd=True>: None
```

This appeared many times, with each ResourceWarning apparently being raised from a different place, most of which had nothing to do with a console. This suggests it was triggered by a spontaneous run of the garbage collector, which should only happen if the file object was part of a reference cycle. However, it's easy to create reference cycles by accident, such as by storing references to exceptions, or any of the other unusual stuff that pytest does to display detailed failure messages.

If Python 3.14 no longer runs the `__del__` method on an object before an object that references it, even when there's no reference cycle between *those* two objects, that seems like a pretty serious bug. But I can't see any better explanation either.

mhsmith (approved):

PR #2434, opened 2025-08-08
Add http retry handler for unit tests

rmartin16:
## Changes
- Introduce more robust retry handler for HTTP requests during unit tests
- It may be reasonable to use such a handler more broadly for calls with httpx

## PR Checklist:

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -6,7 +6,7 @@ import plistlib
 import tarfile
 import zipfile
 from email.message import EmailMessage
-from http import HTTPMethod, HTTPStatus
+from http import HTTPStatus
 from pathlib import Path
 
 import httpx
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -333,17 +335,46 @@ def file_content(path: Path) -> str | bytes | None:
 
 def assert_url_resolvable(url: str):
     """Tests whether a URL is resolvable with retries; raises for failure."""
-    transport = httpx.HTTPTransport(
-        # Retry for connection issues
-        retries=3,
```
rmartin16, line 338: This retry only applies to connection attempts. `httpx` maintainers decided not to introduce more sophisticated retry handling in to their clients....so third party packages (or custom transports) are necessary to fill this void.

Issue #2337, opened 2025-06-12
Delegate ownership of PyScript, Shoelace and Bootstrap to Toga / GUI Toolkit


kavi2du:
### What is the problem or limitation you are having?

Briefcase currently owns the configuration of frontend web dependencies PyScript, Shoelace, and Bootstrap, via hardcoded values in the Briefcase Web Template. This tightly couples Toga (and other Web GUI Toolkits) to Briefcase and the Web Template, creating several limitations:

- GUI toolkits like Toga cannot define or upgrade their own runtime stack without changes in Briefcase/Web Template.  
- PyScript configuration is entangled with Web Template logic, preventing clear version control or modular updates.  
- Toolkit-level concerns (runtime versions, UI styling) are mixed with packaging-level concerns (wheels, app paths), which hinders maintainability and growth of the BeeWare ecosystem.

### Describe the solution you'd like

@caydnn and I propose that we introduce a clear separation of responsibilities between Toga (GUI Toolkit) and Briefcase (Packager) by delegating ownership of runtime dependency configuration and insertion content to Toga/GUI-Toolkit.

Toga and 3rd Party GUI Toolkits would contain a `deployment/` directory containing:

- Base/Template `pyscript.toml` file (similar to the file currently included in the Briefcase Web Template)  
- A `config.toml` file containing a PyScript version for Briefcase configuration  
- `/inserts` directory containing HTML script inserts  
  - In Toga’s case this includes PyScript, Bootstrap and Shoelace headers  
- `/static` directory containing static CSS code  
  - For Toga this will be the current `/static` directory including `toga.css`

During the build process, Briefcase would gather and integrate these resources with `index.html` and other files from the Briefcase Web Template.

### Expected Outcomes

- GUI Toolkits/Toga will have more control over the frontend dependencies  
- Briefcase remains a packager with less focus on defining the internals of a GUI Toolkit  
- Toga maintenance improved through owning dependencies—for example, taking advantage of new features or fixes from dependency updates will now require a single PR to Toga rather than an additional PR to the Briefcase Web Template

Further details for our solution can be found in the comments below.

### Describe alternatives you've considered

- **Considered entirely decoupling PyScript from Briefcase**  
  Not feasible as this will require Toga/GUI Toolkits to package their own PyScript wheels. This would stray from their intended scope.

- **Embedding PyScript version in `pyscript.toml`**  
  Rejected as this is not valid in accordance with PyScript documentation.

[...]

PR #2442, opened 2025-08-11
Support for GUI Toolkit/Toga owned pyscript configurations (Phase 2)

caydnn:
Update Briefcase’s build system to support toolkit-owned pyscript/backend configuration.
Modified platforms/web/static.py to detect and load deploy/config.toml and pyscript.toml as well as validate that only one toolkit provides deployment configuration.
Refs#2337 

## PR Checklist:

### DIFF
--- a/src/briefcase/platforms/web/static.py
+++ b/src/briefcase/platforms/web/static.py
@@ -144,8 +144,7 @@ class StaticWebBuildCommand(StaticWebMixin, BuildCommand):
                     if (
                         len(path.parts) > 1
                         and path.parts[1] == "deploy"
-                        and path.suffix == ".toml"
-                        and path.name == "config"
+                        and path.name == "config.toml"
                     ):
                         self.console.info(f"    Found {filename}")
                         config_counter += 1
--- a/src/briefcase/platforms/web/static.py
+++ b/src/briefcase/platforms/web/static.py
@@ -1,4 +1,5 @@
 import errno
+import shutil
 import subprocess
 import sys
 import webbrowser
@@ -46,8 +47,11 @@ class StaticWebMixin:
     def binary_path(self, app):
         return self.bundle_path(app) / "www/index.html"
 
+    def static_path(self, app):
+        return self.project_path(app) / "static"
+
     def wheel_path(self, app):
-        return self.project_path(app) / "static/wheels"
+        return self.static_path(app) / "wheels"
 
     def distribution_path(self, app):
         return self.dist_path / f"{app.formal_name}-{app.version}.web.zip"
@@ -125,6 +129,93 @@ class StaticWebBuildCommand(StaticWebMixin, BuildCommand):
                     )
                     css_file.write(wheel.read(filename).decode("utf-8"))
 
+    def _gather_backend_config(self, wheels):
+        """Processes multiple wheels to gather a config.toml and a base pyscript.toml file.
+
+        :param wheels: A list of wheel files to be scanned.
+        """
+        config_counter = 0
+        pyscript_config = None
+
+        for wheelfile in wheels:
+            with ZipFile(wheelfile) as wheel:
+                for filename in wheel.namelist():
+                    path = Path(filename)
+                    if (
+                        len(path.parts) > 1
+                        and path.parts[1] == "deploy"
+                        and path.suffix == ".toml"
+                        and path.name == "config"
+                    ):
+                        self.console.info(f"    Found {filename}")
+                        config_counter += 1
+                        # Raise an error if more than one configuration file is supplied.
+                        if config_counter > 1:
+                            raise BriefcaseConfigError(
+                                "Only 1 backend configuration file can be supplied."
+                            )
+                        # Check which backend type is used. Raise error if no backend is present in config.toml
+                        with wheel.open(filename) as config_file:
+                            config_data = tomllib.load(config_file)
+
+                            if "backend" in config_data:
+                                backend = config_data.get("backend")
+
+                                # Currently, only pyscript is supported, will raise an error if another backend is found.
+                                if backend != "pyscript":
+                                    raise BriefcaseConfigError(
+                                        "Only 'pyscript' backend is currently supported for web static builds."
+                                    )
+
+                                pyscript_config = self._gather_backend_config_file(wheel, backend, path)
+
+                            else:
+                                raise BriefcaseConfigError(
+                                    'No backend was provided in config.toml file.'
+                                )
+        # Return a blank pyscript config if no configuration file is found.
+        if (
+            config_counter == 0
+            and pyscript_config is None
+        ):
+            pyscript_config = {}
+
+        return pyscript_config
+
+    def _gather_backend_config_file(self, wheel, backend, path):
+        """Find backend config file (eg: pyscript.toml) from a wheel and save it to project pyscript.toml if found.
+
+        :param wheel: Wheel file to scan for configuration file.
+        :param backend: The backend type as a String (eg "pyscript")
+        :param path: Path to the wheels configuration file (config.toml). This should be in the same directory as the backend configuration file.
+        """
+        backend_counter = 0
+        backend_config = None
+        deploy_dir = path.parent
+
+        for deploy_file in wheel.namelist():
+            deploy_parent = Path(deploy_file).parent
+            if (
+                deploy_parent == deploy_dir
+                and Path(deploy_file).name == f"{backend}.toml"
+            ):
+                backend_counter += 1
+                # Raise an error if more than one pyscript.toml file is found.
+                if backend_counter > 1:
+                    raise BriefcaseConfigError(
+                        "Only 1 pyscript configuration file can be supplied."
+                    )
+                # Save pyscript config file.
+                else:
+                    try:
+                        with wheel.open(deploy_file) as pyscript_file:
+                            backend_config = tomllib.load(pyscript_file)
+                    except tomllib.TOMLDecodeError as e:
+                        raise BriefcaseConfigError(
+                            f"pyscript.toml content isn't valid TOML: {e}"
+                        ) from e
+        return backend_config
+
     def build_app(self, app: AppConfig, **kwargs):
         """Build the static web deployment for the application.
 
@@ -132,33 +223,33 @@ class StaticWebBuildCommand(StaticWebMixin, BuildCommand):
         """
         self.console.info("Building web project...", prefix=app.app_name)
 
-        deploy_path = files("toga_web.deploy")
-        deploy_config_path = deploy_path / "config.toml"
-        deploy_pyscript_path = deploy_path / "pyscript.toml"
+        # deploy_path = files("toga_web.deploy")
+        # deploy_config_path = deploy_path / "config.toml"
+        # deploy_pyscript_path = deploy_path / "pyscript.toml"
 
-        if deploy_config_path.exists():
-            try:
-                with deploy_config_path.open("rb") as f:
-                    deploy_config = tomllib.load(f)
-            except tomllib.TOMLDecodeError as e:
-                raise BriefcaseConfigError(f"Invalid config.toml: {e}") from e
-        else:
-            deploy_config = {}
+        # if deploy_config_path.exists():
+        #     try:
+        #         with deploy_config_path.open("rb") as f:
+        #             deploy_config = tomllib.load(f)
+        #     except tomllib.TOMLDecodeError as e:
+        #         raise BriefcaseConfigError(f"Invalid config.toml: {e}") from e
+        # else:
+        #     deploy_config = {}
 
-        if "backend" in deploy_config and deploy_config["backend"] != "pyscript":
-            raise BriefcaseConfigError(
-                "Only 'pyscript' backend is currently supported for web static builds."
-            )
+        # if "backend" in deploy_config and deploy_config["backend"] != "pyscript":
+        #     raise BriefcaseConfigError(
+        #         "Only 'pyscript' backend is currently supported for web static builds."
+        #     )
 
-        if deploy_pyscript_path.exists():
-            try:
-                extra_pyscript_toml = deploy_pyscript_path.read_text(encoding="utf-8")
-            except Exception as e:
-                raise BriefcaseConfigError(
-                    f"Unable to read deploy/pyscript.toml: {e}"
-                ) from e
-        else:
-            extra_pyscript_toml = ""
+        # if deploy_pyscript_path.exists():
+        #     try:
+        #         extra_pyscript_toml = deploy_pyscript_path.read_text(encoding="utf-8")
+        #     except Exception as e:
+        #         raise BriefcaseConfigError(
+        #             f"Unable to read deploy/pyscript.toml: {e}"
+        #         ) from e
+        # else:
+        #     extra_pyscript_toml = ""
 
         if self.wheel_path(app).exists():
             with self.console.wait_bar("Removing old wheels..."):
@@ -217,15 +308,7 @@ class StaticWebBuildCommand(StaticWebMixin, BuildCommand):
         with self.console.wait_bar("Writing Pyscript configuration file..."):
             # Load any pre-existing pyscript.toml provided by the template. If the file
             # doesn't exist, assume an empty pyscript.toml as a starting point.
-            try:
-                with (self.project_path(app) / "pyscript.toml").open("rb") as f:
-                    config = tomllib.load(f)
-            except tomllib.TOMLDecodeError as e:
-                raise BriefcaseConfigError(
-                    f"pyscript.toml content isn't valid TOML: {e}"
-                ) from e
-            except FileNotFoundError:
-                config = {}
+            config = self._gather_backend_config(self.wheel_path(app).glob("*.whl"))
 
             # Add the packages declaration to the existing pyscript.toml.
             # Ensure that we're using Unix path separators, as the content
@@ -247,17 +330,17 @@ class StaticWebBuildCommand(StaticWebMixin, BuildCommand):
             except AttributeError:
                 pass
 
-            # Parse any deploy pyscript.toml content, and merge it into
-            # the overall content
-            try:
-                extra = tomllib.loads(extra_pyscript_toml)
-                config.update(extra)
-            except tomllib.TOMLDecodeError as e:
-                raise BriefcaseConfigError(
-                    f"Deploy pyscript.toml content isn't valid TOML: {e}"
-                ) from e
-            except AttributeError:
-                pass
+            # # Parse any deploy pyscript.toml content, and merge it into
+            # # the overall content
+            # try:
+            #     extra = tomllib.loads(extra_pyscript_toml)
+            #     config.update(extra)
+            # except tomllib.TOMLDecodeError as e:
+            #     raise BriefcaseConfigError(
+            #         f"Deploy pyscript.toml content isn't valid TOML: {e}"
+            #     ) from e
+            # except AttributeError:
+            #     pass
 
             # Write the final configuration.
             with (self.project_path(app) / "pyscript.toml").open("wb") as f:
--- a/src/briefcase/platforms/web/static.py
+++ b/src/briefcase/platforms/web/static.py
@@ -3,6 +3,7 @@ import subprocess
 import sys
 import webbrowser
 from http.server import SimpleHTTPRequestHandler, ThreadingHTTPServer
+from importlib.resources import files
 from pathlib import Path
 from typing import Any
 from zipfile import ZipFile
@@ -131,6 +132,34 @@ class StaticWebBuildCommand(StaticWebMixin, BuildCommand):
         """
         self.console.info("Building web project...", prefix=app.app_name)
 
+        deploy_path = files("toga_web.deploy")
+        deploy_config_path = deploy_path / "config.toml"
+        deploy_pyscript_path = deploy_path / "pyscript.toml"
+
+        if deploy_config_path.exists():
+            try:
+                with deploy_config_path.open("rb") as f:
+                    deploy_config = tomllib.load(f)
+            except tomllib.TOMLDecodeError as e:
+                raise BriefcaseConfigError(f"Invalid config.toml: {e}") from e
+        else:
+            deploy_config = {}
+
+        if "backend" in deploy_config and deploy_config["backend"] != "pyscript":
+            raise BriefcaseConfigError(
+                "Only 'pyscript' backend is currently supported for web static builds."
+            )
+
+        if deploy_pyscript_path.exists():
+            try:
+                extra_pyscript_toml = deploy_pyscript_path.read_text(encoding="utf-8")
+            except Exception as e:
+                raise BriefcaseConfigError(
+                    f"Unable to read deploy/pyscript.toml: {e}"
+                ) from e
+        else:
+            extra_pyscript_toml = ""
+
         if self.wheel_path(app).exists():
             with self.console.wait_bar("Removing old wheels..."):
                 self.tools.shutil.rmtree(self.wheel_path(app))
@@ -214,6 +243,18 @@ class StaticWebBuildCommand(StaticWebMixin, BuildCommand):
             except tomllib.TOMLDecodeError as e:
                 raise BriefcaseConfigError(
                     f"Extra pyscript.toml content isn't valid TOML: {e}"
+                ) from e
+            except AttributeError:
+                pass
+
+            # Parse any deploy pyscript.toml content, and merge it into
+            # the overall content
+            try:
+                extra = tomllib.loads(extra_pyscript_toml)
+                config.update(extra)
+            except tomllib.TOMLDecodeError as e:
+                raise BriefcaseConfigError(
+                    f"Deploy pyscript.toml content isn't valid TOML: {e}"
                 ) from e
             except AttributeError:
                 pass


caydnn:
Draft work and currently untested. The commented out code is from Kavi's original work that I've tried to repurpose. Kavi will likely merge more work into this PR before we would like to merge it.

kavi2du:
@freakboy3742 Hi Russell, we’ve updated the PR with the phase 3 changes. We’d appreciate it if you could review them before we move forward. Thanks!

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
-                    and path.parts[1] == "static"
-                    and path.suffix == ".css"
-                ):
-                    self.console.info(f"    Found {filename}")
-                    css_file.write(
-                        "\n/*******************************************************\n"
+                if len(path.parts) < 2:
+                    continue
+
+                is_inserts = (path.parts[1] == "inserts") or (
```
freakboy3742: Why handle both `inserts` and `deploy/inserts`? There isn't a legacy consideration here; is there some other use case I'm missing?

I'd also recommend rewriting this check has `path.parts[:2] == ["deploy", "inserts"]`

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+                    len(path.parts) >= 3
+                    and path.parts[1] == "deploy"
+                    and path.parts[2] == "inserts"
+                )
+                if is_inserts and path.name:
```
freakboy3742: What's the case for `path.name` not existing? And why pre-compute `is_inserts` if it's not going to be used later?

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+                        package_key
+                    ] = text
+                    continue
+
+                is_static = (path.parts[1] == "static") or (
+                    len(path.parts) >= 3
+                    and path.parts[1] == "deploy"
+                    and path.parts[2] == "static"
+                )
+                if is_static:
```
freakboy3742: Similar to above - what's the use case for `deploy/static`? It's not a historical usage.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+        """
+        backend_counter = 0
+        backend_config = None
+        deploy_dir = path.parent
+
+        for deploy_file in wheel.namelist():
+            deploy_parent = Path(deploy_file).parent
+            if (
+                deploy_parent == deploy_dir
+                and Path(deploy_file).name == f"{backend}.toml"
```
freakboy3742: While I understand the intention to make this pluggable, there's no reason to believe that any other framework will adopt `framework.toml` as its configuration file. This would be much better handled as a specific `pyscript` configuration loader.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+            ):
+                backend_counter += 1
+                # Raise an error if more than one pyscript.toml file is found.
```
freakboy3742: This isn't quite right - it's going to search all files for *backend.toml*. So if one backend has a `pyscript.toml`, and another has `foobar.toml`, this error won't be triggered. 

The real constraint is the *config* counter (which is implemented above), not a count of pyscript configurations.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+        :param wheel: Wheel file to scan for configuration file.
+        :param backend: The backend type as a String (eg "pyscript")
+        :param path: Path to the wheels configuration file (config.toml). This should be
+            in the same directory as the backend configuration file.
+        """
+        backend_counter = 0
+        backend_config = None
+        deploy_dir = path.parent
+
+        for deploy_file in wheel.namelist():
```
freakboy3742: Why is this iterating over a list of *all* wheels when it's being invoked in a context where you're iterating over all wheels?

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+                        len(path.parts) > 1
+                        and path.parts[1] == "deploy"
+                        and path.name == "config.toml"
+                    ):
+                        self.console.info(f"    Found {filename}")
+                        config_counter += 1
+                        # Raise an error if more than one configuration file is supplied.
+                        if config_counter > 1:
+                            raise BriefcaseConfigError(
+                                "Only 1 backend configuration file can be supplied."
```
freakboy3742: This error could be more helpful, pointing at the specific package that contains the duplicate (and probably the original package where one was found).

src/briefcase/platforms/web/static.py
```diff
@@ -231,14 +491,23 @@ def build_app(self, app: AppConfig, **kwargs):
...
-                    self._process_wheel(wheelfile, css_file=css_file)
+                self._process_wheel(
+                    wheelfile=wheelfile,
+                    inserts=inserts,
+                    static_path=static_root,
+                )
 
-        return {}
+            # Write inserts per target
+            for target in sorted(inserts.keys()):
```
freakboy3742: `for target, insert in inserts.items()` would be more idiomatic.

Also - why does this list need to be sorted? The sorting process takes time; unless there's a reason to sort it, there's no reason to spend that time.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+                if is_inserts and path.name:
+                    source = (
+                        str(Path(*path.parts[2:]))
+                        if path.parts[1] == "inserts"
+                        else str(Path(*path.parts[3:]))
                     )
-                    css_file.write(f" * {package}::{'/'.join(path.parts[2:])}\n")
-                    css_file.write(
-                        " *******************************************************/\n\n"
+                    if ":" not in path.name:
```
freakboy3742: Some inline documentation here would be very helpful for future readers. I can broadly follow what's going on because I know the history; but some explanations along the way of what format things will be in, etc, and what is being done at each step, will aid future debugging.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+                        continue
+                    target, insert = source.split(":", 1)
+                    self.console.info(
+                        f"    {source}: Adding {insert} insert for {target}"
+                    )
+                    try:
+                        text = wheel.read(filename).decode("utf-8")
+                    except UnicodeDecodeError:
+                        self.console.warning(
+                            f"    {source}: non-UTF8 insert; skipping."
```
freakboy3742: Have you actually seen this problem in practice? Otherwise, I think it's safe to say that non-UTF8 templated content can raise an immediate error.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+        target_path = self.project_path(app) / filename
+        if not target_path.exists():
+            self.console.warning(f"  Target {filename} not found; skipping inserts.")
+            return
+
+        with target_path.open("r", encoding="utf-8") as f:
+            file_text = f.read()
+
+        for insert in sorted(inserts.keys()):
+            packages = inserts[insert]
```
freakboy3742: `for insert, packages in inserts.items()` would be more idiomatic.

Also - is there a reason to sort the items?

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+            insert.
+        """
+        # Read the current content
+        target_path = self.project_path(app) / filename
+        if not target_path.exists():
+            self.console.warning(f"  Target {filename} not found; skipping inserts.")
+            return
+
+        with target_path.open("r", encoding="utf-8") as f:
+            file_text = f.read()
```
```suggestion
        target_path.read_text(encoding="utf-8")
```

freakboy3742: You can also combine this statement with the "if exists" check - try to read the content, and catch the `FileNotFound` exception. That's a general principle in Python of "it's better to ask forgiveness than permission".

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+            ]
+
+            html_body = "\n".join(
+                html_banner.format(package=pkg, content=packages[pkg])
+                for pkg in sorted(packages.keys())
+            )
+            css_body = "\n".join(
+                css_banner.format(package=pkg, content=packages[pkg])
+                for pkg in sorted(packages.keys())
+            )
```
freakboy3742: This doesn't seem right - it's using the same content to render both HTML and CSS...

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+
+            replaced = False
+            for pattern_tmpl, repl_tmpl, kind in marker_styles:
+                pattern = re.compile(
+                    pattern_tmpl.format(nsert=insert), flags=re.MULTILINE | re.DOTALL
```
freakboy3742: Well that can't work...
```suggestion
                    pattern_tmpl.format(insert=insert), flags=re.MULTILINE | re.DOTALL
```

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
             for line in content:
                 f.write(line)
 
-    def _process_wheel(self, wheelfile, css_file):
+    def _merge_insert_content(self, inserts, key, path):
```
freakboy3742: Is this method actually used?

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +102,295 @@ def _trim_file(self, path, sentinel):
...
+                    pattern_tmpl.format(nsert=insert), flags=re.MULTILINE | re.DOTALL
+                )
+                if pattern.search(file_text):
+                    body = html_body if kind == "html" else css_body
+                    file_text = pattern.sub(
+                        repl_tmpl.format(insert=insert, content=body),
+                        file_text,
+                    )
+                    replaced = True
+                    break
```
freakboy3742: I'm *very* wary of breaks inside deeply nested for loops - if you want to know why, read up on the causes of the 1990 Martin Luther King Jr Day crash of the US phone system.

To me, either `if not replaced` checks, or restructuring this loop so it's a `while not replaced` would make more sense.

freakboy3742 (changes requested):
This is headed in the right direction, but there's a bunch of simplifications and idiomatic code usage that can be cleaned up. 

I'm also not 100% certain the code *can* work as written - there's at least 2 areas I've flagged that, if they work for you, are likely accidental at best.

Also - I can see you've tried to preserve at least of the historical "static" handling; that's good, but (a) we only need to retain legacy functionality in the form it historically supported - we don't need to add hybrid handling of legacy and new layout; and (b) we need to retain *all* the functionality, including the banner/separator content.

src/briefcase/platforms/web/static.py
caydnn: that's understandable, this wasn't very thought out. Will leave this out for now and leave it for when other backends are supported.

src/briefcase/platforms/web/static.py
caydnn: Sorry Russell, I'm a bit confused with this one. This function is given the wheel in which the previous function found config.toml as the existence of this file indicates pyscript.toml (or another backend configuration file at this stage) will also be present within the wheel. This line here is iterating over a single wheel file to my understanding.
Maybe there is a more efficient and less confusing way to do this

src/briefcase/platforms/web/static.py
kavi2du: The sorting was added here to make sure the inserts are always written in the same order, regardless of how the system happens to return the items which would make the output consistent. It’s not strictly necessary, so I’ll remove it to avoid the extra work.

src/briefcase/platforms/web/static.py
kavi2du: Sure thing, I'll add more inline comments to explain the formats and the steps.

src/briefcase/platforms/web/static.py
freakboy3742: If it's to preserve consistent output order, that's a good reason to sort. I hadn't caught that this was the reason.

src/briefcase/platforms/web/static.py
freakboy3742: Problem exists between my keyboard and chair :-)

I think I misread this as iterating over a list of wheels, not iterating over the list of files in one specific wheel.

src/briefcase/platforms/web/static.py
kavi2du: This method was brought over from the old PR#1285 to handle and merge multiple CSS files into one insert block, but it was overlooked accidentally when wiring it to the rest of the code. We’ve left it commented out for now until we get the main wheels up and running...

kavi2du:
Hi @freakboy3742, the updates are in place now, and the changes are ready for your review.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,251 @@ def _trim_file(self, path, sentinel):
             for line in content:
                 f.write(line)
 
-    def _process_wheel(self, wheelfile, css_file):
+    def _write_inserts(self, app: AppConfig, filename: Path, inserts: dict):
```
freakboy3742: As a type annotation, `dict` should be qualified by what it is a dictionary *of* - in this case, I think it's `dict[str, dict[str, str]]`?

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,251 @@ def _trim_file(self, path, sentinel):
...
+        target_path = self.project_path(app) / filename
+        try:
+            file_text = target_path.read_text(encoding="utf-8")
+        except FileNotFoundError:
+            self.console.warning(f"  Target {filename} not found; skipping inserts.")
+            return
+
+        # Each insert slot and its package contributions are processed in sorted order
+        for insert, packages in sorted(inserts.items()):
+            packages = inserts[insert]
```
freakboy3742: This line isn't required if you're iterating over `items()`.

src/briefcase/platforms/web/static.py
freakboy3742: This makes sense now; but organisationally, this should come before the marker bodies. As is, the "body" handling is split in half.

This would also allow using the computed `body` directly, rather than needing to use strings like `html` as a proxy for "selecting" the right body.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,251 @@ def _trim_file(self, path, sentinel):
...
+            css_body = "\n".join(
+                css_banner.format(package=pkg, content=pkg_content.get("css", ""))
+                for pkg, pkg_content in sorted(packages.items())
+                if pkg_content.get("css")
+            )
+
+            # Find first matching marker in file
+            matched_style = None
+            for pattern_tmpl, repl_tmpl, kind in marker_styles:
+                pattern = re.compile(
```
freakboy3742: Since this pattern doesn't vary with anything in the loop, it can be defined outside the loop (which will be faster - regex compilation is a mildly expensive operation) and then reused on each iteration.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,251 @@ def _trim_file(self, path, sentinel):
...
+                    pattern_tmpl.format(insert=insert),
+                    flags=re.MULTILINE | re.DOTALL,
+                )
+                if matched_style is None and pattern.search(file_text):
+                    matched_style = (pattern, repl_tmpl, kind)
+
+            if matched_style is not None:
```
freakboy3742: Why is this being done as a separate if block? Isn't this the behavior that should be done if a match is found?

Also - a page *could* contain HTML *and* CSS inserts on the same page, as it would be legal to have a `<style>` element on a HTML page. I'm not sure we can use the "bail on first insert found" approach here.

Also

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,251 @@ def _trim_file(self, path, sentinel):
...
+                    try:
+                        text = wheel.read(filename).decode("utf-8")
+                    except UnicodeDecodeError as e:
+                        raise BriefcaseCommandError(
+                            f"{source}: insert must be UTF-8 encoded"
+                        ) from e
+
+                    # Classify insert as HTML vs CSS/JS
+                    t_ext = Path(target).suffix.lower()
+                    kind = "css" if t_ext in {".css", ".js"} else "html"
```
freakboy3742: Is this type of content separation necessary? As noted earlier, it would be legal to have CSS, JS and HTML inserts all on the same `.html` content. Determining filetype by extension is also fragile as an approach.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,251 @@ def _trim_file(self, path, sentinel):
...
+                if parts[:2] == ("deploy", "inserts"):
+                    source = str(Path(*parts[2:]))
+
+                    try:
+                        target, insert = source.split(":", 1)
+                    except ValueError:
+                        self.console.warning(
+                            f"    {source}: missing ':<insert>'; skipping insert."
+                        )
+                        continue
```
freakboy3742: I don't mind having "top of loop shortcut" continue statements; but once you're in body logic, you should be working out how to use if statements. `continue` is a GOTO wearing lipstick, and has all the same problems.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,251 @@ def _trim_file(self, path, sentinel):
...
+
+        for wheelfile in wheels:
+            with ZipFile(wheelfile) as wheel:
+                for filename in wheel.namelist():
+                    path = Path(filename)
+                    if (
+                        len(path.parts) > 1
+                        and path.parts[1] == "deploy"
+                        and path.name == "config.toml"
+                    ):
```
freakboy3742, line 385: This match seems a little loose... `deploy/foobar/config.toml` will match.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,251 @@ def _trim_file(self, path, sentinel):
...
+
+                        # Check which backend type is used. Raise error if no backend is present in config.toml
+                        with wheel.open(filename) as config_file:
+                            config_data = tomllib.load(config_file)
+
+                            if "backend" in config_data:
+                                backend = config_data.get("backend")
+
+                                # Currently, only pyscript is supported, will raise an error if another backend is found.
+                                if backend != "pyscript":
```
freakboy3742: The logic here is all valid AFAICT, but it's getting a bit deep in nesting. 

Rather than process the package when it is found, It might be worth organising this so that you: 
1. Find all packages that have config files, accumulating a list.
2. If the accumulated list has >1 entry, raise an error (reporting all the packages that have config files)
3. If the accumulated list has 1 entry, processing the config in that package
4. If the accumulated list has no entries, perform default handling.

src/briefcase/platforms/web/static.py
```diff
@@ -188,15 +410,7 @@ def build_app(self, app: AppConfig, **kwargs):
...
-            try:
-                with (self.project_path(app) / "pyscript.toml").open("rb") as f:
-                    config = tomllib.load(f)
-            except tomllib.TOMLDecodeError as e:
-                raise BriefcaseConfigError(
-                    f"pyscript.toml content isn't valid TOML: {e}"
-                ) from e
-            except FileNotFoundError:
-                config = {}
+            config = self._gather_backend_config(self.wheel_path(app).glob("*.whl"))
```
freakboy3742: This is bike shedding, but I'd go with `extract_backend_config`; and it doesn't need to be a private method.

src/briefcase/platforms/web/static.py
freakboy3742: Also - this doesn't need to be a private method.

freakboy3742 (changes requested):
A few comments inline, but this is looking good. Most of the comments are about high level structural things where you've got the right logic, but there are ways to reorder the code so that it requires less nesting (and/or less confusing error paths).

The only other issue is whether we actually need to do "content type" separation on inserts. It's entirely legal for a single `.html` page to contain CSS, JS and HTML inserts (or for a JS page to potentially include HTML content); so the substitution should be entirely based on the insert key name and full filename, not qualified by content type.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,251 @@ def _trim_file(self, path, sentinel):
...
+                        .setdefault(package_key, {"html": "", "css": ""})
                     )
-                    css_file.write(wheel.read(filename).decode("utf-8"))
+                    # Append into the right bucket
+                    if pkg_entry[kind]:
+                        pkg_entry[kind] += "\n"
+                    pkg_entry[kind] += text
+                    continue
+
+                # Handle static files under deploy/static
```
freakboy3742: This static handling is different to the historical handling. I flagged this last time, but if we're going to preserve support for `static`, we need to preserve the historical handling. There's no point preserving historical syntax if we're not going to preserve historical behavior.

caydnn:
Ok, it seems the insertion issue is fixed. @kavi2du pointed out that the logic for finding the deploy folder and the logic for finding insertion files was incorrect. This has now been fixed, shoelace elements now display and work in the browser.

We still have to address the matter of handling legacy static files so no need for a review just yet.

src/briefcase/platforms/web/static.py
kavi2du: Hi @freakboy3742, I believe you’re referring to the new deploy/static/ directory structure introduced here. That actually came from our proposed solutions, to move the static folder inside the deploy directory. At the moment, the static folder lives inside the toga_web directory. In the old Toga PR [#1945](https://github.com/beeware/toga/pull/1945), both the index.html and toga.css files were located in the inserts folder.

Caydn and I discussed this and agreed to remove the static folder entirely, placing both index.html and toga.css in the deploy/inserts directory, as in PR1945 and preserve that directory structure and legacy functionality including the old CSS banners. I just wanted to confirm if this approach would address the issue before I push the new changes.

freakboy3742: I evidently misread the significance of that `/static` handling part of your proposal, so apologies for the confusion on this.

There's three options on the table when it comes to `/static` handling:

1. Preserving the historical functionality of the /static directory
2. Preserving the old functionality, but requiring the content to be in a new location
3. Completely removing the old functionality

(1) is the backwards compatible approach. (3) is the backwards incompatible approach, where we tell users they need to transition to inserts. (2) is what you've proposed - something that isn't compatible for old users, and adds new functionality that doesn't really have much utility when the option for using inserts exists.

Based purely on functionality, my inclination would be to just move to (3). toga-web is sufficiently immature that I'm comfortable introducing a backwards incompatibility, and toga-web will the only user of this functionality in practice. 

However, the issue is the transition as we release this set of changes. If we land this code in Briefcase, we won't be able to deploy a published version of toga-web; if we land an updated toga-web PR, we don't be able to use a published version of Briefcase to deploy it. We'd essentially be in a position where there's a need to tightly coordinate releases of 2 projects, and ensure all users update both simultaneously - which isn't something we can guarantee.

So - we need to do (1). 

By way of implementation: my suggestion is to treat the backwards compatibility path *as inserts* - but inserts with fixed names/behaviors.

So - if a wheel provides CSS content in a `/static` folder, that content is read, and treated as an insert into the default CSS file in the template - but if static content is read in this way, a warning is raised that it's legacy handling that will be removed in the future.

Similarly, if the project configuration defines `style_framework`, that triggers the addition of specific inserts that match what the template currently includes automatically, with a warning to the user that the feature will go away in the future.

Does that make sense?

caydnn: Hi Russell, thank you for the detailed explanation. I understand and agree with you here. :)
We will return the functionality to handle static files under `toga_web/static/` or similar.

In the event that the version of toga-web does not supply shoelace inserts and uses static files but the version of the Web Template has insert markers and no hard-coded pyscript version or shoelace declarations, would we need functionality to account for this? Currently, the pyscript version should default to 2024.11.1 in the case it is not supplied but I'm not particularly sure of shoelace.

freakboy3742: We essentially need to handle 2 cases:

1. A legacy layout using `/static`, with no special inserts
2. A "modern" layout using the deploy folder, with all the inserts.

There won't have the case of a package that has both; and we only need to maintain the legacy layout for transition purposes. I'd imagine that when this PR lands, we'll open a PR to remind us to remove the legacy support in 6-12 months.

So - if it's a legacy layout, then we can hardcode the behavior of the current template into Briefcase - that is, the specific Shoelace (or Bootstrap) and PyScript versions that are currently defined in the template can be put into Briefcase.

kavi2du:
Hi @freakboy3742 , the updates for static handling have been pushed, and I believe the changes are now ready for your review.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,365 @@ def _trim_file(self, path, sentinel):
             for line in content:
                 f.write(line)
 
-    def _process_wheel(self, wheelfile, css_file):
-        """Process a wheel, extracting any content that needs to be compiled into the
-        final project.
+    def write_inserts(
+        self, app: AppConfig, filename: Path, inserts: dict[str, dict[str, str]]
```
freakboy3742: As a style thing - if the argument list is long enough that it needs to be split onto a separate line like this, we generally go straight to the "one argument per line" format (i.e., what black does automatically if you add a comma to the end of the argument list).

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,365 @@ def _trim_file(self, path, sentinel):
...
+                ),
+                # CSS/JS
+                (
+                    r"/\*@@ {insert}:start @@\*/.*?/\*@@ {insert}:end @@\*/",
+                    r"/*@@ {insert}:start @@*/\n{content}/*@@ {insert}:end @@*/",
+                    "css",
+                ),
+            ]
+
+            # Pre-compile patterns once per insert
```
freakboy3742: Any reason to not do the compilation at time of definition? Why construct a list, and then process that list?

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,365 @@ def _trim_file(self, path, sentinel):
...
+            if not any_match:
+                self.console.warning(
+                    f"  Slot '{insert}' markers not found in {filename}; skipping."
+                )
+
+        # Save modified content
+        target_path.write_text(file_text, encoding="utf-8")
+
+    def write_pyscript_version(
+        self, app: AppConfig, filename: Path, pyscript_version: str
```
freakboy3742: As above - we prefer the one-arg-per-line format.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,365 @@ def _trim_file(self, path, sentinel):
...
+    ):
+        """Write pyscript version into an existing html file.
+
+        This function looks for markers in the named file and replaces the
+        markers with the pyscript version.
+
+        The markers are in pyscript declarations within the html.
+
+        * Marker: ``<!--@@ pyscript_version @@-->``
```
freakboy3742: Won't this be a destructive operation? If I update toga-web, the `header.index.html` insert will be replaced; if I bump the PyScript version, the marker will have been removed from the file, so the second pass won't update it.

Why not handle the PyScript content as a full header inclusion (and potentially a script inclusion as well) defined by Briefcase, rather than Toga (or the contributing package?

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,365 @@ def _trim_file(self, path, sentinel):
...
+            file_text = file_text.replace(marker, pyscript_version)
+        else:
+            raise BriefcaseConfigError(
+                f"No pyscript markers found in {filename}; pyscript may not be configured correctly."
+            )
+
+        target_path.write_text(file_text, encoding="utf-8")
+
+    def _process_wheel(
+        self, wheelfile, inserts: dict[str, dict[str, dict[str, str]]], static_path
```
freakboy3742: As above - one line per argument for long lists.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,365 @@ def _trim_file(self, path, sentinel):
...
-                        "\n/*******************************************************\n"
-                    )
-                    css_file.write(f" * {package}::{'/'.join(path.parts[2:])}\n")
-                    css_file.write(
-                        " *******************************************************/\n\n"
-                    )
-                    css_file.write(wheel.read(filename).decode("utf-8"))
+                parts = path.parts
+
+                if not (filename.endswith("/") or len(parts) < 2):
```
freakboy3742: Is this extra level of if check needed? Won't the subsequent "legacy CSS" and "deploy insert" if branches do the same check?

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,365 @@ def _trim_file(self, path, sentinel):
...
+                            raise BriefcaseCommandError(
+                                f"{filename}: CSS content must be UTF-8 encoded"
+                            ) from e
+
+                        # Wrap CSS with a source banner showing package and file
+                        rel_inside = "/".join(path.parts[2:])
+                        css_payload = (
+                            "\n/*******************************************************\n"
+                            f" * {package_key}::{rel_inside}\n"
+                            " *******************************************************/\n\n"
```
freakboy3742: Doesn't this header reproduce what the insertion logic does? Instead of having 2 headers, could we bend the content of `package` so that it isn't "just" the package, but is something like `package (legacy static CSS content)`? That should also avoid the key collision problem that you've got protection against on L312-316.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,365 @@ def _trim_file(self, path, sentinel):
...
+                                except UnicodeDecodeError as e:
+                                    raise BriefcaseCommandError(
+                                        f"{filename}: insert must be UTF-8 encoded"
+                                    ) from e
+
+                                # Store insert under the correct target and slot
+                                pkg_map = inserts.setdefault(target, {}).setdefault(
+                                    insert, {}
+                                )
+                                # Append if package already contributed to this slot
```
freakboy3742: How would this manifest? Wouldn't it require two files with the same filename? If the issue is having two `index.html` files in different locations in the `deploy/inserts` tree - can't that problem be avoided by using the full path name, rather than just the filename? (and shouldn't it be that anyway, to ensure that the insert is applied to the right file?)

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,365 @@ def _trim_file(self, path, sentinel):
...
+                                else:
+                                    pkg_map[package_key] = text
+
+                            except ValueError:
+                                self.console.debug(
+                                    f"    {filename}: not an <insert>.<target> name; skipping generic insert handling."
+                                )
+
+                        # CSS inserts
+                        if basename.endswith(".css"):
```
freakboy3742: Why is CSS handling different here? Other than the extension, Isn't CSS Just Another File?

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +100,365 @@ def _trim_file(self, path, sentinel):
...
+        # Gather a backend configuration file from the package.
+        # For now, is a pyscript.toml as no other backend is currently supported.
+        else:
+            with ZipFile(config_package_list[0]) as wheel:
+                # Check which backend type is used.
+                with wheel.open(config_filename) as config_file:
+                    config_data = tomllib.load(config_file)
+
+                    if "backend" in config_data:
+                        backend = config_data.get("backend")
```
freakboy3742: This reads as a `try:except:` - or, at least, a series of "if bad, exception; if bad for other reason; exception" before doing the actual processing of the "good" case.

freakboy3742 (changes requested):
A couple of style notes and suggestions for ways to tighten things up; but this is definitely on the right track.

src/briefcase/platforms/web/static.py
caydnn: Oh that is a good point and one that I didnt consider. I think we'll go with your suggestion :)

src/briefcase/platforms/web/static.py
kavi2du: Sorry about the earlier comment, I had it confused with something else : p. I’ll update the styling to the one-argument-per-line format.

src/briefcase/platforms/web/static.py
caydnn: Hey Russ, Tox appears to collapse this and write_inserts into a single line

kavi2du: Yeah, it was doing the same earlier as well. I used `git commit --no-verify` earlier just to bypass this, once all the other changes in the commit were finalised.

src/briefcase/platforms/web/static.py
kavi2du: The compilation is handled this way because the regex includes the insert slot name (`{insert}`), which is only known at runtime. If we compiled at definition time, we would either need a generic catch all regex (IMO, will introduce a risk of false matches) or still rebuild it later. That's why we went with compiling per slot as it keeps patterns precise, and safe.

freakboy3742: I'm not saying the entire regex should be at the module level - I'm saying you can create the list once, rather than creating a list and then converting it into a second list that has *exactly* the same content, but compiled.
```
            compiled_markers = [
                # HTML
                (
                    re.compile(rf"<!--@@ {insert}:start @@-->.*?<!--@@ {insert}:end @@-->", flags=re.MULTILINE | re.DOTALL),
                    r"<!--@@ {insert}:start @@-->\n{content}<!--@@ {insert}:end @@-->",
                    "html",
                ),
                # CSS/JS
                (
                    re.compile(rf"/\*@@ {insert}:start @@\*/.*?/\*@@ {insert}:end @@\*/", flags=re.MULTILINE | re.DOTALL),
                    r"/*@@ {insert}:start @@*/\n{content}/*@@ {insert}:end @@*/",
                    "css",
                ),
            ]
```

src/briefcase/platforms/web/static.py
freakboy3742: The trick is to add a comma to the end of the line. This tells black/ruff that you want "one line per item" syntax.

src/briefcase/platforms/web/static.py
caydnn: After trying to change this to a `try:except`, I changed this to a quick fail if "backend" was not present in config_data.
Not really sure if this is what you wanted but I think this has made it a bit easier to trace the code

freakboy3742: I don't mind whether the "try/except" and "if/error" approach is used; the primary motivation here is trying to avoid deep nesting. This is one of those places where ruff/black's line length is a proxy for an underlying problem. Every time you indent another level, you lose another 4 characters of line length - and eventually, it's near impossible to say anything non-trivial on a single line of code.

In this case - the body of "if backend in config_data" is "the rest of the method"; but "if backend != pyscript" raises immediately. If you rework this as an initial `backend = config_data.get("backend", None)` (which will return None if the key doesn't exist, you can de-indent most of this block of code by one level, and the error handling becomes a more consistent (and less nested) `if None; if != pyscript; if ...` flow.

caydnn: Thank you for the explanation Russell, I will be keeping this in mind going forward :)
My previous change de-indented one level, but your method is much easier to read and understand and so I have switched to that.

kavi2du:
Hi @freakboy3742, we believe we’ve addressed all the review comments and would appreciate another review before we push the testing changes.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +101,368 @@ def _trim_file(self, path, sentinel):
...
+                " * {package}\n"
+                " -------------------------------------------------->\n"
+                "{content}"
+            )
+            css_banner = (
+                "/**************************************************\n"
+                " * {package}\n"
+                " *************************************************/\n"
+                "{content}"
+            )
```
freakboy3742: It's just occurred to me that these are constants - so we don't need to redefine them on every loop. We can pull them out as module level `HTML_BANNER` and `CSS_BANNER` values.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +101,368 @@ def _trim_file(self, path, sentinel):
...
+        if marker_found:
+            indented_content = indent(
+                dedent(content),
+                marker_found.group(1),
+            )
+            file_text = re.sub(
+                marker, indented_content, file_text, flags=re.DOTALL | re.MULTILINE
+            )
+        # Warning if no markers were found.
+        # NOTE: this is likely due to using an older version of the Web Template.
```
freakboy3742: Is this possible? The version of the template that is used will be linked to the Briefcase version.

On top of that, one of the use cases of a custom template is "I know what I'm doing and I'm hard-coding everything", so there's no guarantee that version markers will exist. A warning is definitely worth it - but a detailed hint isn't really called for IMHO.

If there's concern about ensuring that the Briefcase template has been updated to match the Briefcase version (and I agree that there probably with this set of changes), Briefcase has the concept of a template "epoch" - a version marker in the template that says "this template has been updated for version X". See [the iOS template](https://github.com/beeware/briefcase-iOS-Xcode-template/blob/main/%7B%7B%20cookiecutter.format%20%7D%7D/briefcase.toml#L4) for an example - if you set one of those markers, an old template without this marker won't work with the next Briefcase release.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +101,368 @@ def _trim_file(self, path, sentinel):
...
+
+                    if not rel_inside or rel_inside.endswith("/"):
+                        self.console.debug(
+                            f"    {filename}: skipping, not a valid insert file."
+                        )
+                    else:
+                        try:
+                            dot_idx = rel_inside.index("~")
+                            target = rel_inside[:dot_idx]
+                            insert = rel_inside[dot_idx + 1 :]
```
freakboy3742: Python has a cleaner way to express this sort of operation:
```suggestion
                            target, insert = rel_inside.split("~")
```
(although you may want to also add a `"~" in rel_inside` check (or equivalent) to make sure that badly named inserts won't break things.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +101,368 @@ def _trim_file(self, path, sentinel):
...
+        marker_found = re.search(marker, file_text, flags=re.DOTALL | re.MULTILINE)
+
+        if marker_found:
+            indented_content = indent(
+                dedent(content),
+                marker_found.group(1),
+            )
+            file_text = re.sub(
+                marker, indented_content, file_text, flags=re.DOTALL | re.MULTILINE
+            )
```
freakboy3742: Unless I'm mistaken, this is the exact logic that is done on line ~188-204 for every other insert... so why treat the PyScript insert differently? Why isn't it similar to "legacy" static handling - something where we're in a position provide the insert manually, but it's otherwise something that is inserted in the same way as everything else?

If we play our cards right, we could even get into a situation where Briefcase will insert a PyScript insert by default, but an end user *could* manually define an insert in their wheel that would use the same name and thus override the Briefcase default insert.

freakboy3742: That would also give us a reason to use a name other than "PyScript" for the insert - while I agree that this insert *will* be the PyScript insertion, and we don't have any immediate plans to support any other Python web distro, there's no reason we couldn't set up the naming so that, in future, another web Python distro *could* insert this block - so, using `Python:start` (or similar - I'm open to other suggestions) might be worthwhile.

freakboy3742 (changes requested):
This is looking really good. A couple of fairly cosmetic things, and what I suspect is one more big piece of refactoring around simplifying the PyScript insertion (unless I'm missing something significant).

src/briefcase/platforms/web/static.py
caydnn: This is a good point. I did initially have this as a way for briefcase to provide the default pyscript insert but yes, the logic is the same as `write_inserts`. For now, I'll refactor this into appending an insert into the inserts list and look into the user override.
Are suggesting that the end user would provide the override via their projects `pyproject.toml` and/or `extra_pyscript_toml_content`?

freakboy3742: I wasn't thinking about end-user extension here; more pointing out that although PyScript is the only intended target, hard-coding PyScript into any naming paints us into a corner. If someone wanted to add a "Pyodide" backend, it would be a little weird to configure it using a `PyScript` block.

That said - if you can see a clean path for a `pyproject.toml` entry to be used to provide override configuration, then that sounds like a good addition. I wouldn't consider that a hard requirement for v1 of this feature, though.

src/briefcase/platforms/web/static.py
kavi2du: Thanks for pointing this out, Russell. I’ve moved the banner templates to module level constants (just under the imports) and updated `write_inserts` to reference them.

src/briefcase/platforms/web/static.py
kavi2du: Thanks for the suggestion! I’ve updated this to use `rel_inside.split("~")` and added a guard to handle cases where `"~"` is missing or occurs multiple times.

kavi2du:
Hi @freakboy3742, as we discussed earlier in the meeting, this PR is ready for another review. We've addressed all the comments from the last round, and there are also a few test case updates included for your review.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,333 @@ def _trim_file(self, path, sentinel):
...
+
+            if not any_match:
+                self.console.warning(
+                    f"  Slot '{insert}' markers not found in {filename}; skipping."
+                )
+
+        # Save modified content
+        target_path.write_text(file_text, encoding="utf-8")
+
+    def _write_pyscript_insert(
```
freakboy3742: Minor housekeeping issue - this is no longer writing anything, so the method name and docstring is misleading.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,333 @@ def _trim_file(self, path, sentinel):
...
+        """
+        name_parts = wheelfile.name.split("-")
+        package_key = f"{name_parts[0]} {name_parts[1]}"
+
+        # Warning flag for legacy CSS
+        legacy_css_warning = False
+
         with ZipFile(wheelfile) as wheel:
-            for filename in wheel.namelist():
+            for filename in sorted(wheel.namelist()):
```
freakboy3742, line 347: From a code organization (and testing) perspective, this method is getting very deeply nested. It would be worth refactoring it - the `handle_legacy_css()` and `handle_insert()` would be two logical places to break up this long loop into smaller (less indented) pieces.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,333 @@ def _trim_file(self, path, sentinel):
...
+                            self.console.debug(
+                                f"    {filename}: skipping, filename must match '<target>~<insert>'."
+                            )
+                        else:
+                            try:
+                                target, insert = rel_inside.split("~")
+                            except ValueError:
+                                self.console.debug(
+                                    f"    {filename}: skipping, filename must match '<target>~<insert>'."
+                                )
```
freakboy3742: These two checks are redundant - if `~` is in rel_inside, you *must* be able to split on `~`. 

If the concern is the case of having more than one `~` - you can pass in the number of splits to allow (`.split("~", 1)`, or `.rsplit("~", 1)` if you want to preserve the "last" ~ as the split character.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,333 @@ def _trim_file(self, path, sentinel):
...
+                            else:
+                                self.console.info(
+                                    f"    {filename}: Adding {insert} insert for {target}"
+                                )
+                                try:
+                                    text = wheel.read(filename).decode("utf-8")
+                                except UnicodeDecodeError as e:
+                                    raise BriefcaseCommandError(
+                                        f"{filename}: insert must be UTF-8 encoded"
+                                    ) from e
```
freakboy3742: Is this a realistic concern? This is a case where yes, sure, we can catch the error... but it's so unlikely to occur that the error handling just makes things harder to read, and in the case it's a problem, the error that propagates will point at the problem pretty clearly.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,333 @@ def _trim_file(self, path, sentinel):
...
+                            "treating as insert into briefcase.css; this legacy handling will be removed in the future."
+                        )
+                        legacy_css_warning = True
+
+                    try:
+                        css_text = wheel.read(filename).decode("utf-8")
+                    except UnicodeDecodeError as e:
+                        raise BriefcaseCommandError(
+                            f"{filename}: CSS content must be UTF-8 encoded"
+                        ) from e
```
freakboy3742: As with the other handling - is this a realistic concern?

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,333 @@ def _trim_file(self, path, sentinel):
...
+                        config_package_list.append(wheelfile)
+                        config_filename = filename
+
+        # Return a blank pyscript config if no configuration file is found.
+        if len(config_package_list) == 0:
+            pyscript_config = {}
+        # Raise an error if more than one configuration file is supplied.
+        elif len(config_package_list) > 1:
+            raise BriefcaseConfigError(
+                f"""Only 1 backend configuration file can be supplied.
```
freakboy3742: AP style guide says `1` is always written as the word.
```suggestion
                f"""Only one backend configuration file can be supplied.
```

tests/platforms/web/static/conftest.py
```diff
@@ -16,27 +16,39 @@ def first_app_generated(first_app_config, tmp_path):
...
-existing-key-1 = "value-1"
-existing-key-2 = 2
+        bundle_path / "www/index.html",
+        """<!doctype html>
+<html>
+  <head>
+    <!--@@ head:start @@-->
+    <!--@@ head:end @@-->
+    <!--@@ Python:start @@-->
+    <!--@@ Python:end @@-->
```
freakboy3742: Minor consistency note - this should probably be `python`, so that we can predict insert title case.

tests/platforms/web/static/conftest.py
```diff
@@ -16,27 +16,39 @@ def first_app_generated(first_app_config, tmp_path):
...
         bundle_path / "www/static/css/briefcase.css",
         """
 #pyconsole {
   display: None;
 }
 /*******************************************************************
- ******************** Wheel contributed styles ********************/
+******************** Wheel contributed styles ********************/
+/*@@ CSS:start @@*/
+/*@@ CSS:end @@*/
```
freakboy3742: Similarly - `css` to preserve a consistent insert

tests/platforms/web/static/test_build.py
```diff
@@ -168,27 +224,40 @@ def mock_run(*args, **kwargs):
...
-                "wiggle = false",
-                # A custom array of tables
-                "[[runtimes]]",
-                'src = "https://example.com/pyodide.js"',
-            ]
-        )
-        + "\n"
-    )
+    # Mock some wheels without a config.toml
+    def mock_run(*args, **kwargs):
```
freakboy3742: This looks like it has been significantly reproduced in multiple tests; it looks like there's an opportunity to refactor so there's a common utility method.

src/briefcase/platforms/web/static.py
kavi2du: Ah yes, I’ll keep the explicit UTF-8 decode, but drop the try/except block to avoid extra noise.

src/briefcase/platforms/web/static.py
kavi2du: Thanks for the suggestion! I’ve now split the loop into two helpers, `_handle_legacy_css()` for `/static/*.css` and `_handle_insert()` for `deploy/inserts/<target>~<insert>` files. The main `_process_wheel()` is now just a thin dispatcher. It iterates over sorted filenames, delegates to the right helper, and logs unsupported entries, flattening the nesting.

caydnn:
Hi @freakboy3742, I believe this is pretty much done along with our other PRs (https://github.com/beeware/toga/pull/3666 and https://github.com/beeware/briefcase-web-static-template/pull/21). Whenever you have a chance, we would like to have another review :)

Feel free to try out the features using the method outlined in our issue post (https://github.com/beeware/briefcase/issues/2337)

caydnn:
Also, I believe the failing tests for 3.14 might be false negatives as I have definitely included those lines in `test_build_extract_backend_config.py::test_extract_backend_config`. If I am wrong, please let me know :)

freakboy3742:
> Whenever you have a chance, we would like to have another review :)

ACK - I'm currently snowed under with a work deadline; hopefully I'll get a chance to look at this in the next day or two.

> Also, I believe the failing tests for 3.14 might be false negatives as I have definitely included those lines in `test_build_extract_backend_config.py::test_extract_backend_config`. If I am wrong, please let me know :)

Hrm - that's definitely an odd one. It's possibly a bug in the coverage tool itself; as part of my review, I'll see if I can work out what is going on.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,354 @@ def _trim_file(self, path, sentinel):
...
+        :param path: Path object of the file inside the wheel.
+        :param filename: Filename string inside the wheel.
+        :param package_key: Provenance label (e.g. "name version").
+        :param inserts: Nested dict of inserts keyed by target - insert - package.
+        :param legacy_css_warning: Whether the warning has already been shown.
+        :return: Updated legacy_css_warning flag.
+        """
+        self.console.info(f"    Found {filename}")
+
+        # Show deprecation warning once per wheel
```
freakboy3742: If we're warning, we might as well warn for every usage - with the added bonus that this is a lot simpler regardless.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,354 @@ def _trim_file(self, path, sentinel):
...
+        filename: str,
+        package_key: str,
+        inserts: dict[str, dict[str, dict[str, str]]],
+        legacy_css_warning: bool,
+    ) -> bool:
+        """Handle legacy CSS under /static/*.css and add to briefcase.css.
+
+        :param wheel: Open wheel ZipFile being processed.
+        :param path: Path object of the file inside the wheel.
+        :param filename: Filename string inside the wheel.
```
freakboy3742: You shouldn't need to pass both the path *and* filename. The two are functionally interchangeable.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,354 @@ def _trim_file(self, path, sentinel):
...
+                    )
+
+                # New deploy/inserts handling
+                elif len(parts) >= 3 and parts[1] == "deploy" and parts[2] == "inserts":
+                    self._handle_insert(wheel, parts, filename, package_key, inserts)
+
+                else:
+                    self.console.debug(
+                        f"    {filename}: skipping, not a supported insert."
                     )
```
freakboy3742: This is a bit of a red herring - it's going to log on *every* file that isn't a legacy or new-style insert.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,354 @@ def _trim_file(self, path, sentinel):
...
+        # Return a blank pyscript config if no configuration file is found.
+        if len(config_package_list) == 0:
+            pyscript_config = {}
+        # Raise an error if more than one configuration file is supplied.
+        elif len(config_package_list) > 1:
+            raise BriefcaseConfigError(
+                f"""Only one backend configuration file can be supplied.
+                Initial config.toml found in package: {config_package}
+                Duplicate config.toml found in package: {wheel.filename}"""
+            )
```
freakboy3742, line 399: This is a multiline string, so the indentation on the second and third line will be preserved as-is.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,354 @@ def _trim_file(self, path, sentinel):
...
+        # Gather a backend configuration file from the package.
+        # For now, this is a pyscript.toml as no other backend is currently supported.
+        else:
+            with ZipFile(config_package_list[0]) as wheel:
+                with wheel.open(config_filename) as config_file:
+                    config_data = tomllib.load(config_file)
+
+                    # Gather backend type from config.toml
+                    backend = config_data.get("backend", None)
```
freakboy3742: "backend" isn't a great description of this feature - it's more of an "implementation".

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,354 @@ def _trim_file(self, path, sentinel):
...
+
+                    # Fail if no backend is present in config.toml
+                    if backend is None:
+                        raise BriefcaseConfigError(
+                            "No backend was provided in config.toml file."
```
freakboy3742: Given the mode of failure, it would be helpful to name the wheel where the config.toml was found as part of the error.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,354 @@ def _trim_file(self, path, sentinel):
...
+
+                else:
+                    self.console.debug(
+                        f"    {filename}: skipping, not a supported insert."
                     )
-                    css_file.write(f" * {package}::{'/'.join(path.parts[2:])}\n")
-                    css_file.write(
-                        " *******************************************************/\n\n"
+
+    def extract_backend_config(self, wheels):
```
freakboy3742: Given this is very specifically extracting a pyscript config, it would be appropriate to name it as such; if we ever add a different backend, we can flag that.

src/briefcase/platforms/web/static.py
freakboy3742: Also - given that a pyscript configuration will work out of the box with no extra configuration, we might as well "fail safe" and fall back to `pyscript` as a default value here.

src/briefcase/platforms/web/static.py
```diff
@@ -231,11 +565,21 @@ def build_app(self, app: AppConfig, **kwargs):
                 sentinel=" ******************* Wheel contributed styles **********************/",
             )
```
freakboy3742: This sentinel no longer exists in the template, so the trim isn't needed anymore.

tests/platforms/web/static/test_build__process_wheel.py
```diff
@@ -93,9 +82,267 @@ def test_process_wheel_no_content(build_command, tmp_path):
...
+        zf.writestr("dummy/static/bad.css", bad)
+
+    inserts = {}
+    with pytest.raises(UnicodeDecodeError):
+        build_command._process_wheel(wheelfile=wheel_path, inserts=inserts)
+
+
+def test_handle_legacy_css_warning_once_and_append(
+    build_command, tmp_path, monkeypatch
+):
```
freakboy3742: Two notes here:
1. As this method is testing `handle_legacy_css`, it should be in a `test_build_handle_legacy_css.py` file
2. When black/ruff splits the argument list breaks onto a new line, our preference is to move to "one argument per line" format.

tests/platforms/web/static/test_build__process_wheel.py
```diff
@@ -93,9 +82,267 @@ def test_process_wheel_no_content(build_command, tmp_path):
...
+
+@pytest.mark.parametrize(
+    "entry, expected_skip",
+    [
+        ("dummy/deploy/inserts/", True),  # Top-level directory entry
+        ("dummy/deploy/inserts/assets/", False),  # Nested directory entry
+    ],
+)
+def test_handle_insert_skip_dir_entries(
+    build_command, tmp_path, monkeypatch, entry, expected_skip
```
freakboy3742: As above - one-argument-per-line format.

tests/platforms/web/static/test_build__process_wheel.py
```diff
@@ -93,9 +82,267 @@ def test_process_wheel_no_content(build_command, tmp_path):
...
+    # One warning only
+    legacy_msgs = [m for m in warnings if "legacy '/static' CSS detected" in m]
+    assert len(legacy_msgs) == 1
+
+    # Content appended
+    out = inserts["static/css/briefcase.css"]["css"][key]
+    assert "h1 { x:1; }" in out and "/*extra*/" in out
+
+
+def test_handle_insert_register_valid_file(build_command, tmp_path):
```
freakboy3742: As above - this is testing `_handle_insert()`, so it should be in `test_build__handle_insert.py`.

src/briefcase/platforms/web/static.py
```diff
@@ -96,33 +116,354 @@ def _trim_file(self, path, sentinel):
...
+        self.console.info(f"    Found {filename}")
+
+        rel_inside = "/".join(parts[3:])
+
+        if not rel_inside or rel_inside.endswith("/"):
+            self.console.debug(f"    {filename}: skipping, not a valid insert file.")
+            return
+
+        if "~" not in rel_inside:
+            self.console.debug(
```
freakboy3742: These are both enough of a problem that I'd say they're worth flagging as a warning; debug will be essentially hidden unless you go digging, and they're an indication of a badly formed insert.

src/briefcase/platforms/web/static.py
caydnn: I've added the filename back as using path was causing issues in finding files as forward slashes would be converted to back slashes

caydnn:
Hi @freakboy3742, we have addressed your review comments and fixed up our test cases. This should be good for another review :)

On the note of documentation, I'm unsure if Kavi or I will have time to complete that this week. If we do, we can submit it as another PR if that works for you

src/briefcase/platforms/web/static.py
freakboy3742: Ok - that shouldn't be a problem... where exactly is this manifesting? If the problem is that we need to be able to guarantee forward slashes, you can use `PosixPath` instead of `Path` - that's the subclass of Path that is used on Linux/macOS, but you can also use it on Windows.

Also - is this something that is revealed by a test? If not... it should be.

caydnn: It was revealed when running `test_build_app` at first. It seemed the path would become malformed at some point as str(path) would be "dependency\\\\\\\\static\\\\\\\\style.css". So when attempting the line `css_text = wheel.read(str(path)).decode("utf-8")` it would result in KeyError.

This function is tested in `test_build__handle_legacy_css.py`, however this test only covers the function itself and not process_wheel as a whole.

caydnn: I should also note that I switched back to filename to match `_write_insert`, which also uses filename as well as 'parts' to create 'rel_inside'. If `PosxPath` were to resolve this issue as well, I'd be happy to switch both functions over to using that :)

freakboy3742: Ok - that's how the test is failing, but what's the actual manifestation in practice? The `\\\\` format suggests that something is getting double escaped - something is interpreting a Windows path as a unix path, and then back to a windows path (or similar). 

The core of what I'm getting at here is that Path() as an object *contains* all the information we need. We need to be careful about the exact format - but that's a code hygiene issue. You see the same thing with unicode and byte strings - you need to know, at every step of the process, if you have an array of unicode code points, or an array of bytes - and if it's an array of bytes, it is utf-8, or utf-16, or something else exotic. You don't fix the problem by passing both `bytes` and `str` around - you work out what format you have, what format you need, find out the representation that is richest and requires the least conversion (or the least risky covnersion), and then convert as needed.

In this case - does Zip on Windows return `/` or `\` paths? Can you read a `/` path out of a Windows zip file? We need to establish the ground truth here, and then work a clean solution from that.

caydnn: I see what you mean. Zip should always return `/` regardless of OS. 

After re-reading these functions, it looks like `_handle_insert` doesn't even need 'parts' as it immediately converts them to a PosixPath with `rel_inside = "/".join(parts[3:])` and `_handle_legacy_css` does a similar conversion with `rel_inside = "/".join(path.parts[2:])`

So, I think the best solution with the most control would be to replace the first conversion at `Path(filename)` in `_process_wheel` with a `PosixPath`, as you initially suggested. This would eliminate the need for parts and filename and in both handle functions.

caydnn: This has now been implemented and has resolved any issues with backslash in paths :)

freakboy3742 (approved):
Awesome - nice work!

PR #2475, opened 2025-09-12
Audit SIMXXX

### LINKED ISSUE #2384: Audit `SIMXXX` Ruff errors to resolve desired issues

The `SIM` (Flake8-simplify) checks in Ruff are *very* opinionated, and do not necessarily make for better code. For this reason, they are disabled by default. There are some errors provided by this check that are useful to resolve, some that require serious consideration, and some that should be avoided entirely. As there are some viable fixes within this rule set, it is worth enabling the check temporarily to work through the reasonable errors, and then disable it again to continue to ignore the unreasonable ones.

Work through this as follows:
1. Enable the `SIM` check by uncommenting [this line](https://github.com/beeware/briefcase/blob/1ffd045752d3a3d920a68d920201d890e8637e49/pyproject.toml#L248) in the `pyproject.toml` file.
3. Work through the resulting errors, focusing on the rules listed under "Errors to resovle" below. **Skip the errors listed under "Errors to skip".**
4. Once the errors under "Errors to Resolve" have been resolved or considered, *disable the `SIM` check* by commenting out [this line](https://github.com/beeware/briefcase/blob/1ffd045752d3a3d920a68d920201d890e8637e49/pyproject.toml#L248) in the `pyproject.toml` file.
5. Run `pre-commit run -a` to ensure it is passing with the `SIM` check disabled. 
6. There should be no changes present in `pyproject.toml` when you are ready to submit the PR.
7. Submit the PR.

### Errors to resolve
Of the errors you are likely to encounter here, *only* the following should be addressed. They are separated by which errors are most likely worth resolving, and which require consideration.

Errors that are almost certainly worth resolving include:
* `SIM115` - Use a context manager for opening files. This is valid except when it isn't possible; ask if you are unsure.
* `SIM117` - Use a single `with` statement with multiple contexts instead of nested `with` statements.
* `SIM118` - Use `key in dict` instead of `key in dict.keys()`.
* `SIM201` - Use `x != y` instead of `not x == y`
* `SIM300` - Yoda condition detected `(0 != x)`. 

Errors that require careful consideration include:
* `SIM103` - Return the condition `{condition}` directly.
* `SIM108` - Use ternary operator `{contents}` instead of `if`-`else`-block.
* `SIM905` - Consider using a list literal instead of `str.split`.

### Errors to skip
The following errors should be skipped and not resolved:
* `SIM102` - Use a single `if` statement instead of nested `if` statements
* `SIM105` - Use `contextlib.suppress({exception})` instead of `try`-`except`-`pass`

If you are unsure about anything or you encounter an error not included above, comment here to find out what course of action to take.

None.

### Additional context

Do not delete the `SIM` check or the explanatory comment above it. It should be commented out when you are finished with this issue to disable it again.

cgkoutzigiannis:
Fixes #2384.

I believe the code is easier to read now that SIM117, SIM118, SIM201, and SIM300 have been fixed.
I am not sure regarding opening the file with a context manager on: https://github.com/beeware/briefcase/blob/9e0a40774e235a8bdcc6a97abd2b100b3439c50f/src/briefcase/console.py#L160 so I left it as is for now, let me know if I can fix this.

I am not a big fan of the ternary operator, because I believe being more verbose makes the code easier to read with little cost, so I left SIM108 as is.

I did the same for SIM103.

ruff fixed SIM905 and I think it's better, although not a huge different either way.

## PR Checklist:

### DIFF
--- a/changes/2384.misc.rst
+++ b/changes/2384.misc.rst
@@ -1,0 +1,1 @@
+Resolved the SIMXXX caregory of ruff rules.


mhsmith:
Thanks, I haven't reviewed this in detail yet, but these mostly look like good improvements.

>  self._dev_null = open(os.devnull, "w", encoding="utf-8", errors="ignore") 

This can't be replaced with a context manager, because the file is used outside of the containing function. If Ruff doesn't detect that, then the check can be ignored using a comment on that line.

I know #2384 says otherwise, but is it possible to update the pyproject.toml file so that the rules you've fixed can continue to be checked in the future?

There are also some coverage and test failures which will need to be resolved.

cgkoutzigiannis:
Initially I added the comment on that line, so I will revert to that and enable the checks.

Regarding the coverage I am not sure what's going on since even when I initially cloned the repo I never got 100% coverage.

tests/test_cmdline.py
```diff
@@ -600,7 +600,7 @@ def test_command_overrides(monkeypatch, capsys, console):
     # Invoke a command that is known to have its own custom arguments
     # (In this case, the channel argument for publication)
     cmd, options, overrides = do_cmdline_parse(
-        "publish macos app -C version='1.2.3' -C extra=42".split(),
+        ["publish", "macos", "app", "-C", "version='1.2.3'", "-C", "extra=42"],
```
mhsmith, line 603: > ruff fixed SIM905 and I think it's better, although not a huge different either way.

It would be even better if `do_cmdline_parse` accepted either a list or a string, and automatically converted the string using `shlex.split`.

mhsmith:
It's expected that you won't get 100% coverage when running the tests locally, because some code is platform-specific. The CI runs the tests on all platforms and then merges the results together.

tests/test_cmdline.py
freakboy3742: Given it's a test utility method, and one that is only used in this test file, I'm inclined to think only accepting the list form is more appropriate - it's more explicit, and the extra characters to split the parse call into extra arguments isn't especially onerous.

changes/2384.misc.rst
```diff
@@ -0,0 +1 @@
+Resolved the SIMXXX caregory of ruff rules.
```
freakboy3742, line 1: Typo:
```suggestion
Resolved the SIMXXX category of ruff rules.
```

freakboy3742:
Sorry - had #2516 open - that review was intended for that version of this PR.

mhsmith:
Continued in #2516.

PR #2476, opened 2025-09-13
Resolve b904

### LINKED ISSUE #2381: Resolve `BXXX` series ignored Ruff rules

cgkoutzigiannis:
Fixes #2381 

In most cases I can't see any reason to keep the original error, since the exceptions we are eventually throwing are much more informative.

## PR Checklist:

### DIFF
--- a/changes/2381.misc.rst
+++ b/changes/2381.misc.rst
@@ -1,0 +1,1 @@
+Resolved the B904 ruff rule ruff.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -252,7 +252,7 @@ extend-select = [
 # for future fixes throughout the project.
 ignore = [
     "E501",    # line length
-    "B904",    # use raise ... from ...
+    # "B904",    # use raise ... from ...
 ]
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -252,7 +252,6 @@ extend-select = [
 # for future fixes throughout the project.
 ignore = [
     "E501",    # line length
-    # "B904",    # use raise ... from ...
 ]
 
 
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -159,7 +159,7 @@ def parse_cmdline(args, console: Console | None = None):
         try:
             platform_module = platforms[platform]
         except KeyError:
-            raise InvalidPlatformError(platform, platforms.keys())
+            raise InvalidPlatformError(platform, platforms.keys()) from None
 
         # If the output format wasn't explicitly specified, check to see
         # Otherwise, extract and use the default output_format for the platform.
@@ -184,11 +184,11 @@ def parse_cmdline(args, console: Console | None = None):
             raise InvalidFormatError(
                 requested=output_format,
                 choices=list(output_formats.keys()),
-            )
+            ) from None
         except AttributeError:
             raise UnsupportedCommandError(
                 platform=platform,
                 output_format=output_format,
                 command=options.command,
-            )
+            ) from None
     return Command, extra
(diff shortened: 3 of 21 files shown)


mhsmith:
The [B904 description](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/) says:

> When raising an exception from within an `except` clause, always include a `from` clause to facilitate exception chaining. If the exception is not chained, it will be difficult to trace the exception back to its root cause.

But this is untrue, because a raise without a `from` clause still causes the exception to be chained, just with a slightly different wording in the traceback. This is a perfectly reasonable default, and I see no reason to force the developer to add noise by spelling it out everywhere.

There may be some benefit to adding `from None` in some places, but the examples in this PR mostly fall into two categories:

* `BriefcaseCommandError` or its subclasses, where the full stack trace isn't shown to the user anyway, but it still might be useful to keep it in the log file.
* "Impossible" situations like the ParseError in `linux/__init__.py`, where having the full exception chain is potentially important.

cgkoutzigiannis:
I see what you mean, I will keep the errors on all the BriefcaseCommandError, it's subclasses and on weird cases like ParseError and we will see about the rest.

mhsmith:
Sorry, I think you've misunderstood me. I was saying I don't think we need explicit exception chaining when the implicit chaining is good enough. In other words, B904 would remain ignored, and this PR would be reduced only to adding `from None` where appropriate.

freakboy3742:
> Sorry, I think you've misunderstood me. I was saying I don't think we need explicit exception chaining when the implicit chaining is good enough. In other words, B904 would remain ignored, and this PR would be reduced only to adding `from None` where appropriate.

I'm not sure about that - keeping B904 on means we will be able to forcing every new exception (or exception change) to make an explicit decision about whether chaining is appropriate. Without the automated rule enforcement, we're going to continue to forget to add `from None` when it would be appropriate to do so, because there's no forcing function requiring it. Over time, reviewers will inevitably miss usages, because they're not asked to notice something that isn't being done, rather than noticing that something that *is* being done is being done wrong.

Yes, this means a little more code because the default behavior requires adding `from e` in a bunch of places - but it's not *that* hard to do (or read), so my inclination is that it's worth the extra characters.

mhsmith:
Many of these are BriefcaseCommandErrors, where the traceback wouldn't be shown to the user anyway. But I guess encouraging the use of `from None` when appropriate may help to keep the log files a bit more readable, so I'll accept enforcing the check.

However, I think most if not all of the `KeyErrors` would provide no additional information, and should be replaced with `from None`.

cgkoutzigiannis:
How should we proceed with this?

mhsmith:
There are 12 instances in this PR of `except KeyError as e ... raise ... from e`. I've looked over them all, and I don't think any of them should be chained, so please change them all to `from None`.

Apart from that, the PR looks fine to me.

mhsmith:
Continued in #2515.

PR #2478, opened 2025-09-14
Add Flatpak base documentation

limdingwen:
Add documentation for https://github.com/beeware/briefcase-linux-flatpak-template/pull/64

## PR Checklist:


mhsmith:
Please open the "towncrier" link in the "checks" section below and follow the instructions. A small change like this can use the "misc" category.

freakboy3742 (changes requested):
In addition to the missing changenote, I wonder if this is actually the complete fix here - while we definitely need to document these new options, should we also be adding a default base to the PySide app template? The PR description says this a base can be used to "fix" Pyside apps - but I'm not clear what specifically is "broken". If there's a bug and fix here, then it sounds like we should be modifying the PySide bootstrap as well, not just documenting a workaround.

limdingwen:
I think one needs to check the PySide app template to see if it's broken by default, and if so, add this as the base. No time for me currently. For me, my use case was in a PyQt app, which I think is related. "libgssapi_krb5.so.2" tends to be missing, which is part of what the https://github.com/flathub/com.riverbankcomputing.PyQt.BaseApp/ base fixes.

docs/reference/platforms/linux/flatpak.rst
```diff
@@ -129,6 +129,19 @@ corresponding :attr:`flatpak_sdk` must be defined.
 
 A string, identifying the version of the Flatpak runtime that should be used.
 
+.. attribute:: flatpak_base
+
+An optional string, identifying the `base
+<https://docs.flatpak.org/en/latest/flatpak-builder-command-reference.html#flatpak-manifest#:~:text=Start%20with%20the%20files%20from%20the%20specified%20application.%20This%20can%20be%20used%20to%20create%20applications%20that%20extend%20another%20application.>`__
+to use as an additional base for the Flatpak app, on top of the runtime. Per the
+Flatpak documentation, "this can be used to create applications that extend
+another application".
```
freakboy3742: This is currently failing link validation, probably because of the link format: 
```suggestion
An optional string, identifying the `base
<https://docs.flatpak.org/en/latest/flatpak-builder-command-reference.html#flatpak-manifest>`__
to use as a base for the Flatpak app. A base is a set of extensions layered on top of a
runtime image, providing additional functionality for the packaged application. Some GUI
frameworks (e.g., PyQt) provide a base image to ensure common required resources are
available at runtime.
```

Issue #2465, opened 2025-09-04
Truncated icon link in shortcut when packaging on Windows, leading to default "white paper" icon in Windows taskbar, search, and on shortcut


initialcommit-io:
### Describe the bug

When building/packaging on Windows 11 and installing via msi, the icon path in the shortcut is getting truncated for some reason to:

pData\Roaming\Microsoft\Installer{29E11852-9351-4930-B6BD-A7FBABB7708C}\ProductIcon?,0

The actual ProductIcon is being properly generated and placed here:

C:\Users\asdf\AppData\Roaming\Microsoft\Installer{29E11852-9351-4930-B6BD-A7FBABB7708C}\ProductIcon

However, for some reason, the shortcut icon path is being truncated, resulting in issues displaying the icon in various locations.

So far I'm only able to reproduce this on my app build, but not on the example helloworld app in the tutorial, even when using my same icon .ico file.

### Steps to reproduce

So far I'm only able to reproduce this on my app build, but not on the example helloworld app in the tutorial, even when using my same icon .ico file.

### Expected behavior

The shortcut path to the app icon is not truncated, and the icon displays properly in Windows taskbar, search, and shortcut.

### Screenshots

<img width="425" height="435" alt="Image" src="https://github.com/user-attachments/assets/511cda0d-db8e-497a-a544-f3a136964509" />
<img width="532" height="113" alt="Image" src="https://github.com/user-attachments/assets/85553bc5-54d8-4ec0-afcd-474abe95d492" />
<img width="77" height="70" alt="Image" src="https://github.com/user-attachments/assets/4c3d48dc-0e85-475b-be55-fa03a7fd2451" />

### Environment

- Operating System: Windows 11
- Python version: 3.11
- Software versions:
  - Briefcase: 0.3.25

### Logs

```

```

freakboy3742:
After some debugging on Discord, the problem has been identified:

If the `description` field for a Windows shortcut is > 256 characters, the prefix of the icon link is shortened by the amount that the description field overflows the 256 character limit. In this case, the description field was 272 characters, so the first 16 characters was being truncated.

This was only discovered by rolling back to Wix 3. Wix 3 raises a warning in this condition. Wix 5 is silent on the matter. I can't see any obvious "verbose" or "display warnings" mode that we've missed in wix.exe, so...  thanks Wix. :-) 

So - the immediate workaround is to truncate the description field.

Longer term, there are 2 (well... 3) fixes required:
[...]

PR #2489, opened 2025-09-25
Fix truncated icon

bigsby-exe:
This PR adds a warning message for descriptions over 80 chars this is due to an issue packaging on Windows where the icon was getting truncated.

This is part of 3 PRs to update both the Windows app template and VisualStudio template to truncate descriptions over 256 chars & add a warning to briefcase
https://github.com/beeware/briefcase-windows-VisualStudio-template/pull/68
https://github.com/beeware/briefcase-windows-app-template/pull/72

This resolves issue: #2465 

Unsure if I am doing something wrong but I can't get all the tests to pass even on a clean clone of the main repo.
I have managed to reduce some of the failures by adding tests for the extra checks I have added

## PR Checklist:


src/briefcase/config.py
```diff
@@ -40,6 +40,11 @@ def is_valid_app_name(app_name):
     return not is_reserved_keyword(app_name) and is_valid_pep508_name(app_name)
 
 
+def is_valid_description(description):
+    """Determine if the description is not longer than 80 characters."""
+    return len(description) <= 80
```
freakboy3742: As there's only one place where this logic is needed, and the logic is fundamentally pretty simple, factoring this out as a method is slightly overkill.

src/briefcase/config.py
```diff
@@ -397,6 +402,14 @@ def __init__(
 
         self.test_mode: bool = False
 
+        if not is_valid_description(self.description):
+            raise BriefcaseConfigError(
+                "The app description is too long\n\n"
+                f"The description is {len(self.description)} characters long.\n"
+                "Longer descriptions should use the 'long_description' field.\n"
+                "On some platforms, descriptions may be truncated to fit platform limitations.",
+            )
```
freakboy3742: There's a slight discrepancy here between the message and the way it's handled. The message suggests the content "may" be truncated; however, it will be impossible to reach that point, because any description over 80 characters will raise a hard error.

freakboy3742 (changes requested):
Thanks for the PR. 

Regarding the test failures - the test failure is exactly what it says on the tin. The test is looking for a specific piece of text - and that text doesn't occur in the actual output. It's *almost* there... but there's a missing period.

However, in this case, I think the implementation isn't in quite the right place. As flagged inline, the current wording is a hard error, not a soft warning. The AppConfig object is a good place for hard errors - i.e., you can't use a config *at all* unless the version number is valid, for example. In this case, we *can* proceed - but we need to warn the user that the description will result in potentially unexpected behavior. 

I'd suggest a better approach would be to use the `finalize_app_config` step - that's a step that will be run once, in a context where you have access to the console, so you can output a warning, rather than a hard error.

bigsby-exe:
Thanks for the feedback, I shall try and re-implement

bigsby-exe:
@freakboy3742 is this a better implementation?

freakboy3742:
> @freakboy3742 is this a better implementation?

The approach is good; however, I'd be inclined to make it a check on the create command that applies for *all* platforms, rather than something specific to the Windows backend. While it only has immediate consequences for Windows, someone who is writing an app for Linux/macOS/whatever should be warned that future ports to Windows might be a problem.

As a heads up - the test failure isn't of your making; overnight, OpenSUSE has made a change to their repository that has broken Linux app support. We're still investigating the cause to see if it's something we need to work around, or if it's a case where they've messed up an update (which is something that happens *way* more often than it should with OpenSUSE).

mhsmith:
The OpenSUSE issue has now been resolved.

bigsby-exe:
> > @freakboy3742 is this a better implementation?
> 
> The approach is good; however, I'd be inclined to make it a check on the create command that applies for _all_ platforms, rather than something specific to the Windows backend. While it only has immediate consequences for Windows, someone who is writing an app for Linux/macOS/whatever should be warned that future ports to Windows might be a problem.
> 
> As a heads up - the test failure isn't of your making; overnight, OpenSUSE has made a change to their repository that has broken Linux app support. We're still investigating the cause to see if it's something we need to work around, or if it's a case where they've messed up an update (which is something that happens _way_ more often than it should with OpenSUSE).

Thanks again for the feedback. That all makes sense so I will try and implement it on create for all platforms.
Also thanks for the note on the failing test, that did catch me off guard 🤣

glasnt:
Hi! It looks like this pull request is a little bit old, and isn't mergable due to needing a rebase. 

If you are still interested in having this PR considered for merging, please complete a rebase. If you aren't, please feel free to close this PR. 

(If we don't hear back from you, we will close this PR for you.)

freakboy3742:
This was implemented in #2869. Thanks for the PR, and apologies for the confusion.

PR #2420, opened 2025-07-29
And Venv Management to Dev Command Web Platform

JMah007:
Functionality has been added to `briefcase dev web` to handle the creation of a separate virtual environment for each app within a web project. The implementation is in the `dev` command file to allow the desktop platforms to utilise the isolated venv feature in the future. 

These `venv's` are initialised in the default project folder under `.briefcase`. Implementations for handling the running of `briefcase dev web` and refreshing of a web app's venv have not been done so have been replaced temporarily with the invocation of `UnsupportedCommandError`. The applicable tests have been added to cover these new features and actual demos for the venv management have been tested on windows. 

Do note that this PR builds upon #2419 as the `dev command` must first be enabled on the web platform before the venv can be implemented.

Refs #1735 
Refs #2334
Refs #2419 
## PR Checklist:

- [x] All new features have been tested
- [x] All new features have been documented
- [x] I have read the **CONTRIBUTING.md** file
- [x] I will abide by the code of conduct

PR #2495, opened 2025-10-02
Replaced dist-info tracking with venv creation tracking for dev command

mEp3ii2:
After #2420, an edge case was identified where running briefcase dev followed by briefcase dev web would cause the web dev command to skip dependency installation.
The issue occurs because both commands share the same source directory where the dist-info file is written. When briefcase dev runs first, it writes the dist-info file to the source. Then when briefcase dev web runs with its own isolated virtual environment, it sees the existing dist-info file and incorrectly assumes dependencies are already installed in its venv, even though they aren't.
This PR replaces the dist-info approach with venv-specific tracking:

Virtual environments now track a created attribute that's set when the venv is created or recreated
For non-isolated environments, a marker file (.venv-marker) is stored alongside the venv and contains the Python executable path
Dev venv names now include the Python version (e.g., dev-3.12, dev-3.12-web) to isolate environments across Python versions
Dependencies are installed when venv.created=True or when explicitly requested via --update-requirements

Since the marker is stored in the venv-specific directory rather than the shared source directory, each environment correctly tracks its own dependency installation status.
Refs #2334

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/dev.py
+++ b/src/briefcase/commands/dev.py
@@ -11,7 +11,6 @@ from briefcase.exceptions import BriefcaseCommandError, RequirementsInstallError
 from briefcase.integrations.virtual_environment import VenvContext, virtual_environment
 
 from .base import BaseCommand
-from .create import write_dist_info
 
 
 class DevCommand(RunAppMixin, BaseCommand):
@@ -207,8 +206,8 @@ class DevCommand(RunAppMixin, BaseCommand):
 
         :returns: Name for virtual environment directory
         """
-
-        return "dev"
+        python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+        return f"dev--{python_version}"
 
     def venv_path(self, appname: str) -> Path:
         """Return the path for the app's virtual environment.
@@ -218,7 +217,9 @@ class DevCommand(RunAppMixin, BaseCommand):
         """
         return self.base_path / ".briefcase" / appname / self.venv_name
 
-    def virtual_environment(self, appname: str, isolated: bool) -> virtual_environment:
+    def virtual_environment(
+        self, appname: str, isolated: bool, recreate: bool
+    ) -> virtual_environment:
         """Create and return a virtual environment context for the app.
 
         :param appname: The name of the app to create a venv context for
@@ -231,6 +232,7 @@ class DevCommand(RunAppMixin, BaseCommand):
             console=self.console,
             venv_path=self.venv_path(appname),
             isolated=isolated,
+            recreate=recreate,
         )
 
     def __call__(
@@ -270,21 +272,19 @@ class DevCommand(RunAppMixin, BaseCommand):
         # If one exists, assume that the requirements have already been
         # installed. If a dependency update has been manually requested,
         # do it regardless.
-        dist_info_path = (
-            self.app_module_path(app).parent / f"{app.module_name}.dist-info"
-        )
+
         if not run_app:
             # If we are not running the app, it means we should update requirements.
             update_requirements = True
-
-        with self.virtual_environment(
+        env = self.virtual_environment(
             appname=app.app_name,
             isolated=options.get("isolated", False),
-        ) as venv:
-            if update_requirements or not dist_info_path.exists():
+            recreate=update_requirements,
+        )
+        with env as venv:
+            if env.created or update_requirements:
                 self.console.info("Installing requirements...", prefix=app.app_name)
                 self.install_dev_requirements(app, venv, **options)
-                write_dist_info(app, dist_info_path)
 
             if run_app:
                 if app.test_mode:
--- a/src/briefcase/integrations/virtual_environment.py
+++ b/src/briefcase/integrations/virtual_environment.py
@@ -23,6 +23,7 @@ class VenvContext(Tool):
     def __init__(self, tools: ToolCache, venv_path: Path):
         super().__init__(tools=tools)
         self.venv_path = venv_path
+        self.created = False
 
     @classmethod
     def verify_install(
@@ -68,6 +69,7 @@ class VenvContext(Tool):
                 check=True,
             )
             self.update_core_tools()
+            self.created = True
         except stdlib_subprocess.CalledProcessError as e:
             raise BriefcaseCommandError(
                 f"Failed to create virtual environment at {self.venv_path}"
@@ -184,6 +186,11 @@ class VenvEnvironment:
             tools=self.tools, venv_path=self.venv_path
         )
 
+    @property
+    def created(self) -> bool:
+        """Exposes the created status of the venv context."""
+        return self.venv_context.created
+
     def __enter__(self):
         rel_venv_path = self.venv_path.relative_to(Path.cwd())
         if self.recreate:
@@ -206,11 +213,34 @@ class VenvEnvironment:
 class NoOpEnvironment:
     """A no-op environment that returns a native runner."""
 
-    def __init__(self, tools: ToolCache, console: Console):
+    def __init__(self, tools: ToolCache, console: Console, marker_path: Path):
         self.tools = tools
         self.console = console
+        self.marker_path = marker_path
+        self.created = False
+
+    def check_and_update_marker(self) -> bool:
+        """Check marker file and update if needed.
+
+        :returns: True if this is a new environment of Python executable changed.
+        """
+
+        if not self.marker_path.exists():
+            self.marker_path.parent.mkdir(parents=True, exist_ok=True)
+            self.marker_path.write_text(sys.executable, encoding="utf-8")
+            return True
+        try:
+            existing_executable = self.marker_path.read_text(encoding="utf-8").strip()
+            if existing_executable != sys.executable:
+                self.marker_path.write_text(sys.executable, encoding="utf-8")
+                return True
+        except (OSError, UnicodeDecodeError):
+            self.marker_path.write_text(sys.executable, encoding="utf-8")
+            return True
+        return False
 
     def __enter__(self):
+        self.created = self.check_and_update_marker()
         return self.tools.subprocess
 
     def __exit__(self, exc_type, exc_val, exc_tb):
@@ -238,7 +268,8 @@ def virtual_environment(
     :returns: A context manager for an environment where code can be executed.
     """
     if not isolated:
-        return NoOpEnvironment(tools=tools, console=console)
+        marker_path = venv_path / ".venv-marker"
+        return NoOpEnvironment(tools=tools, console=console, marker_path=marker_path)
 
     if venv_path is None:
         raise BriefcaseCommandError("A virtual environment path must be provided")
--- a/src/briefcase/platforms/web/static.py
+++ b/src/briefcase/platforms/web/static.py
@@ -461,8 +461,8 @@ class StaticWebDevCommand(StaticWebMixin, DevCommand):
 
         :returns: Name for virtual environment directory
         """
-
-        return "dev-web"
+        python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+        return f"dev-web--{python_version}"
 
     def add_options(self, parser):
         super().add_options(parser)


mEp3ii2:
This is just a first pass over. Will need to add tests to check that this new behavior works as expected and will iterate based on test results

src/briefcase/commands/dev.py
```diff
@@ -207,8 +206,8 @@ def venv_name(self) -> str:
 
         :returns: Name for virtual environment directory
         """
-
-        return "dev"
+        python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+        return f"dev--{python_version}"
```
freakboy3742: Assuming this is a typo
```suggestion
        return f"dev-{python_version}"
```

src/briefcase/platforms/web/static.py
```diff
@@ -461,8 +461,8 @@ def venv_name(self) -> str:
 
         :returns: Name for virtual environment directory
         """
-
-        return "dev-web"
+        python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+        return f"dev-web--{python_version}"
```
```suggestion
        return f"dev-web-{python_version}"
```

src/briefcase/integrations/virtual_environment.py
```diff
@@ -238,7 +268,8 @@ def virtual_environment(
     :returns: A context manager for an environment where code can be executed.
     """
     if not isolated:
-        return NoOpEnvironment(tools=tools, console=console)
+        marker_path = venv_path / ".venv-marker"
```
freakboy3742: Making this a hidden file will make it a little harder to visually inspect based on a directory listing.

I'd suggest that it might make more sense to make `venv_path` a *file* for the non-isolated case. That makes it clear that it's not an isolated virtual environment; and provides a basis to check for the "was/wasn't previously run in a non-isolated environment" in `VenvContext.exists()`/`check_and_update_marker()` (and prompt the user to see if they want to destroy the old configuration).

src/briefcase/integrations/virtual_environment.py
```diff
@@ -206,11 +213,34 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                self.marker_path.write_text(sys.executable, encoding="utf-8")
+                return True
+        except (OSError, UnicodeDecodeError):
+            self.marker_path.write_text(sys.executable, encoding="utf-8")
+            return True
+        return False
 
     def __enter__(self):
+        self.created = self.check_and_update_marker()
         return self.tools.subprocess
```
freakboy3742: I presume the reason you've put the `created` check on the context manager is because `subprocess` doesn't have a `created` attribute. Python allows us to dynamically associate any attribute we want - so setting `self.tools.subprocess.created` would be a valid approach here. We can then use the existence of that attribute as a marker for whether subprocess is being used as a No-op virtual environment. Or, if you want to go all out, you could define a NoOpVenvContext that is a facade around subprocess, passing method calls through directly.

freakboy3742 (changes requested):
Looks to be headed in the right direction; a couple of small notes inline.

src/briefcase/integrations/virtual_environment.py
mEp3ii2: I've had a go at implementing a `NoOpVenvContext`, have a look when you've got some time.

src/briefcase/commands/dev.py
```diff
@@ -270,21 +277,19 @@ def __call__(
...
-        with self.virtual_environment(
+        env = self.virtual_environment(
             appname=app.app_name,
             isolated=options.get("isolated", False),
-        ) as venv:
-            if update_requirements or not dist_info_path.exists():
+            recreate=update_requirements,
+        )
+        with env as venv:
+            if env.created or update_requirements:
```
freakboy3742: With the no-op change, it should be possible to make this `if venv.created`, which removes the need to pre-declare `env`.

src/briefcase/integrations/virtual_environment.py
```diff
@@ -203,15 +210,75 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    def __init__(self, tools: ToolCache, console: Console, marker_path: Path):
         self.tools = tools
         self.console = console
+        self.marker_path = marker_path
+        self.noop_context = NoOpVenvContext(tools=tools)
+
+    @property
+    def created(self) -> bool:
+        """Exposes the created stat of the noop context."""
+        return self.noop_context.created
```
freakboy3742: This won't be needed if you test created directly against the `venv` context object.

src/briefcase/integrations/virtual_environment.py
```diff
@@ -184,6 +186,11 @@ def __init__(
             tools=self.tools, venv_path=self.venv_path
         )
 
+    @property
+    def created(self) -> bool:
+        """Exposes the created status of the venv context."""
+        return self.venv_context.created
+
```
freakboy3742: This won't be needed if you test created directly against the `venv` context object.

src/briefcase/integrations/virtual_environment.py
```diff
@@ -203,15 +210,75 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
         self.console = console
+        self.marker_path = marker_path
+        self.noop_context = NoOpVenvContext(tools=tools)
+
+    @property
+    def created(self) -> bool:
+        """Exposes the created stat of the noop context."""
+        return self.noop_context.created
+
+    def check_and_update_marker(self) -> bool:
```
freakboy3742: This is really the analog of `create()`, so it should be on the context object, not the context manager.

mEp3ii2:
@freakboy3742 I've made some further changes but I've noticed that after making changes to the test_call file for dev we now have missing coverage for the create command (lines 74-49) which relate to the dist-info file creation that we have phased out in the dev command. Should I add coverage for this in the create tests, or there another approach you'd prefer?

freakboy3742:
> @freakboy3742 I've made some further changes but I've noticed that after making changes to the test_call file for dev we now have missing coverage for the create command (lines 74-49) which relate to the dist-info file creation that we have phased out in the dev command. Should I add coverage for this in the create tests, or there another approach you'd prefer?

From the look of it, we were implicitly testing the creation of a dist-info for a minimal package; so yes - the fix here is to add a test to `create/test_install_app_code.py` that exercises the minimal dist-info case.

mEp3ii2:
@freakboy3742 I believe that this is ready now. 
My one point of concern is that in the new test method in `create/test_install_app_code.py` I had to use the `\x20` escape sequence here to preserve a trailing space where we check the DOWNLOAD-URL which is blank in this case. The pre-commit hook automatically strips normal trailing spaces, so I tried this not sure if this is the correct approach. That or we could switch to checking each value individually rather then reading the entire file

src/briefcase/commands/dev.py
```diff
@@ -270,21 +277,19 @@ def __call__(
...
             isolated=options.get("isolated", False),
-        ) as venv:
-            if update_requirements or not dist_info_path.exists():
+            recreate=update_requirements,
+        )
+        with env as venv:
+            if venv.created or update_requirements:
                 self.console.info("Installing requirements...", prefix=app.app_name)
                 self.install_dev_requirements(app, venv, **options)
-                write_dist_info(app, dist_info_path)
```
freakboy3742, line 287: Was this removal intentional? We still need the dist_info file in dev - we're just not using it as an install marker any more.

src/briefcase/integrations/virtual_environment.py
```diff
@@ -203,15 +205,79 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    """
+
+    name = "no_op_environment"
+    full_name = "No-Op Environment"
+
+    def __init__(self, tools, marker_path: Path, **kwargs):
+        super().__init__(tools, **kwargs)
+        self.created = False
+        self.venv_path = None
+        self.marker_path = marker_path
```
freakboy3742: Why not consolidate these two? The marker path is a known extension of the venv_path; might as well keep the interface to vent-like objects consistent, always create them with a path, and leave the construction of any internal marker files to the internal implementation.

src/briefcase/integrations/virtual_environment.py
```diff
@@ -203,15 +205,79 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    def check_and_update_marker(self) -> bool:
+        """Check marker file and update if needed.
+
+        :returns: True if this is a new environment or Python executable changed.
+        """
+
+        if not self.marker_path.exists():
+            self.marker_path.parent.mkdir(parents=True, exist_ok=True)
+            self.marker_path.write_text(sys.executable, encoding="utf-8")
+            return True
```
freakboy3742: This correctly evaluates whether the environment has been created, but `venv.created` is never updated.

src/briefcase/integrations/virtual_environment.py
```diff
@@ -238,7 +304,8 @@ def virtual_environment(
     :returns: A context manager for an environment where code can be executed.
     """
     if not isolated:
-        return NoOpEnvironment(tools=tools, console=console)
+        marker_path = venv_path / "venv_path"
+        return NoOpEnvironment(tools=tools, console=console, marker_path=marker_path)
```
freakboy3742: This is the sort of interface I was referring to earlier - The *environment* is at a specific venv_path; the marker file is an internal implementation detail that can be honoured at *any* path.

tests/integrations/virtual_environment/test_NoOpVenvContext__create.py
freakboy3742, line 1: This file name looks like it has a stray space...

tests/integrations/virtual_environment/test_NoOpVenvContext__exists.py
```diff
@@ -0,0 +1,9 @@
+def test_exists_always_returns_true(noop_context):
+    """Exists always returns True for no-op context."""
+    assert noop_context.exists() is True
+
+
+def test_exists_when_marker_missing(noop_context):
+    """Exists returns True even when marker file doesn't exist."""
```
freakboy3742: These two tests don't validate anything different. Either they should be a single test; or there should be some form of environment modification to make the test conditions different.

freakboy3742 (changes requested):
This is looking good; the testing strategy looks on track. Responding to your specific question: the `\x20` trick for testing is very clever, and thoroughly acceptable. For the record, another way to address that would be to construct the overall string as a series of list of single line strings, joined with `\n` - but what you've done definitely works as well.

The only catch with the rest of the PR... it doesn't actually work :-)

A clean project with no `.briefcase` directory doesn't trigger the installation of requirements. It correctly identifies that the environment is new, and creates the marker file - but then the `created` state of the venv is never updated. So - dependencies are never installed into a no-op venv.

I suspect the gap between what is here and what would work is very small; but it will obviously require a change to some tests as well - if only to verify what isn't currently being detected (that `venv.created` is set as a result of writing the marker file)

mEp3ii2:
@freakboy3742 Everything should be all set now, can you please have a look over to confirm its all good.

src/briefcase/commands/dev.py
```diff
@@ -273,15 +259,17 @@ def __call__(
...
 
-        with self.virtual_environment(
-            appname=app.app_name,
+        with self.tools.virtual_environment.create(
+            venv_path=self.venv_path(app.app_name),
             isolated=options.get("isolated", False),
+            recreate=update_requirements,
         ) as venv:
-            if update_requirements or not dist_info_path.exists():
+            if venv.created:
```
freakboy3742, line 270: It turns out that if you pass `recreate` down to the no-op, you can *completely* remove the need for update_requirements handling here, and do everything with the `created` flag.

src/briefcase/commands/dev.py
```diff
@@ -207,7 +209,6 @@ def venv_name(self) -> str:
 
         :returns: Name for virtual environment directory
         """
-
         return "dev"
```
freakboy3742, line 210: It turns out the Python version numbering isn't really needed here - the venv path implies the Python version; and if you have 2 different Python environments with the same version, you need to re-install anyway. So - we might as well keep this simple and re-create on *any* environment change.

freakboy3742 (approved):
Thanks for those updates - everything is now working as expected.

In the process of doing a fine-grained review, I realised that we actually got the Tool abstraction slightly wrong in the original venv PR - the Tool should be the virtual environment factory, not the VenvEnvironment. With that change, a bunch of other simplifications and cleanups are possible. I've pushed an update to make this change.

Really excited to see this fix - the way requirements work in dev environments has been a hard to explain behavior for a long time, and this change makes it really straight forward and less fragile. Thanks!

Issue #2191, opened 2025-03-12
Custom install / uninstall scripts


mhsmith:
We're currently discussing using Briefcase as a backend for the [conda constructor](https://github.com/conda/constructor) tool. To do this would require several new features, but the following two are closely connected:

* Custom scripts to run during installation and uninstallation.
* Boolean options that the user can choose during installation and uninstallation.
  * It must also be necessary to set these on the command line for a non-interactive install / uninstall, e.g. in an organizational device management system.

The following is a sketch of how Briefcase could support this in a generic way. Like most of the existing pyproject.toml settings, these could be set either at the app level, or per platform. 

We're only considering the 3 desktop platforms at the moment. And in the case of macOS, only the .pkg format would support these features.

----

Supporting the scripts themselves would be the simpler part: add settings to select their paths in the source directory. Since the scripts would require some kind of interpreter on the target machine, their supported formats would inevitably be platform-specific, but should at least include .bat for Windows, and .sh for Linux and macOS.

There would be 4 types of scripts: preinstall, postinstall, preuninstall and postuninstall. "Pre" and "post" here refer to the creation or deletion of the installed directory, so only postinstall and preuninstall would be run with that directory present.

The scripts would not be able to take input: they could only print to stdout and stderr, and report success or failure with an exit code.

----

The boolean options would be specified as a list of items, each of which has the following properties:

* ID (string)
* Title (string)
* Description (string)
* Command-line flag (string)
* Default (boolean)
* Enabled (string: see below)

All of these are self-explanatory except for "enabled", as this may depend on the value of other options, or details of the installation environment. We can therefore express it as a string containing a boolean expression in a simple language supporting only the following elements:

* Operators `and`, `or` and `not`
* Parentheses
* IDs of other options, or IDs pre-defined by Briefcase. 

At installer build time, Briefcase would parse this expression with the Python `ast` module, and transform it into whatever code is necessary to implement it in each installer format.
 
Currently the only pre-defined ID we expect to need is an indication of whether an installation is in admin mode. For example, an option which should only be enabled in single-user mode would have `enabled = "not admin_mode"`.

At install / uninstall time, the user's chosen options are passed to the scripts using environment variables with the name `BRIEFCASE_` followed by the option's ID, and the value `0` or `1`. This is likely more convenient than parsing a command line in a .sh or .bat script.

freakboy3742:
This all broadly make sense; a couple of questions/clarifications.

[...]

PR #2496, opened 2025-10-03
Add handling for post-install and pre-uninstall scripts

freakboy3742:
Refs #2191.

Adds support for post-install and pre-uninstall scripts; at present, only for Windows MSI installers. Pre/post install scripts are required to be batch files (as there's no guarantee powershell has been enabled on any given target machine - and disabling powershell seems to be a common default configuration).

Also add support for optional features in an installer; these are (at present) required to be booleans.

Lastly, adds acknowledgement of the project license to the installer process. This requires a text to RTF conversion; a naive txt2rtf converter has been implemented.

---
At present, this is undocumented; I'm pushing it for some initial feedback. It requires the use of the beeware/briefcase-windows-app-template#73 (branch `pre-post-script`) template modifications. For testing purposes, this is a sample app configuration:

```
[tool.briefcase.app.scripttest.windows]
...
template_branch="pre-post-script"

post_install_script = "script/post_install.bat"
pre_uninstall_script = "script/pre_uninstall.bat"
# installer_path = "_installer"

[[tool.briefcase.app.scripttest.install_option]]
name = "foo"
description = "Foo the thing"
description = "Foo the thing"
default = true

[[tool.briefcase.app.scripttest.install_option]]
name = "bar"
title = "Bar the thing"
description = "Make the Bar the thing"
default = false
```

The two script files should be in the repo relative to the pyproject.toml. They can be any batch file; I suggest content similar to the following for testing:
```
@echo off
echo POST INSTALL SCRIPT
echo ARGS: %*
timeout 10
```

## PR Checklist:


src/briefcase/platforms/windows/__init__.py
```diff
@@ -132,6 +175,58 @@ def _cleanup_app_support_package(self, support_path):
...
+        This includes any post-install or pre-uninstall scripts, plus converting the
+        LICENSE file into an RTF file.
+
+        :param app: The config object for the app
+        """
+        if post_install := getattr(app, "post_install_script", None):
+            post_install_script = self.base_path / post_install
+            if post_install_script.suffix != ".bat":
+                raise BriefcaseCommandError(
+                    "Windows post-install scripts must be batch files."
```
mhsmith: We could be more explicit about what we're expecting. This applies to both script types.

```suggestion
                    "Windows post-install scripts must be .bat files."
```

mhsmith:
> I’ve simplified the script itself to “always a batch script”, because as far as I can make out, powershell won’t be an option. On all the Windows machines I have access to, I get errors if I try to run powershell scripts - and the errors are “the admin hasn’t allowed PS1 scripts”. 

OK, makes sense.

> If that’s even close to being a default config for Windows, it’s going to cause major problems for installers - because if the post-install script fails, that seems to be a trigger for borking your entire Windows install.

The constructor post-install script will be complex, and there are many things that could go wrong with it. It should at least be possible for the user to recover by doing an uninstall.

> I’ve ommitted anything to do with logging or environment variables. Command line arguments are the only viable entry points I can see

Is that because WiX doesn't provide any way to set environment variables? `--option=value` is not going to be easy to parse in a .bat file. We previously talked about launching the script via a "bootstrap" .bat file, maybe programmatically-generated, which converts its arguments into environment variables and then launches the user's script.

> and we can’t manage logging at the level of Wix

You mean, there's no way to see the script's stdout and stderr?

> so everythign is going to be in the territory of “what gets executed” - at which point, why not leave that in the hands of the person using the install script.

I'm not really sure what this means.

> The option configuration itself is also simpler - We need an ID and a description and a default; but title and command line flag aren’t really options we can do anything with (at least, not in the MSI context).

Agreed that the flag isn't relevant here.

The reason for distinguishing "title" and "description" was to support a UI like this, where the title is one line and the description is a more detailed paragraph:

Screenshot 2025-10-05 at 17 03 46

WiX also supports something it calls CustomizeDlg or "feature tree", which is more scalable to large numbers of options:

Screenshot 2025-10-05 at 16 55 25

But I guess for this PR we can just implement the title using check boxes, and leave the description for later.

> “Enabled” is also a bit of a hairy one - I’m not entirely sure I follow the use case for that one; and it’s going to be very difficult (afaict) to do anything that dynamic during the install process on anything other than “install for all users/admin”.

Yes, let's not worry about that just now.

mhsmith:
> ```
> [tool.briefcase.app.scripttest.install_option.foo]
> description = "Foo the thing"
> default = true
> 
> [tool.briefcase.app.scripttest.install_option.bar]
> description = "Bar the thing"
> default = false
> ```

I see that the option order is preserved as a result of Python dicts being ordered. But this is explicitly [not guaranteed by TOML itself](https://toml.io/en/v1.0.0#table), which could cause problems if generating a pyproject.toml file with a non-Python tool. So I'd be more comfortable with making this an array, as in https://github.com/beeware/briefcase/issues/2191#issuecomment-2719503147.

freakboy3742:
> > If that’s even close to being a default config for Windows, it’s going to cause major problems for installers - because if the post-install script fails, that seems to be a trigger for borking your entire Windows install.
> 
> The constructor post-install script will be complex, and there are many things that could go wrong with it. It should at least be possible for the user to recover by doing an uninstall.

Agreed - but we are somewhat limited by our tools here. MSI doesn't give very robust options for error recovery. 

The worse outcome I've hit is when software becomes uninstallable. From my testing today, the worst offender seems to be when there's something wrong with the WIX script itself (so - the command that WIX says to execute is corrupted in some way), or when the *pre-uninstall* script fails for some reason. 

The former is reasonably easy to avoid in practice - it's common during development, but shouldn't be in production. Failures in the post-install script raise an error, but result in "successful" install from the perspective of the registry/MSI database, AFAICT. But a failure in the pre-uninstall script means there's no way to get past the pre-uninstall script and actually do the uninstall. But I also don't see any good ways to catch that error or work around it.
 
> > I’ve ommitted anything to do with logging or environment variables. Command line arguments are the only viable entry points I can see
> 
> Is that because WiX doesn't provide any way to set environment variables? `--option=value` is not going to be easy to parse in a .bat file. We previously talked about launching the script via a "bootstrap" .bat file, maybe programmatically-generated, which converts its arguments into environment variables and then launches the user's script.

It's not as difficult as you might think, especially when you know the format the arguments will be presented in.

<details>
  <summary>Example arg parsing</summary>

```
@echo off
setlocal enabledelayedexpansion
echo CUSTOM POST INSTALL
echo ARGS %*

SET ALL_USERS=0
SET FOO=0
SET BAR=0
SET WHIZ=0
SET BANG=0

:parse_args
if "%~1"=="" goto :done_parsing
SET "arg=%~1"

if "!arg:~0,2!"=="--" (
    SET prev="!arg:~2!"
) else (
    if !prev!=="allusers" (
        SET ALL_USERS=!arg!
    ) else if !prev!=="foo" (
        SET FOO=!arg!
    ) else if !prev!=="bar" (
        SET BAR=!arg!
    ) else if !prev!=="whiz" (
        SET WHIZ=!arg!
    ) else if !prev!=="bang" (
        SET BANG=!arg!
    ) else (
        echo Unknown flag !prev!
    )
    SET prev=""
)

shift
goto :parse_args
:done_parsing

echo ALL_USERS=%ALL_USERS%
echo FOO=%FOO%
echo BAR=%BAR%
echo WHIZ=%WHIZ%
echo BANG=%BANG%

```
 
</details>

> > and we can’t manage logging at the level of Wix
> 
> You mean, there's no way to see the script's stdout and stderr?

It displays a console window, so you see the console output - but WIX doesn't provide a way to do the equivalent of a unix `tee` on the executing script. That's part of the problem with diagnosing the problems with the script - the MSI log doesn't include any of the output of the script that was executed.
 
> > so everythign is going to be in the territory of “what gets executed” - at which point, why not leave that in the hands of the person using the install script.
> 
> I'm not really sure what this means.

I'm suggesting that the approach here essentially needs to be that the post_install script needs to take responsibility for logging. 
 
> The reason for distinguishing "title" and "description" was to support a UI like this, where the title is one line and the description is a more detailed paragraph:

That makes sense; it poses a problem for layout because we have to do pixel-based positioning of text in the dialog. That may be less of an issue with the other WiX CustomizeDlg - some more investigation may be required.

> But I guess for this PR we can just implement the title using check boxes, and leave the description for later.

Done.

> I see that the option order is preserved as a result of Python dicts being ordered. But this is explicitly [not guaranteed by TOML itself](https://toml.io/en/v1.0.0#table)

Also done - I forgot this wasn't guaranteed by TOML. I've updated the PR description with the new expected format.

freakboy3742:
I've also pushed an update that should enable us to accomodate externally packaged apps.

The LICENSE.rtf is no longer written into the packaged source; and the installer scripts are now written into an `extras/_installer` directory. The `extras` directory is added as an additional source root when building MSI content, in addition to the package path (which is src by default, but will be the external package path if configured as such). The `_installer` par of the name is configurable with the `installer_path` app configuration item; that name needs to *not* exist in the packaged source. `_installer` is a relatively safe name in that regard; if there is a collision, Wix will raise an error.

src/briefcase/platforms/windows/__init__.py
```diff
@@ -132,6 +171,61 @@ def _cleanup_app_support_package(self, support_path):
 """
         )
 
+    def install_app_resources(self, app: AppConfig):
+        """Install Windows-specific app resources.
```
mhsmith, line 248: This function isn't called when `external_package_path` is set.

src/briefcase/platforms/windows/__init__.py
```diff
@@ -132,6 +171,61 @@ def _cleanup_app_support_package(self, support_path):
...
+                install_scripts_path.mkdir(exist_ok=True, parents=True)
+                self.tools.shutil.copyfile(
+                    pre_uninstall_script,
+                    install_scripts_path / "pre_uninstall.bat",
+                )
+
+        # Convert the license to RTF, then output it into the project.
+        # This assumes the license is pure text; which for now, we can guarantee.
+        # When PEP639 support is added, we will need to adapt.
+        license_content = (self.base_path / "LICENSE").read_text()
```
mhsmith: This hard codes the license file path, which is supposed to be configurable in pyproject.toml.

Although somehow, we seem to have missed documenting this anywhere – this should also be fixed.

freakboy3742: Good point - I forgot that we're currently honouring the PEP621 status of licenses, which allows for a file definition. 

And yes, this should be documented - getting to PEP639 compliance would be a good excuse to do that.

src/briefcase/platforms/windows/__init__.py
freakboy3742: dammit... 

Thinking about the problem though... *should* we be installing app resources for external apps?

Looking purely at the Windows installer - one of the resource is `icon.ico` - which is used as the product icon for the add/remove program entry and the application shortcut. The same will also be true for any document type registrations associated with the app. 

More generally - there will be *some* resources that end up installed into the unused "app" tree for an external app. It doesn't hurt if we install something that ultimately isn't used, but it will if we don't install a resource that is needed for an installer. 

On that basis, I'm going to add an `install_app_resources` step to the external apps creation process.

docs/reference/platforms/windows/index.rst
```diff
@@ -166,6 +182,51 @@ However, if you need to override this default value, you can define
...
+
+.. attribute:: title
+
+A short human-readable label describing the option. Any value provided will be converted
+into a string for display to the user.
+
+.. attribute:: description
+
+A longer description of the purpose of the option. Any value provided will be converted
+into a string for display to the user.
```
mhsmith: We don't generally accept non-strings in string settings, or if we do, it's an implementation detail which we don't promise to maintain. Why would these be any different?

src/briefcase/config.py
```diff
@@ -191,6 +191,67 @@ def validate_document_type_config(document_type_id, document_type):
...
+    install_options = {}
+    known_names = set()
+    if config:
+        for i, config_item in enumerate(config):
+            try:
+                name = config_item["name"]
+            except KeyError:
+                raise BriefcaseConfigError(
+                    f"Install option {i} does not define a `name`."
+                )
```
mhsmith: This will need a `from` for the recently-added Ruff rule.

src/briefcase/config.py
```diff
@@ -191,6 +191,67 @@ def validate_document_type_config(document_type_id, document_type):
         pass
 
 
+def validate_install_options_config(config):
+    """Validate that a install options are valid and complete, and convert into ."""
```
mhsmith: I guess this is supposed to say "into a single dict", in which case it should explain why cookiecutter  makes this necessary.

docs/reference/platforms/windows/index.rst
```diff
@@ -166,6 +182,51 @@ However, if you need to override this default value, you can define
...
+~~~~~~~~~~~~~~~~~
+
+.. currentmodule:: install_option
+
+Windows MSI installers are able to present a panel of optional features to the user
+as part of the installation process. These features are binary flags which can then be
+used by a post-install script to perform additional installation behaviors.
+
+Installer options are defined using a TOML array of tables - each option is in
+group named ``[[ toga.briefcase.app.<app name>.install_option ]]``. Each option
```
mhsmith:
```suggestion
Installer options are defined using a TOML array of tables - each option is in a
group named ``[[ toga.briefcase.app.<app name>.install_option ]]``, which
```

docs/reference/platforms/windows/index.rst
```diff
@@ -166,6 +182,51 @@ However, if you need to override this default value, you can define
...
+A Boolean describing the initial value of the option in the GUI. If not provided,
+defaults to ``False``.
+
+Using installer options
+-----------------------
+
+When an installer option is defined, the value of the option will be passed to the
+post-install script as a command line argument. If your installer defines an option
+with a name of ``foo``, the argument ``--foo=1`` will be passed to your post-install
+script if the option is selected by the user ``--foo=0`` will be passed if the option
```
mhsmith: > It's not as difficult as you might think, especially when you know the format the arguments will be presented in.
>
> Example arg parsing

That's not too bad, but it's still a lot of boilerplate that every script will need to work out for itself, with lots of opportunities to make a mistake. Since the developer is going to want this information in variables anyway, can we save them the effort by setting the variables ourselves? If WiX doesn't support it, we could generate a wrapper script.

docs/reference/platforms/windows/index.rst
```diff
@@ -166,6 +182,51 @@ However, if you need to override this default value, you can define
...
+is *not* selected. The ``--allusers`` option will always be passed to post-install
+scripts; this option describes whether the app has been installed for all users, or
+just for the current user.
```
mhsmith:
```suggestion
scripts; this option will be 1 if the app has been installed for all users, or 0
for just the current user.
```

tests/platforms/windows/app/create/test_create.py
mhsmith, line 1: There's a merge conflict on this file.

docs/reference/platforms/windows/index.rst
freakboy3742: No good reason; it made the error handling slightly easier (in that there wasn't any). I'll make that change.

src/briefcase/config.py
```diff
@@ -191,6 +191,83 @@ def validate_document_type_config(document_type_id, document_type):
...
+    install_options = {}
+    known_names = set()
+    if config:
+        for i, config_item in enumerate(config):
+            try:
+                name = config_item["name"]
+                if not isinstance(name, str):
+                    raise BriefcaseConfigError(
+                        f"Name for install option {i} is not a string."
+                    ) from None
```
mhsmith: `from None` is unnecessary here because it's not inside an `except` block. There are several other instances in this function.

freakboy3742: Yup - got a bit overzealous there. Removed.

docs/reference/platforms/windows/index.rst
freakboy3742: I've just updated the templates to include a wrapper script. The user's scripts will now have environment variables set for each of the options, plus ALLUSERS.

This also potentially gives us an entry point for controlling both non-batch execution, and for capturing logs (since the actual install script will be a subprocess/call)

docs/reference/platforms/windows/index.rst
```diff
@@ -217,13 +217,14 @@ defaults to ``False``.
...
-script if the option is selected by the user ``--foo=0`` will be passed if the option
-is *not* selected. The ``--allusers`` option will always be passed to post-install
-scripts; this option will be 1 if the app has been installed for all users, or 0 if
-it has only been installed for the current user.
+When an installer option is defined, the value of the option will be made available to
+the post-install script as an environment variable whose name is the capitalized version
+of the option name. If your installer defines an option with a name of ``foo``, an
+environment variable of ``FOO`` will be defined, with a value of 1 if the option has
+been selected by the user, and 0 will be defined if the option has not been selected.
+The ``ALLUSERS`` environment variable will also be set; this option will be 1 if the app
```
mhsmith:
```suggestion
been selected by the user, and 0 if the option has not been selected.
The ``ALLUSERS`` environment variable will also be set; its value will be 1 if the app
```

docs/reference/platforms/windows/index.rst
```diff
@@ -217,13 +217,14 @@ defaults to ``False``.
...
-post-install script as a command line argument. If your installer defines an option
-with a name of ``foo``, the argument ``--foo=1`` will be passed to your post-install
-script if the option is selected by the user ``--foo=0`` will be passed if the option
-is *not* selected. The ``--allusers`` option will always be passed to post-install
-scripts; this option will be 1 if the app has been installed for all users, or 0 if
-it has only been installed for the current user.
+When an installer option is defined, the value of the option will be made available to
+the post-install script as an environment variable whose name is the capitalized version
+of the option name. If your installer defines an option with a name of ``foo``, an
+environment variable of ``FOO`` will be defined, with a value of 1 if the option has
```
mhsmith: To prevent clashes with other environment variables, It would be safer to prefix these variables with `OPTION_` in the same way as the MSI properties.

I don't think we should prefix ALLUSERS as OPTION_ALLUSERS as well, because it's a different kind of option that isn't declared in pyproject.toml, and is controlled by the end-user in a different way in both MSI and PKG installers. Also, if we keep all the variable names identical between the post-install script and the msiexec command-line, that would be one less thing to remember.

freakboy3742: I considered both `FOO` and `OPTION_FOO`; I essentially got to near a coin flip, and thought that since the variables are user-specified, they can avoid collisions if they need to. However, I guess that collisions have as much to do with the user's computer as the developers, so anything we can do to make the variables unambiguous is probably a good move.

Agreed about ALLUSERS though - for all the reasons you flagged.

mhsmith (approved):

PR #2498, opened 2025-10-05
Modify Dev Web Command to Support Editable Installs For Pure Python Requirements 

JMah007:
This PR implements editable installation support for the web development command, significantly improving iteration speed by avoiding wheel recompilation on every source code change.

Changes have been made to the way requirements are installed via the `dev web` command. The function for installing requirements overrides the implementation from the base dev. Remote dependencies are installed via pip as per usual whilst local dependencies are now identified and then installed as editable in order to generate .pth files to later symlink pure python modules to the server.

Ref #2334 

## PR Checklist:

### DIFF
--- a/src/briefcase/platforms/web/static.py
+++ b/src/briefcase/platforms/web/static.py
@@ -11,8 +11,10 @@ from briefcase.console import Console
 from briefcase.exceptions import (
     BriefcaseCommandError,
     BriefcaseConfigError,
+    RequirementsInstallError,
     UnsupportedCommandError,
 )
+from briefcase.integrations.virtual_environment import VenvContext
 
 if sys.version_info >= (3, 11):  # pragma: no-cover-if-lt-py311
     import tomllib
@@ -473,6 +475,114 @@ class StaticWebDevCommand(StaticWebMixin, DevCommand):
             default=True,
             help="Run without creating an isolated environment (not recommended for web).",
         )
+
+    def _is_local_dependency(self, req):
+        """Checks if a dependency is local in order to install it as editable."""
+        is_local = False
+
+        clean_req = req.strip()
+
+        # Unix path
+        if (
+            clean_req.startswith("./")
+            or clean_req.startswith("../")
+            or clean_req.startswith("/")
+            or
+            # Windows path
+            clean_req.startswith(".\\")
+            or clean_req.startswith("..\\")
+            or clean_req.startswith("\\")
+            or
+            # Windows absolute C:\ or similar
+            (len(clean_req) > 1 and clean_req[1] == ":")
+        ):
+            if Path(clean_req).exists():
+                is_local = True
+        return is_local
+
+    def install_dev_requirements(self, app: AppConfig, venv: VenvContext, **options):
+        """Install the requirements for the web app dev. Local dependencies are editable
+        installed.
+
+        This will always include test requirements, if specified.
+
+        :param app: The config object for the app
+        :param venv: The context object used to run commands inside the virtual
+            environment.
+        """
+
+        # Install src code as editable
+        venv.run(
+            [sys.executable, "-m", "pip", "install", "-e", str(self.app_path(app))],
+            check=True,
+        )
+
+        requires = app.requires if app.requires else []
+        if app.test_requires:
+            requires.extend(app.test_requires)
+        if not requires:
+            self.console.info("No application requirements")
+            return
+
+        local_requires = []
+        remote_requires = []
+        for req in requires:
+            if self._is_local_dependency(req):
+                local_requires.append(req)
+            else:
+                remote_requires.append(req)
+
+        if remote_requires:
+            with self.console.wait_bar("Installing remote dev requirements..."):
+                try:
+                    venv.run(
+                        [
+                            sys.executable,
+                            "-u",
+                            "-X",
+                            "utf8",
+                            "-m",
+                            "pip",
+                            "install",
+                            "--upgrade",
+                            *(["-vv"] if self.console.is_deep_debug else []),
+                            *remote_requires,
+                            *app.requirement_installer_args,
+                        ],
+                        check=True,
+                        encoding="UTF-8",
+                    )
+                except subprocess.CalledProcessError as e:
+                    raise RequirementsInstallError() from e
+
+        if local_requires:
+            with self.console.wait_bar(
+                "Installing local dev requirements as editable..."
+            ):
+                for req in local_requires:
+                    clean_req = req.strip()
+                    try:
+                        venv.run(
+                            [
+                                sys.executable,
+                                "-u",
+                                "-X",
+                                "utf8",
+                                "-m",
+                                "pip",
+                                "install",
+                                "-e",
+                                clean_req,
+                                *(["-vv"] if self.console.is_deep_debug else []),
+                                *app.requirement_installer_args,
+                            ],
+                            check=True,
+                            encoding="UTF-8",
+                        )
+                    except subprocess.CalledProcessError as e:
+                        raise RequirementsInstallError() from e
+
+        return
 
     def run_dev_app(self, app: AppConfig, **options):
         raise UnsupportedCommandError(


JMah007:
@mEp3ii2 Heres my first attempt at the solution. Let me know if you spot anything unusual.

JMah007:
@freakboy3742 May I have a review and comments on how close this approach is to the ideal solution.

src/briefcase/platforms/web/static.py
```diff
@@ -474,6 +476,114 @@ def add_options(self, parser):
...
+        :param app: The config object for the app
+        :param venv: The context object used to run commands inside the virtual
+            environment.
+        """
+
+        # Install src code as editable
+        venv.run(
+            [sys.executable, "-m", "pip", "install", "-e", str(self.app_path(app))],
+            check=True,
+        )
```
freakboy3742: Is this actually needed? The app code should be included by virtue of being on the PYTHONPATH - is there a need to `pip install` it?

If it is, it isn't really a "dev requirements" thing - it should probably be part of an abstracted "install app code" step (which may well be a no-op on the base dev command - or might wrap the `dist-info` creation step).=

src/briefcase/platforms/web/static.py
```diff
@@ -474,6 +476,114 @@ def add_options(self, parser):
             help="Run without creating an isolated environment (not recommended for web).",
         )
 
+    def _is_local_dependency(self, req):
```
freakboy3742: Pretty sure this will be matching the same thing a `_is_local_path()` used by the create command.

freakboy3742 (changes requested):
A couple of notes inline; but I think the broad approach you've taken here makes sense.

The question I'd have is whether this should be promoted to the general `dev` command, not just `dev web`. `run` installs can't be editable - but `install_dev_requirements` is isolated to the dev command, so I can't think of any downside to defaulting to an editable install on all `dev` commands.

JMah007:
@freakboy3742 Ive made those changes. Please give them a review when you can.

src/briefcase/commands/dev.py
```diff
@@ -75,8 +76,23 @@ def add_options(self, parser):
             help="Run the app in test mode",
         )
 
+    def install_app_code(
+        self, app: AppConfig, venv: VenvContext, dist_info_path: str, **options
```
freakboy3742: One-line-per-arg format, please

src/briefcase/commands/dev.py
```diff
@@ -92,27 +108,65 @@ def install_dev_requirements(self, app: AppConfig, venv: VenvContext, **options)
...
+                        encoding="UTF-8",
+                    )
+                except subprocess.CalledProcessError as e:
+                    raise RequirementsInstallError() from e
+
+        if local_requires:
+            with self.console.wait_bar(
+                "Installing local dev requirements as editable..."
+            ):
+                for req in local_requires:
```
freakboy3742: I can see what you're doing here, but it's going to have some interesting side effects.

Firstly, each subprocess call take time to spin up, so this is going to be a little slower than a single call that installs all dependencies in one hit.

However, more importantly - there is actually a functional difference between `pip install a b` and `pip install a; pip install b`. Part of the install process is a dependency resolver, where it finds the downstream dependencies of `a` and `b`, and ensures that they're all mutually compatible. The resolver math only operates fully (and consistently) when run as a single install pass, because otherwise it doesn't have the full context of everything that needs to be mutually compatible.

So - the installation of dependencies needs to remain a single `pip install` call, rather than 2+ calls. 
What is needed here is a single call of `pip install a b -e ../path/to/c -e ../path/to/d`.

src/briefcase/platforms/web/static.py
freakboy3742: You've done the refactoring as requested here, but the original question hasn't been addressed - have you confirmed this is actually needed at all?

src/briefcase/commands/dev.py
```diff
@@ -75,8 +76,23 @@ def add_options(self, parser):
...
+        self, app: AppConfig, venv: VenvContext, dist_info_path: str, **options
+    ):
+        """This method creates dist-info metadata so Python recognizes the app as an
+        installed package without requiring pip install. For desktop platforms, source
+        code is accessible via PYTHONPATH at runtime.
+
+        :param app: The config object for the app
+        :param venv: The context object used to run commands inside the virtual
+            environment.
+        :param dist_info_path: The path to the .dist-info directory for the app
```
freakboy3742: dist_info_path shouldn't be needed as a separate argument; it can be entirely derived from `app`. This likely indicates that the evaluation of `dist_info_path` needs to be turned into a property.

freakboy3742 (changes requested):
Some good refactoring here, but there's a couple of edge cases to consider (and one outstanding question).

src/briefcase/commands/dev.py
JMah007: Is it tied close enough to a specific app that it needs to be made into a property or should I just declare the dist_info_path in the method before writing the file instead of passing it in? I get that the former is more clean.

freakboy3742: It's very much an app specific thing - you give me the app config, and I can tell you the name (and content) of the dist_info file.

src/briefcase/platforms/web/static.py
JMah007: Looking at it again i realized its not needed as the src code is made available to web via PYTHONPATH as well.

JMah007:
@freakboy3742 I have now removed install_app_code() function as the source code is already made available via PYTHONPATH for both base and static dev. I am just not sure now if the dist info file still needs to be written as the src code isnt actually installed via pip anymore?

Apart from that its ready for another review

src/briefcase/commands/dev.py
```diff
@@ -92,27 +94,36 @@ def install_dev_requirements(self, app: AppConfig, venv: VenvContext, **options)
...
-            except subprocess.CalledProcessError as e:
-                raise RequirementsInstallError() from e
+        sorted_requires = []
+        for req in requires:
+            if _is_local_path(req):
+                sorted_requires.extend(["-e", req.strip()])
+            else:
+                sorted_requires.append(req)
+
+        if sorted_requires:
```
freakboy3742: Is this `if` possible? if requires exists, surely `sorted_requires` exists?

src/briefcase/commands/dev.py
```diff
@@ -92,27 +94,36 @@ def install_dev_requirements(self, app: AppConfig, venv: VenvContext, **options)
...
-                    ],
-                    check=True,
-                    encoding="UTF-8",
-                )
-            except subprocess.CalledProcessError as e:
-                raise RequirementsInstallError() from e
+        sorted_requires = []
+        for req in requires:
+            if _is_local_path(req):
+                sorted_requires.extend(["-e", req.strip()])
```
freakboy3742: Why `strip`? The value as provided should be correct; it's a quoted value, and any input should be used as-is.

JMah007:
@freakboy3742 I've fixed those 2 issues. Initially I added strip() thinking the user might enter dependencies in app.requires with trailing spaces by mistake.

src/briefcase/commands/dev.py
```diff
@@ -105,14 +114,15 @@ def install_dev_requirements(self, app: AppConfig, venv: VenvContext, **options)
...
-                        *requires,
+                        *sorted_requires,
                         *app.requirement_installer_args,
                     ],
                     check=True,
                     encoding="UTF-8",
                 )
             except subprocess.CalledProcessError as e:
                 raise RequirementsInstallError() from e
+        return
```
freakboy3742: There's no need for a bare return 
```suggestion
```

JMah007:
@freakboy3742 tests ready for a review

tests/commands/dev/test_install_dev_requirements.py
```diff
@@ -14,7 +14,14 @@ def test_install_requirements_no_error(dev_command, first_app, logging_level):
     # Configure logging level
     dev_command.console.verbosity = logging_level
 
-    first_app.requires = ["package-one", "package_two", "packagethree"]
+    first_app.requires = [
```
freakboy3742, line 17: This case would be worth splitting into two tests:

1. The basic "inclusion" test - does `install_dev_requirments` prepend `-e` *at all*?
2. A more comprehensive "does install_dev_requirements work for all known cases" - including:
  - relative paths of multiple depths
  - absolute paths
  - anything else on [this page of `pip`'s documentation](https://pip.pypa.io/en/stable/cli/pip_install/#examples) that is potentially "interesting" from a parsing perspective.

JMah007: All fixed

tests/commands/dev/test_install_dev_requirements.py
```diff
@@ -31,7 +35,80 @@ def test_install_requirements_no_error(dev_command, first_app, logging_level):
...
+        "src/mypackage",
+        # Absolute paths
+        "/absolute/path",
+        # Windows paths
+        "folder\\windows",
+        ".\\windows\\current",
+        "..\\windows\\parent",
+        "C:\\absolute\\windows",
+        # One regular package as well
+        "package1",
```
freakboy3742: What about SCM-based installs? I said you don't have to test hg and svn, but you should still be testing *an* example of an SCM install (in various legal forms). 

Also - tarball installs are legal as well.

I mention these two specifically because (a) they're on the pip documentation page I mentioned, and (b) they're edge cases. Tarballs can't be installed editable, but they *are* installable; VCS repos *can* be installed editable, but the URL format is slightly different.

JMah007: Im a little confused here. So SCM requirements can be both local and remote but briefcase currently doesnt append -e to both local and remote SCM based requiremnts. 

As for Tarball local requirements they will have -e appended to them even though pip doesnt support editable installs for tarball packages. 

Do i just adjust the tests to expect that all SCM based installs dont have -e appended to them and for tarball do i test local req should have -e appended still whilst remote req shouldnt?


Am just testing to make sure it appends -e to what appears as "local" disregarding other rules?

freakboy3742: What I'm flagging here is that a comprehensive test would reveal a problem in the current implementation.

Tarballs *can't* have `-e` prepended, because that's not legal; and SCM URLs need a different form if they're going to be installed `-e` (They need the `#egg` handling suffix). However, right now, I think both of these would come through as editable.

So - we need a test that verifies that a tarball or URL requirement is installed non-editable. It's only pure filesystem paths that can be installed editable. And we need to fix the implementation so that these test cases pass.

In future, it might be possible to add support for editable SCM URLs - but that's going to need a lot more work, rewriting the URL format.

JMah007: So should i be fixing the implementation by creating a temporary new function that will install SCM and tarballs as non editable just to make the tests pass for now?

freakboy3742: Yes, we should be fixing the implementation - but your comment about a "temporary new function" doesn't make any sense to me.

I'm saying, for right now, the correct behavior is to only install editable if the requirement is a local path, and *not* a tarball or zip file. Any SCM reference should *not* be installed editable. The test suite should verify all these cases (and interesting variants of these cases); the implementation should then pass with that test suite.

I doubt any additional methods are needed; but the `_is_local_path()` check will need to be augmented with additional "and not req.endswith('.tar.gz') ... (etc)" logic. It should already (AFAICT) be identifying the difference between a URL and a local path.

In future, we might add logic to make URL installs editable; but that can be safely tackled as future work.

JMah007:
@freakboy3742 I applied the fixes but I cant run tox for coverage report as I keep getting import errors. Basically I added extra parameters for testing is_local_path() and dummy requirements with the different extensions test_install_requirements_path_formats()

freakboy3742:
> @freakboy3742 I applied the fixes but I cant run tox for coverage report as I keep getting import errors. Basically I added extra parameters for testing is_local_path() and dummy requirements with the different extensions test_install_requirements_path_formats()

What error are you seeing? The only thing I can think that might be an issue is a stale Tox environment - you might find adding `-r` to your tox command will help - that tells tox to rebuild the environment, which might flush out a problem.

If that doesn't fix it, let me know what you're running (and where), and the error you're seeing and I might be able to suggest a fix.

JMah007:
> > @freakboy3742 I applied the fixes but I cant run tox for coverage report as I keep getting import errors. Basically I added extra parameters for testing is_local_path() and dummy requirements with the different extensions test_install_requirements_path_formats()
> 
> What error are you seeing? The only thing I can think that might be an issue is a stale Tox environment - you might find adding `-r` to your tox command will help - that tells tox to rebuild the environment, which might flush out a problem.
> 
> If that doesn't fix it, let me know what you're running (and where), and the error you're seeing and I might be able to suggest a fix.

I tried this but it did not work. The error says cannot import RCEdit from integrations file

freakboy3742:
> I tried this but it did not work. The error says cannot import RCEdit from integrations file

Can't import it, or can't run it? If it can't import, then there is something very weird going on. If it's unable to *run* it, that's a different problem entirely. The failure to *run* could be the sort of thing you'd see if you're running the tests on Windows, and the system was unable to download or execute the binary due to OS-level security issues. Are you able to share a full stack trace or error log?

JMah007:
> > I tried this but it did not work. The error says cannot import RCEdit from integrations file
> 
> Can't import it, or can't run it? If it can't import, then there is something very weird going on. If it's unable to _run_ it, that's a different problem entirely. The failure to _run_ could be the sort of thing you'd see if you're running the tests on Windows, and the system was unable to download or execute the binary due to OS-level security issues. Are you able to share a full stack trace or error log?


======================================================= short test summary info ======================================================== 
ERROR tests/integrations/rcedit - ImportError: cannot import name 'RCEdit' from 'briefcase.integrations.rcedit' (C:\Users\jaede\OneDrive\Desktop\DeliverableBriefcase\ImplementTestsBriefcase\briefcase\.tox\py\Lib\site-packages\briefcase\integrations\rcedit.py)
ERROR tests/platforms/windows/app/test_build.py
ERROR tests/platforms/windows/app/test_create.py
ERROR tests/platforms/windows/app/test_dev.py
ERROR tests/platforms/windows/app/test_mixin.py
ERROR tests/platforms/windows/app/test_open.py
ERROR tests/platforms/windows/app/test_package.py
ERROR tests/platforms/windows/app/test_run.py
ERROR tests/platforms/windows/visualstudio/test_create.py
ERROR tests/test_cmdline.py



I am running `tox -e py`

freakboy3742:
That's definitely an odd one. It looks like the entire Tox environment is corrupted. It's clearly not a problem in the test environment - if `tox -r -e py` isn't working for you, it might be worth completely deleting the `.tox` directory and trying again. Another thing to try is to see if anyone else in your team is able to reproduce the problem.

JMah007:
> I've cleaned up the tests a little, so now they pass on Windows and Unix; I've also slightly modified the impact on the create command - `is_local_file()` needs to _just_ check for local file status, because a local file reference to a tarball _is_ valid for the purposes of a normal app install.
> 
> Otherwise - this is great, and will be a significant improvement to Briefcase workflows for those using it for development purposes (especially for Toga).

Thank you for correcting that. Glad to have helped.

PR #2502, opened 2025-10-08
WIP: Move Briefcase to MkDocs

kattni:
Getting this going for testing purposes.

## PR Checklist:
 

kattni:
@freakboy3742 This is ready for review.

Notes:
* Please review the content on the various index pages, as it's primarily new content produced by me, and you may have Thoughts.

kattni:
The tables are terrible, yes. I agree the Toga updates would be useful here.

Everything else should be good to go.

docs/en/about/faq.md
```diff
@@ -0,0 +1,40 @@
+# Frequently Asked Questions
+
+## What version of Python does Briefcase support?
+
+Python 3.10 or higher.
```
mhsmith, line 5: For some reason this is rendering with a bigger gap than the other questions:

Screenshot 2025-10-23 at 11 36 49

docs/en/reference/platforms/android/gradle.md
```diff
@@ -0,0 +1,323 @@
...
+<col style="width: 5%" />
+<col style="width: 5%" />
+<col style="width: 7%" />
+<col style="width: 11%" />
+<col style="width: 7%" />
+<col style="width: 10%" />
+</colgroup>
+<thead>
+<tr>
+<th colspan="11"><a href="/reference/platforms/#platform-support-key">Host Platform Support</a></th>
```
mhsmith: Here, and in all the similar tables, the link is broken because it points at the root of the domain. I assume this would also affect a live deployment because of the `/latest` or `/stable` in the URL.

docs/en/about/faq.md
kattni: This is a side-effect of the paragraph marker. When the title is exactly the width of the content, the paragraph marker renders below it. If you hover over it, you should see the paragraph marker appear. I have in other cases suppressed the marker, and I can dig up what I did if you would like me to do that here, however, there are always going to be instances of this. Alternatively, I can bump the FAQ questions to H3, and that should shrink it down enough to eliminate the issue here.

docs/en/reference/platforms/android/gradle.md
kattni: This has been resolved.

As a side note, the tables will be redone eventually with a much better setup. This is a temporary solution.

docs/en/about/faq.md
mhsmith: Changing the font size would fix the problem in some places and create it in others. I tried various ways to fix it in CSS, but the question mark will always be treated as a line break opportunity unless you disable wrapping entirely. 

So I guess any general solution would have to come from MkDocs. But since this is the top of a fairly prominent page, how about we work around it by making the title slightly shorter, like "What versions of Python are supported?"

kattni: Done.

I suppressed the paragraph marker, though I can't recall how I did it. It wasn't an MkDocs solution, it was CSS or HTML. Again, it's not ideal.

docs/en/reference/platforms/android/gradle.md
mhsmith: Thanks, the links work now on the sub-pages like app.md, but they go one level too far up in the index.md pages.

kattni: Good catch. Everything works locally regardless, so I'm having to guess here and rely on the RTD build. I tested all of the links this time on the RTD build, and they appear to be working.

mhsmith:
Thanks, that all looks good, but unfortunately there's one last conflict from #2496 that will need to be resolved.

freakboy3742:
> Thanks, that all looks good, but unfortunately there's one last conflict from #2496 that will need to be resolved.

I've just merged and fixed that one.

PR #2514, opened 2025-10-18
briefcase dev: Don't overwrite existing environment variables

### LINKED ISSUE #2497: Add ability to disable PYTHONDEVMODE in dev command.

When then `dev` command runs, it enforced PYTHONDEVMODE=1 in the environment. This is useful for debugging purposes, but there are known issues with PYTHONDEVMODE and some binary packages - e.g:

* https://github.com/docker/compose/issues/8176
* https://github.com/SAML-Toolkits/python3-saml/issues/360

There should be a way to *not* enable PYTHONDEVMODE when running `briefcase dev`.

`briefcase dev --no-py-dev-mode` should run dev mode exactly the same as usual, but *without* setting PYTHONDEVMODE=1 in the environment.

## Do nothing. 
 
PYTHONDEVMODE *should* work everywhere; it's a bug when it doesn't. However, pragmatically, we need to do *something*.

## Use the developer's environment.

Instead of explicitly setting PYTHONDEVMODE, ensure that if the user defines PYTHONDEVMODE, the setting is preserved when dev mode runs. However, PYTHONDEVMODE is useful for catching some errors (especially asyncio errors), and many developers won't know it exists.

## Richer Python environment options

Provide a richer set of options for the whole of the Python interpreter. The Python interpreter has a bunch of configurability; we could expose all those options. However, pragmatically, dev mode should be aiming to emulate the environment that will be seen at runtime. PYTHONDEVMODE is a debugging addition; it's not seen at runtime, but useful in "dev" conditions. Full configurability of the dev environment isn't a realistic representation of what will happen with packaged apps (e.g., packaged apps will *always* be UTF-8, so there's no point exposing PYTHONUTF8 as a configurable option).

### Additional context

Most of the changes will be needed in `briefcase/commands/dev.py`:

* Add a new `--no-py-dev-mode` option, storing `dev_mode=True` by default, but storing False when it is provided. This is similar to how `--no-run` is handled now.
* Take PYTHONDEVMODE out of the default environment, and make adding the option conditional on whether the flag is present in the environment.
* Add test (in `tests/commands/dev/test_run_dev_app.py`) that verifies the environment is processed as expected when the flag is provided.

NMertsch:
`briefcase dev`: Only set `PYTHONUNBUFFERED=1`, `PYTHONDEVMODE=1`, and `PYTHONUTF8=1` if these variables are not already set.

#2497 is only about `PYTHONDEVMODE`, but I think this generic implementation is less surprising to users. I might lack context, though.

(I fixed two very minor typos along the way, but don't think this is worth a towncrier fragment.)

Fixes #2497 

## PR Checklist:

### DIFF
--- a/changes/2520.misc.1.rst
+++ b/changes/2520.misc.1.rst
@@ -1,0 +1,1 @@
+The `RUF ruff rule set <https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf>`_ was enabled and all findings were fixed.
--- a/changes/2520.misc.3.rst
+++ b/changes/2520.misc.3.rst
@@ -1,0 +1,1 @@
+The `PT ruff rule set <https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt>`_ was enabled and all findings were fixed.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -240,6 +240,8 @@ extend-select = [
     "ASYNC",  # flake8-async
     "C4",     # flake8-comprehensions
     "I",      # isort
+    "RUF",    # ruff-specific rules
+    "PT",     # flake8-pytest-style
     # The SIM rules are *very* opinionated, and don't necessarily make for better code.
     # They may be worth occasionally turning on just to see if something could actually
     # use improvement.
--- a/src/briefcase/bootstraps/base.py
+++ b/src/briefcase/bootstraps/base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Collection
 from pathlib import Path
 from typing import Any, TypedDict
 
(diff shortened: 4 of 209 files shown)


freakboy3742 (changes requested):
Thanks - this looks great. The only thing it's missing is documentation of the new feature. A new section on the [command reference for `dev`](https://briefcase.readthedocs.io/en/stable/reference/commands/dev.html) would make sense to me.

NMertsch:
> Thanks - this looks great. The only thing it's missing is documentation of the new feature. A new section on the [command reference for dev](https://briefcase.readthedocs.io/en/stable/reference/commands/dev.html) would make sense to me.

Thank you. I added the section in commit c97b096f20296f82fed8648223c7b168480f9b91, deployed by CI [here](https://briefcase--2514.org.readthedocs.build/en/2514/reference/commands/dev.html#environment-variables) (once flathub decides to work again). Please let me know if you want any changes (or add them yourself if that feels easier).

freakboy3742 (approved):
Thanks for that - this looks great. Thanks for the PR!

PR #2516, opened 2025-10-18
Resolve SIMXXX

### LINKED ISSUE #2384: Audit `SIMXXX` Ruff errors to resolve desired issues

NMertsch:
Handle [flake8-simplify ruff rules](https://docs.astral.sh/ruff/rules/#flake8-simplify-sim).

This PR build on the stale PR #2475 by @cgkoutzigiannis. I added the following:
* Rebase to `main`
* 670feba6: Fix failing test (close file before moving it)
* e29472d8, e29472d8: Use strings instead of lists for cmdline test inputs ([requested](https://github.com/beeware/briefcase/pull/2475#discussion_r2346407120) by @mhsmith))
* a3abfff8: Activate `SIM` rules, ignore [SIM108](https://docs.astral.sh/ruff/rules/if-else-block-instead-of-if-exp/) (enforce ternary operator, too opinionated)
* Handle remaining SIM rules:
    * b7f96a13: Refactor nested `if`'s ([SIM102](https://docs.astral.sh/ruff/rules/collapsible-if/))
    * 48cfe755: Refactor `return True/False` pattern ([SIM103](https://docs.astral.sh/ruff/rules/needless-bool/))
    * 081c7552: Use `contextlib.suppress` ([SIM105](https://docs.astral.sh/ruff/rules/suppressible-exception/), reverted in 59b8b53614c9d8a1bf46c5dc899dcf0120a11ee9)
    * edbb8639: Use `open` with context manager ([SIM115](https://docs.astral.sh/ruff/rules/open-file-with-context-handler/))

Fixes #2384.
Closes #2475. 

## PR Checklist:

- [x] I have read the **CONTRIBUTING.md** file (see #2475 for @cgkoutzigiannis's confirmation)
- [x] I will abide by the code of conduct (see #2475 for @cgkoutzigiannis's confirmation)

### DIFF
--- a/src/briefcase/integrations/file.py
+++ b/src/briefcase/integrations/file.py
@@ -288,7 +288,7 @@ class File(Tool):
         :param response: ``httpx.Response``
         :param filename: full filesystem path to save data
         """
-        # `temp_file` is used in the `finally` block, so make sure it's assigned 
+        # `temp_file` is used in the `finally` block, so make sure it's assigned
         # before the `try`.
         temp_file = tempfile.NamedTemporaryFile(  # noqa: SIM115 (use context manager)
             dir=filename.parent,


NMertsch:
I don't think the CI failure is caused by this PR (`HTTPSConnectionPool(host='flathub.org', port=443): Read timed out`).

src/briefcase/integrations/file.py
```diff
@@ -288,13 +288,13 @@ def _fetch_and_write_content(self, response: httpx.Response, filename: Path):
         :param response: ``httpx.Response``
         :param filename: full filesystem path to save data
         """
-        with tempfile.NamedTemporaryFile(
-            dir=filename.parent,
-            prefix=f"{filename.name}.",
-            suffix=".download",
-            delete=False,
-        ) as temp_file:
-            try:
```
mhsmith: If `NamedTemporaryFile` fails with an `OSError`, then `temp_file` will not be assigned, and the `finally` block will chain an unhelpful `NameError` onto the `OSError`. 

The simplest solution is probably to go back to the original version without the `with`, and leave a `noqa` comment with an explanation.

mhsmith:
Please don't force push a PR after it's been reviewed, as this makes the review history difficult to follow. If you need to update from the main branch, do a merge.

src/briefcase/integrations/file.py
NMertsch: Thank you, I didn't consider this case.

Sorry for messing up the commit history. I fixed this in 865ebd2e. I hope the explanation I added makes sense:

```diff
-         temp_file = tempfile.NamedTemporaryFile(
+         # Don't use a context manager to separate operations and related errors:
+         # Create, open, write, close, move, delete.
+         temp_file = tempfile.NamedTemporaryFile(  # noqa: SIM115 (use context manager)
```

NMertsch:
> Please don't force push a PR after it's been reviewed, as this makes the review history difficult to follow. If you need to update from the main branch, do a merge.

Thank you, that makes total sense. I didn't consider this perspective.

> if you could roll back the SIM105 change set, I'd be happy to merge this.

Done.

src/briefcase/integrations/file.py
```diff
@@ -288,13 +288,16 @@ def _fetch_and_write_content(self, response: httpx.Response, filename: Path):
         :param response: ``httpx.Response``
         :param filename: full filesystem path to save data
         """
+        # Don't use a context manager to separate operations and related errors:
+        # Create, open, write, close, move, delete.
```
mhsmith:
```suggestion
        # `temp_file` is used in the `finally` block, so make sure it's assigned 
        # before the `try`.
```

mhsmith:
Merge conflicts with your other PRs will also need to be resolved now.

PR #2520, opened 2025-10-19
Enable RUF and PT ruff rules

NMertsch:
Enable the [RUF](https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf) and [PT](https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt) ruff rule sets and handle all findings.

These changes were not requested. If you don't want them, please just reject the PR. If you want only a subset, or all changes spread over multiple PRs, please let me know.

Here are all code changes:

RUF - ruff-specific rules
* 1595e70a: [RUF001](https://docs.astral.sh/ruff/rules/ambiguous-unicode-character-string/) - Label ambiguous non-ASCII characters in code
* d35c8b38: [RUF003](https://docs.astral.sh/ruff/rules/ambiguous-unicode-character-comment/) - Normalize or label ambiguous non-ASCII characters in comments
* 88afbc93: [RUF005](https://docs.astral.sh/ruff/rules/collection-literal-concatenation/) - Replace collection concatenations with unpacking
* fad6a61f: [RUF012](https://docs.astral.sh/ruff/rules/mutable-class-default/) - Annotate mutable class attributes as immutable (e.g. `dict` as `Mapping`)
* ec3a2d65: [RUF013](https://docs.astral.sh/ruff/rules/implicit-optional/) - Replace `x: str = None` with `x: str | None = None`
* 90c6c331: [RUF015](https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/) - Replace `list(...)[0]` with `iter(next(...))` 
* 2b890eaf: [RUF019](https://docs.astral.sh/ruff/rules/unnecessary-key-check/) - Simplify "dict key is missing or empty" checks
* 08f6b7c1: [RUF043](https://docs.astral.sh/ruff/rules/pytest-raises-ambiguous-pattern/) - Escape `.` in Regex patterns
* 5aacfa87: [RUF059](https://docs.astral.sh/ruff/rules/unused-unpacked-variable/) - Prefix unused unpacked variables with underscores
* d7ff87f6: [RUF100](https://docs.astral.sh/ruff/rules/unused-noqa/) - Remove two unused `noqa` directives

PT - flake8-pytest-style
* 296fa2a4: [PT001](https://docs.astral.sh/ruff/rules/pytest-fixture-incorrect-parentheses-style/) - Remove empty parentheses on `pytest.fixture` calls
* 29da6fc0: [PT006](https://docs.astral.sh/ruff/rules/pytest-parametrize-names-wrong-type/) - Use tuple instead of str for `pytest.mark.parametrize` parameter list
* a7add3aa: [PT007](https://docs.astral.sh/ruff/rules/pytest-parametrize-values-wrong-type/) - Use list instead of tuple for `pytest.mark.parametrize` cases
* e1661bd2: [PT011](https://docs.astral.sh/ruff/rules/pytest-raises-too-broad/) - Check for error messages in `pytest.raises()`
* 563d61c6: [PT012](https://docs.astral.sh/ruff/rules/pytest-raises-with-multiple-statements/) - Simplify `pytest.raises()` blocks
* 574a4fb4: [PT014](https://docs.astral.sh/ruff/rules/pytest-duplicate-parametrize-test-cases/) - Remove duplicate test cases in `pytest.mark.parametrize`
* 1479207f: [PT018](https://docs.astral.sh/ruff/rules/pytest-composite-assertion/) - Split `assert ... and ...` statements
* be79a8cd: [PT022](https://docs.astral.sh/ruff/rules/pytest-useless-yield-fixture/) - Replace `yield` in pytest fixture without teardown with `return`

## PR Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
       - id: docformatter
         args: [--in-place, --black]
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.0
+    rev: v0.14.1
     hooks:
       - id: ruff-format
       - id: ruff-check
--- a/changes/2381.misc.rst
+++ b/changes/2381.misc.rst
@@ -1,0 +1,1 @@
+Resolved ruff rule ``B904`` - raise-without-from-inside-except.
--- a/changes/2517.misc.rst
+++ b/changes/2517.misc.rst
@@ -1,0 +1,1 @@
+Updated setuptools-scm from 9.2.0 to 9.2.1.
--- a/changes/2518.misc.rst
+++ b/changes/2518.misc.rst
@@ -1,0 +1,1 @@
+Updated httpx-retries from 0.4.3 to 0.4.5.
--- a/changes/2519.misc.rst
+++ b/changes/2519.misc.rst
@@ -1,0 +1,1 @@
+Updated coverage[toml] from 7.10.7 to 7.11.0.
--- a/changes/2521.misc.rst
+++ b/changes/2521.misc.rst
@@ -1,0 +1,1 @@
+The ``pre-commit`` hook for ``ruff-pre-commit`` was updated to its latest version.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,13 +95,13 @@ dependencies = [
 # Extras used by developers *of* briefcase are pinned to specific versions to
 # ensure environment consistency.
 dev = [
-    "coverage[toml] == 7.10.7",
+    "coverage[toml] == 7.11.0",
     "coverage-conditional-plugin == 0.9.0",
-    "httpx-retries == 0.4.3",
+    "httpx-retries == 0.4.5",
     "pre-commit == 4.3.0",
     "pytest == 8.4.2",
     "pytest-xdist == 3.8.0",
-    "setuptools_scm == 9.2.0",
+    "setuptools_scm == 9.2.1",
     "tox == 4.31.0",
 ]
 # Docs are always built on a specific Python version; see RTD and tox config files,
@@ -254,7 +254,6 @@ extend-select = [
 # for future fixes throughout the project.
 ignore = [
     "E501",    # line length
-    "B904",    # use raise ... from ...
 ]
 
 
--- a/src/briefcase/cmdline.py
+++ b/src/briefcase/cmdline.py
@@ -159,7 +159,7 @@ def parse_cmdline(args, console: Console | None = None):
         try:
             platform_module = platforms[platform]
         except KeyError:
-            raise InvalidPlatformError(platform, platforms.keys())
+            raise InvalidPlatformError(platform, platforms.keys()) from None
 
         # If the output format wasn't explicitly specified, check to see
         # Otherwise, extract and use the default output_format for the platform.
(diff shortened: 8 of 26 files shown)


freakboy3742 (changes requested):
This looks good; however, I'd argue against adding the RSE ruleset. I understand the braces are redundant, but I think the bracket less syntax is confusing unless you fully understand the underlying behavior, as it means the raise doesn't explicitly read as an empty instantiation.

As a general rule, multiple smaller PRs are preferable to one big one; but in this case, just dropping the RSE rule (and related changes) is fine (unless undoing the RSE changes is easier to do in new PRs).

src/briefcase/integrations/base.py
```diff
@@ -57,7 +57,7 @@ class Tool(ABC):
 
     name: str
     full_name: str
-    supported_host_os: set[str] = {"Darwin", "Linux", "Windows"}
+    supported_host_os: Collection[str] = {"Darwin", "Linux", "Windows"}
```
rmartin16, line 60: hmm...the usefulness of [RUF012](https://docs.astral.sh/ruff/rules/mutable-class-default/) seems limited in BeeWare's source. Type correctness isn't enforced; so, this only helps if the developer's IDE escalates conflicts.

Specifically here, though, this isn't really an improvement; a `Collection` is just as mutable as a `set`.....maybe if we made these all `frozensets`.

The usefulness is most present for `dict` to `Mapping`, though; at least, the `Mapping` type implies immutability.

rmartin16: Looks like I'm wrong about `Collection`; its interface doesn't include mutability methods.

I still think this strategy is opaque, though; only adept typing users will understand why these specific types are being used. I'd much rather see natively immutable types (although, I don't think a frozen dict is natively available....so maybe there just aren't great solutions....🤷🏼).

freakboy3742: I don't feel especially strongly about typing in this case. If we've got an automated tool that is doing something more correct-ish, then I'm inclined to lean on that; but if you think it will be especially onerous, I won't push back on reverting the use of this rule.

rmartin16: idk...i think it is clever insofar as achieving the stated purpose of the ruff rule. It isn't inherently problematic and in lieu of a better non-typing solution, I guess it's ultimately fine.

freakboy3742: Yeah - I'm definitely keeping an eye on ty to see if it becomes a viable option.

NMertsch: For what it's worth, PyCharm adds yellow squiggly lines when using these attributes for other purposes than annotated, and I assume other IDE-integrated type checkers do the same.

Alternatives:
* Revert this change
* Use `ClassVar[set[str]]` (wrong, these sets aren't meant to be mutated)
* Use frozensets instead of sets (I don't like this for subjective reasons)
* Use tuples instead of sets, keep dict annotated as Mapping (they are only used for membership tests anyway, and duplicates in a small tuple would be obvious to spot manually)

If you want a change, just let me know :)

mhsmith: > PyCharm adds yellow squiggly lines when using these attributes for other purposes than annotated

If by "other purposes" you mean "mutation", then that seems like enough of a benefit to justify the change.

NMertsch:
> in this case, just dropping the RSE rule (and related changes) is fine

Done.

PR #2522, opened 2025-10-20
Add the ability to specify bluetooth permission

depaolim:
Add the ability to specify bluetooth permission for android

Allows you to specify the enabling of permissions for using Bluetooth on the Android device

Inspired by #1599

## PR Checklist:

### DIFF
--- a/docs/en/reference/platforms/macOS/index.md
+++ b/docs/en/reference/platforms/macOS/index.md
@@ -218,13 +218,11 @@ For more details on macOS document type declarations, see the following web page
 
 ## Permissions
 
-### `macOS`
-
 Briefcase cross platform permissions map to a combination of [`info`][] and [`entitlement`][] keys:
 
-- [`permission.microphone`][permissionmicrophone]: an [`info`][] entry for `NSMicrophoneUsageDescription`; and an [`entitlement`][] of `com.apple.security.device.audio-input`
 - [`permission.bluetooth`][permissionbluetooth]: an [`info`][] entry for `NSBluetoothAlwaysUsageDescription`; and an [`entitlement`][] of `com.apple.security.device.bluetooth`
 - [`permission.camera`][permissioncamera]: an [`info`][] entry for `NSCameraUsageDescription`; and an [`entitlement`][] of `com.apple.security.device.camera`
+- [`permission.microphone`][permissionmicrophone]: an [`info`][] entry for `NSMicrophoneUsageDescription`; and an [`entitlement`][] of `com.apple.security.device.audio-input`
 - [`permission.coarse_location`][permissioncoarse_location]: an [`info`][] entry for `NSLocationUsageDescription` (ignored if [`permission.background_location`][permissionbackground_location] or [`permission.fine_location`][permissionfine_location] is defined); plus an entitlement of `com.apple.security.personal-information.location`
 - [`permission.fine_location`][permissionfine_location]: an [`info`][] entry for `NSLocationUsageDescription`(ignored if [`permission.background_location`][permissionbackground_location] is defined); plus an [`entitlement`][] of `com.apple.security.personal-information.location`
 - [`permission.background_location`][permissionbackground_location]: an [`info`][] entry for `NSLocationUsageDescription`; plus an [`entitlement`][] of `com.apple.security.personal-information.location`


depaolim:
Apologies, I'm not sure what's causing the build failures. If you can provide any hints or steps to reproduce the issue locally, I'd be happy to take a look and try to fix it.

rmartin16:
There was a significant AWS outage affecting Docker among many others. I re-ran the jobs for you.

timrid:
A few weeks ago I investigated a little regards bleak and briefcase on macOS and [added some documetation in an bleak PR](https://github.com/hbldh/bleak/pull/1832/files#diff-880df77e57299cd2cb0a89ad26cfbd91223f910eb41ef1574bd1c123f887559e).  

I am also currently working on iOS support for bleak (see https://github.com/hbldh/bleak/issues/1833). But it will take some time until I have tested the permissions with an iOS device. However, I think it will be very similar to macOS.

AFAIK For Windows and Linux there is no similar permissions concept for bluetooth.

depaolim:
> Thanks for the PR. The implementation and tests for this look perfectly good for Android; but we need to do a little more investigation before we can accept this as a change in Briefcase.
> 
> Briefcase's top level configuration for permissions is explicitly cross platform. That means we need at least a cursory investigation into whether a simple `bluetooth` permission will be sufficient on other platforms - iOS, macOS, Windows, and Linux.
> 
> We don't necessarily need to you _implement_ permissions for other platforms - but we need to know that we have sufficient granularity in our cross-platform permissions to capture the necessary features.

Your position makes perfect sense.

depaolim:
At the moment, I haven't found a straightforward way to use Bluetooth in a cross-platform manner with BeeWare, similar to how the GUI works with Toga. I've written separate experimental code snippets for Linux and Android. For Linux (and Windows), I used the python's socket implementation, while on Android I used objects from the "android.bluetooth" package. However, I haven't found a simple way to abstract this and write unified code that works across Linux, Windows, Android (and possibly iOS).
My use case is quite simple: I need to open a socket using a Bluetooth address, then perform recv/send operations. I've read several tickets discussing the use of Bleak or other solutions, but I haven't found a truly BeeWare-like approach (unlike the excellent abstractions BeeWare provides in other areas).
Do you have any suggestions or guidance on this?

freakboy3742:
> At the moment, I haven't found a straightforward way to use Bluetooth in a cross-platform manner with BeeWare, similar to how the GUI works with Toga... Do you have any suggestions or guidance on this?

At present, no. You're not the first person to ask about Bluetooth support, and it's an obvious area for a mobile app (or desktop app, for that matter) to support. However, I haven't done a lot of work with Bluetooth myself, so I don't have any particular advice to give. As you've noted, Bleak is usually the suggestion that comes up; but I haven't used it myself, and I'm aware that it doesn't offer an iOS solution at present (although evidently @timrid is looking into that).

depaolim:
> > At the moment, I haven't found a straightforward way to use Bluetooth in a cross-platform manner with BeeWare, similar to how the GUI works with Toga... Do you have any suggestions or guidance on this?
> 
> At present, no. You're not the first person to ask about Bluetooth support, and it's an obvious area for a mobile app (or desktop app, for that matter) to support. However, I haven't done a lot of work with Bluetooth myself, so I don't have any particular advice to give. As you've noted, Bleak is usually the suggestion that comes up; but I haven't used it myself, and I'm aware that it doesn't offer an iOS solution at present (although evidently @timrid is looking into that).

Thanks for your feedback!
(I was thinking that another challenge with adding Bluetooth support to BeeWare in the future is that unit testing Bluetooth behavior in CI pipelines is difficult, since smartphone emulators, as far as I know, don’t support Bluetooth.)

src/briefcase/platforms/android/gradle.py
```diff
@@ -245,6 +245,12 @@ def permissions_context(self, app: AppConfig, x_permissions: dict[str, str]):
         # Default feature usage for all Android apps
         features = {}
 
+        if x_permissions["bluetooth"]:
+            permissions["android.permission.BLUETOOTH"] = True
```
timrid: This will generate the following in the `AndroidManifest.xml`:

```
<uses-permission android:name="android.permission.BLUETOOTH" />
<uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
<uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
<uses-permission android:name="android.permission.BLUETOOTH_SCAN" />
```

But `android.permission.BLUETOOTH` and `android.permission.BLUETOOTH_ADMIN` are only available up to API version 30. So you have to specify `maxSdkVersion` like this:

```
<uses-permission android:name="android.permission.BLUETOOTH" android:maxSdkVersion="30" />
<uses-permission android:name="android.permission.BLUETOOTH_ADMIN" android:maxSdkVersion="30" />
```

See: https://developer.android.com/develop/connectivity/bluetooth/bt-permissions#declare-android12-or-higher

---

Additionally the bluetooth permissions are slightly mixed with the location permissions, because bluetooth *can* be used to get your location. If you *do not* want to use it for locations (`x_permissions["fine_location"]` is not defined), you have to specify:

On API Level 30 or above:
```
<uses-permission android:name="android.permission.BLUETOOTH_SCAN" android:usesPermissionFlags="neverForLocation" />
```

On API Level below 30:
```
<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" android:maxSdkVersion="30"/>
```

---

Additionally the permission for `BLUETOOTH_ADVERTISE` is missing.

freakboy3742: Thanks for those details - sounds like there might be a bit more design work needed here, on top of what is needed to confirm compatibility with other platforms. 

Regarding the fine-location handling - can those two co-exist? i.e., can I add both BLUETOOTH_SCAN and ACCESS_FINE_LOCATION definitions, or do I need to pick one based on the target SDK level? I guess that really means "can I add BLUETOOTH_SCAN on pre-SDK30 projects, even if it's a no-op?"

timrid: As far as I understand [this example from the Android docs](https://developer.android.com/develop/connectivity/bluetooth/bt-permissions#assert-never-for-location) yes, it is possible to declare `BLUETOOTH_SCAN` in pre-SDK30 projects.

---

There is a bleak Android port using briefcase called [bleekware](https://github.com/MarkusPiotrowski/bleekWare). In the docs they [recommend](https://github.com/MarkusPiotrowski/bleekWare?tab=readme-ov-file#how-to-use-it) the following permissions:

```xml
<uses-permission android:name="android.permission.BLUETOOTH" android:maxSdkVersion="30" />
<uses-permission android:name="android.permission.BLUETOOTH_ADMIN" android:maxSdkVersion="30" />
<uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" android:maxSdkVersion="30" />
<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" android:maxSdkVersion="30" />
<uses-permission android:name="android.permission.BLUETOOTH_SCAN" android:usesPermissionFlags="neverForLocation" />
<uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
```

depaolim: Thanks for the details. I added ACCESS_COARSE_LOCATION and ACCESS_FINE_LOCATION, but the android:maxSdkVersion and android:usesPermissionFlags attributes aren't being generated.
It is necessary to modify:
* android-gradle-template [for permission in cookiecutter.permissions](https://github.com/beeware/briefcase-android-gradle-template/blob/1abe20417fa00e959de536295c6bd686c3d0b302/%7B%7B%20cookiecutter.format%20%7D%7D/app/src/main/AndroidManifest.xml#L3)
* [permissions context dictionary](https://github.com/depaolim/briefcase/blob/3a47f0bbba2b4ccaa7b89315df5c92f6a1977729/src/briefcase/platforms/android/gradle.py#L249)

Of course, there's another elephant in the room: compatibility with other platforms needs to be confirmed.

timrid:
I just verified that bluetooth is working with iOS 26.1 on an iPhone 16 by using this branch https://github.com/hbldh/bleak/pull/1853. For permission handling I had to add:
```
[tool.briefcase.app.bleakbleexplorer.iOS]
info."NSBluetoothAlwaysUsageDescription" = "This is your description shown to the user. Explain why the app needs Bluetooth access."
```

image


---

On macOS Sequoia 15.5 I had to add the same as on iOS:

```
[tool.briefcase.app.bleakbleexplorer.macOS]
info."NSBluetoothAlwaysUsageDescription" = "This is your description shown to the user. Explain why the app needs Bluetooth access."
```

image

src/briefcase/platforms/android/gradle.py
freakboy3742: Given that Android evidently has other properties tied to permissions, I would suggest the approach here will be to modify the format that Android uses to generate permissions - instead of a dictionary of boolean values, the values should probably be a dictionary of property values:
```
permission."android.permission.ACCESS_FINE_LOCATION" = {"android:maxSdkVersion"="30"}
permission."android.permission.BLUETOOTH_SCAN" = {"android:usesPermissionFlags"="neverForLocation"}
permission."android.permission.BLUETOOTH_CONNECT" = {}
```
which can then be rolled out as attributes on the XML element. For backwards compatibility purposes, a value of `true` would need to be interpreted as `{}`

Briefcase's permissions context would be built up in a similar way, with the appropriate version and permission defaults described above.

As for other platforms: based on the details from @timrid below, it looks like there's only one permission needed for iOS - NSBluetoothAlwaysUsageDescription. Based on what I'm reading, the same permission is needed for macOS (based on both [Apple](https://developer.apple.com/documentation/BundleResources/Information-Property-List/NSBluetoothAlwaysUsageDescription) and [Bleak's documentation](https://bleak.readthedocs.io/en/latest/backends/macos.html#permissions). There doesn't appear to be any permission requirements on Linux and Windows mentioned in Bleak's documentation, so I guess that means we don't have to do anything there.

depaolim: I added the ability to specify attributes.
I also had to modify the Android template; see https://github.com/beeware/briefcase-android-gradle-template/pull/105. (The template change appears backwards compatible, so it may be approved before this PR is approved)
I'm still a bit confused about how the False case is handled, see https://github.com/depaolim/briefcase/blob/17fda62124dfd74ed5ec71c71abccc7d9931bfcd/tests/platforms/android/gradle/test_create.py#L456

iOS and macOS are still missing.

docs/en/reference/configuration.md
```diff
@@ -379,6 +379,10 @@ Applications may also need to declare the permissions they require. Permissions
 
 Briefcase maintains a set of cross-platform permissions:
 
+#### `permission.bluetooth`
+
+Permission to connect to an external device via Bluetooth.
```
freakboy3742, line 384: This is all that is needed for configuration; however, there is also a need to document the specific manifestation of permissions in the platform guide for android/gradle. 

There's two relevant sections:
1. The `permissions` attribute should clarify how dictionary values are interpreted
2. The "Permissions" section should detail the default mapping of the Bluetooth permission on Android.

src/briefcase/platforms/android/gradle.py
```diff
@@ -239,41 +239,59 @@ def permissions_context(self, app: AppConfig, x_permissions: dict[str, str]):
...
             features["android.hardware.camera.external"] = False
             features["android.hardware.camera.autofocus"] = False
 
         if x_permissions["microphone"]:
-            permissions["android.permission.RECORD_AUDIO"] = True
+            permissions["android.permission.RECORD_AUDIO"] = {}
 
         if x_permissions["fine_location"]:
-            permissions["android.permission.ACCESS_FINE_LOCATION"] = True
+            permissions["android.permission.ACCESS_FINE_LOCATION"] = {}
```
freakboy3742, line 279: This highlights an additional edge case: if I specify `permission.bluetooth` and either `permission.fine_location` or `permission.coarse_location`, then:

1. We need to validate that the maxSdkVersion *isn't* applied to the location permissions
2. It's probably safe to assume that I'm OK using bluetooth for location. Presumably we should have a value other than `neverForLocation` for the `BLUETOOTH_SCAN` permission.

src/briefcase/platforms/android/gradle.py
freakboy3742: You're right... the existing template *isn't* handling the `False` case for permissions. Looking at the context, I suspect this is a case of *feature* needing an explicit True/False value (and using both of those); but Permission is relying on "if you don't need it, don't define it" logic. 

So - that would be good to resolve this issue while we're at it.

Based on the detail from @timrid, it looks like the iOS and macOS cases should be reasonably straightforward - would you feel comfortable adding those in as well? I appreciate you may not be in a position to practically test them, but if you can get a stub implementation and docs in place, we can do the testing at our end.

changes/2522.feature.md
freakboy3742, line 1: We've moved to Markdown for all our docs; so this should now be a .md file.

freakboy3742 (changes requested):
This looks good; I've flagged a couple issues related to documentation and an edge case of permissions handling.

The other detail: if we're going to fix the `permission = False` handling, we should also add a second changenote (2522.bugfix.md) that describes the fact that we're now handling permission disabling correctly.

depaolim:
Thanks for the feedback, I'll try to get to it next week

src/briefcase/platforms/android/gradle.py
depaolim: I'm sorry, I can't quite understand how you intend to handle the False case: is it better to completely eliminate the test case for False or is it better to modify the template to handle it?

I added the implementation and documentation for iOS and macOS, please check if it is correct

freakboy3742: I'd say the template is the right place to handle it. 

For example - Briefcase might turn on one particular permission as part of a default; the user could then turn that off by setting the specific permission to `false`, which would then be part of the template context.

docs/en/reference/platforms/android/gradle.md
```diff
@@ -279,6 +296,17 @@ Briefcase cross platform permissions map to `<uses-permission>` declarations in
...
+Specifying a [`permission.bluetooth`][permissionbluetooth] permission will result in the following `<uses-permission>` declarations in the app's `AppManifest.xml`:
+
+```xml
+<uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" android:maxSdkVersion="30"/>
+<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" android:maxSdkVersion="30"/>
+<uses-permission android:name="android.permission.BLUETOOTH" android:maxSdkVersion="30"/>
+<uses-permission android:name="android.permission.BLUETOOTH_ADMIN" android:maxSdkVersion="30"/>
+<uses-permission android:name="android.permission.BLUETOOTH_CONNECT"/>
+<uses-permission android:name="android.permission.BLUETOOTH_SCAN" android:usesPermissionFlags="neverForLocation"/>
+```
```
freakboy3742: This should be in 'human readable' form, rather than a XML snippet.
```suggestion
* `android.permission.ACCESS_COARSE_LOCATION`, with an attribute declaration of `android:maxSdkVersion="30"`. If `permission.coarse_location` is defined, the attribute declaration will be omitted
* `android.permission.ACCESS_FINE_LOCATION", with an attribute declaration of `android:maxSdkVersion="30"`. If `permission.coarse_location` is defined, the attribute declaration will be omitted
* `android.permission.BLUETOOTH", with an attribute declaration of `android:maxSdkVersion="30"`
* `android.permission.BLUETOOTH_ADMIN", with an attribute declaration of `android:maxSdkVersion="30"`)
* `android.permission.BLUETOOTH_CONNECT`
* `android.permission.BLUETOOTH_SCAN`, with an attribute declaration of `android:usesPermissionFlags="neverForLocation"`. If `permission.fine_location` or `permission.coarse_location` is defined, the attribute declaration will be omitted.
```

freakboy3742 (changes requested):
One tweak to documentation, plus the update to the template, and I think we'll be good to go!

freakboy3742 (approved):
Thanks for those updates - this looks great!

depaolim:
> Thanks for those updates - this looks great!

Happy to contribute to this fantastic project and thank you for your valuable suggestions

Issue #2383, opened 2025-07-09
Resolve `E501` ignored Ruff rule


kattni:
### What is the problem or limitation you are having?

When Ruff was enabled, for review purposes, the identified errors were globally ignored, with the intention of resolving them in subsequent PRs.

### Describe the solution you'd like

This repo currently ignores the following rules from the `EXXX` series:
* `E510`

Work through the issues as follows:
1. Comment out the error number in the `ignore` list found [here](https://github.com/beeware/briefcase/blob/1ffd045752d3a3d920a68d920201d890e8637e49/pyproject.toml#L256).
2. Run `pre-commit run -a` to run the checks.
3. Work through each of the errors in the result until they are all resolved.
4. `pre-commit run -a` should pass without the errors in the `ignore` list.
5. Remove the `E501` rule from the `ignore` list.
6. If this is the last of the errors in the `ignore` list, remove the [`ignore` list](https://github.com/beeware/briefcase/blob/1ffd045752d3a3d920a68d920201d890e8637e49/pyproject.toml#L251-L267) completely, including the comment above.
7. Submit the PR.

### Describe alternatives you've considered

None.

### Additional context

The `SIMXXX` checks are explicitly disabled in the configuration. Do not remove the commented out `SIM` check, and avoid resolving the `SIMXXX` errors.

bomtall:
Hi, I am looking at this issue at EuroPython 2025 sprint weekend

NMertsch:
For those looking into this:

black can fix some of these cases with the experimental [string_processing](https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html#labels-string-processing) feature:

```shell
black --enable-unstable-feature string_processing --preview --line-length 88 .
```

This will (among other things) splint long string at spaces, with the space at the beginning of the next line:

```python
"Some long string"
" with spaces."
```

@freakboy3742 would prefer the space at the end of the previous line:

```python
"Some long string "
"with spaces."
```

This can probably be done with a regex-replace.

JohananOppongAmoateng:
I am picking this up to continue

JohananOppongAmoateng:
leaving this for others to continue working on

xavya-dev:
Hi team, I would like to work on this issue. I have just raised a PR with a partial fix for the test_signing.py file. I will continue to work on fixing the remaining files.

damiankas1122-crypto:
[...]

PR #2526, opened 2025-10-22
Limit line length in `src/briefcase` comments (ruff E501)

NMertsch:
Issue #2383 requests to resolve [ruff rule E501](https://docs.astral.sh/ruff/rules/line-too-long/) by limiting all lines to 88 characters (and probably adding some `noqa` comments).

I'd like to split it into multiple PRs because it is a large task and the resolutions might trigger some (justified) discussions around style preferences.

This PR only addresses the **comments** in `src/briefcase`, which reduces the number of violations from 520 to 446. Please let me know what you think, and if you are interested in more PRs like this.

(I started with comments because I assume these fixes are less controversial than actual code changes.)

Refs #2383 

## PR Checklist:

### DIFF
--- a/changes/2526.misc.rst
+++ b/changes/2526.misc.rst
@@ -1,0 +1,1 @@
+Comments in ``src/briefcase`` now have a maximum line length of 88 characters (ruff rule ``E501``).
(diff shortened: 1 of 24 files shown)


freakboy3742 (approved):
Incremental improvement is still improvement; If this gets us ~100 issues closer to being able to enable E501, then it's worth it. 

I'd suggest a similar incremental approach is worthwhile for the code changes as well - a PR that fixes "all line lengths in the platform module" (or similar) will make the review process a lot easier.

PR #2533, opened 2025-10-24
Fix line length in `briefcase/*.py` and `briefcase/integrations` (ruff rule E501)

NMertsch:
Activate the ruff rule [E501 - line too long](https://docs.astral.sh/ruff/rules/line-too-long/) in `briefcase/*.py` and `briefcase/integrations` and handle all findings.

I'd especially appreciate feedback on the following commits:
* 2bbef0cff4a4d59c6c10f521533f93ced24afca5 - I added backslashes in multiline strings to not change the behavior. However, I'd prefer different line breaks, or using [textwrap](https://docs.python.org/3/library/textwrap.html) for long warning and error messages.
* 9d714d97761556e9aaba34005c69b3a115d86572 - I marked three lines with `noqa: 501` because I didn't find a satisfying solution. I don't think a larger code change is worth the benefit of reducing these line lengths.

## PR Checklist:

### DIFF
--- a/src/briefcase/exceptions.py
+++ b/src/briefcase/exceptions.py
@@ -202,6 +202,10 @@ class MissingAppMetadata(BriefcaseCommandError):
 
 class MissingPlatformSupport(BriefcaseCommandError):
     def __init__(self, python_version_tag, platform, host_arch, is_32bit, item):
+        self.python_version_tag = python_version_tag
+        self.platform = platform
+        self.host_arch = host_arch
+        self.is_32bit = is_32bit
         platform_name = f"{'32 bit ' if is_32bit else ''}{platform}"
         super().__init__(
             f"Unable to download {platform_name} {item} for Python "
--- a/src/briefcase/integrations/java.py
+++ b/src/briefcase/integrations/java.py
@@ -117,7 +117,7 @@ class JDK(ManagedTool):
                 else:
                     install_message = f"""
 *************************************************************************
-** WARNING: JAVA_HOME does not point to Java {cls.JDK_MAJOR_VER}                  **
+** WARNING: JAVA_HOME does not point to a Java {cls.JDK_MAJOR_VER} JDK            **
 *************************************************************************
 
     Android requires a Java {cls.JDK_MAJOR_VER} JDK, but the location pointed to by the
--- a/tests/integrations/java/test_JDK__verify.py
+++ b/tests/integrations/java/test_JDK__verify.py
@@ -263,7 +263,7 @@ def test_invalid_jdk_version(mock_tools, host_os, java_home, tmp_path, capsys):
 
     # Warning is shown for invalid JAVA_HOME
     output = capsys.readouterr()
-    assert "WARNING: JAVA_HOME does not point to Java 17" in output.out
+    assert "WARNING: JAVA_HOME does not point to a Java 17 JDK" in output.out
     assert output.out.endswith("****\n")
     assert output.err == ""
 
@@ -504,7 +504,7 @@ def test_successful_jdk_download(
     # Console output contains a warning about the bad JDK location
     output = capsys.readouterr()
     assert output.err == ""
-    assert "** WARNING: JAVA_HOME does not point to Java 17" in output.out
+    assert "** WARNING: JAVA_HOME does not point to a Java 17 JDK" in output.out
 
     # Download was invoked
     mock_tools.file.download.assert_called_with(


src/briefcase/cmdline.py
```diff
@@ -63,7 +63,7 @@ def parse_cmdline(args, console: Console | None = None):
 
     description_max_pad_len = max(len(cmd.command) for cmd in COMMANDS) + 2
     command_description_list = "\n".join(
-        f"  {cmd.command}{' ' * (description_max_pad_len - len(cmd.command))}{cmd.description}"
+        f"  {cmd.command}{' ' * (description_max_pad_len - len(cmd.command))}{cmd.description}"  # noqa: E501 (line too long)
```
mhsmith: Maybe the padding could be done by using the "field width" feature of the formatting mini-language. Otherwise, `description_max_pad_len - len(cmd.command)` could be broken out into a separate variable.

src/briefcase/console.py
```diff
@@ -489,7 +492,7 @@ def _build_log(self, command) -> str:
 
         # Capture env vars removing any potentially sensitive information
         sanitized_env_vars = "\n".join(
-            f"\t{env_var}={value if not SENSITIVE_SETTING_RE.search(env_var) else '********************'}"
+            f"\t{env_var}={value if not SENSITIVE_SETTING_RE.search(env_var) else '********************'}"  # noqa: E501 (line too long)
```
mhsmith: The if-else expression could be taken out of the f-string and written as a separate wrappable expression, appended to the f-string with `+`.

src/briefcase/console.py
```diff
@@ -520,7 +523,7 @@ def _build_log(self, command) -> str:
             #     directory (e.g. /usr/local).
             # sys.prefix is updated (from base_prefix's value) to the virtual env's
             #     site-specific directory.
-            f"Virtual env:     {hasattr(sys, 'real_prefix') or sys.base_prefix != sys.prefix}\n"
+            f"Virtual env:     {hasattr(sys, 'real_prefix') or sys.base_prefix != sys.prefix}\n"  # noqa: E501 (line too long)
```
mhsmith: Older versions of virtualenv are obsolete now, so I think all references to `real_prefix` can be removed.

src/briefcase/exceptions.py
```diff
@@ -136,7 +136,8 @@ def __init__(self, tool: str, env_var: str):
             msg=f"""\
 Briefcase cannot install {tool} on this machine.
 
-Install {tool} manually and specify the installation directory in the {env_var} environment variable.
+Install {tool} manually and specify the installation directory in the {env_var} \
+environment variable.
```
mhsmith: > I added backslashes in multiline strings to not change the behavior. However, I'd prefer different line breaks, or using [textwrap](https://docs.python.org/3/library/textwrap.html) for long warning and error messages.

`textwrap` would be a good idea in the future, in the place where these messages are printed. To leave that option open, we shouldn't wrap them manually in the place where they're created. In which case, I think the most readable option would be to not use a multiline string at all, i.e.:

```py
            msg=(
                f"Briefcase cannot install {tool} on this machine."
                f"\n\n"
                f"Install {tool} manually and specify the installation "
                f"directory in the {env_var} environment variable."
            )
```

This also has the advantage of not obscuring the code indentation.

src/briefcase/config.py
```diff
@@ -475,19 +480,23 @@ def __init__(
...
         if not is_valid_app_name(self.app_name):
             raise BriefcaseConfigError(
-                f"{self.app_name!r} is not a valid app name.\n\n"
-                "App names must not be reserved keywords such as 'and', 'for' and 'while'.\n"
-                "They must also be PEP508 compliant (i.e., they can only include letters,\n"
-                "numbers, '-' and '_'; must start with a letter; and cannot end with '-' or '_')."
+                f"""\
+{self.app_name!r} is not a valid app name.
+
+App names must not be reserved keywords such as 'and', 'for' and 'while'.
```
mhsmith: Suggest not using a multiline string for this kind of situation either, i.e.:
* Go back to the previous code.
* Remove the embedded newlines, other than the paragraph break after the first sentence.
* Adjust the line breaks in the code to comply with the length limit, without changing the resulting concatentated string.

src/briefcase/cmdline.py
freakboy3742: Even shortening `description_max_pad_len` to `pad` will get you to < 80. But agreed that field width padding is a better option here.

NMertsch: Thanks, I learned something new today. Fixed in b53228e9.

src/briefcase/console.py
NMertsch: Thank you. Resolved in d41552f8.

src/briefcase/console.py
NMertsch: I don't quite understand what you mean, but hope 3a9c3de4 is a valid resolution.

src/briefcase/exceptions.py
NMertsch: Thanks, resolved in 1f98642c.

src/briefcase/config.py
NMertsch: Thank you. Reverted in a77c1356, I hope 62e36e5a is the resolution you described.

pyproject.toml
```diff
@@ -256,16 +256,17 @@ extend-select = [
...
     "SIM105",  # use `contextlib.suppress` (too opinionated)
     "SIM108",  # replace if-else with ternary operator (too opinionated)
 ]
 
+[tool.ruff.lint.per-file-ignores]
+# E501: line too long, to be fixed in future changes
+"tests/*" = ["E501"]
+"src/briefcase/bootstraps/*" = ["E501"]
+"src/briefcase/commands/*" = ["E501"]
+"src/briefcase/platforms/*" = ["E501"]
```
freakboy3742, line 268: With #2548 and #2549 being merged, this should be updated:
```suggestion
[tool.ruff.lint.per-file-ignores]
# E501: line too long, to be fixed in future changes
"tests/*" = ["E501"]
"src/briefcase/bootstraps/*" = ["E501"]
```

src/briefcase/console.py
freakboy3742: The idea is sound - but it doesn't need to be a full function. It can be assigned to a `sanitized_value` variable, and then used in the f-string.

src/briefcase/exceptions.py
freakboy3742: There's a couple of other instances of this behavior later in the same file.

src/briefcase/integrations/java.py
```diff
@@ -117,7 +117,7 @@ def verify_install(cls, tools: ToolCache, install: bool = True, **kwargs) -> JDK
                 else:
                     install_message = f"""
 *************************************************************************
-** WARNING: JAVA_HOME does not point to a Java {cls.JDK_MAJOR_VER} JDK                  **
+** WARNING: JAVA_HOME does not point to a Java {cls.JDK_MAJOR_VER} JDK **
```
freakboy3742: This won't be right, because the length of the line can vary.

src/briefcase/exceptions.py
freakboy3742: There's also some uses in android_sdk and windows_sdk

src/briefcase/integrations/java.py
mhsmith: I think the previous code was assuming the version was 2 digits long; it's fine to keep assuming that for now.

src/briefcase/console.py
mhsmith: What I meant was something like this:
```py
f"\t{env_var}=" + (
    value if not SENSITIVE_SETTING_RE.search(env_var) else '********************'
)
```
But depending on wrapping, a separate variable might be better.

mhsmith:
@NMertsch: in future, please don't manually resolve conversations on your own PR. Let a reviewer do that, or GitHub's automatic resolutions from applying suggestions.

freakboy3742:
> @NMertsch: in future, please don't manually resolve conversations on your own PR. Let a reviewer do that, or GitHub's automatic resolutions from applying suggestions.

I'm fairly sure that was me - I did a review pass today, and cleaned up the items that seemed to be resolved.

mhsmith:
I'd already unresolved the conversations a couple of days ago, but didn't comment at that time.

src/briefcase/integrations/java.py
NMertsch: Thanks, my mistake. Reverted in dbddd17cad3591f7a850d7b8ac7f0ffb000a73b9.

NMertsch:
> @NMertsch: in future, please don't manually resolve conversations on your own PR. Let a reviewer do that, or GitHub's automatic resolutions from applying suggestions.

Thank you, will do so in the future. I didn't consider that perspective and thought it would save you time.

src/briefcase/exceptions.py
NMertsch: > we shouldn't wrap them manually in the place where they're created.

> There's a couple of other instances of this behavior later in the same file. There's also some uses in android_sdk and windows_sdk.

There are many multiline strings with wrapped lines in the briefcase codebase. I added backslashes to keep the lines below 89 characters and keep the original line breaks.

What do you expect me to do?

src/briefcase/console.py
NMertsch: > It can be assigned to a `sanitized_value` variable, and then used in the f-string.

No, this is inside a comprehension, so I can't create a separate variable. See a non-comprehension solution below.

> What I meant was something like this: [...]

The full result would be this, and the line would still be too long:

```python
        sanitized_env_vars = "\n".join(
            f"\t{env_var}=" + (
                value if not SENSITIVE_SETTING_RE.search(env_var) else '*************************'
            )
            for env_var, value in sorted(command.tools.os.environ.items())
        )
```

We could remove five asterisks to meet the line length limit. I personally find the separate function from 3a9c3de much nicer, but if you want, I can change it. 

Another option would be a loop instead of a comprehension:

```python
        sanitized_env_lines = []
        for env_var, value in sorted(command.tools.os.environ.items()):
            if SENSITIVE_SETTING_RE.search(env_var):
                sanitized_env_lines.append(f"{env_var}=********************")
            else:
                sanitized_env_lines.append(f"{env_var}={value}")
```

Tbh, this feels like bikeshedding to me.

NMertsch:
I feel like this PR has cost you and me too much time already, given that it is just about the formatting of existing and working code which already passed a code review before.
Sorry, I thought this would be a "quick win", but I currently don't have the energy left to discuss code formatting details in my free time.

Maybe it is best to either close this PR as-is, or revert the undesired changes, re-ignore the rule in `pyproject.toml`, and merge only the changes you want?

src/briefcase/console.py
mhsmith: You're right about a variable not being possible in a comprehension, sorry for the confusion. Outside of an f-string, the if-else expression could be wrapped onto multiple lines. However, I agree that the function approach is more readable here.

src/briefcase/exceptions.py
mhsmith: I think he's suggesting that the few other multiline strings that this PR touches, could be coded in the same way as this one.

mhsmith:
Unfortunately I think it's inevitable that enforcing a shorter line length will require some code to be restructured in order to remain as readable as before. If you don't want to continue right now, that's fine, but let's leave the PR open in case you or anyone else wants to finish it later.

src/briefcase/console.py
mhsmith: Actually, I've gone with the if-else expression after all.

src/briefcase/exceptions.py
mhsmith: I've removed all the added backslashes, sometimes by reformatting as above, and sometimes by rewording or reflowing the string.

pyproject.toml
mhsmith: I've enabled the E501 check in `commands`, but not in `platforms`, because #2549 didn't resolve all the long lines in that folder. That can be left to a future PR.

src/briefcase/integrations/java.py
mhsmith: Fixed by rewording the text.

src/briefcase/exceptions.py
```diff
@@ -197,41 +200,33 @@ def __init__(self, app_bundle_path):
...
 
 
-class MissingSupportPackage(BriefcaseCommandError):
-    def __init__(self, python_version_tag, platform, host_arch, is_32bit):
-        self.python_version_tag = python_version_tag
-        self.platform = platform
-        self.host_arch = host_arch
-        self.is_32bit = is_32bit
+class MissingPlatformSupport(BriefcaseCommandError):
+    def __init__(self, python_version_tag, platform, host_arch, is_32bit, item):
```
freakboy3742, line 204: The exception should be retaining the properties that constructed it so it can be interrogated later.
```suggestion
    def __init__(self, python_version_tag, platform, host_arch, is_32bit, item):
        self.python_version_tag = python_version_tag
        self.platform = platform
        self.host_arch = host_arch
        self.is_32bit = is_32bit
```

src/briefcase/integrations/java.py
```diff
@@ -117,7 +117,7 @@ def verify_install(cls, tools: ToolCache, install: bool = True, **kwargs) -> JDK
                 else:
                     install_message = f"""
 *************************************************************************
-** WARNING: JAVA_HOME does not point to a Java {cls.JDK_MAJOR_VER} JDK                  **
+** WARNING: JAVA_HOME does not point to Java {cls.JDK_MAJOR_VER}                  **
```
freakboy3742: "Java" and "A Java JDK" are two different things, as "Java" can (and usually does) refer to the runtime only.
```suggestion
** WARNING: JAVA_HOME does not point to a Java {cls.JDK_MAJOR_VER} JDK            **
```

tests/integrations/java/test_JDK__verify.py
```diff
@@ -263,7 +263,7 @@ def test_invalid_jdk_version(mock_tools, host_os, java_home, tmp_path, capsys):
 
     # Warning is shown for invalid JAVA_HOME
     output = capsys.readouterr()
-    assert "WARNING: JAVA_HOME does not point to a Java 17 JDK" in output.out
+    assert "WARNING: JAVA_HOME does not point to Java 17" in output.out
```
freakboy3742:
```suggestion
    assert "WARNING: JAVA_HOME does not point to a Java 17 JDK" in output.out
```

tests/integrations/java/test_JDK__verify.py
```diff
@@ -504,7 +504,7 @@ def test_successful_jdk_download(
     # Console output contains a warning about the bad JDK location
     output = capsys.readouterr()
     assert output.err == ""
-    assert "** WARNING: JAVA_HOME does not point to a Java 17 JDK" in output.out
+    assert "** WARNING: JAVA_HOME does not point to Java 17" in output.out
```
freakboy3742:
```suggestion
    assert "** WARNING: JAVA_HOME does not point to a Java 17 JDK" in output.out
```

NMertsch:
Thank you for pushing this over the line.

> I've enabled the E501 check in commands, but not in platforms, because #2549 didn't resolve all the long lines in that folder. That can be left to a future PR.

I added noqa comments to these cases, but apparently pre-commit removed them. Sorry for that, I'm not used to pre-commit modifying code.

src/briefcase/exceptions.py
mhsmith: Interrogated where? I couldn't find any references.

src/briefcase/integrations/java.py
mhsmith: The reason why this line was too long to begin with is that it has a 20-character placeholder that "expands" into a 2-character value. This makes it difficult to line up the asterisk box. As it stands, it will render like this:
```
```
These kinds of messages are very awkward to maintain, so I've proposed a general solution: #2559.

PR #2534, opened 2025-10-25
Updates to support Constructor

mhsmith:
* Allow external x86-64 apps to be packaged on any architecture, with a warning. This makes it possible to test on Parallels on an ARM64 Mac.

* Since Constructor creates Start menu shortcuts using the [menuinst](https://github.com/conda/menuinst) tool, provide a way to disable creation of Start menu shortcuts by setting `external_package_executable_path = ""`. This is implemented in the corresponding template PRs:
  * https://github.com/beeware/briefcase-windows-app-template/pull/77
  * https://github.com/beeware/briefcase-windows-VisualStudio-template/pull/71

* Miscellaneous documentation improvements.

## PR Checklist:


docs/en/reference/platforms/windows/index.md
```diff
@@ -164,7 +167,7 @@ However, if you need to override this default value, you can define [`version_tr
 
 ## Installer options
 
-Windows MSI installers are able to present a panel of optional features to the user as part of the installation process. These features are binary flags which can then be used by a post-install script to perform additional installation behaviors.
+Windows MSI installers are able to present a panel of optional features to the user as part of the installation process. These features are binary flags which can then be used by a [`post_install_script`][] to perform additional installation behaviors.
```
freakboy3742:
```suggestion
Windows MSI installers are able to present a panel of optional features to the user as part of the installation process. These features are binary flags which can then be used by a [post-install script][post_install_script] to perform additional installation behaviors.
```

tests/platforms/windows/app/create/test_create.py
```diff
@@ -28,24 +28,39 @@ def test_unsupported_host_os(create_command, host_os):
...
+    first_app_config.external_package_path = "external"
+    create_command.verify_host()
+    stdout, stderr = capsys.readouterr()
+    assert "WARNING: Possible architecture mismatch" in stdout
+    assert stderr == ""
+
+    # The warning is not repeated on cloned commands.
+    create_command.is_clone = True
+    create_command.verify_host()
+    assert capsys.readouterr() == ("", "")
```
freakboy3742: This feels like it should be a second test, rather than piggybacking on test_unsupported_arch

freakboy3742 (changes requested):
All does what it says on the tin; my only question is whether `external_package_executable_path` is the right control point for this feature. 

Would it not be better to have a separate `install_start_menu` item, defaulting to the same value as `console_app`? In the current setup, you will get a Start menu item for every command line tool as well - which is unlikely to be the desired behavior. And, as Constructor has highlighted, it's not inconceivable that a GUI app might want to take control of creating the start menu item as part of a post-install script.

tests/platforms/windows/app/create/test_create.py
mhsmith: Done.

mhsmith:
> Would it not be better to have a separate `install_start_menu` item

Good idea: I've given it the generic name of `install_launcher`, since it could be implemented on other platforms in the future.

freakboy3742 (approved):
👍

PR #2552, opened 2025-11-03
Add anchor for stable/0.3.25 build validation.

freakboy3742:
With the arrival of the docs landing page, we needed to rebuild the old stable/0.3.25 docs to pick up the new header bar/theme.

However, those builds are failing because of [a broken link to `gradle.html#android.build_gradle_dependencies`](https://app.readthedocs.org/projects/briefcase/builds/30174618/#292613919--935).

The `.html` suffix no longer exists - but that isn't a problem; but the anchor has changed to `build_gradle_dependencies`. However, we can add a temporary link so that the build will complete.

This also takes the chance to clean up the homepage to simplify the headings. This matches a change made in beeware/rubicon-objc#670, and the structure that is already in Toga.

Once this lands, it should be possible to trigger a re-build of the 0.3.25 and stable docs on RTD. We should be able to remove the anchor immediately afterwards - it's only needed so the rebuild succeeds.

## PR Checklist:

### DIFF
--- a/changes/2552.misc.rst
+++ b/changes/2552.misc.rst
@@ -1,0 +1,1 @@
+A docs anchor was added for backwards compatibility purposes.


kattni (approved):
Thanks for sorting out the fix.

PR #2563, opened 2025-11-14
Add helper for formatting long warning messages

### LINKED ISSUE #2559: Add helper for formatting long warning messages

This has several problems:

   ```
   ``` 

### Additional context

otekraden:
* Fixes #2559

I solved Issue #2559 by creation the function "format_message_to_asterisk_box", that formatting message to asterisk box representation.

It takes this args:
message (str): The message to format.
title (str, optional): The title of the box. Defaults to None.
width (int, optional): The width of the box. Defaults to 80.
border_char (str, optional): The character to use for the border.
Defaults to "*".
padding (int, optional): The number of spaces to pad the border.
Defaults to 2.

Also I added tests for it. Function it seems working good.

But I was not sure in what file to put function.
Now its location and tests by this paths:
src/briefcase/utils.py
tests/test_utils.py

It is my very first attempt of contribution. Please don beat me hard)

I am waiting for your feedback.

## PR Checklist:

### DIFF
--- a/changes/2563.misc.md
+++ b/changes/2563.misc.md
@@ -1,1 +1,1 @@
-``Warning_banner`` method added to ``Console`` class.
+An internal helper to display warning banners was added.
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -251,64 +251,68 @@ class Console:
         return self._log_impl.export_text()
 
     @staticmethod
-    def _dedent_and_wrap(text, wrap_width, is_title=False):
-        """Dedent text, split text into paragraphs and wrap each paragraph to lines of
-        the specified width."""
+    def _dedent_and_wrap(text, wrap_width, border=0):
+        """Dedent text, split text into paragraphs and wrap to a width.
 
+        If a border is specified, that border width will be preserved on the left *and
+        right* of the text. The right hand side won't be printed, but it will be
+        accommodated in any line wrapping.
+
+        Text that is already indented relative to the rest of the text will have that
+        indentation preserved.
+
+        If text is indented, it will *not* undergo line breaks on spaces and hyphens, on
+        the assumption that it will be an environment variable or command line.
+        """
         # Dedent and remove common leading empty lines
         text = textwrap.dedent(text).strip("\n")
-
-        # Add a warning prefix for title
-        if is_title:
-            text = f"WARNING: {text.strip()}"
 
         # Split text into text lines
         input_lines = text.split("\n")
 
-        # create paragraphs from lines
+        # Create paragraphs from lines. Use "None" for a paragraph break.
+        # Preserve any line that starts with spaces as-is. Any other line
+        # is accumulated into the last paragraph
         paragraphs = []
         for line in input_lines:
-            # add empty paragraph
             if not line:
+                # Preserve an empty line as a paragraph break
                 paragraphs.append(None)
-                continue
-
-            # calc relative indentation; always zero for title
-            rel_indent = 0 if is_title else len(line) - len(line.lstrip())
-
-            # add new paragraph
-            # if it first, if last empty or have relative indentation
-            if any((not paragraphs, rel_indent)) or not paragraphs[-1]:
-                # add new paragraph
-                paragraphs.append([line.strip(), rel_indent])
-
-            # append line to last paragraph
+            elif line[0] == " ":
+                # If the line starts with indentation, preserve it
+                paragraphs.append(line.rstrip())
+            elif not (paragraphs and paragraphs[-1] and paragraphs[-1][0] != " "):
+                # This is either the first paragraph, or the current last paragraph is
+                # either a paragraph break, or indented literal paragraph. This means
+                # the current line is the start of a new paragraph.
+                paragraphs.append(line.strip())
             else:
-                paragraphs[-1][0] += f" {line.strip()}"
+                # Accumulate onto the most recent paragraph.
+                paragraphs[-1] = f"{paragraphs[-1]} {line.strip()}"
 
         # Wrap each paragraph to lines
         output_lines = []
         for p in paragraphs:
-            # If the paragraph is not empty, wrap it to the specified width
             if p:
-                output_lines.extend(
-                    # add indent to each line in wrapping
-                    f"{' ' * p[-1]}{line}"
-                    for line in
-                    # wrapping paragraph to width minus rel indent
-                    textwrap.wrap(p[0], width=wrap_width - p[1])
-                )
-
-            # Otherwise, add an empty line
+                if p[0] == " ":
+                    # Preserve an indented line as a literal
+                    output_lines.append(f"{' ' * border}{p}")
+                else:
+                    # Break the paragraph into lines.
+                    output_lines.extend(
+                        f"{' ' * border}{line}"
+                        for line in textwrap.wrap(p, width=wrap_width - (border * 2))
+                    )
             else:
+                # Insert a paragraph break.
                 output_lines.append("")
-        # Return the list of lines
+
         return output_lines
 
     def warning_banner(
         self,
-        title: str,
-        message: str,
+        title: str | None = None,
+        message: str | None = None,
         width: int = 80,
     ) -> str:
         """The title or message can be provided as a single or as multiline string. Any
@@ -324,52 +328,24 @@ class Console:
         :param width: The total width of the box in characters. Defaults to 80.
         :return: The formatted message enclosed in a bordered box.
         """
+        BORDER_LINE = "*" * width
+        lines_array = ["", BORDER_LINE]
 
-        # character to use for the box border
-        border_char = "*"
-
-        # If message and title are both empty rase ValueError
-        if not any((message, title)):
-            raise ValueError("Message or title must be provided")
-        # If width is not an integer, raise TypeError
-        if not isinstance(width, int):
-            raise TypeError("Width must be an integer")
-
-        # Create border line
-        border_line = border_char * width
-        # create lines array with opening line of the box
-        lines_array = [border_line]
-
-        # if title exists, format title in the box
         if title:
-            # title must be a string
-            if not isinstance(title, str):
-                raise TypeError("Title must be a string")
+            # Remove 6 from the width to allow for "** " and " **" wrappers
+            title_lines = self._dedent_and_wrap(f"WARNING: {title}", width - 6)
 
-            # get title wrapped to lines
-            title_lines = self._dedent_and_wrap(title, width - 6, True)
-
-            # center each line of the title and add to the box
+            # Center each line of the title and add to the box
             for line in title_lines:
-                line = line.center(width - 4)
-                lines_array.append(f"{border_char * 2}{line}{border_char * 2}")
+                line = line.center(width - 6)
+                lines_array.append(f"** {line} **")
 
-            # closing line of title in the box
-            lines_array.append(border_line)
+            lines_array.append(BORDER_LINE)
 
-        # If message is not empty, add it to the box
         if message:
-            # message must be a string
-            if not isinstance(message, str):
-                raise TypeError("Message must be a string")
-            # get message wrapped to lines
-            msg_lines = self._dedent_and_wrap(message, width)
+            msg_lines = self._dedent_and_wrap(message, width, border=2)
 
-            # add message lines to the box
-            lines_array.extend(msg_lines)
-
-            # closing line of message
-            lines_array.append(border_line)
+            lines_array.extend(["", *msg_lines, "", BORDER_LINE, ""])
 
         # merge lines into a single string and send warning to console
         self.warning("\n".join(lines_array))
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -205,13 +205,13 @@ class AndroidSDK(ManagedTool):
         if android_home:
             if android_sdk_root and android_sdk_root != android_home:
                 tools.console.warning_banner(
-                    title="ANDROID_HOME and ANDROID_SDK_ROOT are inconsistent",
-                    message=f"""
+                    "ANDROID_HOME and ANDROID_SDK_ROOT are inconsistent",
+                    f"""
                         The ANDROID_HOME and ANDROID_SDK_ROOT environment variables
                         are set to different paths:
 
-                            ANDROID_HOME:     {android_home}
-                            ANDROID_SDK_ROOT: {android_sdk_root}
+                          ANDROID_HOME:     {android_home}
+                          ANDROID_SDK_ROOT: {android_sdk_root}
 
                         Briefcase will ignore ANDROID_SDK_ROOT and only use the path
                         specified by ANDROID_HOME.
@@ -294,13 +294,12 @@ class AndroidSDK(ManagedTool):
                         f"""
                             The Android SDK specified by {sdk_source_env} at:
 
-                            {sdk_root_env}
+                              {sdk_root_env}
 
-                            does not contain Command-Line Tools
-                            version {cls.SDK_MANAGER_VER}.
-                            Briefcase requires this version to be installed to use
-                            an external Android SDK.
-                            Use Android Studio's SDK Manager to install it.
+                            does not contain Command-Line Tools version
+                            {cls.SDK_MANAGER_VER}. Briefcase requires this version to be
+                            installed to use an external Android SDK. Use Android
+                            Studio's SDK Manager to install it.
 
                             Briefcase will proceed using its own SDK instance.
                         """,
@@ -312,15 +311,15 @@ class AndroidSDK(ManagedTool):
                         The location pointed to by the {sdk_source_env} environment
                         variable:
 
-                        {sdk_root_env}
+                          {sdk_root_env}
 
-                        doesn't appear to contain an Android SDK with
-                        the Command-line Tools installed.
+                        doesn't appear to contain an Android SDK with the Command-line
+                        Tools installed.
 
-                        If {sdk_source_env} is an Android SDK, ensure it is
-                        the root directory of the Android SDK instance such that
+                        If {sdk_source_env} is an Android SDK, ensure it is the root
+                        directory of the Android SDK instance such that:
 
-                        ${sdk_source_env}{os.sep}{sdk.sdkmanager_path.relative_to(sdk.root_path)}
+                            ${sdk_source_env}{os.sep}{sdk.sdkmanager_path.relative_to(sdk.root_path)}
 
                         is a valid filepath.
 
@@ -498,8 +497,8 @@ its output for errors.
             )
         except (OSError, subprocess.CalledProcessError) as e:
             self.tools.console.debug(str(e))
-            self.tools.console.warning_banner(
-                "", f"Failed to install cmdline-tools;{self.SDK_MANAGER_VER}"
+            self.tools.console.warning(
+                f"Failed to install cmdline-tools;{self.SDK_MANAGER_VER}"
             )
             return False
         return True
@@ -721,8 +720,7 @@ connection.
             self.tools.console.warning_banner(
                 "Unexpected emulator ABI",
                 f"""
-                    The system image {system_image!r}
-                    does not match the architecture of
+                    The system image {system_image!r} does not match the architecture of
                     this computer ({self.emulator_abi}).
 
                     Briefcase will proceed assuming the emulator is correctly
@@ -1286,15 +1284,15 @@ Review the emulator output above for:
 
 Ensure your Android SDK is up-to-date by running:
 
-    $ briefcase upgrade {AndroidSDK.name}
+  $ briefcase upgrade {AndroidSDK.name}
 
 To review Google's general troubleshooting steps for the emulator, visit:
 
-    https://developer.android.com/studio/run/emulator-troubleshooting
+  https://developer.android.com/studio/run/emulator-troubleshooting
 
 You can also start the emulator manually by running:
 
-    $ {emulator_command}
+  $ {emulator_command}
 """
 
         failed_startup_error_msg = f"{{prologue}}\n{general_error_msg}"
--- a/tests/console/Console/test_warning_banner.py
+++ b/tests/console/Console/test_warning_banner.py
@@ -1,248 +1,220 @@
 import pytest
 
-test_title = "ANDROID_HOME and ANDROID_SDK_ROOT are inconsistent"
-test_message = """
+TEST_TITLE = "ANDROID_HOME and ANDROID_SDK_ROOT are inconsistent"
+TEST_MESSAGE = """
     The ANDROID_HOME and ANDROID_SDK_ROOT environment variables
     are set to different paths:
 
-        ANDROID_HOME:     /briefcase/tests/console/Console/test_warning_banner.py
-        ANDROID_SDK_ROOT: /home/anton/briefcase/tests/console/Console/\
-test_warning_banner.py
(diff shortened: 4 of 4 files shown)


src/briefcase/utils.py
```diff
@@ -0,0 +1,68 @@
...
+        message (str): The message to format.
+        title (str, optional): The title of the box. Defaults to None.
+        width (int, optional): The width of the box. Defaults to 80.
+        border_char (str, optional): The character to use for the border.
+            Defaults to "*".
+        padding (int, optional): The number of spaces to pad the border.
+            Defaults to 2.
+
+    Returns:
+        str: The formatted message.
```
freakboy3742: We use the [Sphinx docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html); check any other source file for examples of use.

src/briefcase/utils.py
```diff
@@ -0,0 +1,68 @@
...
+            # Center each line within the available space
+            padded_line = line.center(inner_width)
+            # add line to the box
+            lines.append(f"**{padded_line}**")
+
+        # closing line of title in the box
+        lines.append(border_line)
+
+    # split the message into paragraphs
+    paragraphs = message.split("\n")
```
freakboy3742: We should probably use `textwrap.dedent()` here as a convenience as well.

src/briefcase/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    for paragraph in paragraphs:
+        if paragraph.strip():  # Non-empty paragraph
+            # Wrap paragraph to lines to fit the width of the box
+            wrapped_lines = textwrap.wrap(paragraph, width=width)
+            lines.extend(wrapped_lines)
```
freakboy3742: If you check the existing banner warnings, the body message is indented relative to the actual text - so the width should be 8 chars less than the specified width, and there should be a 4 character left pad at the start of every line.

freakboy3742 (changes requested):
Thanks for the contribution! The broad strokes of this look like it's on the right track - but there's a few details to sort out. Some of them I've flagged inline; but there are a couple of bigger picture items.

Firstly, this is currently failing CI because of a town crier check; see [our contribution guide](https://briefcase.beeware.org/en/stable/how-to/contribute/code.html#submitting-a-pull-request) for details on what that means.

Secondly - we try to avoid generic "utils" modules. In this case, the code is a very specific addition to console handling - so it should be part of the console module. Given that this will always be used in the context of a `console.warning()`, it would make sense to me for this to be a standalone method on `Console` - something like `console.warning_banner(title, body, width=80)`. It may help to keep the "formatting" part of the code separate so that `console._banner()` returns a string, which is then output by `console.warning_banner()` (and provides an opportunity for `error_banner()` if we ever need it).

otekraden:
@freakboy3742 
Thank you for your feedback. 
As expected, it turned out to be not so easy. 
I've seen that the CI process crashes into an error, but I haven't figured out how to fix it yet. 
I will take all your comments into account and finalize this pull request in the near future.

src/briefcase/utils.py
otekraden: Got it! 
I applied this format.

src/briefcase/utils.py
otekraden: Sorry, but in my current understanding of the logic of my function, I don't see how to use the textwrap.dedent() method. Could you please show me how to integrate it?

otekraden:
I have attempted to make all the necessary changes based on the instructions above.
The function has been transformed into a static method "warning_banner" in the Console primitives section.
Tests have been relocated to the Console tests folder.
Currently, CI towncrier works fine, but there seems to be an issue deeper in the app building process.

otekraden:
I rewrote Console method from static to instance method.

src/briefcase/console.py
```diff
@@ -250,6 +250,75 @@ def export_log(self):
...
+        :type title: str, optional
+        :param width: The total width of the box in characters. Defaults to 80.
+        :type width: int, optional
+        :param border_char: Character to use for the box border. Defaults to "*".
+        :type border_char: str, optional
+        :param padding: Number of spaces between the border and text content. Defaults
+            to 2.
+        :type padding: int, optional
+        :return: The formatted message enclosed in a bordered box.
+        :rtype: str
```
freakboy3742: Apologies for the miscommunication here; but we don't need the type documentation - because they are (or should be) in the type annotations for the method itself.
```suggestion
        :param message: The message to format inside the box.
        :param title: The title of the box. If provided, appears centered at the top.
        :param width: The total width of the box in characters. Defaults to 80.
        :param border_char: Character to use for the box border. Defaults to "*".
        :param padding: Number of spaces between the border and text content. Defaults
            to 2.
        :return: The formatted message enclosed in a bordered box.
```

src/briefcase/console.py
```diff
@@ -250,6 +250,75 @@ def export_log(self):
         """Export the text of the entire log; the log is also cleared."""
         return self._log_impl.export_text()
 
+    def warning_banner(
+        self,
+        message: str,
+        title: str = "",
+        width=80,
+        border_char="*",
+        padding=2,
```
freakboy3742: Missing type annotations:
```suggestion
        width: int = 80,
        border_char: str = "*",
        padding: int = 2,
```

src/briefcase/console.py
```diff
@@ -250,6 +250,75 @@ def export_log(self):
...
+        border_line = border_char * width
+        # create lines array with opening line of the box
+        lines = [border_line]
+
+        # if title exists, format title in the box
+        if title:
+            # Wrap the title to lines to fit the width of the box
+            wrapped_title_lines = textwrap.wrap(
+                title.strip(),
+                width=width - (padding * 2) - 4,
```
freakboy3742: I'm not sure there's a use case for configurable padding here - 1 space of padding in the title will always be the default.

src/briefcase/utils.py
freakboy3742: What I'm suggesting is that being able to call:
```
        console.warning_banner(
            "Some banner title"
            """\
                This is the content that I want to display. It is long
                but I don't want to have to manually word wrap\n
                \n
                But manual line breaks should be preserved.
            """
        )
```

We *could* do this by requiring the user to wrap the `"""` string in `textwrap.dedent()`... or we could do that automatically on every usage.

freakboy3742 (changes requested):
Thanks for those cleanups. The CI failure you saw was unrelated - we had some disk space issues in our CI configuration. We've fixed those now, and I've re-run the affected tests, and they're passing now.

I've left a couple more review comments; the only other suggestion is that it could use a couple more tests.

In particular, this is a situation where parameterized tests would be helpful. There's a general pattern of a set of input. Then a need to test how that input renders at different widths - widths that are longer than the text, the same width as the text, less than the width of the text, and so on.

Lastly, it would be useful to actually use this API somewhere. It doesn't need to update *every* usage of banner-like output, but if you pick a module (like the Android integrations), and replace every warning banner with the equivalent usage of this utility, we can see that it will work in practice.

otekraden:
I have made the following changes: 
- added the ability to enter a message and title as a multiline or single line. At the same time, it is possible to manually split texts into paragraphs using "\\\n";
- did parametric tests;
- corrected docstring and arg types.

kattni:
@otekraden Hello! Thank you for submitting the requested changes. As you can see, the CI is failing. We maintain 100% code coverage in testing. The current CI issue appears to be related to the PR now being less-than 100% coverage. You can click on the failure to see exactly where the issue is. Please get this PR passing before we review it again. If you have questions or need help, please post here. Thanks!

otekraden:
@kattni @freakboy3742 
Hello! I need your help with tests coverage. 
I tried everything but string 324 in console.py not covered((
 
image

I will be grateful for any hint.

tests/console/Console/test_warning_banner.py
```diff
@@ -0,0 +1,212 @@
...
+
+source_title = f"""\
+                \\nINFO: I TRIED TO SOLVE THIS ISSUE #{issue_idx}.\\n\\n
+                IT SEEMED TO ME RATHER SIMPLE,
+                BUT I GOT VERY GOOD EXPERIENCE\
+"""
+
+
+@pytest.mark.parametrize(
+    ("test_name", "message", "title", "width", "expected"),
```
freakboy3742: There's no need to include a descriptive test name - your test doesn't use it anyway:
```suggestion
    ("message", "title", "width", "expected"),
```

If you *do* want to provide a test identifier, use [`pytest.param()`](https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids) to define each group of test data; the ID in that case should be something short (like "test-1" or "empty-message-title") that can be used to run a specific test. This isn't *required*, but it can be a nice convenience.

tests/console/Console/test_warning_banner.py
```diff
@@ -0,0 +1,212 @@
...
+name = "Anton"
+issue_idx = 2559
+
+source_msg = f"""\
+            Hi there! My name is {name}, and I'm a Python developer.\\n\\n
+            This is my very first
+            experience in open source
+            development.
+            I like your project, and I'm excited to contribute to it.\
+"""
```
freakboy3742: Is the use of an f-string testing anything specific here? I'm fairly sure this can be simplified to a static string definition.

tests/console/Console/test_warning_banner.py
```diff
@@ -0,0 +1,212 @@
...
+    [
+        ("Test 1: Empty message and title", "", "", 80, ""),
+        (
+            "Test 2: Default width (80) with title and message",
+            source_msg,
+            source_title,
+            80,
+            """\
+********************************************************************************
+**                                                                            **
```
freakboy3742: This extra blank link in the title is a little odd. It wouldn't match any existing uses of warning banners.

tests/console/Console/test_warning_banner.py
```diff
@@ -0,0 +1,212 @@
...
+    else:
+        msg = console.warning_banner(message, title=title, width=width)
+
+    # # Debug output if needed
+    # print('\n\n' + test_name)
+    # print("Generated:")
+    # print(msg)
+    # print("Expected:")
+    # print(expected)
+
```
freakboy3742: There's no need to retain this debug code.
```suggestion
```

src/briefcase/console.py
```diff
@@ -250,6 +250,97 @@ def export_log(self):
...
+        :param border_char: Character to use for the box border. Defaults to "*".
+        :return: The formatted message enclosed in a bordered box.
+        """
+
+        # If message and title are both empty or width is not an integer,
+        # return empty string
+        if not any((message, title)) or not isinstance(width, int):
+            return ""
+
+        def dedent_and_split(text):
```
freakboy3742: Rather than defining this as an inline function, it should be a private method on the base Console class (i.e. , `_dedent_and_split()`.

I'd also suggest that we can add an extra affordance around the handling of input strings. As written, I *think* this requires that the input string has a trailing `\` on the first line (i.e., `warning_banner(f"""\`. That's the sort of detail that will be consistently forgotten in practice. Since the effect of the `\` is to suppress the newline on the first character, it might be worth adding a `text.strip('\n')` so that any leading or trailing newlines are removed from the input string, ensuring that we're in complete control of the spacing around the message in the box.

src/briefcase/console.py
```diff
@@ -250,6 +250,97 @@ def export_log(self):
...
+        :param title: The title of the box. If provided, appears centered at the top.
+        :param width: The total width of the box in characters. Defaults to 80.
+        :param border_char: Character to use for the box border. Defaults to "*".
+        :return: The formatted message enclosed in a bordered box.
+        """
+
+        # If message and title are both empty or width is not an integer,
+        # return empty string
+        if not any((message, title)) or not isinstance(width, int):
+            return ""
```
freakboy3742: Wouldn't it make more sense to raise an *error* if the arguments aren't the right type? 

It's also worth asking if any error handling needed *at all*. This isn't a public API - it's an internal utility; we don't have to be *rock solid* in the API we expose, because we're in control of how the method is used.

src/briefcase/console.py
```diff
@@ -250,6 +250,97 @@ def export_log(self):
...
+
+        # Create border line
+        border_line = border_char * width
+        # create lines array with opening line of the box
+        lines = [border_line]
+
+        # if title exists, format title in the box
+        if title:
+            if not isinstance(title, str):
+                return ""
```
freakboy3742: If we're doing error handling, we should be raising an error, not failing silently.

But again - is any error handling needed?

src/briefcase/console.py
```diff
@@ -250,6 +250,97 @@ def export_log(self):
...
+                        f"{border_char * 2}{''.center(inner_width)}{border_char * 2}"
+                    )
+
+            # closing line of title in the box
+            lines.append(border_line)
+
+        # If message is not empty, add it to the box
+        if message:
+            if not isinstance(message, str):
+                return ""
```
freakboy3742: As above - is this error handling needed; and if it is, shouldn't it be error handling?

freakboy3742 (changes requested):
This is looking good! A few review comments inline.

Regarding the coverage failure - the error is telling you that a specific line of code isn't being exercised by the test suite. The fix is to add a test case that will exercise that line of code - in this case, the case of a message that isn't a string. 

That said - as noted by my review comments, that line of code probably isn't needed, because we can assume the method is being used by a "reasonable person" who is working with Briefcase. It doesn't necessarily need to be robust enough to handle any random input from an end user.

The last suggestion - at present, this PR adds a new feature, and does a good job testing; however, it would be good to see a few examples of the API being used in practice. You don't need to update *every* banner in the codebase - but if you pick one module (say, the Android SDK integrations in `src/briefcase/integrations/android_sdk.py`) and replace all banners with usage of this API, you can prove that the API is "fit for purpose".

otekraden:
Good afternoon,

I've completed all the requested code edits. 
My apologies for the delay — I was occupied  at my primary job.

While implementing the changes, I encountered an issue with "tox": I couldn't reach 100% test coverage. I discovered that running the "tox p" command after making edits allows the coverage to be recalculated correctly.

I tried to use created feature in the Android SDK integrations, but tests falling in errors((

freakboy3742:
> I've completed all the requested code edits. My apologies for the delay — I was occupied at my primary job.

No problem at all - we're grateful for any time you can volunteer.
 
> While implementing the changes, I encountered an issue with "tox": I couldn't reach 100% test coverage. I discovered that running the "tox p" command after making edits allows the coverage to be recalculated correctly.
 
I'm not sure I understand what you're describing here - you'll need to provide some more details here. The current test suite isn't reporting any coverage problems.

> I tried to use created feature in the Android SDK integrations, but tests falling in errors((

This is to be expected. You've made changes that alter the output of the Android SDK by switching to the new banner API; the tests that are failing are reporting that the output of some commands has changed. Once you've confirmed that the changes are expected, you'll need to update the tests to reflect the new output.

otekraden:
@freakboy3742 
I switched Android SDK to use new "warning_banner" method.

src/briefcase/integrations/android_sdk.py
```diff
@@ -204,27 +204,22 @@ def sdk_path_from_env(cls, tools: ToolCache) -> tuple[str | None, str | None]:
...
-    specified by ANDROID_HOME.
+    specified by ANDROID_HOME.\\n\\n
 
     You should update your environment configuration to either not set
     ANDROID_SDK_ROOT, or set both environment variables to the same
     path.
-
-*************************************************************************
-"""
+""",
```
freakboy3742: The title being expressed as a triple string over three lines is... very odd; and the need for large numbers of `\\n` characters in the input for dialogs clearly indicates that the implementation isn't right. There may be places where a manual \n is required, but they should be the *extreme* exception, not the rule. 

At the very least, a blank line in the input should be interpreted as a blank line in the output, without any need for special handling. My immediate impression is that the simple "replace '\n' with ' '" isn't right; we should only be replacing a newlines where the next character isn't a newline or whitespace. That would preserve both the "layout" blank lines that are common at present, as well as the ANDROID_HOME indented example here.

freakboy3742: Also - if the indentation isn't being preserved in output, then we can simplify and remove it from the input as well. Part of the reason of introducing this wizard is so that we can put the message "in context", rather than needing to have it in code the way we want it to be displayed.

otekraden:
@freakboy3742 Hello! I am making another attempt to complete work on this task.
I took into account all the above comments and made adjustments to the message generation logic.
Now the method calls look much better without using special characters like "\\n".

src/briefcase/integrations/android_sdk.py
```diff
@@ -274,36 +269,28 @@ def verify_install(
 
             if sdk.exists():
                 if sdk_source_env == "ANDROID_SDK_ROOT":
-                    tools.console.warning(
+                    tools.console.warning_banner(
+                        "WARNING: Using Android SDK from ANDROID_SDK_ROOT",
```
freakboy3742: This usage should be something like:
```suggestion
                    tools.console.warning_banner(
                        title="Using Android SDK from ANDROID_SDK_ROOT",
                        message=f"""\
                            Briefcase is using the Android SDK specified by the 
                            ANDROID_SDK_ROOT environment variable.

                            Android has deprecated ANDROID_SDK_ROOT in favor of 
                            the ANDROID_HOME environment variable.

                            Update your environment configuration to set 
                            ANDROID_HOME instead of ANDROID_SDK_ROOT to ensure 
                            future compatibility.                    
                        """
                    )
```

Note that if we're including "WARNING: " in every message, we can hard code that text.

otekraden:
I've fixed the method calls in "android_sdk".
I'm sorry that I didn't immediately show the correct calls with code alignment.
I was worried about the big length of the lines, so I shifted the texts to the left. 

In addition, I encoded the "WARNING:" prefix of the headers. 

I really hope that now my solution looks like what you require.

src/briefcase/console.py
```diff
@@ -250,6 +250,130 @@ def export_log(self):
...
+        message: str,
+        width: int = 80,
+    ) -> str:
+        """The title or message can be provided as a single or as multiline string. Any
+        common leading whitespace from each line will be removed.
+
+        To separate text into paragraphs you can use:
+        - blank line;
+        - relative indentation.
+        Lines with zero relative indentation will be merged to upper paragraph.
```
otekraden, line 324: I rewrote the logic of splitting the text into paragraphs.

The width of each paragraph is equal to the specified border width minus the relative indentation.

otekraden:
@freakboy3742 The result of the method is now as follows:



Paragraphs take up the full width of the box minus relative indentation.

src/briefcase/console.py
```diff
@@ -250,6 +250,130 @@ def export_log(self):
...
+
+        # character to use for the box border
+        border_char = "*"
+
+        # If message and title are both empty rase ValueError
+        if not any((message, title)):
+            raise ValueError("Message or title must be provided")
+        # If width is not an integer, raise TypeError
+        if not isinstance(width, int):
+            raise TypeError("Width must be an integer")
```
freakboy3742: There's no need to do this sort of error checking. It's an internal tool; we can rely on developers to use it correctly.

src/briefcase/console.py
```diff
@@ -250,6 +250,130 @@ def export_log(self):
...
+        Lines with zero relative indentation will be merged to upper paragraph.
+
+        :param title: The title of the box. If provided, appears centered at the top.
+        :param message: The message to format inside the box.
+        :param width: The total width of the box in characters. Defaults to 80.
+        :return: The formatted message enclosed in a bordered box.
+        """
+
+        # character to use for the box border
+        border_char = "*"
```
freakboy3742: Unless we're going to need to change this, it's a lot clearer to just use `*` where it's needed.

src/briefcase/console.py
```diff
@@ -250,6 +250,130 @@ def export_log(self):
...
+        # Create border line
+        border_line = border_char * width
+        # create lines array with opening line of the box
+        lines_array = [border_line]
+
+        # if title exists, format title in the box
+        if title:
+            # title must be a string
+            if not isinstance(title, str):
+                raise TypeError("Title must be a string")
```
freakboy3742: Again - we can rely on developers to use this right.

src/briefcase/console.py
```diff
@@ -250,6 +250,130 @@ def export_log(self):
...
+        # If width is not an integer, raise TypeError
+        if not isinstance(width, int):
+            raise TypeError("Width must be an integer")
+
+        # Create border line
+        border_line = border_char * width
+        # create lines array with opening line of the box
+        lines_array = [border_line]
+
+        # if title exists, format title in the box
```
freakboy3742: Be wary of using excessive comments - they end up being more confusing than helpful.

src/briefcase/console.py
```diff
@@ -250,6 +250,130 @@ def export_log(self):
...
+
+        # Add a warning prefix for title
+        if is_title:
+            text = f"WARNING: {text.strip()}"
+
+        # Split text into text lines
+        input_lines = text.split("\n")
+
+        # create paragraphs from lines
+        paragraphs = []
```
freakboy3742: This is an area where comments *would*  have been helpful - because "paragraphs" isn't a list of paragraphs - it's a list of lists, where each element is a line, and the "rel_indent" value. 

However, this entire logic can be a lot simpler.

src/briefcase/console.py
```diff
@@ -250,6 +250,130 @@ def export_log(self):
...
+    def _dedent_and_wrap(text, wrap_width, is_title=False):
+        """Dedent text, split text into paragraphs and wrap each paragraph to lines of
+        the specified width."""
+
+        # Dedent and remove common leading empty lines
+        text = textwrap.dedent(text).strip("\n")
+
+        # Add a warning prefix for title
+        if is_title:
+            text = f"WARNING: {text.strip()}"
```
freakboy3742: This is the wrong place to add WARNING - add it as part of the *input* not the wrapping mechanism.

changes/2563.misc.md
```diff
@@ -0,0 +1 @@
+``Warning_banner`` method added to ``Console`` class.
```
freakboy3742: This is markdown, not RST (i.e., use single quotes).

tests/console/Console/test_warning_banner.py
```diff
@@ -0,0 +1,248 @@
...
+    specified by ANDROID_HOME.
+
+    You should update your environment configuration to either
+    not set ANDROID_SDK_ROOT, or set both environment variables
+    to the samepath.
+"""
+
+
+@pytest.mark.parametrize(
+    ("message", "title", "width", "expected"),
```
freakboy3742: Given the function takes "title" then "message", it makes sense to preserve that ordering in tests.

tests/console/Console/test_warning_banner.py
```diff
@@ -0,0 +1,248 @@
...
+
+ Briefcase will ignore ANDROID_SDK_ROOT and only use the path specified by
+ ANDROID_HOME.
+
+You should update your environment configuration to either not set
+ANDROID_SDK_ROOT, or set both environment variables to the samepath.
+********************************************************************************
+""",
+        ),
+        # Custom width (60) with title and message
```
freakboy3742: This test case isn't really testing anything new - it's "just another width". Test cases should always be exercising something new.

freakboy3742 (approved):
Thanks for those updates. I've pushed some updates to simplify the implementation and remove some safety checks that aren't needed, and slightly tweak the presentation for some content (e.g, adding padding around the message content, and preserving indented lines, as they're an indicator of shell literal content).

This will be a big improvement for future developers - thanks!

PR #2567, opened 2025-11-20
Document that the install directory is configurable.

### LINKED ISSUE #2553: Allow the user to choose install location

It is currently not possible as a user to choose the exact install location, and the install directory is forced into one of two option based on the scope the user chooses.

As a user, after choosing the install scope, ideally I would like to be able to choose the install location as below:

Image

Of course the install location should still be limited to where the user is allowed to install into.

I have not considered any alternative solutions since this is not an uncommon feature in installers today.

### Additional context

_No response_

freakboy3742:
Documents the change in the MSI installers making the install directory configurable ~~, and the new environment variable exposed into the post-install script.~~

The actual change is entirely in templates:
* https://github.com/beeware/briefcase-windows-app-template/pull/78
* https://github.com/beeware/briefcase-windows-VisualStudio-template/pull/72

Fixes #2553

## PR Checklist:

### DIFF
--- a/docs/en/reference/platforms/windows/index.md
+++ b/docs/en/reference/platforms/windows/index.md
@@ -191,7 +191,7 @@ A Boolean describing the initial value of the option in the GUI. If not provided
 
 ### Using installer options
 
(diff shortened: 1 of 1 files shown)


freakboy3742:
I see it is "uncharitable thoughts towards SuSE" season again...

PR #2575, opened 2025-11-22
Multi app list

### LINKED ISSUE #2574: In dev/run, provide multiple app list as option, rather than an error

If a project defines a single app, `briefcase dev` and `briefcase run` will start that app.

If a project defines more than one app, `briefcase dev` and `briefcase run` will raise an error. It is necessary to use `--app` to specify the app to run/dev. 

If a project has more than one app, and no app is explicitly specified, a list of options should be presented to the user to select (similar to how the list of simulators is shown to the user if no simulator is selected).

The default app should be the first one in the file.

If `--no-input` is specified, an error should be raised.

Always run the first app, but show a warning to the user in this case.

Don't expose an "menu" option, but allow the user (or the project configuration) to specify a "default" app as a user-level option (see also #2279)

### Additional context

The implementation of this shouldn't require much more than a call to `self.console.selection_question()`, with the list of options being constructed from the list of available apps. 

The options should be a dictionary of `{app name: formal name}`.

moondial-pal:
Adds an interactive app-selection menu to briefcase dev and briefcase run when multiple applications are defined, with proper handling of --no-input.

Previously, these commands raised an error when multiple apps existed without --app; this update guides users by presenting a selectable list instead.

Fixes #2574 

## PR Checklist:

### DIFF
--- a/tests/commands/dev/test_call.py
+++ b/tests/commands/dev/test_call.py
@@ -152,26 +152,7 @@ def test_no_args_one_app(dev_command, first_app):
     ]
 
 
-def test_no_args_two_apps(dev_command, first_app, second_app):
-    """If there are two apps and --no-input is provided, an error is raised."""
-    # Add two apps
-    dev_command.apps = {
-        "first": first_app,
-        "second": second_app,
-    }
-
-    # Simulate --no-input on the command line
-    options, _ = dev_command.parse_options(["--no-input"])
-
-    # In non-interactive mode, invoking the run command raises an error
-    with pytest.raises(BriefcaseCommandError):
-        dev_command(**options)
-
-    # Finalization will not occur
-    assert dev_command.actions == []
-
-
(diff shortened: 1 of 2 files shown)


moondial-pal:
Took a stab at this issue...failing a bunch of CI tests. Looking forward to feedback...I think the tests I updated are incomplete. I'll push an update soon to see if it will satisfy the CI.

moondial-pal:
I need to work on this. I'll re-submit when it is complete.

freakboy3742:
FWIW - from a quick overview, you're on the right track in the implementation; the CI failures are because you've added new code paths that aren't being tested. 

If you want to keep the PR open, and indicate that it's not ready for review yet, but you *are* intending to keep working on it, you can convert the PR to a draft.

moondial-pal:
Converted to draft - Going to keep working on this.

freakboy3742 (approved):
Awesome! I've made a couple of minor cosmetic tweaks, but otherwise, this is fantastic. Thanks for the patch!

PR #2582, opened 2025-11-26
Add support for uninstall options

### LINKED ISSUE #2573: Add support for uninstall options

#2496 added the option for pre-install and post-uninstall scripts. That PR also included the ability to define install options that can be selected during the installation sequence. 

However, there is no ability to configure options for *uninstallation*. 

It should be possible to define `uninstall_options`, comparable to `install_options`; those options should be displayed during uninstallation, and passed to the pre-uninstall script.

-

### Additional context

At least for now, this is only supported by MSI installers.

However, even there, there's some question about whether this is actually possible.

Apparently, Windows installer *doesn't* expose any dialog panels as part of a standard uninstall - it runs uninstalls in "quiet" mode. (Ref [[1]](https://github.com/orgs/wixtoolset/discussions/7637#discussioncomment-6900255), [[2]](https://stackoverflow.com/a/22321156), [[3]](https://stackoverflow.com/a/4857789)). 

In order to display *any* dialogs during an uninstall, you need to run the *original* MSI file.

freakboy3742:
A companion for #2496 - adds user-configurable pre-uninstall options as part of the uninstall process.

The GUI for these options will only be visible if the user re-runs the MSI installer. This appears to be a "feature" of the Windows "remove software" tooling - even if you modify the uninstaller registry entry to call `msiexec /qf /I...`, the GUI is not displayed. The default value for options will be honoured, though.

As with #2496 - most of the heavy lifting is actually done by the templates, not this PR:
* beeware/briefcase-windows-app-template#79
* beeware/briefcase-windows-visualstudio-template#74

Fixes #2573.

## PR Checklist:

### DIFF
--- a/docs/en/reference/platforms/windows/index.md
+++ b/docs/en/reference/platforms/windows/index.md
@@ -173,7 +173,7 @@ Installer and uninstaller options are defined using a TOML array of tables. Up t
 
 ### `install_option.name` / `uninstall_option.name`
 
-An identifier for the option. This name must be a valid Python identifier; the list of options specified for an app must be unique when converted into upper case (i.e., you cannot have `value` and `VALUE` in the same configuration). Names mus be unique across both `install_options` *and* `uninstall_options`
+An identifier for the option. This name must be a valid Python identifier; the list of options specified for an app must be unique when converted into upper case (i.e., you cannot have `value` and `VALUE` in the same configuration). Names must be unique across both `install_options` *and* `uninstall_options`.
 
 ### `install_option.title` / `uninstall_option.title`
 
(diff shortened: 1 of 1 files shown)


docs/en/reference/platforms/windows/index.md
```diff
@@ -165,34 +165,38 @@ Briefcase will attempt to convert your [`version`][] into a valid MSI value by e
...
 
 A Boolean describing the initial value of the option in the GUI. If not provided, defaults to `False`.
 
 ### Using installer options
 
 When an installer option is defined, the value of the option will be made available to the post-install script as an environment variable. For example, if your installer defines an option with a name of `foo`, an environment variable of `OPTION_FOO` will be defined, with a value of 1 if the option has been selected by the user, and 0 if the option has not been selected. The `ALLUSERS` environment variable will also be set; its value will be 1 if the app has been installed for all users, or 0 if it has only been installed for the current user.
 
+### Using uninstaller options
+
+Uninstaller options are handled in a similar way to installer options, but they are provided to the pre-uninstall script. An uninstall option of `foo` will result in an environment variable named `UNINSTALL_FOO` being defined.
```
mhsmith: This naming seems inconsistent: either both install and uninstall scripts should use `OPTION_FOO`, or one should use `INSTALL_FOO` and the other `UNINSTALL_FOO`. 

Making them both use `OPTION_FOO` seems easier to remember. I guess the reason you didn't do that was because the install and uninstall properties are both in the same namespace. It's possible that could be worked around by making the default values conditional on whether we're running in install or uninstall mode, but an even easier solution would be to require no duplication between the install and uninstall option names.

freakboy3742:
> End users usually delete installers after installation, so this makes the whole feature fairly useless.

Agreed that it's a pretty bad situation. 
 
> Ideas around the registry entry:
> 
> * I think the uninstall option is /x rather than /i; does that make any difference?

The default configuration outputs a /x registry entry for uninstall. Switching to /i (by manually editing the registry entry) doesn't alter the behavior of the Uninstall button.

> * Is it possible the previous value of the registry entry is being cached in memory? A reboot should clear that. Or [this approach](https://stackoverflow.com/a/10324938) could be used to set the key to the correct value from the start.

I've tried manually changing the registry entry and then rebooting; that doesn't change the behavior.

> Alternatively, could we change the [`ARPNOREMOVE`](https://learn.microsoft.com/en-us/windows/win32/msi/arpnoremove) and ARPNOMODIFY properties so that the app has a Modify menu item rather than an Uninstall one?

I've tried this, and... enabling ARPNOREMOVE also disables the Modify button. I can't work out any way to enable *only* the Modify button. 

I guess that leaves open the option of having both enabled, with "modify" being used as "uninstall with options"... but that's not exactly intuitive either. It's probably the best of a bad set of options, though...

docs/en/reference/platforms/windows/index.md
freakboy3742: You're correct that avoiding the collision was the primary motivating factor. I'm not wild about `INSTALL_FOO`, and conditional default values is going to be complicated to make work, so I'll add code to require unique option names.

docs/en/reference/platforms/windows/index.md
```diff
@@ -173,7 +173,7 @@ Installer and uninstaller options are defined using a TOML array of tables. Up t
 
 ### `install_option.name` / `uninstall_option.name`
 
-An identifier for the option. This name must be a valid Python identifier; the list of options specified for an app must be unique when converted into upper case (i.e., you cannot have `value` and `VALUE` in the same configuration).
+An identifier for the option. This name must be a valid Python identifier; the list of options specified for an app must be unique when converted into upper case (i.e., you cannot have `value` and `VALUE` in the same configuration). Names mus be unique across both `install_options` *and* `uninstall_options`
```
mhsmith:
```suggestion
An identifier for the option. This name must be a valid Python identifier; the list of options specified for an app must be unique when converted into upper case (i.e., you cannot have `value` and `VALUE` in the same configuration). Names must be unique across both `install_options` *and* `uninstall_options`.
```

PR #2592, opened 2025-12-05
Verify system requirements when running `dev` on Linux

### LINKED ISSUE #1137: system_requires aren't verified when building system packages locally

freakboy3742:
Fixes #1137.

During live tutorials, a common failure mode for Linux users is to *not* have the required system dependencies installed; this then manifests as obscure errors during the `briefcase create` process when pygobject can't be compiled because of a pycairo dependency. However, unless you know what you're looking at, it's an extremely impenetrable error message.

Historically, this would have been difficult to resolve because `dev` wasn't platform aware; however, as of #2419, it is. This means we're now in a position to use the platform-specific tooling of the `system` backend to check system dependencies.

This PR modifies the `briefcase dev linux system` command (which is the default `briefcase dev` on Linux) to invoke the system package checks prior to installing development dependencies. Since a fresh Briefcase Hello World that specifies a Toga backend will have all the platform dependencies defined, this means you'll get a notification that you need to `apt install` (or whatever) a specific set of packages if they're not available.

It also reorganizes the Linux system tooling; the  "Passive" and "Mostly Passive" commands might have meant something historically, but that distinction no longer exists - there are no uses of "Passive". This means we can have a base "System" Mixin that only operates locally, and is really only used by dev, but has all the tools to verify local system behavior; and a "Docker" mixin that extends "System" adding Docker-specific checks and flags.

## PR Checklist:

### DIFF
--- a/tests/platforms/linux/system/test_dev.py
+++ b/tests/platforms/linux/system/test_dev.py
@@ -65,7 +65,7 @@ def test_dev_system_app_starts(dev_command, first_app, tmp_path):
     assert (
         Path(popen_args[0][0])
         .parts[-3]
-        .startswith("dev.cp-" if sys.platform == "win32" else "dev.cpython-")
+        .startswith("dev.cp3" if sys.platform == "win32" else "dev.cpython-3")
     )
     assert Path(popen_args[0][0]).parts[-2] == (
         "Scripts" if sys.platform == "win32" else "bin"


freakboy3742:
This is currently failing 4 tests because the CI environment doesn't contain system packages that are flagged as runtime requirements. They are both optional (apps will *run* without libcanberra, for example); but since they're listed as requirements, we should be installing them before trying to run the app. beeware/.github#273 should address this; once that has been merged, a re-run of those 4 CI tasks should pass.

mhsmith:
The .github PR has fixed the Toga tests, but not the Pyside ones.

freakboy3742:
The PySide failures were caused because although you can *install* `libglib2.0-0`, Debian *actually* installs `libglib2.0-0t64`. 

However... I can't work out why that's on the list to be installed. The Qt requirements are currently based [on a list from Qt's docs](https://doc.qt.io/qt-6/linux-requirements.html) - but reading that list closer, it seems like that's actually a list of *development* requirements not runtime requirements. The PySide requirements installed in CI is a lot more conservative - and has clearly been working fine, including starting an app. 

Even that list looks like it might be overkill - I just span up a fresh VM, disabled *all* the requirements in the pyproject.toml file, and the app started fine. Comparing with the list of requirements that is installed on CI, there's *some* overlap, but there's a lot more in the bootstrapped list. 

I'm going to have to tinker with this a bit more.

freakboy3742:
The changes in #2594 should correct the PySide6 issue.

rmartin16:
I was trying to test this but it seems that Briefcase is enforcing system python much more strictly now. Wasn't sure if that was the intention.
```
tmp/beeware/helloworld via 🐍 v3.10.15 (briefcase-3.10) 
❯ briefcase run

The version of Python being used to run Briefcase ('3.10.15 (main, Oct 10 2024, 10:31:10) [GCC 11.4.0]') is not the system python3 ('3.10.12 (main, Nov  4 2025, 08:48:33) [GCC 11.4.0]').
```

freakboy3742:
> I was trying to test this but it seems that Briefcase is enforcing system python much more strictly now. Wasn't sure if that was the intention.

So.. it *was* the intention... part of the bigger "you have to use a system provided Python" requirement. 

I know we've spoken about loosening that requirement, but I have read that as "you should be able to use a system provided Python that isn't the system python3".

I presume you're using uv or pyenv (or similar) to get a Python version that is the same major/minor?

rmartin16:
> I presume you're using uv or pyenv (or similar) to get a Python version that is the same major/minor?

Right; pyenv has been managing my Python installs for many years now.

freakboy3742:
I've merged the Qt changes from #2594 into this branch; and I've also added logic to use `apt-cache` to convert virtualised packages (like `libqt6gui6`, or any other `-t64` package) to their "native" form. 

This is currently running on the beeware/.github#274 branch, as that installs the simplified qt6 packages.

The only outstanding issue from #2594 should be the state of requirements on Ubuntu 22.02; I've been able to run with the requirements listed in this PR, so I'm not sure what is causing the extra warnings that @rmartin16 was seeing there.

.github/workflows/ci.yml
```diff
@@ -175,7 +175,8 @@ jobs:
   verify-projects:
     name: Verify project
     needs: [ package, unit-tests ]
-    uses: beeware/.github/.github/workflows/app-create-verify.yml@main
+    # uses: beeware/.github/.github/workflows/app-create-verify.yml@main
+    uses: beeware/.github/.github/workflows/app-create-verify.yml@qt-requires
```
mhsmith: Just leaving this here as a reminder to revert this change after the .github PR is merged but before this PR is merged.

.github/workflows/ci.yml
```diff
@@ -188,7 +189,8 @@ jobs:
   verify-apps:
     name: Build app
     needs: [ package, unit-tests ]
-    uses: beeware/.github/.github/workflows/app-build-verify.yml@main
+    # uses: beeware/.github/.github/workflows/app-build-verify.yml@main
+    uses: beeware/.github/.github/workflows/app-build-verify.yml@qt-requires
```
mhsmith: See above.

mhsmith (commented):
Looks reasonable to me. @rmartin16: are you able to test it?

rmartin16:
Trying to test as much as I can tonight.

The first thing I've encountered is an order of operations issue. When I run `briefcase dev` and get the `missing system dependencies` message, after I install the requirements and re-run `briefcase dev`, I get an error for `ModuleNotFoundError: No module named 'PySide6'`. So, despite the missing dependencies error, Briefcase later believes the app's requirements have already been installed.

freakboy3742:
> Trying to test as much as I can tonight.
> 
> The first thing I've encountered is an order of operations issue. When I run `briefcase dev` and get the `missing system dependencies` message, after I install the requirements and re-run `briefcase dev`, I get an error for `ModuleNotFoundError: No module named 'PySide6'`. So, despite the missing dependencies error, Briefcase later believes the app's requirements have already been installed.

Ah - that's an interesting one. I put the system dependencies check as part of the installation of requirements so that it wouldn't be required on every run... but that then means the check is performed *after* the virtual environment is created, and the pre-existence of the virtual environment is used to determine if the requirements should be created.

I can think of a couple of possible fixes for this; let me poke around and see what works cleanly.

src/briefcase/bootstraps/pyside6.py
```diff
@@ -79,26 +79,10 @@ def pyproject_table_linux_system_debian(self):
...
-    "libice6",
-    "libglib2.0-0",
     "libgl1",
-    "libegl1",
-    "libdbus-1-3",
+    "libqt6dbus6",
+    "libqt6gui6",
+    "libxcb-cursor0",
 ]
 """
```
rmartin16, line 93: For ubuntu 22.04, I still needed to install these packages to run the app:
```
apt install \
    libxkbcommon-x11-0 \
    libxcb-icccm4 \
    libxcb-shape0 \
    libxcb-keysyms1
```
I even saw this with `briefcase run --target ubuntu:22.04`. I wouldn't be surprised if 22.04 is just missing dependencies on the qt packages.

rmartin16:
> Ah - that's an interesting one. I put the system dependencies check as part of the installation of requirements so that it wouldn't be required on every run... but that then means the check is performed _after_ the virtual environment is created, and the pre-existence of the virtual environment is used to determine if the requirements should be created.
> 
> I can think of a couple of possible fixes for this; let me poke around and see what works cleanly.

Also hit the same for `briefcase run`.

rmartin16:
ok; other than ubuntu 22.04, i didn't have problems with other Ubuntu or Debian versions. Also, worked on arm64 debian.

src/briefcase/bootstraps/pyside6.py
freakboy3742: Huh - that looks like an oddity of 22.04. Debian 11/bookworm [definitely lists those packages as dependencies](https://packages.debian.org/bookworm/libqt6gui6), but it looks like [the dependency list for 22.04 is different](https://launchpad.net/ubuntu/jammy/amd64/libqt6gui6/6.2.4+dfsg-2ubuntu1). I'll add them in with a note.

freakboy3742:
> > Ah - that's an interesting one. I put the system dependencies check as part of the installation of requirements so that it wouldn't be required on every run... but that then means the check is performed _after_ the virtual environment is created, and the pre-existence of the virtual environment is used to determine if the requirements should be created.
> > I can think of a couple of possible fixes for this; let me poke around and see what works cleanly.

Ok - I've fixed this for `dev`; if installing requirements fails, it now cleans the dev venv. That means the next run will recreate the venv, which means another attempt to install requirements will happen.
 
> Also hit the same for `briefcase run`.

Yeah - that's a pre-existing problem though. The exact conditions have changed slightly because we're now testing *runtime* requirements as well; but there's always been the issue that requirements are only installed during `create`, and if `create` fails, the directory structure still exists, so running `briefcase run` doesn't detect that requirements are out of date. This is really a manifestation of #807.

.github/workflows/ci.yml
freakboy3742: Reverted.

.github/workflows/ci.yml
freakboy3742: Reverted.

PR #2593, opened 2025-12-05
Use isolated development environments for dev mode

### LINKED ISSUE #1735: Use an isolated environment for `briefcase dev`

At present, `briefcase dev` uses the currently active virtual environment to run a project in developer mode.

However:
* if you don't have an active virtual environment, app dependencies will be installed into the global environment. 
* If you use the same virtual environment for multiple projects, it's easy for dependencies to conflict. 
* The development environment will include packages (including briefcase itself, and setuptools_scm) that may not exist in the packaged runtime environment, which can lead to inconsistencies.

`briefcase dev` should create and use a standalone virtual environment for development mode. This would ensure that each app has unique and isolated dependencies. 

Do nothing. 

### Additional context

The most notable impact of this approach is that dependencies installed in the briefcase environment will no longer be visible to development mode. This complicates the process of adding dependencies to a project; however, it would be consistent with the behavior of `briefcase run` - so arguably this is a good thing, as it would avoid the "package X is in my environment, why doesn't `briefcase run` work" questions. 

If this change is made; we may want to consider to additional related changes:
1. a `--no-isolation` option that reverts to the current "use the briefcase environment" behavior
2. a new `briefcase add` command to manage adding new dependencies

An alternative to (2) would be #1367 (or similar) - i.e., provide better integration with other tools that manage dependencies. 

It will also impact on #881/#1714-style build tracking; but likely in a positive way, as isolation will remove some edge cases in keeping the environment current.

freakboy3742:
Fixes #1735 

Modifies the `dev` command to use an isolated virtual environment by default. This means an app in dev mode runs in a "clean" virtual environment, only containing the dependencies needed to run the app. 

The old behavior of using the current venv can be restored by using the `--no-isolation` flag.

This PR also changes the name of the dev mode environment to include a platform identifier; this is a convenience so that it is possible for a project checkout to be shared across multiple Python versions and OS versions. This is mostly a convenience for the BeeWare team - I have a testbed app that I need to test on multiple OS versions and Python versions, so it helps to have a separate venv for each one, rather than re-creating every time you change OS or Python version. Most users will only ever have a single isolated development venv per project.

This will require a change to the BeeWare tutorial to adjust to the fact that `pip install` will no longer change the runtime environment for dev mode. This is also enough of a change to expectations about Briefcase that it warrants our next release being 0.4.0.

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/dev.py
+++ b/src/briefcase/commands/dev.py
@@ -234,7 +234,7 @@ class DevCommand(RunAppMixin, BaseCommand):
 
         The environment name is platform and Python version specific, so
         that multiple OSes and Python versions can share a `.briefcase`
-        folder. The name is based on the extension module filename that the 
+        folder. The name is based on the extension module filename that the
         platform uses (e.g., cpython-313-darwin).
 
         :returns: Name for virtual environment directory


src/briefcase/commands/dev.py
```diff
@@ -222,9 +232,15 @@ def get_environment(self, app: AppConfig):
     def venv_name(self) -> str:
         """Returns the name of the virtual environment directory.
 
+        The environment name is platform and Python version specific, so
+        that multiple OSes and Python versions can share a `.briefcase`
+        folder. The name is based on the binary wheel name that the platform
+        uses (e.g., cpython-313-darwin).
```
mhsmith: Wheel filenames are different to this.

```suggestion
        folder. The name is based on the extension module filename that the 
        platform uses (e.g., cpython-313-darwin).
```

src/briefcase/platforms/web/static.py
```diff
@@ -779,22 +779,13 @@ class StaticWebDevCommand(StaticWebMixin, DevCommand):
     def venv_name(self) -> str:
         """Returns the name of the virtual environment directory.
 
+        The web environment uses a different development venv a web deployment has
```
mhsmith:
```suggestion
        The web environment uses a different venv because a web deployment has
```

PR #2594, opened 2025-12-06
Simplify Qt runtime requirements on Debian.

freakboy3742:
Simplify the Qt runtime requirements for Debian. This is born out of the additional runtime requirement checks added #2592.

This makes 2 changes to Qt dependencies:

1. It switches to using direct libqt6 dependencies (plus 2 other packages needed for desktop integration)
2. It introduces a separate configuration for Debian bookworm.

The former means the list of dependencies is much shorter, and Qt specific. Strictly, the actual Qt libraries will be provided by the binary PySide6 wheel - that's why we don't need to specify `libqt6widgets6`. But for the cost of some duplicated qt6 libraries (which are quite likely already installed anyway), we get all the required dependencies of libqt6 specified by Debian's own packaging system.

The latter is needed because of the 64-bit time transition. Debian Trixie introduces t64 suffixes on any package that uses `time_t`; although you can *install* `libqt6widgets6`, the actual package that is installed is `libqt6widgets6t64`, and the name *doesn't* round trip with dkpg checks (which is what #2592 uses to verify requirements are installed). The alternative to this would be to switch Debian testing to Trixie, and drop support for Bookworm for PySide; the duplicated config seemed less invasive.

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,23 +175,20 @@ jobs:
   verify-projects:
     name: Verify project
     needs: [ package, unit-tests ]
-    # uses: beeware/.github/.github/workflows/app-create-verify.yml@main
-    uses: beeware/.github/.github/workflows/app-create-verify.yml@qt-requires
+    uses: beeware/.github/.github/workflows/app-create-verify.yml@main
     with:
       runner-os: ${{ matrix.runner-os }}
       framework: ${{ matrix.framework }}
     strategy:
       fail-fast: false
       matrix:
-        # framework: [ "toga", "pyside6", "pygame", "console" ]
-        # runner-os: [ "macos-15-intel", "macos-latest", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest" ]
-        framework: [ "pyside6" ]
-        runner-os: [ "ubuntu-24.04", "ubuntu-24.04-arm" ]
+        framework: [ "toga", "pyside6", "pygame", "console" ]
+        runner-os: [ "macos-15-intel", "macos-latest", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest" ]
 
   verify-apps:
     name: Build app
     needs: [ package, unit-tests ]
-    uses: beeware/.github/.github/workflows/app-build-verify.yml@qt-requires
+    uses: beeware/.github/.github/workflows/app-build-verify.yml@main
     with:
       # Builds on Linux must use System Python; otherwise, fall back to version all GUI toolkits support
       python-version: ${{ startsWith(matrix.runner-os, 'ubuntu') && 'system' || '3.12' }}
@@ -200,7 +197,5 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # framework: [ "toga", "pyside6", "pygame", "console" ]
-        # runner-os: [ "macos-15-intel", "macos-15", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest" ]
-        framework: [ "pyside6" ]
-        runner-os: [ "ubuntu-24.04", "ubuntu-24.04-arm" ]
+        framework: [ "toga", "pyside6", "pygame", "console" ]
+        runner-os: [ "macos-15-intel", "macos-15", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest" ]


src/briefcase/bootstraps/pyside6.py
```diff
@@ -79,26 +79,18 @@ def pyproject_table_linux_system_debian(self):
...
 system_runtime_requires = [
-    # Derived from https://doc.qt.io/qt-6/linux-requirements.html
-    "libxext6",
-    "libxrender1",
-    "libx11-xcb1",
-    "libxkbcommon-x11-0",
-    "libxcb-image0",
+    "libgl1",
+    "libqt6dbus6",
+    "libqt6gui6t64",
```
rmartin16, line 84: Encounter a build error for Ubuntu 25.04, 25.10, or 26.04 as well as Debian trixie and sid:
```
Docker|  > [5/6] RUN apt-get update -y &&     apt-get install --no-install-recommends -y build-essential pkg-config git libgl1 libqt6dbus6 libqt6gui6t64 libxcb-cursor0:
Docker| 
Docker| 1.795 Building dependency tree...
Docker| 1.928 Reading state information...
Docker| 1.943 Package libqt6gui6t64 is not available, but is referred to by another package.
Docker| 1.943 This may mean that the package is missing, has been obsoleted, or
Docker| 1.943 is only available from another source
Docker| 1.943 However the following packages replace it:
Docker| 1.943   libqt6gui6
Docker| 1.943 
Docker| 1.948 E: Package 'libqt6gui6t64' has no installation candidate
```
Noticeably, though, Ubuntu 22.04, 24.04 and Debian Bookworm work.

src/briefcase/bootstraps/pyside6.py
```diff
@@ -79,26 +79,18 @@ def pyproject_table_linux_system_debian(self):
...
-    "libxcb-randr0",
-    "libxcb-xfixes0",
-    "libxcb-sync1",
-    "libxcb-icccm4",
-    "libxcb-keysyms1",
-    "libfontconfig1",
-    "libsm6",
-    "libice6",
-    "libglib2.0-0",
+]
```
rmartin16, line 86: While this builds on Ubuntu 22.04, the app won't run on 22.04.
```
root@ubuntu-22-04-wXBKb:/# pysidetest 
Detected locale "C" with character encoding "ANSI_X3.4-1968", which is not UTF-8.
Qt depends on a UTF-8 locale, and has switched to "C.UTF-8" instead.
If this causes problems, reconfigure your locale. See the locale(1) manual
for more information.
qt.qpa.plugin: From 6.5.0, xcb-cursor0 or libxcb-cursor0 is needed to load the Qt xcb platform plugin.
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: xcb, wayland-brcm, wayland-egl, wayland, eglfs, offscreen, vkkhrdisplay, minimal, linuxfb, vnc, minimalegl.

Aborted (core dumped)

```

One of these dependencies fixes that though.
```
root@ubuntu-22-04-wXBKb:/# apt install \
    libxcb-cursor0 \
    libxcb-xinerama0 \
    libxkbcommon-x11-0 \
    libxcb-icccm4 \
    libxcb-shape0 \
    libxcb-keysyms1 \
    libxcb-render-util0
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
libxcb-render-util0 is already the newest version (0.3.9-1build3).
libxcb-render-util0 set to manually installed.
libxcb-cursor0 is already the newest version (0.1.1-4ubuntu1).
The following additional packages will be installed:
  libxcb-xkb1
The following NEW packages will be installed:
  libxcb-icccm4 libxcb-keysyms1 libxcb-shape0 libxcb-xinerama0 libxcb-xkb1 libxkbcommon-x11-0
0 upgraded, 6 newly installed, 0 to remove and 0 not upgraded.
```

mhsmith:
Wouldn't it be easier to combine this with #2592 so they can verify each other in CI?

freakboy3742:
> Wouldn't it be easier to combine this with #2592 so they can verify each other in CI?

I pulled this into a separate PR because it felt like a separate concern; but you're right - there's definitely potential for a "both pass in isolation, neither pass in combination" confluence of events. I'll merge this change into #2592.

src/briefcase/bootstraps/pyside6.py
freakboy3742: Hrm. Thanks for those details. I'll go back to the drawing board on this one - I think we might need to do something with verifying the "virtual" libqt6gui6 package.

src/briefcase/bootstraps/pyside6.py
freakboy3742: That's weird - it complains about libxcb-cursor0, even though it's included...

freakboy3742: I've done some checking; out of that list, `libxcb-cursor0` and `libxcb-xinerama0` are the only two that aren't direct dependencies of `libqt6gui6` - all the others *should* be installed if you've got `libqt6gui6` installed. 

However, I can also run an app on 22.04 with `libxcb-xinerama0` missing (`libxcb-cursor0` is definitely required). 

I've moved the body of this work over to #2592; if you're able to provide any testing on that branch, it would be appreciated.

PR #2601, opened 2025-12-11
Restore missing section heading.

### LINKED ISSUE #2600: min_os_version documentation header has been lost
### Describe the bug

https://briefcase.beeware.org/en/stable/reference/configuration/

Before:

Image

After:

Image

### Additional context

If this happened during the Mkdocs transition, then we should check whether any other headings were lost in the same way. @kattni: FYI

freakboy3742:
Fixes #2600 

Restores the missing `min_os_version` section heading. Not sure how that went missing, but it's an easy fix.

I've audited the remaining setting names; the only discrepancies I found are settings that have been added in newer versions.

## PR Checklist:

### DIFF
--- a/changes/2600.misc.md
+++ b/changes/2600.misc.md
@@ -1,0 +1,1 @@
+A missing section heading was restored.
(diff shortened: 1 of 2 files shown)


kattni (approved):
Thanks for the fix!

PR #2627, opened 2026-01-10
docs: add signing identity examples and link to how-to guides

kshivani06:
I have updated the documentation for `--identity` in the package reference guide to provide better clarity on valid inputs for different platforms.
* **Indicative Examples:** Added examples for macOS (40-character SHA-1 hex strings/certificate names) and Windows (thumbprints/subject names) to show the "shape" of valid values.
* **Focused Content:** Maintained a concise reference guide by linking directly to the platform-specific how-to guides for detailed signing instructions rather than duplicating them.
* **Path Verification:** Verified that relative links correctly point to `docs/en/how-to/code-signing/` 
* **Scope Management:** Omitted any mention of distribution complexities or CI/GitHub Actions to keep the reference guide strictly focused on tool usage.

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,7 +175,7 @@ jobs:
   verify-projects:
     name: Verify project
     needs: [ package, unit-tests ]
-    uses: beeware/.github/.github/workflows/app-create-verify.yml@arch-pygame
+    uses: beeware/.github/.github/workflows/app-create-verify.yml@main
     with:
       runner-os: ${{ matrix.runner-os }}
       framework: ${{ matrix.framework }}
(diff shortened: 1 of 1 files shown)


changes/2627.doc
freakboy3742, line 1: A few details here for future reference:
* A changenote file should not be empty
* This should have a `.md` extension
* In this case, it's probably a `misc`, not a `doc`. Yes, it's changing documentation; but the purpose of changenote files is to be a first draft of the release notes. A relatively minor (however valuable) addition to docs isn't something that really needs to be in the release notes of the next version.

kshivani06: Thanks for the feedback and for the final tweaks! I've noted the point about using .misc for minor doc updates instead of .doc for future reference.

PR #2639, opened 2026-01-24
Add option to reset all runtime permissions on Android

timrid:
This PR adds the option `--reset-permissions` to reset all runtime permissions of an app on Android before it is launched.

I am currently working on porting the [BLE library bleak](https://github.com/hbldh/bleak) for Android with briefcase and am also creating integration tests for this. In the integration tests, I create a virtual Bluetooth peripheral using [bumble](https://github.com/google/bumble) and [connect it to the Android emulator](https://google.github.io/bumble/platforms/android.html) (netsim, to be precise) to test the reading/writing of characteristics, etc. However, I also test the Bluetooth permissions by confirming the Bluetooth Permission dialog via [uiautomator2](https://github.com/openatx/uiautomator2).

However, when I test this locally and confirm the dialog once, it remains permanently stored in the emulator. The tests run because the permissions have already been granted, but the code coverage is worse. Therefore, I need the option `--reset-permissions` to have a reproducible starting point for the app.

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,12 +80,12 @@ jobs:
           #   experimental: true
     steps:
     - name: Checkout
-      uses: actions/checkout@v6.0.1
+      uses: actions/checkout@v6.0.2
       with:
         fetch-depth: 0
 
     - name: Set up Python
-      uses: actions/setup-python@v6.1.0
+      uses: actions/setup-python@v6.2.0
       with:
         python-version: ${{ matrix.python-version }}
         allow-prereleases: true
@@ -134,12 +134,12 @@ jobs:
     if: always() && contains('success,failure', needs.unit-tests.result)
     steps:
     - name: Checkout
-      uses: actions/checkout@v6.0.1
+      uses: actions/checkout@v6.0.2
       with:
         fetch-depth: 0
 
     - name: Setup Python
-      uses: actions/setup-python@v6.1.0
+      uses: actions/setup-python@v6.2.0
       with:
         # Use minimum version of python for coverage to avoid phantom branches
         # https://github.com/nedbat/coveragepy/issues/1572#issuecomment-1522546425
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,7 @@ jobs:
           echo "VERSION=${GITHUB_REF_NAME#v}" | tee -a $GITHUB_ENV
 
       - name: Set up Python
-        uses: actions/setup-python@v6.1.0
+        uses: actions/setup-python@v6.2.0
         with:
           python-version: "3.x"
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
       - id: docformatter
         args: [--in-place, --black]
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.13
+    rev: v0.14.14
     hooks:
       - id: ruff-format
       - id: ruff-check
--- a/automation/pyproject.toml
+++ b/automation/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
     # keep versions in sync with ../pyproject.toml
-    "setuptools==80.9.0",
+    "setuptools==80.10.1",
     "setuptools_scm==9.2.2",
     "setuptools_dynamic_dependencies @ git+https://github.com/beeware/setuptools_dynamic_dependencies",
 ]
--- a/changes/2637.feature.md
+++ b/changes/2637.feature.md
@@ -1,0 +1,1 @@
+Post-install scripts for MSI installers now expose an `INSTALLER_PATH` environment variable.
--- a/changes/2641.misc.md
+++ b/changes/2641.misc.md
@@ -1,0 +1,1 @@
+The tests were updated for changes to new-line outputting in Rich 14.3.0 when using a progress bar in a non-interactive session.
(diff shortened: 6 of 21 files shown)


kattni:
@timrid Hello! Thanks for the contribution. The tests are failing due to an issue that was fixed earlier today. You'll need to rebase/merge `main` to get the fix. Be aware that you also need to update `rich` locally, or the tests will continue to fail locally. Let us know if you have any issues with this. Thanks!

mhsmith:
I'm not sure if this feature would be useful to enough people to justify adding it to Briefcase. Can't you achieve the same thing by running the `adb` command manually before calling Briefcase?

timrid:
> I'm not sure if this feature would be useful to enough people to justify adding it to Briefcase. Can't you achieve the same thing by running the `adb` command manually before calling Briefcase?

Calling `adb shell pm reset-permissions` unfortunately only works if the emulator is already running and the app is installed, but the app hasn't been started yet. If I do this while the app is already running and the app already had permissions that are now being revoked, calling `adb shell pm reset-permissions` leads to an abrupt termination of the app.

Actually this was my initial approach to call `adb shell pm reset-permissions` from the app itself. For controlling uiautomator2 from the Android app itself, I have already made a FastAPI server that runs on the host and is connected to the Android emulator via `--reverse-port`. Thus, I can manipulate the Android emulator itself via ADB from the pytests running in the Android emulator (e.g., to confirm permission dialogs). But if I also call `adb shell pm reset-permissions` there, then I see that the app terminates abruptly without any log messages.


An alternative approach is to start briefcase twice:
1. `briefcase run android -ru`: Start the Android emulator, install and run the app
2. `adb shell pm reset-permissions`: Reset permissions
3. `briefcase run android -no-update`: Rerun all the tests

For me, this is inconvenient, since I have to call briefcase twice in a script and also work manually with ADB. I could save myself both if this were built into briefcase.

Of course, I can understand that this is more of a niche. But perhaps this could also be used in other libraries for e.g. NFC or camera access.

freakboy3742:
I think there is value in having an easy option for "make this test environment as clean as possible" - especially for permissions. Yes, this *can* be done manually... but given it's such an obvious use case, it would be worth making the simple stuff trivial.

Part of my reason for saying this is that I can *never* remember the syntax and environment setup for running my local copy of ADB. At the very least, having a convenient "start ADB" entry point that does all the environment setup would be helpful.

mhsmith:
There are other adb commands that test scripts might want to run before running the app, such as uninstalling it, or wiping its data directory. So I thought of generalizing this using something like `briefcase run --Xadb-before=<command>`, following the same pattern as `--Xemulator` (#2652). 

But many of these commands would fail if the app wasn't already installed, and Briefcase would have no idea whether to treat the failure as fatal or not. This `--reset-permissions` option may have the same problem.

So I think it would make more sense to generalize this in the direction of #747, providing a way to call arbitrary adb commands separately from a `briefcase run`. Then test scripts can make their own decisions about whether a failure is fatal, perhaps by examining the command's output.

The command could look something like this:
* `briefcase tool run android_sdk adb [-d DEVICE_OR_AVD] ARGS ...`

Like `briefcase run`, this would prompt for a device selection if `-d` isn't present, and start an emulator if one is requested and it isn't already running.

freakboy3742:
> There are other adb commands that test scripts might want to run before running the app, such as uninstalling it, or wiping its data directory. 

This is true; but they're also "generic" things that you might want any app to do. Purging the data and cache directories in particular seems like the sort of thing you might want even a desktop app to do.

You could argue that starting the emulator, then installing and starting an app are things that you could "easily" do with the appropriate `adb` commands. However, Briefcase provides abstractions for these operations so a novice user doesn't need to learn all those details - they can focus on the actual task they care about: running their app.

I'd suggest that a novice user would understand that apps have persistent permission state, and persistent file storage state. However, the same user wouldn't have the first clue how to drive `adb` to achieve those things on Android. For common use cases, they shouldn't have to care, either. These are high-level concepts that transcend a specific platform's tooling. 

For me, the value of Briefcase is that end-users don't need to care about these details; they can focus on what they care about - running the app, without needing to learn and remember a dozen esoteric `adb` (or `simctl`, or any other tool) commands. To that end, having common options to flush permissions, file storage, etc absolutely makes sense to me, as they *could* be implemented on many platforms (or would be reasonable no-ops on platforms that don't support them).

> So I think it would make more sense to generalize this in the direction of #747, providing a way to call arbitrary adb commands separately from a `briefcase run`. 

I completely agree #747 would be a worthwhile addition to Briefcase; but I don't think it's a replacement for what has been proposed here.

If we went down the "assisted adb-use" path, I feel like we'd ultimately need to add a "cheat sheet" of common ADB commands in Briefcase's docs, because Android docs are *not* easy to consume at the best of times (nor are iOS's, for that matter), *especially* for novice users. And if we need to have a cheat sheet... why not encode the cheat sheet as a convenience?

mhsmith:
OK, that's a fair argument.

mhsmith:
OK, I don't know why there are so many references on the internet to a `-p` option of `pm reset-permissions`. This does not appear in the `pm` help message on API level 36, and based on my experiements, `reset-permissions` ignores its arguments and removes all permissions from all apps, not just the one being developed. This could be extremely annoying if used on the developer's personal device – for example, it will disable all notifications.

If we don't fix this before the next Briefcase release, this feature will need to be reverted. It looks like the appropriate package-specific command is actually `pm revoke`.

timrid:
Oh, that wouldn't be good, of course.

I can take a look at it next week (I'm currently on vacation) and if it's actually as you say, I'll work out an approach with adb revoke.

PR #2640, opened 2026-01-24
Add guidance for third-party bootstraps

### LINKED ISSUE #2342: Add option to aid discoverability of third-party Briefcase bootstraps

Via @tronical on #2340:

We currently encourage third-parties to publish their own Briefcase bootstrap plugins, rather than shipping them with Briefcase. This is good for Briefcase *maintainers*, because we don't have to stay on top of every GUI toolkit in the world; but it's a problem for the maintainers of those GUI toolkits and frameworks, because there's limited discoverability of options that aren't in Briefcase itself.

The selection of GUI toolkits should have a "second to last" option "Other frameworks (select to see options)". This menu option would then list other frameworks with know Briefcase plugins, and if you select one of them, provide installation instructions. 

There should also be a mild warning that the plugins on this secondary menu are not Briefcase's responsibility.

The current implementation provides a link to documentation for other frameworks. We could preserve this as a documentation issue; but given that the user is in the tool and has a question, we might as well solve the problem where they are.

### Additional context

_No response_

moondial-pal:
Add an “Other frameworks…” option to the GUI bootstrap selection shown during
`briefcase new`.

Users can now explicitly select a community provided GUI framework option,
receive guidance about support expectations, and exit cleanly with next steps.

Fixes #2342 

## PR Checklist:


kattni:
Hello! Thank you for the contribution. The majority of the failing tests are due to an issue unrelated to your PR. The fix has been submitted and merged into `main`. You'll need to `rebase` or `merge` `main` to get to a point where you can see which tests are failing related to your contribution. Please let us know if you need any help with that process. Thanks!

Please be aware that you'll also need to upgrade `rich` in your local environment or the tests will continue to fail locally for you

freakboy3742 (changes requested):
Thanks for the PR; however, from the look of it, this doesn't actually address the problem that was described in the ticket.

This PR *does* add another menu item for "other frameworks"; however, it doesn't provide any real information beyond that. The menu prompt says "select to see options"; the implication of this is that you will... see options. As currently implemented, all you get is text that tells you other options might exist, which is not really any better than the prompt that is already provided.

The purpose of the feature request is to help resolve the discovery issue - to let people know that there are other options they *could* install, *and* how to install them.

For example, right now, `Toga Positron` is a plugin option. If you select option 6, that option should be listed; if you select it, you should be told to install `toga-positron`. 

If `toga-positron` *is* installed, that option shouldn't appear in "other options". 

The expectation would be that we'd accept contributions from third-parties that want their plugins listed, and we prime that list with the plugins that are already listed in documentation.

Let me know if you need any more clarification on this.

moondial-pal:
Understood, thanks for the feedback.
I'll continue working on this to add a secondary menu that lists known third-party bootstrap plugins, provides install instructions, and hides options that are already installed. I’ll follow up once I have something closer to that behavior.

moondial-pal:
@kattni Thanks for the heads up! I’ve rebased the branch onto the current main as suggested and confirmed the full test suite passes locally. I’m still seeing CI failures on the PR, so I’m taking another look in case there’s anything else I missed, please let me know if you spot anything on your end.

moondial-pal:
It’s my understanding that community GUI bootstraps are considered installed
only if they expose a `briefcase.bootstraps` entry point. I may be missing
something, but PPB does not appear to publish such an entry point currently.

I tested both PPB and pygame-ce locally: pygame-ce was detected as installed
and correctly hidden from the community submenu, while PPB continued to
appear. I’ve left PPB in the list because it is documented as a known
community GUI bootstrap.

freakboy3742:
> It’s my understanding that community GUI bootstraps are considered installed only if they expose a `briefcase.bootstraps` entry point. I may be missing something, but PPB does not appear to publish such an entry point currently.
> 
> I tested both PPB and pygame-ce locally: pygame-ce was detected as installed and correctly hidden from the community submenu, while PPB continued to appear. I’ve left PPB in the list because it is documented as a known community GUI bootstrap.

Your understanding is correct; and yes, it looks like PPB isn't actually publishing a bootstrap at this point.

I'll need to follow up with the PPB team; IIRC, the entry was added during (or related to activity at) PyCon US sprints; it's possible the PR wasn't finalized or merged. If it doesn't work, there's no point leaving it on the list.

moondial-pal:
Got it. I'll go ahead and remove ppb from the list for this PR. 
I can submit an issue/pr to update the docs for third-party plugins to remove ppb from there as well, or should we leave that as is?

freakboy3742:
> Got it. I'll go ahead and remove ppb from the list for this PR. I can submit an issue/pr to update the docs for third-party plugins to remove ppb from there as well, or should we leave that as is?

Might as well remove it from the docs as well - if it doesn't work, there's no point highlighting it.

moondial-pal:
@freakboy3742 Sorry I fat fingered the request for review. I had not made the requested change yet. Apologies.
I think that was the previous change request from the initial PR submission.
I believe the requested changes are now fully addressed:
- Secondary submenu for third-party GUI bootstraps
- Install guidance shown
- Installed plugins hidden
- PPB removed from the list

Thanks again for the guidance, ready for re-review when you are.

src/briefcase/commands/new.py
```diff
@@ -73,6 +74,20 @@ class NewCommand(BaseCommand):
...
+    # Community GUI bootstraps that are known to Briefcase.
+    # A bootstrap is only considered "installed" if it exposes a
+    # `briefcase.bootstraps` entry point.
+    KNOWN_COMMUNITY_BOOTSTRAPS: ClassVar[list[dict[str, str]]] = [
+        {
+            "entry_point": "pygame_ce",
+            "display_name": "pygame-ce",
+            "package": "pygame-ce",
+            "description": "Community edition fork of pygame.",
+        },
```
freakboy3742: Toga-positron should be on this list as well.

src/briefcase/commands/new.py
```diff
@@ -464,6 +479,12 @@ def create_bootstrap(
             override_value=project_overrides.pop("bootstrap", None),
         )
 
+        if selected_bootstrap == self.OTHER_FRAMEWORKS:
+            self._show_other_frameworks_menu(bootstraps)
+            raise BriefcaseCommandError(
+                "Install a community GUI bootstrap plugin and re-run `briefcase new`."
+            )
```
freakboy3742: Raising BriefcaseCommandError means the command exits as an error, generating a log. Selecting a different GUI framework is "normal behavior", so it shouldn't be generating a log; and while we need to draw the attention of the user, the user hasn't done anything *wrong*, so it shouldn't surface in red as an error.

I'd suggest once the user selects a plugin, the text should be displayed as a warning, and read something like:
```
Pygame-ce is provided by a community plugin. To use this plugin, run:

    python -m pip install pygame-ce

then re-run `briefcase new`.
```

I'd also suggest that the output should all be part of the "other frameworks menu" code - if only because the case where you have installed all known plugins, you currently get an "there are no additional plugins; please install a plugin" message.

src/briefcase/commands/new.py
```diff
@@ -73,6 +74,20 @@ class NewCommand(BaseCommand):
...
 
+    OTHER_FRAMEWORKS = "Other frameworks (select to see options)"
+
+    # Community GUI bootstraps that are known to Briefcase.
+    # A bootstrap is only considered "installed" if it exposes a
+    # `briefcase.bootstraps` entry point.
+    KNOWN_COMMUNITY_BOOTSTRAPS: ClassVar[list[dict[str, str]]] = [
+        {
+            "entry_point": "pygame_ce",
+            "display_name": "pygame-ce",
```
freakboy3742:
```suggestion
            "display_name": "Pygame-ce",
```

src/briefcase/commands/new.py
```diff
@@ -491,32 +512,99 @@ def build_gui_context(
...
         ordered = OrderedDict(sorted(bootstraps.items()))
 
+        none_bootstrap = ordered.pop("None")
+
         # Ensure the first 3 options are: Toga, PySide6, Pygame
-        ordered.move_to_end("Pygame", last=False)
-        ordered.move_to_end("PySide6", last=False)
-        ordered.move_to_end("Toga", last=False)
+        for name in reversed(("Toga", "PySide6", "Pygame")):
+            ordered.move_to_end(name, last=False)
```
freakboy3742: Why this change? You've taken three lines of explicit code, and used a pre-computed list... which is then processed into reverse order. Why not provide the list in reversed order?

src/briefcase/commands/new.py
```diff
@@ -491,32 +512,99 @@ def build_gui_context(
...
+        installed = set(bootstraps.keys())
+
+        available = [
+            plugin
+            for plugin in self.KNOWN_COMMUNITY_BOOTSTRAPS
+            if plugin["entry_point"] not in installed
+        ]
+
+        self.console.warning()
+        self.console.warning(
```
freakboy3742: We generally don't use console commands as "one per line of output" - we combine them all into a single call that includes any required newlines.

freakboy3742 (changes requested):
Generally looks good; the review comments are mostly about user ergonomics, and the exact presentation of error message.

The one bigger question is about the overall workflow. The GUI framework question has always been the last question; however, with this new question, it means that you answer all the questions about your app... and then get told to re-run briefcase new. I think it would make sense to make the framework question the *first* question, so that the user doesn't lose anything if the wizard needs to be re-run.

freakboy3742:
Also - we generally don't rebase PRs unless something catastrophic goes wrong. We use squash commits, so multiple merge commits or noisy "no really fixed this time" commits in a PR won't appear in the final product; and when you rebase, any comments on rebased lines get lost.

moondial-pal:
I really appreciate the patience and time taken for guidance on these reviews. 
I'll work on these suggestions/comments.

moondial-pal:
Minor housekeeping note: commit 8d02559 has an incorrect subject line due to an autocorrect/terminal history mishap when re-running git add + git commit after pre-commit.
Intended subject: “Update change note to reflect final GUI workflow changes”.
No code behavior changes beyond the change note content.

src/briefcase/commands/new.py
```diff
@@ -491,32 +537,123 @@ def build_gui_context(
...
+                ordered[name] = remaining.pop(name)
+
+        # Add everything else (still sorted because remaining started sorted).
+        ordered.update(remaining)
+
+        # Insert "Other frameworks..." as second-to-last
+        ordered[self.OTHER_FRAMEWORKS] = None
+
+        # Re-insert None as the final option
+        ordered["None"] = none_bootstrap
```
freakboy3742: Again - I'm confused why you've done this refactor. This section was previously 6 lines of code, including all the re-organization code. It's now 15... so that we can add *one* more item in a specific location. Why not use:
```
ordered[self.OTHER_FRAMEWORKS] = None
```
just before the move_to_end() calls, followed by
```
ordered.move_to_end(self.OTHER_FRAMEWORKS)
```
just before the `move_to_end("None")` call?

On top of that - the test failures are being caused by this change, because the extra complexity means there are code paths that aren't being tested. You've also deleted a large block of docstring that explains what the subsequent code is doing.

src/briefcase/commands/new.py
```diff
@@ -491,32 +537,123 @@ def build_gui_context(
...
+        self.console.warning(
+            self.console.textwrap(
+                "\n"
+                f"{display_name} is provided by a community plugin.\n"
+                "To use this, run:\n\n"
+                f"    python -m pip install {package}\n\n"
+                "then re-run `briefcase new`."
+            )
+        )
+        raise SystemExit(0)
```
freakboy3742: "SystemExit" is a pretty extreme option here; the way we do this with any other command is a new Exception type that is caught in the mainline. A subclass of BriefcaseWarning will give you the ability to output the final message as a warning, and control the exit code.

src/briefcase/commands/new.py
```diff
@@ -464,10 +487,33 @@ def create_bootstrap(
...
+        self,
+        selected_bootstrap: str,
+        bootstraps: dict[str, type[BaseGuiBootstrap]],
+        context: dict[str, str],
+    ) -> BaseGuiBootstrap:
+        """Instantiate the selected GUI bootstrap."""
+        bootstrap_class = bootstraps[selected_bootstrap]
         return bootstrap_class(console=self.console, context=context)
 
+    def create_bootstrap(
```
freakboy3742: Is `create_bootstrap()` being used any more? Why do we need to retain it?

src/briefcase/commands/new.py
```diff
@@ -464,10 +487,33 @@ def create_bootstrap(
             override_value=project_overrides.pop("bootstrap", None),
         )
 
-        bootstrap_class = bootstraps[selected_bootstrap]
+        if selected_bootstrap == self.OTHER_FRAMEWORKS:
+            self._show_other_frameworks_menu(bootstraps)
+
+        return selected_bootstrap, bootstraps
```
freakboy3742: This is a bit of a red flag for me - if a method is named "select bootstrap", then it should be returning the selected bootstrap. This is especially true when AFAICT the only reason bootstraps is being returned is so that the boostrap *class* can be extracted from the bootstrap list, keyed by the bootstrap this method has selected. Why not... just return the bootstrap class? Then there's not even a need for an `instantiate` method, because instantiation is... creating an instance of the class.

src/briefcase/commands/new.py
```diff
@@ -73,6 +74,31 @@ class NewCommand(BaseCommand):
...
+    # Community GUI bootstraps that are known to Briefcase.
+    # A bootstrap is only considered "installed" if it exposes a
+    # `briefcase.bootstraps` entry point.
+    KNOWN_COMMUNITY_BOOTSTRAPS: ClassVar[list[dict[str, object]]] = [
+        {
+            "entry_point": [
+                "Toga Positron (Django server)",
+                "Toga Positron (Site-specific browser)",
+                "Toga Positron (Static server)",
+            ],
```
freakboy3742: It seems like there's a lot of complexity here to support the fact that there are three entry points, but only one plugin. The entry points aren't actually being *displayed* as part of *this* menu process; we only want to say "there is another plugin available". 

The only reason I can see that this list is being retained is so that it can be compared against the bootstrap list... but if we have the package name, we can check directly against importlib whether it's installed.

This also protects against cosmetic changes in the name of the entry point (and against the fact that in the near future, there will be 2-3 more entry points associated with toga-positron).

freakboy3742 (changes requested):
We're getting there, but there's a lot more complexity here than is required. When you're in a situation where you're having to go to extraordinary lengths to work around data in a format that isn't convenient, it's worth stepping back to ask whether the data actually needs to be in that format.

moondial-pal:
Thanks for the detailed feedback, that’s a big help.

I’ll work on simplifying this and making the ordering more explicit and minimal. I’ll restore the docstring and align the flow so the method returns the selected bootstrap class directly.

Noted on SystemExit as well, I’ll switch this to a warning style exit consistent with other commands.

I’m correcting course and appreciate the guidance. Working on this PR has been and continues to be a great learning opportunity. 
I’ll rework this and push an update as soon as I can.

moondial-pal:
Marking this as draft while I finish simplifying the selection logic and aligning the tests with the updated behavior. I’ll move it back to ready once everything is clean and passing.

moondial-pal:
Working on CI fixes.

freakboy3742:
Apologies for the issues with the docs builds - we launched a new version of the website on Friday, and we've been shaking out loose links ever since. 

I think we've got most of them sorted out at this point; but some of the docs build problems require merging with main, so if you're seeing problems, try that as a fix first. If you're still seeing problems, let us know and we can dig into the problem for you.

moondial-pal:
Hi @freakboy3742, this is ready for review. I believe I have addressed your feedback and the diff is larger than usual given the new community framework wizard behavior we discussed. Let me know if anything needs adjusting.

freakboy3742:
@moondial-pal Thanks for the update - I hope to be able to give this a review in the next couple of days.

src/briefcase/commands/new.py
```diff
@@ -73,6 +87,26 @@ class NewCommand(BaseCommand):
     output_format = ""
     description = "Create a new Briefcase project."
 
+    OTHER_FRAMEWORKS = "Other frameworks (select to see options)"
+
+    # A plugin is treated as "installed" if its distribution package is installed,
+    # regardless of how many Briefcase entry points it provides.
+    KNOWN_COMMUNITY_PLUGINS: ClassVar[list[dict[str, str]]] = [
```
freakboy3742: This data structure is only ever accessed by iteration, or as lookup for a package name. Why not store it as a dictionary keyed by package name?

src/briefcase/commands/new.py
```diff
@@ -491,31 +520,98 @@ def build_gui_context(
...
+                ),
+            )
+
+        self.console.warning(self.console.textwrap("\n" + intro + "\n"))
+
+        options = {
+            plugin["package"]: (
+                f"{plugin['display_name']} — {plugin['description']}"
+                if plugin.get("description")
+                else plugin["display_name"]
```
freakboy3742: We're in control of the list of plugins here, so we can guarantee they all have display names and descriptions.

src/briefcase/commands/new.py
```diff
@@ -491,31 +520,98 @@ def build_gui_context(
...
+        }
+
+        chosen = self.console.selection_question(
+            intro=(
+                "Select a community GUI bootstrap to see installation instructions.\n\n"
+                "Installed plugins are not shown."
+            ),
+            description="Community GUI Framework",
+            default=next(iter(options.keys())),
+            options=options,
```
freakboy3742, line 570: This menu list of options, but doesn't provide a "bail out" option - what happens if I get to this menu and I don't want any of the available options? We need to have a "No GUI framework" option that is the equivalent of a "No GUI framework" from the first GUI menu.

src/briefcase/commands/new.py
```diff
@@ -541,8 +637,12 @@ def new_app(
         self.console.prompt()
         self.console.prompt("Let's build a new Briefcase app!")
 
+        # Ensure we always have a dict before popping keys
+        project_overrides = project_overrides or {}
+
+        bootstrap_class = self.select_bootstrap(project_overrides)
         context = self.build_app_context(project_overrides)
```
freakboy3742, line 619: This changes the order of questions - which is what we wanted - but doesn't address the fact that the first question on the "build_app_context" question list starts "First, we need ...". The question text will need to be tweaked.

src/briefcase/commands/new.py
freakboy3742: The "None" option should also be the default, as there's no reason to preference the first alphabetical framework.

src/briefcase/commands/new.py
```diff
@@ -491,31 +520,98 @@ def build_gui_context(
...
+        )
 
-        return bootstrap_choices
+        not_installed_plugins = [
+            plugin
+            for plugin in self.KNOWN_COMMUNITY_PLUGINS
+            if not is_package_installed(plugin["package"])
+        ]
+
+        if not not_installed_plugins:
```
freakboy3742: `not not_installed` is a bit awkward; `not uninstalled...` would read a bit better.

src/briefcase/commands/new.py
```diff
@@ -491,31 +520,98 @@ def build_gui_context(
...
+        # Pull preferred items to the front (in explicit order) if present.
+        ordered = [name for name in preferred if name in ordered] + [
+            name for name in ordered if name not in preferred
+        ]
 
-        # Option None should always be last
-        ordered.move_to_end("None")
+        ordered.append(self.OTHER_FRAMEWORKS)
+        if "None" in bootstrap_names:
+            ordered.append("None")
```
freakboy3742: This is now the third time I've provided this feedback ([Ref 1](https://github.com/beeware/briefcase/pull/2640#discussion_r2761242062), [ref 2](https://github.com/beeware/briefcase/pull/2640#discussion_r2777034304)): why are you changing this section of code? 

Unless I'm missing something, the existing code does everything that is needed, and is a *lot* less complicated. The updated implementation involves multiple sorts, appended lists, and optional comparisons... when all we need is *what is already here* - sort the inputs, ensure 3 specific options are first, and "Other" and "None" are last. This doesn't require a complex sorting. The code that is *already here* does everything that is needed.

If there's a something that I'm missing here, please let me know - but otherwise, please revert this to the original code. It works and is easy to understand. It doesn't need to be fixed unless there is *actually something wrong with it*.

freakboy3742 (changes requested):
Thanks for those updates. I think the broad structure is getting closer; but there's some things that are still odd about data structure use, and the specifics of the experience the user has when they run the wizard.

moondial-pal:
Understood, I'll revert the ordering section and address the other feedback. Thanks for your patience.

moondial-pal:
Checking what may have caused failures on my end.

moondial-pal:
The pre-commit checks passed locally but the CI caught a line length issue among other things. Circling back to make sure test coverage is 100%.

src/briefcase/commands/new.py
```diff
@@ -73,6 +74,29 @@ class NewCommand(BaseCommand):
...
+    KNOWN_COMMUNITY_BOOTSTRAPS: ClassVar[dict[str, dict[str, object]]] = {
+        "toga-positron": {
+            "entry_point": [
+                "Toga Positron (Django server)",
+                "Toga Positron (Site-specific browser)",
+                "Toga Positron (Static server)",
+            ],
+            "display_name": "Positron",
+            "description": "A Toga base for apps whose GUI is provided by a web view "
+            "(i.e., Electron-like apps, but for Python).",
```
freakboy3742: This is legal Python, but it's really difficult to visually parse, because it's not clear that `"(i.e., ..."`   isn't the start of a new dictionary key.
```suggestion
            "description": (
                "A Toga base for apps whose GUI is provided by a web view "
                "(i.e., Electron-like apps, but for Python)."
            ),
```

moondial-pal:
Great point, I can see how that reads ambiguously. Working on getting the platform tests green...I think my dev environment got a bit tangled working across two machines.

src/briefcase/commands/new.py
```diff
@@ -496,10 +543,14 @@ def _gui_bootstrap_choices(self, bootstraps):
         # Sort the options alphabetically first
         ordered = OrderedDict(sorted(bootstraps.items()))
 
+        # Insert the Other Frameworks sentinel before ordering
+        ordered[self.OTHER_FRAMEWORKS] = None
```
freakboy3742: If you insert it at the right time, there's no need to sort anything.

src/briefcase/commands/new.py
```diff
@@ -541,8 +664,12 @@ def new_app(
         self.console.prompt()
         self.console.prompt("Let's build a new Briefcase app!")
 
+        # Ensure we always have a dict before popping keys
+        project_overrides = project_overrides or {}
+
+        selected_bootstrap, bootstraps = self.select_bootstrap(project_overrides)
         context = self.build_app_context(project_overrides)
-        bootstrap = self.create_bootstrap(context, project_overrides)
+        bootstrap = self._instantiate_bootstrap(selected_bootstrap, bootstraps, context)
```
freakboy3742: Do we actually need a separate method for this? It's a method with a 2 line implementation that isn't re-used.

src/briefcase/commands/new.py
```diff
@@ -517,6 +568,78 @@ def _gui_bootstrap_choices(self, bootstraps):
...
+    ) -> None:
+        """Show a submenu of known community bootstraps and display installation
+        instructions."""
+        installed = set(bootstraps.keys())
+
+        def entry_point_names(plugin: dict[str, object]) -> list[str]:
+            entry_point = plugin["entry_point"]
+            if isinstance(entry_point, str):
+                return [entry_point]
+            return list(entry_point)
```
freakboy3742: We're in control of the source of data here - so we can make it fit the shape we need. If we make the entry points in the list of community bootstraps *always* a list, we don't need this extra handling.

freakboy3742 (changes requested):
We're definitely getting closer. Most of the implementation now looks good; I've pushed a couple of cosmetic tweaks to the actual message text, but the basic flow and implementation now makes sense.

The one part that needs to be cleaned up now is an artefact of the refactor, plus the testing that flows on from that.

In the old code, the interface to bootstraps was `create_bootstrap()`, which handled both selection and instantiation of the bootstrap. This PR has split that up - which makes sense... but the API for `create_boostrap()`... which is then never called - and the testing is still focussed around calling `create_bootstrap()`. 

If a method isn't being used, we don't need to retain it :-)

On top of that, `_instantiate_bootstrap()` is 2 line implementation of fairly obvious logic that isn't re-used anywhere. I think it's safe to put that content directly into the single place it's used.

This means that the content that is currently in `test_create_bootstrap()` should really be in a `test_select_bootstrap()` method - this would be consistent with the fact that most of the test cases in `test_create_bootstreap()` were really tests of the selection process, not the instantiation process.
I'd suggest that instead of introducing a new `_instantiate_boostrap()` method, we repurpose the existing `create_bootstrap()` name to fill that purpose - that is rename `_instantiate_bootstrap()` to `create_bootstrap()`. 

This means a change to tests as well. `test_creat_splitting `test_create_bootstrap()` into two parts - one testing `test_select_bootstrap() to test the selection process, and `test_create_bootstrap()` - since that capability was previously being tested

I've pushed an update doing the API simplification; however, this means the tests will now fail. If you update the tests to match what I've described here, we should be good to go

moondial-pal:
The continued guidance is much appreciated. I'll get to work on the changes.

src/briefcase/commands/new.py
```diff
@@ -73,6 +74,31 @@ class NewCommand(BaseCommand):
...
     description = "Create a new Briefcase project."
 
+    OTHER_FRAMEWORKS = "Other frameworks"
+
+    # Community GUI bootstraps that are known to Briefcase.
+    # A bootstrap is only considered "installed" if it exposes a
+    # `briefcase.bootstraps` entry point.
+    KNOWN_COMMUNITY_BOOTSTRAPS: ClassVar[dict[str, dict[str, object]]] = {
+        "toga-positron": {
+            "entry_point": [
```
freakboy3742: In the process of reviewing the tests for this, something occurred to me. We've recently added a new positron backend for FastAPI. However, this code works fine even though it doesn't know about that new plugin. 

Why? Because there's no way to install one of these plugins without installing the others.  If *any* of the packages are installed, *all* of them must be.

However, even that check is fragile - if we change the spelling of the plugin (say, change "Static server" to "Static web server" in the label), this code will break.

A much better check is to just look for the installable package. If `toga-positron` isn't installed, then *none* of the entry points will be registered; if `toga-positron` *is* installed, then *all* the entry points must be registered.

src/briefcase/commands/new.py
```diff
@@ -541,8 +627,17 @@ def new_app(
...
 
+        # Ensure we always have a dict before popping keys
+        project_overrides = project_overrides or {}
+
+        selected_bootstrap, bootstraps = self.select_bootstrap(project_overrides)
         context = self.build_app_context(project_overrides)
-        bootstrap = self.create_bootstrap(context, project_overrides)
+
+        # Instantiate the bootstrap and add the bootstrap context.
+        bootstrap = bootstraps[selected_bootstrap](
```
freakboy3742: This works, but it can clearly be simplified. If all we need `bootstraps` is so we can return `bootstraps[selected_bootstrap]`... why not just return that result?

tests/commands/new/test_create_bootstrap.py
freakboy3742, line 1: If there isn't a `create_bootstrap` method, there shouldn't be a `test_create_bootstrap` test.

tests/commands/new/test_select_bootstrap.py
```diff
@@ -0,0 +1,444 @@
...
+from briefcase.bootstraps import (
+    ConsoleBootstrap,
+    EmptyBootstrap,
+    PygameGuiBootstrap,
+    PySide6GuiBootstrap,
+    TogaGuiBootstrap,
+)
+
+
+def test_question_sequence_bootstrap_context(
```
freakboy3742: This test is confirming that if the __init__ method defined *in this test* copies the context when it is invoked. That's not a useful test.

tests/commands/new/test_select_bootstrap.py
```diff
@@ -0,0 +1,444 @@
...
+    selected, bootstraps = new_command.select_bootstrap(
+        project_overrides={"bootstrap": "Custom GUI"}
+    )
+    bootstrap = bootstraps[selected](console=new_command.console, context=app_context)
+
+    assert isinstance(bootstrap, GuiBootstrap)
+    assert passed_context == app_context
+
+
+def test_question_sequence_toga(new_command):
```
freakboy3742, line 15: This isn't testing a question sequence - it's testing that an override works.

tests/commands/new/test_select_bootstrap.py
```diff
@@ -0,0 +1,444 @@
...
+
+    context = {"app_name": "myapplication", "author": "Grace Hopper"}
+
+    selected, bootstraps = new_command.select_bootstrap(
+        project_overrides={"bootstrap": "Toga"}
+    )
+    bootstrap = bootstraps[selected](console=new_command.console, context=context)
+
+    assert isinstance(bootstrap, TogaGuiBootstrap)
+    assert bootstrap.context == context
```
freakboy3742: Why is the instantiation of the bootstrap needed? It's not part of what the `select_bootstrap` command is doing.

tests/commands/new/test_new_app.py
```diff
@@ -100,11 +100,6 @@ def test_new_app(
         default_config={"replay_dir": str(tmp_path / "data/templates/.replay")},
     )
 
-    # Bootstrap post-generate step is invoked with correct path
```
freakboy3742, line 103: Removing a test like this is concerning - we're no longer asserting that a key behavior is invoked.

freakboy3742 (approved):
Thanks for those updates.

In the process of reviewing it, I realised there's a much easier way to deal with the "what plugins are installed" problem - but since you've already done a bunch of passes based on past advice, I figured I'd clean up my own mess here. Details inline.

I've also updated a bunch of tests - in the process of migrating to select_bootstrap, you've changed the nature of some tests, retained testing of features that aren't actually part of what the new method does, and in one case, removed an assertion of key behavior. I've corrected those issues.

However - with those changes, this is good to land! Thanks for all your work on this!

moondial-pal:
Thank you for the patience and guidance throughout this, and for cleaning up after giving thorough feedback. I've taken note of the test discipline issues in particular, testing the right behavior for the right method is something I'll carry into future PRs. I really appreciate the opportunity to continue contributing and learning.

PR #2642, opened 2026-01-25
Update expected new-lines in tests for Rich 14.3

### LINKED ISSUE #2641: Rich 14.3.0 has broken many tests
This appeared in CI on January 24 between [this run](https://github.com/beeware/briefcase/actions/runs/21306031577) and [this one](https://github.com/beeware/briefcase/actions/runs/21320524500). Comparing the package list from "Unit tests (ubuntu-24.04, 3.10)":

```
@@ -37,7 +37,7 @@
 python-slugify==8.0.4
 pyyaml==6.0.3
 requests==2.32.5
-rich==14.2.0
+rich==14.3.0
 setuptools==80.10.1
 setuptools-scm==9.2.2
 six==1.17.0
```

Rich 14.3.1 was released less than 24 hours later, but [did not fix the problem](https://github.com/beeware/briefcase/actions/runs/21334328851).

Many of the failures are down to the output having fewer newlines than expected, but in some of them the actual output appears to have changed. 

Several of the entries in the [14.3.0 release notes](https://github.com/Textualize/rich/releases/tag/v14.3.0) look like they might be relevant, but I'm not familiar enough with this area to suggest a fix.

@rmartin16 @kattni: FYI

rmartin16:
## changes
- Fixes #2641 

## PR Checklist:

### DIFF
--- a/changes/2641.misc.md
+++ b/changes/2641.misc.md
@@ -1,0 +1,1 @@
+The tests were updated for changes to new-line outputting in Rich 14.3.0 when using a progress bar in a non-interactive session.


rmartin16:
To summarize the changes in Rich 14.3.0:

The Wait Bar is technically implemented as a Rich Progress Bar. Rich [updated](https://github.com/Textualize/rich/pull/3905/changes) the Progress bar to no longer output a new-line when `stop()` is called while the Progress Bar is "disabled". Briefcase disables the Wait Bar when the console session is non-interactive.

Since the tests run in a non-interactive session, all of the tests asserting on new-lines now see at least one less new-line.

rmartin16:
I think it should be possible via `dependency-groups` to ensure that Briefcase devs have `rich>=14.3.0` installed without enforcing that for all Briefcase users. I just haven't messed around with `dependency-groups` in any capacity....and I'm not immediately sure what the best way to do that is.

freakboy3742:
> I think it should be possible via `dependency-groups` to ensure that Briefcase devs have `rich>=14.3.0` installed without enforcing that for all Briefcase users. I just haven't messed around with `dependency-groups` in any capacity....and I'm not immediately sure what the best way to do that is.

Adding a lower pin to `dependency-groups` would definitely fix the problem; I guess my question is whether we should just add it to the base requirement. We're already requiring 12.6; requiring 14.3 doesn't seem too extreme.

rmartin16:
I played around a little bit with `dependency-groups` and didn't quite get there.....but I feel like it should be possible to just force only the `dev` or `test` groups to require `rich>=14.3`.

> We're already requiring 12.6; requiring 14.3 doesn't seem too extreme.

maybe...i just worry about the conflict between Briefcase and a user's app's dependencies; Rich is really popular. That may be mostly mitigated with the isolated dev environment....as long as most people don't need to disable it for some reason.

PR #2664, opened 2026-02-01
Add option to revoke permissions on Android

timrid:
This PR adds the option `--revoke-permission` to revoke individual permissions from an app before it starts and removes the `--reset-permissions` option from https://github.com/beeware/briefcase/pull/2639.

This is necessary, because in https://github.com/beeware/briefcase/pull/2639#issuecomment-3818536437 it was discovered that the `-p` option in `reset-permissions` has no effect at all. The `-p` option is often mentioned on the internet (e.g. [here](https://gist.github.com/Pulimet/5013acf2cd5b28e55036c82c91bd56d8#file-adbcommands-L86)), but this option does not actually exist and ADB does not report any errors when using `-p`. But this means that the permissions of *all* apps are always reset. This can be problematic if you are not testing in an emulator but on a real device, as it will also affect all other installed apps.

Revoking permissions for individual apps is only possible using `adb shell pm revoke`. Unlike `adb shell pm reset-permissions`, however, you must explicitly specify the permissions to be revoked, thats why the command line parameter changed from `--reset-permissions` to `--revoke-permission`.

## PR Checklist:

### DIFF
--- a/changes/2664.feature.md
+++ b/changes/2664.feature.md
@@ -1,1 +1,1 @@
-Android apps can now revoke permissions before startup via `--revoke-permission`
+Android apps can now revoke permissions before startup via `--revoke-permission`.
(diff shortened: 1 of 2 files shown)


mhsmith:
Thanks, I'll look at this within the next few days.

mhsmith (approved):
Thanks! I tested this with toga/examples/camera and it worked correctly.

PR #2676, opened 2026-02-09
Add note on Liquid Glass quirks to docs

davidfokkema:
Some people (read me) might be running into problems that their PySide6 application suddenly looks terrible on macOS 26 with newer versions of Briefcase. It is because the latest binary GUI stub has been compiled with macOS 26 support and PySide6 has a lot of liquid glass UI bugs. As noted [here](https://www.qt.io/blog/qt-on-macos-26-tahoe) you can set the `UIDesignRequiresCompatibility` key to revert back to the style of previous macOS releases. This PR adds a note to the macOS-specific notes in the docs.

## PR Checklist:

### DIFF
--- a/changes/2676.doc.md
+++ b/changes/2676.doc.md
@@ -1,1 +1,0 @@
-Added note on reverting Liquid Glass to docs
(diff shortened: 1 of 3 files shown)


davidfokkema:
Ah, I didn't realize! Thanks for making the text more generic.

PR #2679, opened 2026-02-12
Resolve E501 ignored Ruff rule

JohananOppongAmoateng:
Resolved E501 errors for platforms

#2383 

## PR Checklist:

### DIFF
--- a/src/briefcase/platforms/macOS/__init__.py
+++ b/src/briefcase/platforms/macOS/__init__.py
@@ -1232,7 +1232,7 @@ password:
                                 f"Notarization was rejected: {summary}\n"
                                 + "\n".join(
                                     f"""
-* ({issue.get("severity", "?")}) {issue.get("path")} [{issue.get("architecture", "unknown architecture")}]
+    * ({issue.get("severity", "?")}) {issue.get("path")} [{issue.get("architecture", "unknown architecture")}]
       {issue.get("message")}
       {issue.get("docUrl", "(No additional help available)")}"""  # noqa: E501
                                     for issue in response.get("issues", [])


pyproject.toml
```diff
@@ -270,9 +270,9 @@ ignore = [
 
 [tool.ruff.lint.per-file-ignores]
 # E501: line too long, to be fixed in future changes
-"tests/*" = ["E501"]
-"src/briefcase/bootstraps/*" = ["E501"]
-"src/briefcase/platforms/*" = ["E501"]
+#"tests/*" = ["E501"]
+#"src/briefcase/bootstraps/*" = ["E501"]
+
```
kattni: You won't be able to remove these ignores until all of the issues are resolved.

kattni (dismissed):
Hello! Thanks for submitting this. You need to add a change note, as described here: https://briefcase.beeware.org/en/latest/how-to/contribute/how/change-note/ That will take care of the towncrier failure. Please see the inline comment below as well.

changes/2679.misc.md
```diff
@@ -0,0 +1 @@
+Resolved E501 errors for platforms
```
freakboy3742: The 'voice' for change notes should be in the form of something you'd see in a release note describing the "great new feature!!", not a mechanical description of the change from a code perspective:
```suggestion
E501 linting issues have been resolved for the platforms folder.
```

This is less important for a misc changenote, as these notes aren't actually surfaced in the final release notes - but it's worth staying in practice for when you need a "real" release note.

JohananOppongAmoateng:
Hi @kattni  and @freakboy3742 , the failing ci tests is due to an AssertionError in the notarization tests. The regex expects indentation before the * (Error) and * (Warning) lines, but the actual output doesn’t include that spacing.

Would it be fine if I update the test regex in tests/platforms/macOS/app/package/test_resume_notarization.py to match the actual output (i.e., remove the indentation requirement and allow flexible whitespace)?

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -674,27 +672,24 @@ def select_identity(
...
+            self.console.info(f"""
 In future, you could specify this signing identity by using:
 
     $ briefcase {self.command} macOS {self.output_format} {ident_option} {identity} ...
 
 or
 
-    $ briefcase {self.command} macOS {self.output_format} {ident_option} "{identity_name}" ...
+    $ briefcase {self.command} macOS {self.output_format}
+    {ident_option} "{identity_name}" ...
```
freakboy3742: This is help that is intended to be displayed on a single line; breaking it across multiple lines will be potentially confusing:
```suggestion
    $ briefcase {self.command} macOS {self.output_format} {ident_option} "{identity_name}" ...
```

Preserving the single line format may require manually disabling the E501 rule on this one single line using the `# noqa: E501` syntax.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1020,17 +1015,16 @@ def notarize(
...
-            f"""
+        self.console.warning(f"""
 Briefcase will now wait for Apple to approve the notarization request.
 This can take some time - in some cases, hours.
 
 If notarization is interrupted, you can resume by running:
 
-    briefcase package macOS {self.output_format} {format_args} {identity_args} --resume {submission_id}
+    briefcase package macOS {self.output_format} {format_args}
+    {identity_args} --resume {submission_id}
```
freakboy3742: As above - this should be preserved on a single line:
```suggestion
    briefcase package macOS {self.output_format} {format_args} {identity_args} --resume {submission_id}
```

src/briefcase/platforms/iOS/xcode.py
```diff
@@ -563,10 +561,8 @@ def run_app(
             label = "app"
 
         self.console.info(
-            (
-                f"Starting {label} on an {device} running iOS {iOS_version} "
-                f"(device UDID {udid})"
-            ),
+            f"Starting {label} on an {device} running iOS {iOS_version} "
+            f"(device UDID {udid})",
```
freakboy3742: Although the brackets are strictly unnecessary here, they do help with readability, as it's clear that we're dealing with a single string as an argument. Without the brackets, there's one character difference between "two string arguments" and "one string argument that is a continuation over 2 lines":
```suggestion
            (
                f"Starting {label} on an {device} running iOS {iOS_version} "
                f"(device UDID {udid})"
            ),
```

freakboy3742 (changes requested):
Thanks for those updates.

In this case, the CI failures have been caused because the output message has changed as a result of the line length reformatting. In this case, the original line length is actually the desired behavior; so the fix is to keep the old regex, and return the message in question to the original line length. This will also require making a single-line exception for the E501 rule using the `#noqa` syntax.

freakboy3742 (approved):
Awesome - Thanks for the contribution!

PR #2689, opened 2026-02-16
Update version pattern

abdnh:
The PEP 440 version regex used by Briefcase is outdated. The canonical format is now described in [packaging.python.org](https://packaging.python.org/en/latest/specifications/version-specifiers/#appendix-parsing-version-strings-with-regular-expressions) and is provided by `packaging`.

## Changes

- Updated `PEP440_CANONICAL_VERSION_PATTERN_RE` to use `packaging.version.VERSION_PATTERN`.
- Removed the `test_invalid_app_version` test since most strings are now valid versions according to the latest specification.

## PR Checklist:

## Context

We're exploring using Briefcase for [Anki](https://apps.ankiweb.net/) and ran into an issue because Anki's calendar versioning scheme is not recognized by Briefcase: 
https://github.com/ankitects/anki/issues/4557

### DIFF
--- a/changes/2689.bugfix.md
+++ b/changes/2689.bugfix.md
@@ -1,0 +1,1 @@
+Updated version specification to conform to the latest Python standards.


freakboy3742:
Thanks for the PR, and the research into the underlying cause.

There's a couple of hidden issues with the "replace the regex" approach you've taken here. Most notably, the full regex accepts `v1.2.3` as valid input, which will break other parts of Briefcase that expect a "clean", numbers-only version number to exist. A leading `v` *isn't* valid when generating a macOS Info.plist version number, for example.

It's also problematic to delete *all* the "invalid" test cases. While it's true that there are a lot of previously invalid version numbers that are now valid, there are still *some* invalid version numbers, and we would still need to validate those.

However - I think the better approach is to go all-in on using `Version` as an internal representation. That saves a bunch of effort other places where we need parts of a version number, and means we don't need to reproduce any sort of validation logic. I've implemented this approach in #2694.

PR #2694, opened 2026-02-17
Use packaging for version parsing

### LINKED ISSUE #2693: Valid version numbers are rejected by config validation
### Describe the bug

Reported via ankitects/anki#4557

Briefcase is currently rejecting some version numbers that are currently valid. The originally reported case is calver version numbers with a leading zero (e.g., `25.09.2`); but there's a number of other deviations from the original PEP440 version numbering scheme that are now considered correct.

### How to reproduce the bug

1. Create an application with a version number of "25.09.2"
2. Run any Briefcase command (e.g., `briefcase create`)
3. Get an error
```
Briefcase configuration error: Version number (25.09.2) is not valid.

Version numbers must be PEP440 compliant; see https://www.python.org/dev/peps/pep-0440/ for details.
```

### Minimum example code

```shell

```

### Screenshots

_No response_

### Environment details

- Operating system and version: All (tested on macOS 15.4)
- Python version: all (tested on 3.14)
- Software versions:
  - Briefcase: 0.3.26

### Logs

```

```

### Additional context

We should probably defer to `packaging.version.Version` rather than implementing our own version parsing.

freakboy3742:
The regex defined in PEP 440 is no longer an accurate definition for all valid version schemes; `packaging.version` provides the canonical version number parsing mechanism.

This PR replaces the custom PEP 440 validator with the `packaging.version.Version` parser. This means we now accept a wider range of version numbers as valid inputs (including `v1.2.3`, `1.2RC1` and `25.09.2`), but internally provides a clean, canonical, parsed version for use by the rest of Briefcase.

Fixes #2693.

## PR Checklist:

### DIFF
--- a/tests/config/test_GlobalConfig.py
+++ b/tests/config/test_GlobalConfig.py
@@ -13,29 +13,33 @@ VALID_VERSIONS = [
     ("v1.2.3", "1.2.3"),  # v prefix
     ("V1.2.3", "1.2.3"),  # v prefix, capitalized
     ("2025.9.5", "2025.9.5"),  # Calver
-    ("2025.09.05", "2025.09.05"),  # Calver with leading zero
+    ("2025.09.05", "2025.9.5"),  # Calver with leading zeros
     ("1.2.3.dev42", "1.2.3.dev42"),  # dev
     ("1.2.3a42", "1.2.3a42"),  # alpha
     ("1.2.3b42", "1.2.3b42"),  # beta
     ("1.2.3rc42", "1.2.3rc42"),  # release candidate
-    ("1.2.3.post42", "1.2.3post42"),  # post release
-    ("1.2.3.post37.dev42", "1.2.3post37.dev42"),  # dev post release
+    ("1.2.3.post42", "1.2.3.post42"),  # post release
+    ("1.2.3.a42", "1.2.3a42"),  # alpha with period
+    ("1.2.3.b42", "1.2.3b42"),  # beta with period
+    ("1.2.3.rc42", "1.2.3rc42"),  # release candidate with period
+    ("1.2.3post42", "1.2.3.post42"),  # post release without period
+    ("1.2.3.post37.dev42", "1.2.3.post37.dev42"),  # dev post release
     ("1.2.3.DEV42", "1.2.3.dev42"),  # dev, capitalized
     ("1.2.3A42", "1.2.3a42"),  # alpha, capitalized
     ("1.2.3B42", "1.2.3b42"),  # beta, capitalized
     ("1.2.3RC42", "1.2.3rc42"),  # release candidate, capitalized
-    ("1.2.3.POST42", "1.2.3post42"),  # post release, capitalized
+    ("1.2.3.POST42", "1.2.3.post42"),  # post release, capitalized
     ("1.2.3alpha42", "1.2.3a42"),  # alpha, full spelling
     ("1.2.3beta42", "1.2.3b42"),  # beta, full spelling
     ("1.2.3c42", "1.2.3rc42"),  # release candidate, alternate spelling
     ("1.2.3pre42", "1.2.3rc42"),  # release candidate, alternate spelling
     ("1.2.3preview42", "1.2.3rc42"),  # release candidate, alternate spelling
-    ("1.2.3-42", "1.2.3post42"),  # post release, alternate spelling
-    ("1.2.3-r42", "1.2.3post42"),  # post release, alternate spelling
+    ("1.2.3-42", "1.2.3.post42"),  # post release, alternate spelling
+    ("1.2.3-r42", "1.2.3.post42"),  # post release, alternate spelling
     ("1.2.3a", "1.2.3a0"),  # implied alpha version
     ("1.2.3b", "1.2.3b0"),  # implied beta version
     ("1.2.3rc", "1.2.3rc0"),  # implied release candidate version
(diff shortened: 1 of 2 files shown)


tests/commands/base/test_parse_config.py
```diff
@@ -91,7 +92,7 @@ def test_parse_config(base_command):
     assert repr(base_command.global_config) == "<Sample project v1.2.3 GlobalConfig>"
     assert base_command.global_config.project_name == "Sample project"
     assert base_command.global_config.bundle == "org.beeware"
-    assert base_command.global_config.version == "1.2.3"
+    assert base_command.global_config.version == Version("1.2.3")
```
mhsmith, line 95: Should add some tests for non-canonical versions to verify that the original problem has been solved.

src/briefcase/config.py
```diff
@@ -420,12 +378,14 @@ def __init__(
...
-        if not is_pep440_canonical_version(self.version):
+        try:
+            self.version = Version(version)
+        except InvalidVersion:
             raise BriefcaseConfigError(
                 f"Version number ({self.version}) is not valid.\n\n"
                 "Version numbers must be PEP440 compliant; "
                 "see https://www.python.org/dev/peps/pep-0440/ for details."
-            )
+            ) from None
```
mhsmith, line 390: As you said, the PEP no longer reflects the current situation, so it would be better to link to the live PyPA spec. Suggested wording:
```py
f"{self.version!r} is not a valid Python package version. For details, see "
f"https://packaging.python.org/en/latest/specifications/version-specifiers/"
```

src/briefcase/config.py
```diff
@@ -537,14 +497,20 @@ def __init__(
...
+            else:
+                self.version = Version(version)
+        except InvalidVersion:
             raise BriefcaseConfigError(
                 f"Version number for {self.app_name!r} ({self.version}) is not valid."
                 f"\n\n"
                 "Version numbers must be PEP440 compliant; "
                 "see https://www.python.org/dev/peps/pep-0440/ for details."
-            )
+            ) from None
```
mhsmith, line 512: See wording update in previous comment – and this would also be an opportunity to eliminate the duplication.

tests/commands/base/test_parse_config.py
freakboy3742: I've added the test to the base config parser, rather than this file - but done.

tests/config/test_GlobalConfig.py
```diff
@@ -4,6 +4,52 @@
...
+# Examples of valid versions; input, parsed
+VALID_VERSIONS = [
+    ("1", "1"),  # Single digit
+    ("1.0", "1.0"),  # Two digits
+    ("1.2.3", "1.2.3"),  # Semver digits
+    ("1.2.3.4", "1.2.3.4"),  # Keep going...
+    ("v1.2.3", "1.2.3"),  # v prefix
+    ("V1.2.3", "1.2.3"),  # v prefix, capitalized
+    ("2025.9.5", "2025.9.5"),  # Calver
+    ("2025.09.05", "2025.09.05"),  # Calver with leading zero
```
mhsmith: This suggests that the leading zeros will be retained, which I [don't think is actually the case](https://github.com/pypa/packaging/issues/443). Changing the two tests as follows would reveal this:
```py
assert str(config.version) == expected
```

src/briefcase/config.py
```diff
@@ -377,14 +377,16 @@ def __init__(
         self.license = license
         self.requires_python = requires_python
 
-        # Version number is PEP440 compliant:
+        # Version number is compliant with PEP440 (and related updates) compliant:
```
mhsmith:
```suggestion
        # Version number is compliant with PEP440 (and related updates):
```

notatallshaw:
> The regex defined in PEP 440 is no longer an accurate definition for all valid version schemes; `packaging.version` provides the canonical version number parsing mechanism.

FWIW, the living specification is the definition for versions: https://packaging.python.org/en/latest/specifications/version-specifiers/. I don't think the regex in the historical PEP 440 document is wrong or different in the living specification, but it is for parsing the version scheme and doesn't do any of the significant amount of normalization work the specification defines.

Packaging attempts to be a sans-IO reference implementation of many of the packaging specifications, but there can be bugs. I did a lot of work that landed in 26.0 on specifiers to both correct behavior, and work with the community to define the correct behavior, and Henry and I did a lot of work on performance for both versions and specifiers, in-particular Henry improved the performance of the version regex while still keeping it correct.

Feel free to report any issues you find.

freakboy3742:
> > The regex defined in PEP 440 is no longer an accurate definition for all valid version schemes; `packaging.version` provides the canonical version number parsing mechanism.
> 
> FWIW, the living specification is the definition for versions: https://packaging.python.org/en/latest/specifications/version-specifiers/. I don't think the regex in the historical PEP 440 document is wrong or different in the living specification, but it is for parsing the version scheme and doesn't do any of the significant amount of normalization work the specification defines.

Thanks for that clarification. That pretty much aligns with what I understood the situation to be; but it's good to have that confirmed.

> Feel free to report any issues you find.

Will do :-)

tests/config/test_GlobalConfig.py
freakboy3742: Good catch. I hadn't caught that I was normalising on both ends, so the "expected" value wasn't representative of what is actually expected. Fixed, along with adding a couple of extra tests for some normalisation cases that weren't there before.

PR #2696, opened 2026-02-17
Reusable helper for formatting boxed console messages

### LINKED ISSUE #2559: Add helper for formatting long warning messages

cdandeneau:
## Summary

This PR introduces a reusable helper for formatting boxed console messages and migrates the JAVA_HOME warnings in the Java integration to use it.

### What changed

- Added `format_message_box(message, title=None)` in `briefcase.console` to:
  - generate consistently sized `*`-border boxes,
  - support an optional title line,
  - wrap and indent body text automatically.

- Replaced two manually aligned triple-quoted warning blocks in `JDK.verify_install()` with `format_message_box()`:
  1. JAVA_HOME points to the wrong Java major version.
  2. JAVA_HOME does not point to a JDK.

- Added tests for the helper to verify:
  - title alignment with borders,
  - wrapped/indented body content formatting.

## Why

This avoids brittle hand-spacing in warning strings, keeps surrounding code indentation clean, and makes variable-content warnings (like Java versions/paths) format correctly without manual padding.

## Testing

- `ruff check src/briefcase/console.py src/briefcase/integrations/java.py tests/console/test_format_message_box.py`
- `pytest -q tests/console/test_format_message_box.py`
- `pytest -q tests/integrations/java/test_JDK__verify.py`

## Issue

Fixes #2559.


freakboy3742:
Thanks for the PR, but I'm going to close this.

Firstly - there's an existing PR in flight for this issue (#2563) that is nearing completion. The discussion history of #2559 points this out. It would not be fair to the original contributor to void their work by accepting this PR.

Secondly - the implementation has some interesting choices, including the introduction of a submodule for no obvious reason.

Lastly - it has included an empty changenote. That's not what our contribution guide requests.

I have reason to suspect that this PR has been generated with the assistance of an LLM. We don't prohibit the use of LLMs, but we *do* ask that you declare when they are used so that we are able to make an assessment on the provenance of the code.

Issue #2698, opened 2026-02-17
Add publish implementation for Google Play Store


freakboy3742:
### What is the problem or limitation you are having?

Briefcase has a `publish` command; however, it is only a stub implementation for Android Gradle targets.

Google provides the [Google Play Developer API](https://developers.google.com/android-publisher) to manage the submission and publication process. It should be possible to use this API to automate submitting an Android app to the Play Store.

### Describe the solution you'd like

`briefcase publish android -c playstore` should publish an app to the Google Play store. 

### Describe alternatives you've considered

Continue to require manual Play Store submission, potentially removing the `publish` command.

### Additional context

This would be a good follow on from #2687.

Ideally, the submission process would also include creating the full Play Store submission, including descriptions, screenshots etc. It's not clear how much of this can be automated, and how much will always require a user to log in and fill in details, but if there's content being provided, the first goal should be to be able to provide that content locally, stored in CI.

It's also not clear if the process will be different for *first* submission and *subsequent* submissions.

Implementing this also requires a fix for #1268.

The Google Developer API [has Python bindings](https://pypi.org/project/google-api-python-client).

[Gradle Play Publisher](https://github.com/Triple-T/gradle-play-publisher) is an unofficial Gradle plugin to automate submission.

[Fastlane](https://docs.fastlane.tools) is a tool that provides a pre-wrapped user experience; it could be considered a competitor to `briefcase publish`. It could also be considered as an implementation mechanism - however, it is a Ruby-based tool, and only works on macOS, so it might not be viable for Android projects.

Issue #2697, opened 2026-02-17
Add publish implementation for iOS/macOS App Store


freakboy3742:
### What is the problem or limitation you are having?

Briefcase has a `publish` command; however, it is only a stub implementation for macOS and iOS targets.

Apple provides the [App Store Connect API](https://developer.apple.com/documentation/appstoreconnectap), as well as various commands in Xcode to manage the archival and submission process. It should be possible to use these commands to automate submitting an iOS and macOS app to their respective App Stores.

### Describe the solution you'd like

`briefcase publish macOS -c appstore` should automate the process of uploading assets and release configurations to the macOS App Store. 

`briefcase publish iOS -c appstore` should do the same, but for iOS apps. 

### Describe alternatives you've considered

Continue to require manual App Store submission, potentially removing the `publish` command.

### Additional context

This would be a good follow on from #2687.

Ideally, the submission process would also include creating the full App Store submission, including descriptions, screenshots etc. It's not clear how much of this can be automated, and how much will always require a user to log in and fill in details, but if there's content being provided, the first goal should be to be able to provide that content locally, stored in CI.

It's also not clear if the process will be different for *first* submission and *subsequent* submissions.

Some pieces of the puzzle that I know about:

Archiving a project
```
xcodebuild archive \
  -scheme "YourSchemeName" \
  -configuration Release \
  -archivePath ./build/YourApp.xcarchive \
  CODE_SIGN_IDENTITY="Apple Distribution: Your Name (TEAMID)" \
  DEVELOPMENT_TEAM="YOURTEAMID"
```

Exporting an IPA from the archive and upload to the App Store:
```
xcodebuild -exportArchive \
  -archivePath ./build/YourApp.xcarchive \
  -exportOptionsPlist ExportOptions.plist \
  -exportPath ./build/export
```

where `ExportOptions.plist` contains:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>method</key>
    <string>app-store</string>
    <key>teamID</key>
    <string>YOURTEAMID</string>
    <key>uploadBitcode</key>
    <false/>
    <key>uploadSymbols</key>
    <true/>
    <key>signingStyle</key>
    <string>automatic</string>
    <key>destination</key>
    <string>upload</string
</dict>
</plist>
```

[...]

Issue #2687, opened 2026-02-16
Add plugin interface for publication channels


filiplajszczak:
### What is the problem or limitation you are having?

The `publish` command exists in stub form across all platforms, but `publish_app()` is not implemented.

### Describe the solution you'd like

Add a [plugin interface](https://briefcase.beeware.org/en/latest/reference/plugins/) for publication channels using dynamic discovery via [entry points](https://packaging.python.org/en/latest/specifications/entry-points/), following the existing pattern for platforms and debuggers. (Discussed in #2678.)

### Entry point registration

Channels register scoped by platform and format, following the pattern used by `briefcase.formats.*`:

```toml
[project.entry-points."briefcase.publication_channels.web.static"]
pythonanywhere = "briefcase_pythonanywhere:PythonAnywhereChannel"
```

### Base class and command API

- A `BasePublicationChannel` ABC with a `publish_app(app, command, **options)` method.
- A `Protocol` type (`PublishCommandAPI`) defines the stable API surface exposed to plugins (`console`, `tools`, `project_path()`, `distribution_path()`), starting minimal and growing as needed.

### Changes to `PublishCommand`
- Dynamic channel discovery via `get_publication_channels(platform, format)` using `importlib.metadata.entry_points()`, same as debuggers and platforms.
- Auto-chain: `publish` triggers `package` (which triggers `build`) if needed. All channels go through packaging — the plugin reads from the packaged artifact as the source of truth for what gets deployed.
- Single installed channel is used automatically; multiple channels require `--channel`.
- `briefcase publish -h` dynamically lists available channels.
- Remove `s3` and `ios_appstore` placeholders.

### Describe alternatives you've considered

Hardcoding publication channels directly in briefcase (e.g., adding PythonAnywhere support as built-in code). This would work but doesn't scale — each new deployment target would require changes to briefcase core and add optional dependencies. (Discussed in #2678.)

### Additional context

Out of scope for this ticket (follow-on work):
- Concrete channel implementations (separate packages and/or PRs)
- Default channel config and API token storage in user config (#2279)

freakboy3742:
Thanks for all the design work on this.

[...]

PR #2701, opened 2026-02-18
Add plugin interface for publication channels (#2687)

filiplajszczak:
Introduce a `BasePublicationChannel` ABC and `PublishCommandAPI` protocol for publishing apps to distribution channels via `briefcase publish`, using dynamic discovery through entry points following the existing pattern for platforms and debuggers.

Publication channel plugins:
- Register scoped by platform and format via entry points (e.g., `briefcase.publication_channels.ios.xcode`)
- Implement `publish_app(app, command, **options)` to perform publication
- Single installed channel auto-selects; multiple require `--channel`

Changes to `PublishCommand`:
- Dynamic channel discovery via `importlib.metadata.entry_points()`
- Auto-chain: publish triggers package if distribution artefact is missing
- Channel classes are instantiated before use
- Remove hardcoded `s3` default and iOS/web `publish_app` stubs

Add placeholder channels for iOS App Store (#2697) and Google Play Store (#2698) that raise `BriefcaseCommandError` when invoked.

Document the publication channel plugin interface in the plugins reference and link the publish command to it. Update publish command reference to reflect the new `--channel` flag behavior.

## Consideration: Mock(spec=...) for test doubles

In this PR I followed the project's established pattern of hand-rolled dummy subclasses, which work well and are readable. But as the plugin interfaces grow (platforms, debuggers, publication channels all share the entry point →
class → instance pipeline), it might be worth considering `Mock(spec=BasePublicationChannel)` as an alternative. The `spec` argument enforces interface parity automatically, catching mismatches between the test double and the real class without manual upkeep. [pytest-mock](https://pypi.org/project/pytest-mock/) and its `mocker` fixture make this particularly convenient in pytest.
[My ramblings on using `Mock(spec=...)` for test doubles](https://filip.lajszczak.dev/testing-dependency-injection-with-mocks--notes-on-the-margin-of-chapter-3-of-architecture-patterns-with-python.html).

## PR Checklist:


src/briefcase/commands/publish.py
```diff
@@ -10,56 +14,71 @@ class PublishCommand(BaseCommand):
...
+                    f"{self.platform} {self.output_format}: "
+                    f"{', '.join(sorted(channels))}.\n\n"
+                    f"Specify a channel with --channel."
+                )
+        else:
+            resolved_channel = channels[channel]()
+
         # Confirm host compatibility, that all required tools are available,
         # and that all app configurations are finalized.
         self.finalize(apps=self.apps.values())
```
freakboy3742: A Briefcase project can contain multiple apps; this will always publish all apps. This is effectively the same problem as we saw with #2635, fixed in #2651 - the fix will be almost identical here.

src/briefcase/publication_channels/base.py
```diff
@@ -0,0 +1,42 @@
...
+
+    This defines the minimal set of attributes and methods that a plugin
+    can rely on from the ``command`` parameter passed to ``publish_app()``.
+    """
+
+    console: Console
+    tools: ToolCache
+    dist_path: Path
+
+    def distribution_path(self, app: AppConfig) -> Path: ...  # pragma: no cover
```
freakboy3742: I get why this `#pragma` is needed... rather than adding it to the specific method, could we add this as a global coverage exclusion on `@runtime_checkable`, as no runtime protocol will ever be covered by tests.

freakboy3742:
Regarding the specific question about mock usage: the existing pattern that you’ve observed/followed is mostly born out of the fact that a simple mock by itself doesn’t give us enough tracking - or, at least, if we did use a mock, we’d end up with a mock that is almost as complex as the Dummy that we’ve implemented, because there needs to be side effects to some function calls etc.

That said - there are places in the code where we do use mocks; and for “simple interface adherence”, a mock might well be a better approach. Given that a Channel plugin is a relatively simple interface with a single method and some properties, it might be worth using them here. If you’re up for the experiment, then I wouldn’t oppose using them here (or in a follow up).

filiplajszczak:
Thanks for the review! Both items addressed:
- Added the `-a/--app` flag to `publish`, following the same pattern as the other commands. I also extracted the app resolution logic into a   `BaseCommand.resolve_apps()` helper and migrated `create`, `update`, `build`, `package`, and `open` to use it as the pattern was duplicated in all five.
- Added `@runtime_checkable` to the global `exclude_lines` and removed the per-method `# pragma: no cover`.

On naming: I was torn between `channel` and `publication_channel`. Went with `publication_channel` since it's less generic and more descriptive. Those are names you read  far more often than you type. The shorter `channel` is used for the CLI flag and local variables where the context already makes the meaning clear.

On `Mock(spec=...)`: I'll experiment with that in a separate PR once this one is wrapped up.

docs/en/reference/commands/publish.md
```diff
@@ -2,7 +2,7 @@
 
 **COMING SOON**
```
freakboy3742: I guess we should drop this now... :-)

pyproject.toml
```diff
@@ -194,6 +194,12 @@ static = "briefcase.platforms.web.static"
 app = "briefcase.platforms.windows.app"
 visualstudio = "briefcase.platforms.windows.visualstudio"
 
+[project.entry-points."briefcase.publication_channels.ios.xcode"]
```
freakboy3742: In order to find the plugin, this capitalisation needs to be:
```suggestion
[project.entry-points."briefcase.publication_channels.iOS.Xcode"]
```

That points out an inconsistency in other registrations - `briefcase.formats.iOS` should be registering `Xcode = "briefcase.platforms.iOS.xcode"` (and similarly for macOS, and the commented out tvOS, watchOS...)

src/briefcase/publication_channels/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def get_publication_channels(
+    platform: str,
+    output_format: str,
+) -> dict[str, type[BasePublicationChannel]]:
+    """Load built-in and third-party publication channels for a platform/format.
+
+    :param platform: The target platform (e.g., "ios")
+    :param output_format: The output format (e.g., "xcode")
```
freakboy3742: Platform identifiers and output formats have capitalisation:
```suggestion
    :param platform: The target platform (e.g., "iOS")
    :param output_format: The output format (e.g., "Xcode")
```

src/briefcase/commands/publish.py
```diff
@@ -10,74 +14,100 @@ class PublishCommand(BaseCommand):
...
-        for app_name, app in self.apps.items():
-            binary_file = self.binary_path(app)
+        for app_name_key, app_obj in apps_to_publish.items():
+            binary_file = self.binary_path(app_obj)
             if not binary_file.exists():
                 raise BriefcaseCommandError(
-                    f"Application {app_name} has not been built. "
+                    f"Application {app_name_key} has not been built. "
                     "Build (and test!) the app before publishing."
                 )
```
freakboy3742: This block of code circumvents the "cascading" call to `package` on L53/54.

freakboy3742 (changes requested):
Nice refactor on the apps-to-X handling!.

I've applied the refactor renaming `publication_channels`->`channels`; I get what you're saying about code being read more often than it's written. However, the name `channel` *should* be clear, as there's no other "channel" concept in Briefcase; and retaining the "full" name in class names retains the disambiguation in the once place where the classes are actually implemented. I've applied a couple of other minor fixes that I found along the way.

A couple of things fell out of live testing.

Firstly, there's a registration inconsistency with `iOS` and `ios` - this is an area where we've historically had some confusion, but in this case, the registered names need to be the `platform_name` and `output_format` as registered by the classes. I've rolled this into the changes that I've made.

However, even with this change, there's some issues using the plugin interface:

1. As flagged inline, the implementation of `__call__` for the Publish command includes a "does the app exist" check that circumvents the "if no distribution path, package the app" logic. I think this should be a straightforward case of deleting that block of code.
2. Even with that block gone, Android apps can't be built, because `distribution_path()` requires knowledge of the packaging format. 

I flagged this during the design process; but it's evidently going to a problem out of the box. Android projects have three potential distribution formats (aab, apk, and debug-apk). Play Store publication requires an AAB artefact; but in the more general case, the publication process would need to be aware of *which* packaging artefact has been selected.

iOS apps won't publish at present, because there's no implementation of `package` for iOS. However, that's part of the bigger "how to publish to the App Store" problem, rather than something specific to this PR.

Longer term, there's probably a need to factor out packaging format plugins in the same way as channel plugins are defined here. We don't have to implement that change here, but we should at least give some thought to how the two will interact.

filiplajszczak:
If I understand correctly, the remaining Android issue is that `distribution_path` needs `packaging_format` which is only set during packaging — so the cascade can't run yet. Do we need to cover that in this PR, or just raise with meaningful message?

freakboy3742:
> If I understand correctly, the remaining Android issue is that `distribution_path` needs `packaging_format` which is only set during packaging — so the cascade can't run yet. Do we need to cover that in this PR, or just raise with meaningful message?

I don't think it's that far from being fully resolved. If we replicate the argument handling from `package`, and ensure the package format is annotated on the app before the existence check of the distribution path is confirmed, we should be done. I'll push an update with this change shortly.

Longer term, we can look at factoring out the packaging formats as plugins in the same way as we've factored out publication plugins. Packaging will require a slightly more complicated API (as there's more steps involved) - but we already have a bunch of built-in implementations to prove out that the interface is complete; From the perspective of the publish command, I'm not sure anything changes beyond needing to know what packaging formats exist, which one has been selected, and how to get the filename for the distribution artefact.

freakboy3742 (approved):
Looks good; as promised, I've pushed one last piece that adds enough support for packaging formats that the Android backend is able to "successfully not package" an app. 

I've also added an `-u` option, on the basis that updating before publication isn't an unreasonable thing to do if we're also cascading the package process.

I'm happy with the code at this point, so I'm marking this as accepted. I'd appreciate some eyeballs on my most recent update to confirm whether you see any problems with the approach; assuming you don't, I'll merge and call this done!

filiplajszczak:
Looks good to me!

Added some docs cross-linking between publish, the plugin interface, and the platform how-to guides. Also extracted the "no working channels" note into a shared snippet.

The main `index.md` doesn't mention publishing yet. Might be worth adding once a working channel (e.g., PythonAnywhere) lands.

filiplajszczak:
@freakboy3742 That name is much more future-proof. Good catch!

PR #2708, opened 2026-02-26
fix passing packaging_format to package_command in publish

filiplajszczak:
When _publish_app triggers packaging, it calls self.package_command() without forwarding packaging_format. This was introduced in cfc9774e which added packaging_format to _publish_app but missed forwarding it to the package_command call.

It's a tail of work on #2687

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/publish.py
+++ b/src/briefcase/commands/publish.py
@@ -82,7 +82,10 @@ class PublishCommand(BaseCommand):
 
         if update or not self.distribution_path(app).exists():
             state = self.package_command(
-                app, update=update, packaging_format=packaging_format, **options
+                app,
+                update=update,
+                packaging_format=packaging_format,
+                **options,
             )
 
         self.verify_app(app)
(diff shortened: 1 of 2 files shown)


src/briefcase/commands/publish.py
```diff
@@ -81,7 +81,9 @@ def _publish_app(
         app.packaging_format = packaging_format
 
         if update or not self.distribution_path(app).exists():
-            state = self.package_command(app, update=update, **options)
+            state = self.package_command(
+                app, update=update, packaging_format=packaging_format, **options
```
freakboy3742: We try to avoid the "all on one line" argument format; if a list of arguments is long enough to need to be pulled onto a separate line, it's worth making them one-argument-per-line:
```suggestion
                app,
                update=update,
                packaging_format=packaging_format,
                **options,
```

tests/commands/publish/test_call.py
```diff
@@ -220,6 +220,22 @@ def test_publish_app_invalid(publish_command, first_app, second_app):
         publish_command(**options)
 
 
+def test_publish_passes_packaging_format_to_package(
+    publish_command, first_app_unpackaged
```
freakboy3742: As above with arguments:
```suggestion
    publish_command,
    first_app_unpackaged,
```

PR #2723, opened 2026-03-06
Add option to configure Android ABIs

### LINKED ISSUE #808: Add option to configure Android ABIs
Generated APKs and AABs currently include the following ABIs:
* armeabi-v7a
* arm64-v8a
* x86_64

We can't drop armeabi-v7a yet, because there are still too many 32-bit devices in use (see https://github.com/chaquo/chaquopy/issues/709). However, we were OK to drop 32-bit x86 because we set our minimum API level to 26 (https://github.com/beeware/briefcase-android-gradle-template/pull/49), and 64-bit emulator images have been available since level 21. Since BeeWare developers will need to use a 64-bit emulator image anyway to test on the current version of Android, requiring them to do the same on the other versions is no great inconvenience. And omitting x86 reduces the APK size and makes it faster to build and install.

Anyway, there are legitimate reasons for developers to want to change the set of available ABIs, so we should make this configurable in pyproject.toml. See https://github.com/beeware/briefcase-android-gradle-template/pull/52#discussion_r935822715 for discussion of possible approaches.

JohananOppongAmoateng:
This PR adds support for configuring Android ABIs through the android_abis setting in pyproject.toml.
  Fixes #808

## PR Checklist:

### DIFF
--- a/docs/en/reference/platforms/android/gradle.md
+++ b/docs/en/reference/platforms/android/gradle.md
@@ -186,8 +186,7 @@ The following options can be added to the `tool.briefcase.app.<appname>.android`
 
 ### `android_abis`
 
-A list of strings that explicitly specifies the Android ABIs (platforms) the app will support. This controls the `abiFilters` parameter in the `ndk` block of the Android app's `defaultConfig` in `build.gradle`(depending on the project template). If not specified,
-the NDK block will be omitted, falling back to the default behavior.
+A list of strings that explicitly specifies the Android ABIs (platforms) the app will support. This controls the `abiFilters` parameter in the `ndk` block of the Android app's `defaultConfig` in `build.gradle`(depending on the project template). If not specified, the NDK block will be omitted, falling back to the default ABIs supported by Briefcase (currently `arm64v8a` and `x86_64`, with `armeabi-v7a` also included on Python 3.11 and earlier).
 
 For example, specifying:
 


changes/2723.feature.md
```diff
@@ -0,0 +1 @@
+Add option to configure Android ABIs
```
freakboy3742: Changenotes should be framed in the way they'd appear in release notes, describing the "Amazing new feature!!"
```suggestion
Android apps can now specify the ABIs that they target.
```

JohananOppongAmoateng:
@freakboy3742 i made the changes for the template too and made a pr

docs/en/reference/platforms/android/gradle.md
```diff
@@ -184,6 +184,29 @@ You may specify multiple `--reverse-port` arguments; each one specifies a single
 
 The following options can be added to the `tool.briefcase.app.<appname>.android` section of your `pyproject.toml` file.
 
+### `android_abis`
+
+A list of strings that explicitly specifies the Android ABIs (platforms) the app will support. This controls the `abiFilters` parameter in the `ndk` block of the Android app's `defaultConfig` in `build.gradle`(depending on the project template). If not specified,
+the NDK block will be omitted, falling back to the default behavior.
```
freakboy3742:
```suggestion
A list of strings that explicitly specifies the Android ABIs (platforms) the app will support. This controls the `abiFilters` parameter in the `ndk` block of the Android app's `defaultConfig` in `build.gradle`(depending on the project template). If not specified, the NDK block will be omitted, falling back to the default ABIs supported by Briefcase (currently `arm64v8a` and `x86_64`, with `armeabi-v7a` also included on Python 3.11 and earlier).
```

freakboy3742 (approved):
Thanks for those updates - with one minor tweak to the docs (to clarify the default behavior), this looks great!

I'll wait until the emergency v0.4.1 release has been finalized before merging.

Issue #2189, opened 2025-03-10
Signed android apk app does not install


alencina-faa:
### Describe the bug

I create, build, run and package (apk) an [app](https://github.com/alencina-faa/Leaf-Severity-Calculator_evolution/tree/android) with briefcase without errors. 
I signed apk with tree methods but apk never installed

### Steps to reproduce

**Method 1:** signed apk following [steps described here](https://briefcase.readthedocs.io/en/latest/how-to/publishing/android.html) (I had to change java to java17 folder) and these warnings were returned:

```
Warning:
The signer's certificate is self-signed.
The SHA1 algorithm specified for the -digestalg option is considered a security risk and is disabled.
The SHA1withRSA algorithm specified for the -sigalg option is considered a security risk and is disabled.
```
Apk does not installed.
adb log:
`adb: failed to install .\LeafSeverityCalculator-0.0.1.apk: Failure [INSTALL_PARSE_FAILED_NO_CERTIFICATES: Failed to collect certificates from /data/app/vmdl271308034.tmp/base.apk: META-INF/UPLOAD-K.SF has invalid digest for res/Hd.xml in /data/app/vmdl271308034.tmp/base.apk]`

**Method 2:**
Assuming a problem with SHA1 warnings' modify the key generation following [**New defaults for DSA keys in jarsigner and keytool**](https://www.oracle.com/java/technologies/javase/8u151-relnotes.html#:~:text=New%20defaults%20for%20DSA%20keys%20in%20jarsigner%20and%20keytool) from Oracle.
The key generation code is:
```
If (-Not (Test-Path "$env:HOMEPATH/.android")) { New-Item -Path "$env:HOMEPATH\.android" -ItemType Directory }
>> & "$env:LOCALAPPDATA\BeeWare\briefcase\Cache\tools\java17\bin\keytool.exe" -keyalg DSA -deststoretype pkcs12 -genkey -v -storepass android -keystore "$env:HOMEPATH\.android\upload-key-leafseveritycalculator_DSA.jks" -keysize 2048 -dname "cn=Upload Key" -alias upload-key -validity 10000
```
Wereas the signing code is:
[...]

PR #2724, opened 2026-03-07
feat: Add code signing for Android release artefacts

### LINKED ISSUE #1268: Add code signing for Android release artefacts

Briefcase currently supports code signing on macOS and Windows. Code signing is also required for Android release artefacts. 

`briefcase package android --identity <identity>` should accept a path to a keystore, and use that keystore to sign the app.

`briefcase package android` should prompt the user to do one of:
1. Ad-hoc sign the app (i.e., do nothing explicit with signatures, as is done currently)
2. Look for any file named `*.jks` in the project folder, the `.android` subfolder of the project folder, or the user's `~/.android` folder, and provide those files as candidate keystores to use.
3. Create a new keystore named `org.example.myapp.jks` in a `.android` folder of the project directory, and use that keystore.

Continue to require manual signing. The process of [generating a keystore](https://briefcase.readthedocs.io/en/latest/how-to/code-signing/android.html) and [signing the app](https://briefcase.readthedocs.io/en/latest/how-to/publishing/android.html); this is a simplification process.

### Additional context

_No response_

JohananOppongAmoateng:
Add code signing for Android release artefacts Closes #1268 

## PR Checklist:


docs/en/reference/platforms/android/gradle.md
```diff
@@ -358,4 +380,40 @@ For advice on how to deal with this situation, see the [Chaquopy FAQ](https://ch
...
 ### Signing of `briefcase package` artefacts
 
-While it is possible to use <span class="title-ref">briefcase package android</span> to produce an APK or AAB file for distribution, the file is *not* usable as-is. It must be signed regardless of whether you're distributing your app through the Play Store, or via loading the APK directly. For details on how to manually sign your code, see the instructions on [signing an Android App Bundle][sign-the-android-app-bundle].
+When building an AAB or a Release APK, Briefcase will automatically sign the artefact using a Java keystore (`.jks`) file.
+
+On the first run, Briefcase will search for existing keystores in the following locations:
+
+* The project root directory
+* The project's `.android/` subfolder
+* `~/.android/`
```
freakboy3742: `~` is UNIX-specific terminology - something like:
```suggestion
* The `.android/` folder in the user's home directory
```
would be preferred here.

docs/en/reference/platforms/android/gradle.md
```diff
@@ -358,4 +380,40 @@ For advice on how to deal with this situation, see the [Chaquopy FAQ](https://ch
...
+```
+
+To produce an *unsigned* artefact (for example, when signing is handled externally in a CI pipeline), use `--adhoc-sign`:
+
+```console
+$ briefcase package android --adhoc-sign
+```
+
+> [!IMPORTANT]
+> Keep your keystore file secure and backed up. If you lose it, you will not be able to publish updates to your app on the Play Store.
```
freakboy3742: This isn't the syntax we use for callouts. See the [admonitions and notes](https://briefcase.beeware.org/en/stable/how-to/contribute/style/docs-style-guide/#admonitions-and-notes) section of our style guide.

docs/en/reference/platforms/android/gradle.md
```diff
@@ -358,4 +380,40 @@ For advice on how to deal with this situation, see the [Chaquopy FAQ](https://ch
...
+
+If keystores are found, you will be presented with a list to choose from. If none are found — or if you choose to create a new one — Briefcase will generate a new keystore at `<project-root>/.android/<bundle-identifier>.jks` using `keytool` from your JDK installation.
+
+You can bypass the interactive selection by supplying the path to an existing keystore on the command line:
+
+```console
+$ briefcase package android --identity /path/to/my-release-key.jks
+```
+
+You can also supply the alias and passwords non-interactively:
```
freakboy3742: This is the first mention of alias and passwords; this might need some clarification:
```suggestion
If you are using a keystore that has been created with an alias or password, you can also supply those properties at the command line:
```

Also - what's the behavior if you interactively select an keystore from a list, but that keystore requires an alias or password?

src/briefcase/platforms/android/gradle.py
```diff
@@ -623,26 +634,298 @@ def forward_ports(
...
+                "(defaults to the keystore password if not specified)"
+            ),
+            required=False,
+        )
+
+    @property
+    def _keytool(self) -> Path:
+        """Path to the keytool executable bundled with the JDK."""
+        ext = ".exe" if self.tools.host_os == "Windows" else ""
+        return self.tools.java.java_home / "bin" / f"keytool{ext}"
```
freakboy3742: These keytool calls seem out of place here. They should be part of the `android_sdk` integrations. (Although the signing tools ship with java, it's very much tied to android signing, so we might as well keep it there). 

I'd suggest pulling these tools into a standalone object (a separate `signing` namespace) so `android_sdk.signing.create_keystore()` is how you create the initial keystore, etc.

src/briefcase/platforms/android/gradle.py
```diff
@@ -623,26 +634,298 @@ def forward_ports(
...
+                identity=identity,
+                keystore_alias=keystore_alias,
+                keystore_password=keystore_password,
+                key_password=key_password,
+            )
+            gradle_args += [
+                f"-Pandroid.injected.signing.store.file={signing_config.keystore_path!s}",
+                f"-Pandroid.injected.signing.store.password={signing_config.store_password}",
+                f"-Pandroid.injected.signing.key.alias={signing_config.alias}",
+                f"-Pandroid.injected.signing.key.password={signing_config.key_password}",
```
freakboy3742, line 769: Are these standard configuration points, or do we need to modify the Android gradle template to support these?

src/briefcase/platforms/android/gradle.py
```diff
@@ -623,26 +634,298 @@ def forward_ports(
...
             prefix=app.app_name,
         )
-        with self.console.wait_bar("Bundling..."):
-            build_type, build_artefact_path = {
-                "aab": ("bundleRelease", "bundle/release/app-release.aab"),
-                "apk": ("assembleRelease", "apk/release/app-release-unsigned.apk"),
-                "debug-apk": ("assembleDebug", "apk/debug/app-debug.apk"),
-            }[app.packaging_format]
 
+        build_type, build_artefact_path = {
```
freakboy3742, line 694: Why have you removed the `wait_bar` usage here?

freakboy3742 (dismissed):
Thanks for the PR. I've given a partial review; the general shape of things looks good, but there's some reorganisation that will be required, and I'm not 100% certain about some of the details in the implementation (specifically, if the `-P` arguments you're referring to are standard features, or something we'll need to accomodate). 

I haven't done a full review of the test cases yet, mostly because they'll see a bunch of changes from the reorganisation.

JohananOppongAmoateng:
@freakboy3742  i will address  all your concerns later bogged down with work now

src/briefcase/platforms/android/gradle.py
JohananOppongAmoateng: These work with the android gradle plugin so we dont need to modify the gradle template

src/briefcase/platforms/android/gradle.py
JohananOppongAmoateng: fixed it

docs/en/reference/platforms/android/gradle.md
mhsmith: Resolved by adding the text "If any required property is not provided, Briefcase will prompt for it".

src/briefcase/platforms/android/gradle.py
mhsmith: Although these properties aren't documented, it looks like they've always been the way in which [Android Studio passes signing settings to Gradle](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/builder-model/src/main/java/com/android/builder/model/InjectedProperties.kt?q=%22android.injected.signing.store.file%22) when building from the GUI, so they should be safe.

src/briefcase/platforms/android/gradle.py
```diff
@@ -623,26 +623,111 @@ def forward_ports(
...
+            "aab": ("bundleRelease", "bundle/release/app-release.aab"),
+            "apk": ("assembleRelease", "apk/release/app-release-unsigned.apk"),
```
mhsmith: When the signing options are passed, this filename changes to app-release.apk – see https://github.com/beeware/briefcase/issues/2189#issuecomment-2721878049.

docs/en/reference/platforms/android/gradle.md
```diff
@@ -122,7 +122,31 @@ Android allows for some customization of the colors used by your app:
 
 The following options can be provided at the command line when producing Android projects:
 
-### run
+### package { #android-package }
+
+#### `--keystore-alias <alias>`
```
mhsmith: For consistency, this and its references elsewhere should be called `key-alias`.

docs/en/reference/platforms/android/gradle.md
```diff
@@ -122,7 +122,31 @@ Android allows for some customization of the colors used by your app:
...
+
+The password for the keystore. If not provided, Briefcase will prompt for it.
+
+#### `--key-password <password>`
+
+The password for the signing key itself. If not provided, it defaults to the keystore password.
+
+#### `-i <path>` / `--identity <path>`
+
+The path to a `.jks` keystore file to use for signing. If not provided, Briefcase will search for existing keystores in the project folder (`.android/`), and in the `.android` folder in the user's home directory, and offer to create a new one if none are found.
```
mhsmith: I know we're trying to be consistent with macOS and Windows, but Android doesn't really have an equivalent to those platforms' `--identity` options, so reusing the same name is confusing. Suggest renaming it to `--keystore`. @freakboy3742: what do you think?

Either way, a note about Android should be added to the `--identity` section of package.md.

docs/en/reference/platforms/android/gradle.md
```diff
@@ -358,4 +382,43 @@ For advice on how to deal with this situation, see the [Chaquopy FAQ](https://ch
...
 ### Signing of `briefcase package` artefacts
 
-While it is possible to use <span class="title-ref">briefcase package android</span> to produce an APK or AAB file for distribution, the file is *not* usable as-is. It must be signed regardless of whether you're distributing your app through the Play Store, or via loading the APK directly. For details on how to manually sign your code, see the instructions on [signing an Android App Bundle][sign-the-android-app-bundle].
+When building an AAB or a Release APK, Briefcase will automatically sign the artefact using a Java keystore (`.jks`) file.
+
+On the first run, Briefcase will search for existing keystores in the following locations:
+
+* The project root directory
+* The project's `.android/` subfolder
+* The `.android/` folder in the user's home directory
```
mhsmith: @freakboy3742: Was this based on some existing convention? My impression is that the `.android/` folder in the user's home directory mostly contains ephemeral files automatically managed by the Android build tools, so encouraging the user to put a permanent key in there seems like a bad idea. Encouraging them to put it in the project directory is also risky, as this will likely lead them to commit it to source control.

So I suggest we remove all assumptions about the keystore location, and require the user always to specify it explicitly, including when creating it.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1415,6 +1422,217 @@ def start_emulator(
         return device, full_name
 
 
+@dataclass
+class AndroidSigningConfig:
+    keystore_path: Path
+    alias: str
+    store_password: str
+    key_password: str
```
mhsmith, line 1437: The order and names of these fields should match the documentation of the command-line arguments (see comments in gradle.md).

docs/en/reference/platforms/android/gradle.md
```diff
@@ -122,7 +122,31 @@ Android allows for some customization of the colors used by your app:
...
+
+#### `--key-password <password>`
+
+The password for the signing key itself. If not provided, it defaults to the keystore password.
+
+#### `-i <path>` / `--identity <path>`
+
+The path to a `.jks` keystore file to use for signing. If not provided, Briefcase will search for existing keystores in the project folder (`.android/`), and in the `.android` folder in the user's home directory, and offer to create a new one if none are found.
+
+If provided, these properties will be used to sign the artefact non-interactively. If any required property is not provided, Briefcase will prompt for it in an interactive session, or fail if input is disabled.
```
mhsmith: "Properties" is an implementation detail. 

```suggestion
If provided, these options will be used to sign the artefact non-interactively. If any required option is not provided, Briefcase will prompt for it in an interactive session, or fail if input is disabled.
```

docs/en/reference/platforms/android/gradle.md
```diff
@@ -358,4 +382,43 @@ For advice on how to deal with this situation, see the [Chaquopy FAQ](https://ch
...
+$ briefcase package android --adhoc-sign
+```
+
+/// note | Note
+
+Keep your keystore file secure and backed up. If you lose it, you will not be able to publish updates to your app on the Play Store.
+
+///
+
+Debug APKs (produced with `-p debug-apk`) are signed automatically by the Android SDK debug key and do not go through this signing flow.
```
mhsmith: This should link to https://developer.android.com/studio/publish/app-signing#debug-mode.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1415,6 +1422,217 @@ def start_emulator(
...
+                        "-keyalg",
+                        "RSA",
+                        "-keysize",
+                        "2048",
+                        "-validity",
+                        "10000",
+                        "-storepass",
+                        store_password,
+                        "-keypass",
+                        key_password,
```
mhsmith, line 1532: When I experimented with this command, I got the message "Warning:  Different store and key passwords not supported for PKCS12 KeyStores. Ignoring user-specified -keypass value." And when I run `keytool -list` on an older keystore, I get the message "The JKS keystore uses a proprietary format. It is recommended to migrate to PKCS12". So it looks like we should only support key passwords for reading existing keystores, not creating new ones.

It appears that the most common filename extension for PKCS12 files is .p12, and we should pass the `-storetype` option to make sure the format matches the name.

docs/en/reference/platforms/android/gradle.md
```diff
@@ -122,7 +122,31 @@ Android allows for some customization of the colors used by your app:
 
 The following options can be provided at the command line when producing Android projects:
 
-### run
+### package { #android-package }
```
mhsmith, line 125: A more logical order for these options would be --identity, --keystore-password, --keystore-alias, --key-password (but see the suggested renamings in the other comments).

docs/en/reference/platforms/android/gradle.md
```diff
@@ -122,7 +122,31 @@ Android allows for some customization of the colors used by your app:
...
+
+#### `--keystore-alias <alias>`
+
+The alias of the signing key within the keystore. If not provided, Briefcase will prompt for it (defaulting to the app name).
```
mhsmith: There's no point in documenting a "press enter to accept default" situation, this will just confuse the user into thinking that they can get away with not giving an alias at all.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1415,6 +1422,217 @@ def start_emulator(
...
+        :param key_password: The key password; defaults to store_password
+        :returns: An AndroidSigningConfig for the new keystore
+        """
+        keystore_path = base_path / ".android" / f"{app.bundle_identifier}.jks"
+
+        if keystore_alias is None:
+            keystore_alias = self.tools.console.text_question(
+                description="Key alias",
+                intro="Enter an alias for the signing key.",
+                default=app.app_name,
```
mhsmith: A better default would be "mykey", since that's the default used by keytool itself.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1415,6 +1422,217 @@ def start_emulator(
...
+            )
+
+        if store_password is None:
```
mhsmith, line 1496: The question order should match the documentation of the command-line arguments (see comments in gradle.md).

src/briefcase/integrations/android_sdk.py
```diff
@@ -1415,6 +1422,217 @@ def start_emulator(
...
+                        "-genkeypair",
+                        "-v",
+                        "-keystore",
+                        str(keystore_path),
+                        "-alias",
+                        keystore_alias,
+                        "-keyalg",
+                        "RSA",
+                        "-keysize",
+                        "2048",
```
mhsmith, line 1526: This should link to https://developer.android.com/build/building-cmdline#sign_cmdline.

src/briefcase/platforms/android/gradle.py
```diff
@@ -623,26 +623,111 @@ def forward_ports(
...
+            "debug-apk": ("assembleDebug", "apk/debug/app-debug.apk"),
+        }[app.packaging_format]
+
+        gradle_args = [build_type]
+
+        extra_gradle = getattr(app, "build_gradle_extra_content", "") or ""
+        if "signingConfig" in extra_gradle:
+            if identity is not None:
```
mhsmith: This error should appear when passing any keystore options, not just `--identity`.

src/briefcase/platforms/android/gradle.py
```diff
@@ -623,26 +623,111 @@ def forward_ports(
...
+
+For better security, consider removing the signing block from
+build_gradle_extra_content and using briefcase's built-in signing instead:
+
+    $ briefcase package android --identity /path/to/keystore.jks
+
+""")
+                adhoc_sign = True
+
+        if app.packaging_format != "debug-apk" and not adhoc_sign:
```
mhsmith: In order to upgrade in-place between debug and release builds, both apps must be signed with the same key. So It should be possible to sign the app even in debug-apk mode, even though this is not the default.

src/briefcase/platforms/android/gradle.py
mhsmith: However, for future reference, the code should include the above explanation and link.

src/briefcase/platforms/android/gradle.py
```diff
@@ -623,26 +623,111 @@ def forward_ports(
...
+            with self.console.wait_bar("Selecting signing credentials..."):
```
mhsmith: Either `select_keystore` will show some interactive questions, or it'll return immediately. Either way, I don't think a wait_bar is necessary.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1415,6 +1422,217 @@ def start_emulator(
...
+                raise BriefcaseCommandError("Failed to create keystore.") from e
+
+        self.tools.console.info(
+            f"""
+A new keystore has been created at:
+
+    {keystore_path}
+
+Keep this file secure and backed up. If you lose it, you will not be able to
+publish updates to this app on the Play Store.
```
mhsmith: This is enforced by Android itself, not just the Play Store.

```suggestion
publish updates to this app.
```

mhsmith:
Thanks, I'll look at this as soon as I can.

src/briefcase/platforms/android/gradle.py
mhsmith: The link you've added doesn't explain the property names; please change it to the above source code link, since that appears to be the only authoritative reference.

docs/en/reference/platforms/android/gradle.md
mhsmith: It doesn't look like `--keystore` was ever added. Argparse treats it as an abbreviation of `--keystore-password`, and the help message still says:
```
  -i IDENTITY, --identity IDENTITY
                        Ignored; signing is not supported
```

My suggestion was to remove `--identity` from the Android command completely, since it's not a meaningful word in this context. "Legacy reasons" aren't applicable here, since the option was never supported on Android to begin with. This may require some changes to the command's base class.

docs/en/reference/platforms/android/gradle.md
mhsmith: This hasn't been dealt with: the code still contains references to `base_path / ".android" / f"{app.bundle_identifier}.p12"`.

When a keystore isn't passed on the command line, my suggested interactive flow is:

* Give the user a choice between creating a new keystore, or using an existing keystore.
* In both cases, then ask for its location. If the user chooses to create a keystore but gives an existing location, do not overwrite it, because that could cause major inconvenience.

mhsmith: Here's the current behavior:
```
% briefcase package android

[helloworld] Building Android App Bundle and APK in release mode...


-- Keystore ------------------------------------------------------------------

Select the keystore to use for signing, or create a new one.

  1) Create a new keystore

Keystore: 1


-- Key alias -----------------------------------------------------------------

Enter an alias for the signing key.

Key alias [mykey]: 

Log saved to /Users/msmith/git/beeware/apps/helloworld/logs/briefcase.2026_04_08-23_28_00.package.log

  File "/Users/msmith/.venv/bw-310/bin/briefcase", line 7, in <module>
    sys.exit(main())
  File "/Users/msmith/git/beeware/briefcase/src/briefcase/__main__.py", line 27, in main
    command(**options)
  File "/Users/msmith/git/beeware/briefcase/src/briefcase/commands/package.py", line 194, in __call__
    state = self._package_app(
  File "/Users/msmith/git/beeware/briefcase/src/briefcase/commands/package.py", line 134, in _package_app
    state = self.package_app(app, **full_options(state, options))
  File "/Users/msmith/git/beeware/briefcase/src/briefcase/platforms/android/gradle.py", line 748, in package_app
    signing_config = self.tools.android_sdk.signing.select_keystore(
  File "/Users/msmith/git/beeware/briefcase/src/briefcase/integrations/android_sdk.py", line 1610, in select_keystore
    return self.create_keystore(
  File "/Users/msmith/git/beeware/briefcase/src/briefcase/integrations/android_sdk.py", line 1497, in create_keystore
    store_password = self.tools.console.text_question(
TypeError: Console.text_question() missing 1 required positional argument: 'default'
```

I don't think you've tested this properly. Please make sure you do that before asking for another review.

src/briefcase/integrations/android_sdk.py
mhsmith: For PKCS12 keystores, it looks like `-keypass` is redundant and can be removed.

src/briefcase/integrations/android_sdk.py
mhsmith: This still hasn't been done.

src/briefcase/integrations/android_sdk.py
mhsmith: This still hasn't been done, either here or in `select_keystore`.

src/briefcase/platforms/android/gradle.py
```diff
@@ -717,25 +731,41 @@ def package_app(
...
 
-    $ briefcase package android --identity /path/to/keystore.jks
+    $ briefcase package android --keystore /path/to/keystore.p12
 
 """)
                 adhoc_sign = True
 
-        if app.packaging_format != "debug-apk" and not adhoc_sign:
-            with self.console.wait_bar("Selecting signing credentials..."):
+        if not adhoc_sign:
```
mhsmith, line 739: This would be easier to follow if the `not` was removed and the if/else branches were reversed.

src/briefcase/platforms/android/gradle.py
```diff
@@ -693,21 +693,35 @@ def package_app(
 
         build_type, build_artefact_path = {
             "aab": ("bundleRelease", "bundle/release/app-release.aab"),
-            "apk": ("assembleRelease", "apk/release/app-release-unsigned.apk"),
+            "apk": (
+                "assembleRelease",
+                (
+                    "apk/release/app-release.apk"
+                    if keystore or not adhoc_sign
```
mhsmith, line 700: This is happening too early, because adhoc_sign can still change. Can we move it below the point where `signing_config` has been set, and then change the condition to `if signing_config`?

src/briefcase/platforms/android/gradle.py
mhsmith: I don't know why GitHub isn't showing the file location of these comments on the main PR page, but you can see it on the "Files changed" page.

Issue #1780, opened 2024-05-06
On Windows, renaming a directory that was just created can fail with `PermissionError: [WinError 5] Access denied`


rmartin16:
### Describe the bug

Using `os.rename()` (or `pathlib.Path.rename()`) can fail on Windows if Briefcase does not have exclusive access to all files in the directory.

A common pattern Briefcase uses is to run `shutil.unpack_archive()` and rename the unpacked directory; for instance, Android cmdline-tools and Java's JDK. In between these two operations, a background process may acquire a handle on a file; for instance, a Windows file indexer or an antivirus scanner.

Modern Linux file systems are much more resilient to these types of conflicts; so, I would only expect to see this on Windows.

### Steps to reproduce

```python
import os
from pathlib import Path

Path("orig-dir-1").mkdir()
Path("orig-dir-1/orig-file").touch()
Path("orig-dir-1").rename("new-dir-1")
os.listdir("new-dir-1")

Path("orig-dir-2").mkdir()
Path("orig-dir-2/orig-file").touch()
Path("orig-dir-2/orig-file").open()
Path("orig-dir-2").rename("new-dir-2")
```

```console
❯ python
Python 3.11.3 (tags/v3.11.3:f3909b8, Apr  4 2023, 23:49:59) [MSC v.1934 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> from pathlib import Path
>>>
>>> Path("orig-dir-1").mkdir()
>>> Path("orig-dir-1/orig-file").touch()
>>> Path("orig-dir-1").rename("new-dir-1")
WindowsPath('new-dir-1')
>>> os.listdir("new-dir-1")
['orig-file']
>>>
>>> Path("orig-dir-2").mkdir()
>>> Path("orig-dir-2/orig-file").touch()
>>> Path("orig-dir-2/orig-file").open()
<_io.TextIOWrapper name='orig-dir-2\\orig-file' mode='r' encoding='cp1252'>
>>> Path("orig-dir-2").rename("new-dir-2")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\user\.pyenv\pyenv-win\versions\3.11.3\Lib\pathlib.py", line 1175, in rename
    os.rename(self, target)
PermissionError: [WinError 5] Access is denied: 'orig-dir-2' -> 'new-dir-2'
>>>
```

### Expected behavior

[...]

PR #2725, opened 2026-03-08
Fix Win rename bug using `tenacity`

JohananOppongAmoateng:
Fixes https://github.com/beeware/briefcase/issues/1780
On Windows, renaming a directory that was just created can fail with PermissionError: [WinError 5] Access denied. Based on the work of @spa51273 in #1826

## PR Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,6 +100,7 @@ dependencies = [
     "rich >= 12.6, < 15.0",
     "tenacity >= 8.0, < 9.0",
     "tomli >= 2.0, < 3.0; python_version <= '3.10'",
+    "tomli_w >= 1.0, < 2.0",
 ]
 
 [dependency-groups]


pyproject.toml
```diff
@@ -97,9 +97,11 @@ dependencies = [
     "psutil >= 5.9, < 8.0",
     "python-dateutil >= 2.9.0.post0",  # transitive dependency (beeware/briefcase#1428)
     "httpx >= 0.20, < 1.0",
+    "requests >= 2.28, < 3.0",
     "rich >= 12.6, < 15.0",
+    "tenacity >= 8.0, < 9.0",
     "tomli >= 2.0, < 3.0; python_version <= '3.10'",
-    "tomli_w >= 1.0, < 2.0",
+    "tomli_w >= 1.0, < 2.0"
```
mhsmith: Why add requests and tomli_w?

changes/1780.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added Files integration using tenacity for path dir rename retry
```
mhsmith: This does not adequately explain the change. Please phrase it in a way that a normal Briefcase user would understand.

src/briefcase/integrations/file.py
```diff
@@ -337,3 +338,17 @@ def _fetch_and_write_content(self, response: httpx.Response, filename: Path):
             # exist if the download fails or the user sends CTRL+C.
             with suppress(FileNotFoundError):
                 self.tools.os.remove(temp_file.name)
+
+    @retry(
+        retry=retry_if_exception_type(PermissionError),
+        wait=wait_fixed(0.2),
+        stop=stop_after_attempt(25),
+    )
+    def path_rename(self, old_path: Path, new_path: object):
```
mhsmith: This is already inside a scope called `file`, so `rename` is a clear enough name: `path` is redundant.

src/briefcase/integrations/android_sdk.py
```diff
@@ -441,8 +441,9 @@ def install(self):
                 self.tools.shutil.rmtree(self.cmdline_tools_path)
 
             # Rename the top level zip content to the final name
-            (self.cmdline_tools_path.parent / "cmdline-tools").rename(
-                self.cmdline_tools_path
+            self.tools.file.path_rename(
+                old_path=(self.cmdline_tools_path.parent / "cmdline-tools"),
+                new_path=self.cmdline_tools_path,
```
mhsmith: The arguments of a rename function are obvious, there's no need to use keyword arguments when calling it.

src/briefcase/integrations/android_sdk.py
```diff
@@ -441,8 +441,9 @@ def install(self):
                 self.tools.shutil.rmtree(self.cmdline_tools_path)
 
             # Rename the top level zip content to the final name
-            (self.cmdline_tools_path.parent / "cmdline-tools").rename(
-                self.cmdline_tools_path
+            self.tools.file.path_rename(
+                old_path=(self.cmdline_tools_path.parent / "cmdline-tools"),
```
mhsmith: These parentheses are now unnecessary.

tests/integrations/file/test_Files__rename.py
```diff
@@ -0,0 +1,43 @@
...
+    mock_tools.file.path_rename(tmp_path / "orig-dir-1", tmp_path / "new-dir-1")
+    file_access_thread.join()
+
+    assert "new-dir-1" in os.listdir(tmp_path)
+
+
+def test_rename_path_file_not_found(mock_tools, tmp_path):
+    """A FileNotFoundError is raised immediately without retrying."""
+    with pytest.raises(FileNotFoundError):
+        mock_tools.file.path_rename(tmp_path / "does-not-exist", tmp_path / "new-name")
```
mhsmith: This doesn't prove that it's raised immediately without retrying. To do that you'd need to replace the underlying rename method with a mock, as in the next test.

tests/integrations/file/test_Files__rename.py
```diff
@@ -0,0 +1,43 @@
...
+from pathlib import Path
+from unittest.mock import MagicMock
+
+import pytest
+import tenacity
+
+
+def test_rename_path(mock_tools, tmp_path):
+    def openclose(filepath):
+        handler = filepath.open(encoding="UTF-8")
```
mhsmith, line 21: Please add a docstring explaining what this is doing and why.

tests/integrations/file/test_Files__rename.py
```diff
@@ -0,0 +1,43 @@
...
+        finally:
+            handler.close()
+
+    (tmp_path / "orig-dir-1").mkdir()
+    tl = tmp_path / "orig-dir-1/orig-file"
+    tl.touch()
+    file_access_thread = threading.Thread(target=openclose, args=(tl,))
+
+    file_access_thread.start()
+    mock_tools.file.path_rename(tmp_path / "orig-dir-1", tmp_path / "new-dir-1")
```
mhsmith: If the rename happens before the thread starts, the test will fail with a file not found error. The main thread should sleep before the rename to make sure that doesn't happen.

tests/integrations/file/test_Files__rename.py
```diff
@@ -0,0 +1,43 @@
...
+def test_rename_path_file_not_found(mock_tools, tmp_path):
+    """A FileNotFoundError is raised immediately without retrying."""
+    with pytest.raises(FileNotFoundError):
+        mock_tools.file.path_rename(tmp_path / "does-not-exist", tmp_path / "new-name")
+
+
+def test_rename_path_fail(mock_tools, tmp_path, monkeypatch):
+    """Retries are exhausted when rename repeatedly raises PermissionError."""
+    monkeypatch.setattr(tenacity.nap.time, "sleep", lambda x: True)
+    monkeypatch.setattr(Path, "rename", MagicMock(side_effect=PermissionError))
```
mhsmith: If you replace `sleep` with a mock as well, then you can verify the number of retries and the time between them.

pyproject.toml
JohananOppongAmoateng: tomli_w is a pre-existing dependency i accidentally removed the  comma that is why its showing in the diff

PR #2732, opened 2026-03-10
Use PEP639 license format internally.

### LINKED ISSUE #2146: supports PEP 639

freakboy3742:
Switch to using PEP 639 license metadata internally. 

* Existing projects without a license declaration get a warning, a dummy licence file, and an SPDX license declaration of `LicenseRef-UnknownLicense`
* Existing projects in pre-PEP621 `license` format have the value of `license` written to a dummy file, and a SPDX-compliant value of `license` derived from the contents of the existing `license` field. A value of `LicenseRef-UnknownLicense` is used if a license cannot be determined.
* Existing projects in PEP621 `license.text` format are treated the same as pre-PEP621 `license` format
* Existing projects in PEP621 `license.file` format have the location turned into `license-files`, and a `license` value guessed from the license text. A value of `LicenseRef-UnknownLicense` is used if a license cannot be determined.
* Any value provided in `license-files` (before or after coercion to PEP 639) must exist.

If a dummy license file is needed, it is written into the `build` directory as `build/license_file.<app name>.txt`

As a result of the config parsing changes, we can now assume that full license files are *always* available when building an app. This significantly simplifies the app builds.

Disclosure: This PR was generated with the use of SpecKit and Claude Sonnet 4.6, but has had substantial manual modifications.

Fixes #2146.
Refs beeware/briefcase-template#249 which makes the PEP639 change to the default app template.

## PR Checklist:


tests/config/test_parse_config.py
```diff
@@ -603,37 +652,45 @@ def test_document_types():
...
         },
     }
 
 
-def test_pep621_defaults():
-    config_file = BytesIO(b"""
+def test_pep_621_merge(tmp_path):
+    """PEP 621 [project] fields are merged into the Briefcase config."""
+    # A license file must exist. The definition is split across `project` and
+    # `tool.briefcase` to prove merging happens
```
mhsmith, line 682: Is there a test for a current-format `license` key in the `[project]` section?

changes/2146.feature.md
```diff
@@ -0,0 +1 @@
+License metadata in ``pyproject.toml`` is now parsed and validated using PEP 639 format (i.e., `license = "<SPDX expression>"` and `license-files = [...]`).
```
mhsmith, line 1: The documentation in configuration.rst needs to be updated:
* `license` section in the middle
* "PEP 621" section at the bottom, which should probably be renamed now that the living spec has diverged from the PEP.

src/briefcase/config.py
```diff
@@ -640,6 +646,398 @@ def merge_config(config, data):
...
+The license configuration for '{app_name}' mixes PEP 621 table format
+(`license.file`) with PEP 639 format (`license-files`).
+
+Update your configuration to use PEP 639 format:
+
+    license = "<SPDX expression>"
+    license-files = ["LICENSE"]
+""") from None
+
+    # license_files without a license expression is always an error
```
mhsmith: This contradicts [the spec](https://packaging.python.org/en/latest/specifications/pyproject-toml/#license-files), which says:

> If the license-files key is present and is set to a value of an empty array, then tools MUST NOT include any license files and MUST NOT raise an error. If the license-files key is not defined, tools can decide how to handle license files. For example they can choose not to include any files or use their own logic to discover the appropriate files in the distribution.

src/briefcase/config.py
```diff
@@ -640,6 +646,398 @@ def merge_config(config, data):
...
+        for license_pattern in license_patterns:
+            if license_pattern.lower() in license_text.lower():
+                return license_id
+
+    return default
+
+
+def normalize_license_config(
+    config: AppConfig,
+    app_name: str,
```
mhsmith: This function is extremely difficult to understand. Fixing the indentation as in #2559 would help a lot, but what might be even more helpful is to split each major branch into a separate function so the top-level control flow can be perceived at a glance.

src/briefcase/config.py
mhsmith: Sorry, this comment was meant to be at "Empty license-files list is always an error" below.

tests/config/test_parse_config.py
freakboy3742: There's a test of project/tool section merging, and test of current format `license` handling; but not an explicit test of both. I've added parameterization so we're explicitly testing both.

changes/2146.feature.md
freakboy3742: Good catch - updated.

src/briefcase/config.py
freakboy3742: Huh - that's... inconvenient. 

This is problematic because Linux and Windows projects both *require* the existence of a file.

Since `license` appears to be a required field, I'll modified the code to *allow* an empty/absent license-files definition, but use dummy `"Released under the terms of the {license} license"` text with a warning that the user probably wants to provide an actual license.

src/briefcase/config.py
freakboy3742: That's fair - I'll break this up in to smaller pieces.

src/briefcase/config.py
freakboy3742: On reflection, I think the better approach is to avoid dummy files wherever possible. 

Since an empty or absent `license-files` list is legal, I've taken the approach of always normalising `license-files` to a list - even if it's an empty list, and only populating a file if a "text-like" license value contains more than one line (i.e., trying to capture the `license.text = """<actual license text>"""` case), and raising warnings about the likely migration path. However, if there isn't any content that could be plausibly considered a "full" license, we don't write a file, and we use `license-files = []`.

Then, when the licenses are actually used, we raise an error if the packaging format requires a license, but one hasn't been provided. That means we can create an iOS/Android/macOS project without any license files; but Windows and Linux will both raise an error (because they have a hard requirement during packaging).

docs/en/reference/configuration.md
```diff
@@ -495,17 +503,28 @@ A URL for help related to the document format.
...
 
-The following PEP621 project metadata keys will be used by Briefcase if they are available:
+The following PEP 621 project metadata keys will be used by Briefcase if they are available:
 
 - `version` maps to the same key in Briefcase.
 - `authors` The `email` and `name` keys of the first value in the `authors` setting map to [`author`][] and [`author_email`][].
 - `dependencies` maps to the Briefcase [`requires`][] setting. This is a cumulative setting; any packages defined in the [`requires`][] setting at the `[tool.briefcase]` level will be appended to the packages defined with `dependencies` at the `[project]` level.
 - `description` maps to the same key in Briefcase.
 - `test` in an `[project.optional-dependencies]` section maps to [`test_requires`][]., As with `dependencies`/[`requires`][], this is a cumulative setting.
-- `text` in a `[project.license]` section will be mapped to [`license`][].
```
mhsmith, line 509: `license` and `license_files` should still be included here. Maybe we could say "The following `[project]` keys have the same format as in `[tool.briefcase]`", and then list them along with `version` and `description`.

And we shouldn't be referring to PEP 621 here, because we don't want to encourage people to look at an outdated PEP and get confused by the different license format. Suggest renaming the section to "`[project]` section" or similar, and then referring to "`[project]` metadata" elsewhere, with the square brackets making it clear that we're not referring to the `tool.briefcase` section.

freakboy3742: Agreed that strict PEP references should be avoided, and we need to integrate `license` and `license-files` more coherently. I've pushed an update to this effect.

mhsmith (approved):

PR #2738, opened 2026-03-14
Add support for MIME type definitions on document types

### LINKED ISSUE #2154: Add support for MIME type definitions on document types

The current specification for document types lets you define an extension, description and icon for a file type - but doesn't let you define a MIME type. Some platforms (e.g., Windows) require a MIME type definition for file association purposes.

It should be possible to define a mime type for a document type:
```
[tool.briefcase.app.helloworld.document_type.foobar]
description = "A foo bar document"
extension = "fbr"
icon = "icons/foobar"
url = "https://example.com"
mime_type = "text/plain"
```

Status quo - an autogenerated mime type based on the document type ID and application name. 

### Additional context

We could do *some* validation of the mime type; there is a [set of known registries](https://www.iana.org/assignments/media-types/media-types.xhtml) that should be the only valid starting text for a mime type (e.g., `application/...`, `text/...`).

Once these this is formalised, we should ensure all app templates use the mime type as appropriate.

JohananOppongAmoateng:
Add support for MIME type definitions on document types

Closes #2154

## PR Checklist:

### DIFF
--- a/tests/config/test_validate_document_type_config.py
+++ b/tests/config/test_validate_document_type_config.py
@@ -172,6 +172,11 @@ def test_document_type_macOS_config_with_mimetype_single(valid_document):
     """
     valid_document["mime_type"] = "application/pdf"
     validate_document_type_config("my-app", "ext", valid_document)
+
+    # MIME type hasn't changed
+    assert valid_document["mime_type"] == "application/pdf"
+
+    # macOS mime configs have been added
     assert "LSItemContentTypes" in valid_document["macOS"]
     assert valid_document["macOS"]["LSItemContentTypes"] == ["com.adobe.pdf"]
     assert valid_document["macOS"]["is_core_type"] is True
@@ -189,6 +194,11 @@ def test_document_type_macOS_config_with_mimetype_list(valid_document):
     """
     valid_document["mime_type"] = "text/vcard"
     validate_document_type_config("my-app", "ext", valid_document)
+
+    # MIME type hasn't changed
+    assert valid_document["mime_type"] == "text/vcard"
+
+    # macOS mime configs have been added
     assert "LSItemContentTypes" in valid_document["macOS"]
     assert valid_document["macOS"]["LSItemContentTypes"] == ["public.vcard"]
     assert valid_document["macOS"]["is_core_type"] is True
@@ -206,6 +216,11 @@ def test_document_type_macOS_config_with_unknown_mimetype(valid_document):
     """
     valid_document["mime_type"] = "application/x-custom"
     validate_document_type_config("my-app", "ext", valid_document)
+
+    # MIME type hasn't changed
+    assert valid_document["mime_type"] == "application/x-custom"
+
+    # macOS mime configs have been added
     assert "LSItemContentTypes" not in valid_document["macOS"]
     assert valid_document["macOS"]["is_core_type"] is False
     assert valid_document["macOS"]["LSHandlerRank"] == "Owner"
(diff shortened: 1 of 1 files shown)


tests/config/test_validate_document_type_config.py
```diff
@@ -215,6 +216,41 @@ def test_document_type_macOS_config_with_unknown_mimetype(valid_document):
...
 
 
+@pytest.mark.parametrize(
+    ("invalid_mime_type", "match"),
+    [
+        (1, r"The MIME type associated with document type 'ext' is not a string\."),
+        (
+            "not-a-mime-type",
+            r"The MIME type 'not-a-mime-type' for document type 'ext' "
+            r"is not in 'type/subtype' format\.",
```
freakboy3742: I've raised this in previous reviews - our code standard says that if a string breaks across multiple lines in a way that is ambiguous about whether a string is a single line or multiple lines, we use () to disambiguate:
```suggestion
            (
                r"The MIME type 'not-a-mime-type' for document type 'ext' "
                r"is not in 'type/subtype' format\."
            ),
```

tests/config/test_validate_document_type_config.py
```diff
@@ -171,7 +171,7 @@ def test_document_type_macOS_config_with_mimetype_single(valid_document):
     application/pdf is the only valid MIME type for PDF files.
     """
     valid_document["mime_type"] = "application/pdf"
-    validate_document_type_config("ext", valid_document)
+    validate_document_type_config("my-app", "ext", valid_document)
```
freakboy3742, line 174: If this is (potentially) overwriting the mime type, we should validate that the mime type hasn't changed as a result of validation.

tests/config/test_validate_document_type_config.py
```diff
@@ -188,7 +188,7 @@ def test_document_type_macOS_config_with_mimetype_list(valid_document):
     should still resolve to public.vcard
     """
     valid_document["mime_type"] = "text/vcard"
-    validate_document_type_config("ext", valid_document)
+    validate_document_type_config("my-app", "ext", valid_document)
```
freakboy3742, line 196: Again - we should validate that this hasn't changed the mime type.

src/briefcase/constants.py
```diff
@@ -379,3 +379,19 @@
     PYTHON_RESERVED_WORDS,
     NON_PYTHON_RESERVED_WORDS,
 )
+
+# Valid MIME type registries
+# https://www.iana.org/assignments/media-types/media-types.xhtml
+MIME_TYPES = {
```
freakboy3742: This constant is a little misleading - it contains mime type *registries*, not mime types: 
```suggestion
MIME_TYPE_REGISTRIES = {
```

src/briefcase/config.py
```diff
@@ -152,6 +152,29 @@ def validate_document_type_config(document_type_id, document_type):
...
+    try:
+        mime_type = document_type["mime_type"]
+        if not isinstance(mime_type, str):
+            raise BriefcaseConfigError(
+                f"The MIME type associated with document type "
+                f"{document_type_id!r} is not a string."
+            )
+        try:
+            main_type, _ = mime_type.split("/")
+            if main_type not in MIME_TYPES:
```
freakboy3742: A minor thing - but it helps to maintain consistency in naming. If it's a registry, call it a registry in variables that store it.

tests/config/test_validate_document_type_config.py
freakboy3742: This comment hasn't been addressed.

tests/config/test_validate_document_type_config.py
freakboy3742: This comment hasn't been addressed.

freakboy3742 (approved):
There were two open comments from my previous review that hadn't been addressed; I've added tests to cover those two issues.

Otherwise - this looks good; thanks for the contribution!

Issue #2739, opened 2026-03-15
Add type-checker-only AppConfig subclasses for platform-specific attributes


filiplajszczak:
### What is the problem or limitation you are having?

Platform backends set platform-specific attributes on `AppConfig` objects during `finalize_app_config()`. For example, `LinuxSystemMixin.finalize_app_config()` sets `target_vendor`, `target_codename`, `glibc_version`, etc. These attributes aren't declared on `AppConfig` or `BaseConfig`.

At runtime this works because the attributes are always set before they're read (`finalize_app_config()` runs before any method that accesses the finalized fields). But `ty` sees `app: AppConfig` and reports `unresolved-attribute` for every access to these dynamic fields.

This is currently the largest remaining source of `unresolved-attribute` diagnostics after #2734: **80 on `AppConfig`** and **14 on `BaseConfig`**, out of 350 total diagnostics (ty 0.0.21).

### Describe the solution you'd like

Define type-checker-only `AppConfig` subclasses for each platform under `if TYPE_CHECKING`, then use `cast()` to narrow the type at usage sites. These exist only for the type checker; zero runtime cost.

In each platform module, declare the finalized config shape:

```python
from __future__ import annotations
from typing import TYPE_CHECKING, cast

if TYPE_CHECKING:
    class LinuxSystemAppConfig(AppConfig):
        target_vendor: str
        target_codename: str
        target_vendor_base: str
        target_image: str
        glibc_version: str
        python_version_tag: str
        packaging_format: str
```

A thin helper wraps the `cast()` so the intent reads clearly at each call site:

```python
AppConfigT = TypeVar("AppConfigT", bound=BaseConfig)

def as_platform_config(config_type: type[AppConfigT], app: AppConfig) -> AppConfigT:
    """Narrow an AppConfig to a platform-specific type for type checking 
    purposes. At runtime this is a no-op.
    """
    return cast(AppConfigT, app)
```

At the point where the backend knows the config has been finalized:

```python
def distribution_filename(self, app: AppConfig) -> str:
    app = as_platform_config(LinuxSystemAppConfig, app)
    if app.packaging_format == "deb":
        return f"{app.bundle_name}_{app.version}-{app.revision}..."
```

Bare `cast()` would work the same way, but `as_platform_config` is worth the one extra function: it gives the pattern a name, carries a docstring, and makes it greppable. When someone reads `as_platform_config(LinuxSystemAppConfig, app)` it's immediately obvious what's happening and why.

[...]

PR #2752, opened 2026-03-16
Return finalized apps from finalize() and finalize_app_config()

filiplajszczak:
Refs #2739 gradual typing improvements.

`finalize_app_config()` now returns the app it was given, and `finalize()` collects the results into a `dict[str, AppConfig]` and returns it. All command `__call__` methods use the returned mapping for iteration instead of the original dict they passed in.

The existing mutation-based flow is preserved, but callers now have an explicit data-flow path. This enables future work where `finalize_app_config()` could return a different object, and where callers can rely on the return value rather than mutation.

## PR Checklist:

### DIFF
--- a/tests/commands/base/conftest.py
+++ b/tests/commands/base/conftest.py
@@ -44,7 +44,7 @@ class DummyCommand(BaseCommand):
         self.actions.append(("verify-tools",))
 
     def finalize_app_config(self, app):
-        super().finalize_app_config(app=app)
+        app = super().finalize_app_config(app=app)
         self.actions.append(("finalize-app-config", app.app_name))
         return app
 
--- a/tests/commands/build/conftest.py
+++ b/tests/commands/build/conftest.py
@@ -39,7 +39,7 @@ class DummyBuildCommand(BuildCommand):
         self.actions.append(("verify-tools",))
 
     def finalize_app_config(self, app):
-        super().finalize_app_config(app=app)
+        app = super().finalize_app_config(app=app)
         self.actions.append(("finalize-app-config", app.app_name))
         return app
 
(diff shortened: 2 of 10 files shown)


tests/commands/base/conftest.py
```diff
@@ -46,6 +46,7 @@ def verify_tools(self):
     def finalize_app_config(self, app):
         super().finalize_app_config(app=app)
         self.actions.append(("finalize-app-config", app.app_name))
+        return app
```
freakboy3742, line 49: Should the return value of the `super()` be incorporated in the result here? If the test is passing, it clearly isn't *required*... but in terms of maintaining consistency, it feels like we should maintain it regardless.

tests/commands/build/conftest.py
```diff
@@ -41,6 +41,7 @@ def verify_tools(self):
     def finalize_app_config(self, app):
         super().finalize_app_config(app=app)
         self.actions.append(("finalize-app-config", app.app_name))
+        return app
```
freakboy3742, line 44: Same here - should this be incorporating super()?

changes/2739.misc.md
```diff
@@ -0,0 +1 @@
+``finalize_app_config()`` and ``finalize()`` now return their results, enabling callers to use the returned mapping instead of relying on mutation alone.
```
freakboy3742: Markdown only requires a single quote.
```suggestion
`finalize_app_config()` and `finalize()` now return their results, enabling callers to use the returned mapping instead of relying on mutation alone.
```

tests/commands/base/conftest.py
filiplajszczak: Yeah, that makes sense consistency-wise. I've applied it to all ten dummy commands.

tests/commands/build/conftest.py
filiplajszczak: That and 8 others.

PR #2755, opened 2026-03-19
Note that MSI pre/post install scripts now run with elevated privileges.

### LINKED ISSUE #2722: MSI: pre/post install scripts run without elevation for all user installs
### Describe the bug

When using a post-install or pre-uninstall script with an MSI installer, the custom action runs without elevated privileges when I try to install for "all users". This causes permission errors when the script attempts to write to protected locations such as HKLM registry keys or Program Files.

Note that if I run the installer via an elevated command line, then it works fine as is.

### How to reproduce the bug

1. Define a post_install_script or pre_uninstall_script in pyproject.toml
2. Build an MSI installer with briefcase package windows
3. Double click the installer.
3. Select "All Users" when choosing install scope.
4. Observe that the post-install script runs without elevation, causing permission denied errors for any writes to protected locations.

### Minimum example code

```shell
[tool.briefcase.app.myapp]
post_install_script = "post_install.bat"
```
For the contents of `post_install.bat` below is what I used for the screenshot:
```shell
@echo off
copy NUL "C:\Program Files\test_elevation.txt"
if errorlevel 1 (
    echo [ERROR] Failed to write to Program Files - insufficient privileges
    pause
    exit /b 1
)
exit /b 0
```

### Screenshots

Image

### Environment details

**Environment details:**
```
- Operating system and version: Windows 10/11
- Python version: 3.x
- Software versions:
  - Briefcase: latest
```

### Additional context

**Proposed fix:**

In the WiX template, change the `CustomAction` `Execute` attribute from `"immediate"` to `"deferred"` for both (post and pre install/uninstall).
Finally, for post install script I also needed to modify the `PostInstallAction` from `After="InstallFinalize"` to `Before="InstallFinalize"`, which **seems**  required for deferred actions to execute correctly within the MSI execution sequence (not sure about the direct impact of this but without it I ran into an error).
```
         {% if cookiecutter.post_install_script %}
-            <Custom Action="PostInstallAction" After="InstallFinalize" Condition="NOT Installed" />
+            <Custom Action="PostInstallAction" Before="InstallFinalize" Condition="NOT Installed" />
         {% endif %}

```
The error (without changing this) was:
```
The installer has encountered an unexpected error installing this package. 
This may indicate a problem with this package. The error code is 2762.
```

freakboy3742:
Fixes #2722.

The actual fix for this is in templates (see beeware/briefcase-windows-app-template#85, also ported to the Visual Studio template); this changenote is so the change is visible publicly.

## PR Checklist:


kattni (approved):
Thanks for the update!

PR #2756, opened 2026-03-19
Separate finalization from runtime concerns and introduce `FinalizedAppConfig`.

filiplajszczak:
`finalize()` previously mixed two concerns: platform configuration (`finalize_app_config`) and runtime mode setup (`test_mode`, `debugger`). This change moves runtime attributes to callers, so `finalize()` only handles host/tool verification and app config finalization.

With that separation in place, the `__draft__` magic attribute is replaced by a real class: `FinalizedAppConfig`. Unfinalized apps are `AppConfig`, finalized ones become `FinalizedAppConfig`. This would give the type checker a concrete type to track finalization state, replacing `hasattr`/`delattr` with `isinstance`.

Groundwork for typing app finalization (Refs #2739).

## PR Checklist:

### DIFF
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -33,6 +33,7 @@ import briefcase
 from briefcase import __version__
 from briefcase.config import (
     AppConfig,
+    FinalizedAppConfig,
     GlobalConfig,
     parse_config,
 )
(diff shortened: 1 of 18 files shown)


src/briefcase/commands/base.py
```diff
@@ -782,6 +786,7 @@ def finalize(
                             f"but not 'external_package_path'."
                         )
             finalized[app.app_name] = app
+        self.apps.update(finalized)
```
freakboy3742, line 789: Do we actually need `finalized` at all here? Can we cache finalized apps directly into `self.apps`?

freakboy3742: Answered on the upstream discussion - this is needed to avoid modification during iteration. This likely isn't a *big* problem as we're not adding new keys, but better to be safe.

PR #2760, opened 2026-03-21
Check for Python x86 Interpreter on Apple Silicon Mac

### LINKED ISSUE #2736: iOS App Fails to Run on Simulator from Briefcase Command Line
### Describe the bug

I was going through the Beeware "Hello World" tutorial and attempted to launch the app on the iOS simulator. 
I was at this part of the [tutorial here](https://tutorial.beeware.org/tutorial/tutorial-5/iOS/).

When I ran the command `briefcase run iOS`, the simulator came up, but the app failed to launch. 
After that, I opened Xcode by running the command `briefcase open iOS` and ran the simulator from Xcode.

From Xcode, I was able to run the simulator, see the app launch, and also interact with the app. 

What I expected to happen was to be able to launch the app within the iOS simulator from the briefcase command line (ie by running `briefcase run iOS`). Instead I could only launch the app from the iOS simulator from within Xcode.

### How to reproduce the bug

1. Follow the Beeware HelloWorld tutorial from [the beginning](https://tutorial.beeware.org/tutorial/tutorial-1/)
2. Ran `python -m pip install briefcase`
3. Ran `briefcase new`. Accept all defaults, except used my own name and email address (email address has been redacted from log files uploaded to this issue). 
4. Then, created an iOS app by running `briefcase create iOS`
5. Ran `briefcase build iOS`
6. Ran `briefcase run iOS -vv`. Selected iOS 26.3 and iPhone 16e device.
7. Simulator came up, but app failed to launch in simulator.
8. Then, open Xcode by running `briefcase open iOS`
9. Selected the iPhone 16 simulator, and check that its running iOS 26.3 and press the play button. See that the app launches just fine from Xcode. Screenshot in this issue was from successfully running the app from Xcode.
10. Optionally, interact with the app and make it display the "Hello" dialog box and see that it works just fine launching from Xcode.

### Minimum example code

```shell

```

### Screenshots

Image

### Environment details

- Operating system and version: MacOS Tahoe 26.3.1
- Architecture: Apple M3 chip / ARM64
- Python version: 3.14.3, installed via Homebrew
- Software versions:
  - Briefcase: 0.4.1

### Logs

[Output_From_RunningSimulatorFromXcode.txt](https://github.com/user-attachments/files/25979253/Output_From_RunningSimulatorFromXcode.txt)
[Output_From_InstallingBriefcase.txt](https://github.com/user-attachments/files/25979250/Output_From_InstallingBriefcase.txt)
[Output_From_briefcase_run_iOS_-vv.txt](https://github.com/user-attachments/files/25979252/Output_From_briefcase_run_iOS_-vv.txt)
[Output_From_briefcase_new.txt](https://github.com/user-attachments/files/25979254/Output_From_briefcase_new.txt)
[Output_From_briefcase_create_iOS.txt](https://github.com/user-attachments/files/25979251/Output_From_briefcase_create_iOS.txt)
[Output_From_briefcase_build_iOS.txt](https://github.com/user-attachments/files/25979249/Output_From_briefcase_build_iOS.txt)
[briefcase.2026_03_13-11_27_47.run-EMAIL-REDACTED.log](https://github.com/user-attachments/files/25979255/briefcase.2026_03_13-11_27_47.run-EMAIL-REDACTED.log)

[pyproject-EMAIL-REDACTED.toml.txt](https://github.com/user-attachments/files/25979283/pyproject-EMAIL-REDACTED.toml.txt)

### Additional context

_No response_

vuinguyen:
This PR fixes #2736. 

The problem this PR solves:
- While following the [beeware tutorial](https://tutorial.beeware.org/tutorial/tutorial-5/iOS/), I was unable to launch the app from the iOS simulator from the briefcase command `briefcase run iOS`.
- Upon further research, it was determined that while porting my x86 mac to my current Apple Silicon mac, I ended up with a Python interpreter environment that claimed to be running on both x86 and Apple Silicon. 
- This caused the briefcase command line tools to call the wrong scripts (the one for x86), leading to the failure of the app launch from the iOS simulator.
- This PR solves the underlying problem in that now the briefcase command line tools will raise an error if it is trying to call the Python x86 interpreter while on an Apple Silicon machine.

Changes implemented offline (not in briefcase) and testing:
- Uninstalled all Python interpreters on my machine and reinstalled Python 3.13 from Homebrew.
- Re-created the beeware tutorial project with the newly installed Python 3.13 in the venv. 
Ran the briefcase command `briefcase run iOS` and saw the the app launch in the iOS simulator and successfully interacted with the app in the simulator.

Changes implemented in briefcase: 
- `changes/2736.bugfix.md` file: Change document
- `src/briefcase/platforms/macOS/__init__.py` file: Added a `verify_tools()` function in the `macOSMixin` class to check for an x86 Python interpreter running on Apple Silicon hardware
- `tests/platforms/macOS/test_macOSMixin_verify_tools.py` file: Add a unit test to test the `macOSMixin.verify_tools()` function.

## PR Checklist:

### DIFF
--- a/changes/2736.bugfix.md
+++ b/changes/2736.bugfix.md
@@ -1,1 +1,1 @@
-Briefcase will warn user if using an x86 Python interpreter on an Apple silicon machine
+Briefcase will now prevent the user from using an x86_64 Python interpreter on an Apple Silicon machine.
--- a/src/briefcase/platforms/macOS/__init__.py
+++ b/src/briefcase/platforms/macOS/__init__.py
@@ -107,7 +107,6 @@ class macOSMixin(_MixinBase):
             )
 
         super().verify_tools()
-        return
 
     def is_icloud_synced(self, path: Path) -> bool:
         """Determine if a path is on an iCloud drive.
--- a/tests/platforms/macOS/test_macOSMixin_verify_tools.py
+++ b/tests/platforms/macOS/test_macOSMixin_verify_tools.py
@@ -6,11 +6,8 @@ import pytest
 from briefcase.exceptions import BriefcaseCommandError
 
 
-# test the verify_tools method of the macOSMixin class,
-# to ensure that it raises the expected error when running on Apple Silicon
-# with an x86_64 Python interpreter
-def test_verify_tools_on_macos(dummy_command):
-    """If you're on macOS, you can verify tools."""
+def test_verify_macos_cpu_arch(dummy_command):
+    """Running through Rosetta on macOS will raise an error."""
     # Create a Mock object for the platform module
     dummy_command.tools.platform = MagicMock(spec_set=platform)
 
@@ -20,9 +17,11 @@ def test_verify_tools_on_macos(dummy_command):
 
     with pytest.raises(
         BriefcaseCommandError,
-        match=r"The Python interpreter that is being used to run Briefcase has been "
-        "compiled for x86_64, and is running in emulation mode on Apple "
-        "Silicon hardware. You must use a Python interpreter that has been "
-        "compiled for Apple Silicon, or is a Universal binary.",
+        match=(
+            r"The Python interpreter that is being used to run Briefcase has been "
+            r"compiled for x86_64, and is running in emulation mode on Apple "
+            r"Silicon hardware. You must use a Python interpreter that has been "
+            r"compiled for Apple Silicon, or is a Universal binary."
+        ),
     ):
         dummy_command.verify_tools()


src/briefcase/platforms/macOS/__init__.py
```diff
@@ -93,6 +94,21 @@ class macOSMixin(_MixinBase):
 
     def bundle_package_path(self, app) -> Path:
         return self.binary_path(app)
+    
+    # currently commenting out the check below because the "platform" class
+    # defined in the Python interpreter clashes with the platform variable in line 89
+    # of this file (platform variable defined right underneath the 
+    # "class maxOSMixin(_MixinBase)" declaration line
+    def verify_tools(self):
+        #if platform.machine() == "x86_64" and "ARM64" in platform.version():
```
vuinguyen: I currently have line 103 `if platform.machine() == "x86_64" and "ARM64" in platform.version():` commented out because it was giving me an error. Apparantly, the `platform` class is conflicting with the `platform` variable in line 89. And I can't change the `platform` variable without mediating the ripple effects that such a change could create.

kattni:
I asked Vui to create this PR without running pre-commit because of issues with her dev environment. That's why it has had the opportunity to fail. The attempt was made, but there were problems.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -93,6 +94,21 @@ class macOSMixin(_MixinBase):
 
     def bundle_package_path(self, app) -> Path:
         return self.binary_path(app)
+    
+    # currently commenting out the check below because the "platform" class
+    # defined in the Python interpreter clashes with the platform variable in line 89
+    # of this file (platform variable defined right underneath the 
+    # "class maxOSMixin(_MixinBase)" declaration line
+    def verify_tools(self):
```
vuinguyen, line 97: Where in `briefcase` do we want to call the function macOSMixin.verify_tools()?

src/briefcase/platforms/macOS/__init__.py
freakboy3742: Ah - you're in luck - we've had this problem before :-)

We abstract all the "important" stdlib modules in the `self.tools` module - so `self.tools.platform` will be the `platform` module from the standard library. 

This also means you can use `self.tools.platform = Mock()` in a test setup, rather than needing to do the module-level `monkeypatch` (or similar).

src/briefcase/platforms/macOS/__init__.py
freakboy3742: `verify_tools()` is part of the standard startup mechanism for a Briefcase command. It's part of the `BaseCommand` definition, and it's invoked automatically by every command when the project is finalized (see `finalize()` in `BaseCommand`)

So - there's no need to *explicitly* call `verify_tools()` - it will be called implicitly because it's mixed into a subclass of `BaseCommand`.  However, you *do* need to preserve the inheritance chain, by invoking `super().verify_tools()` (so that any other mixins or subclasses have their implementations invoked).

vuinguyen:
After pairing with @freakboy3742 and @kattni today (thanks so much for your help!), I now have a plan for how to move forward with this PR:

- **Problem 1 to fix: Remove all x86 Python stuff.** This will fix the original issue #2736. 
- **Problem 2 to fix: Have Briefcase dev environment point to Python 3.13, NOT Python 3.14.** This will get commits to this PR working. Commits work only with Python 3.13, NOT Python 3.14
- **Problem 3 to fix: Get all unit tests that have failed in this PR to pass.** This will help in getting this PR reviewed, approved, and eventually merged!

src/briefcase/platforms/macOS/__init__.py
vuinguyen: I made the requested changes.

src/briefcase/platforms/macOS/__init__.py
vuinguyen: Thanks for answering my question! I made the requested changes, and will select Resolve conversation.

vuinguyen: > `verify_tools()` is part of the standard startup mechanism for a Briefcase command. It's part of the `BaseCommand` definition, and it's invoked automatically by every command when the project is finalized (see `finalize()` in `BaseCommand`)
> 
> So - there's no need to _explicitly_ call `verify_tools()` - it will be called implicitly because it's mixed into a subclass of `BaseCommand`. However, you _do_ need to preserve the inheritance chain, by invoking `super().verify_tools()` (so that any other mixins or subclasses have their implementations invoked).

Although I added a call to `super().verify_tools()`, I'm still getting a lot of failed tests. It's possible that I don't have it in the right spot. But I'm in the middle of fixing my Python environment right now and I got feedback that only reviewers can Resolve conversation. So I will Unresolve conversation and circle back to this later.

vuinguyen:
I discovered while uninstalling Homebrew that I actually have a Python 2.7 version on my mac. It's the x86 version one and was installed in 2011. It looks like the version that came from Python.org, or maybe it came already installed on my old mac when I got it, not sure.

But I came across a couple of good sites on uninstalling Python on MacOS.
This one only covers uninstalling Python3 on macOS: 
https://www.askpython.com/python/examples/uninstall-python3-macos

This one covers uninstalling and reinstalling Python on macOS, through Homebrew and through Python.org:
https://coderivers.org/blog/how-to-uninstall-and-reinstall-python-mac/

So, still some more work to do, but I'm learning things and getting closer I think, to a clean development environment.

vuinguyen:
Here's another resource that I used for installing / uninstalling Python, that also describes how to remove it from Finder: https://iboysoft.com/howto/how-to-uninstall-python-on-mac.html

Status: To the best of my knowledge / ability, I've done the following:
 - uninstalled Homebrew, and uninstalled all versions of Python
 - re-installed the Apple Silicon version of Homebrew, which installed into the `/opt/homebrew` directory and added the `/opt/homebrew/bin` path to my `.zprofile`
 - installed `Python3.13` using Homebrew, which installed into the `/opt/homebrew/opt/python@3.13/libexec/bin` path
 - I added that last path into my `.zshrc`, so that now `python` and `python3` points to that path when I do a `which python` command. 
 - Also, doing a `python3 --version` and `python --version` both display `Python 3.13.12`
 - Running `platform.machine()` now gives me `arm64` and running `platform.version()` gives me `'Darwin Kernel Version 25.3.0: Wed Jan 28 20:56:34 PST 2026; root:xnu-12377.91.3~2/RELEASE_ARM64_T8122'`
 
I think this is a good place for me to stop for now, and will pick things back up on Monday.

vuinguyen:
Here's some updates. 

The good news: 

- I got the sample app from the [beeware tutorial](https://tutorial.beeware.org/tutorial/tutorial-5/iOS/) to finally launch from the iOS simulator by running `briefcase run iOS`. See the `attached file to see console output `and the screenshot of iOS simulator. I believe this means that my Python environment is now correct.

[4-17-2026-SuccessfulAppLaunch-iOSSimulator.txt](https://github.com/user-attachments/files/26554190/4-17-2026-SuccessfulAppLaunch-iOSSimulator.txt)

iPhone17-iOS26 4-smaller

- The unit test passes when I run the command:
`tox -e py -- tests/platforms/macOS/test_macOSMixin_verify_tools.py`

The bad news: 
- Some of the unit tests are failing, as you can see from the Github checks in this PR.
- I also encountered some failed tests from running `tox -e py`. See the attached `create.log file` and the `tox-e-py-Output` file (copy-pasted starting from the failed tests).

[briefcase.2026_04_07-16_48_01.create.log](https://github.com/user-attachments/files/26554212/briefcase.2026_04_07-16_48_01.create.log)

I have not made any _code_ changes to briefcase since my last commits, only recreated my virtual environment after I got my Python environment fixed (see comment above).

[4-7-2026-tox-e-py-Output.txt](https://github.com/user-attachments/files/26554214/4-7-2026-tox-e-py-Output.txt)

@freakboy3742 @kattni Any thoughts, suggestions on how to debug, proceed?

Kattni, I can also wait until our Wednesday team meeting to discuss this more with you! 

Thanks for any help!

freakboy3742:
> * I got the sample app from the [beeware tutorial](https://tutorial.beeware.org/tutorial/tutorial-5/iOS/) to finally launch from the iOS simulator by running `briefcase run iOS`. See the `attached file to see console output `and the screenshot of iOS simulator. I believe this means that my Python environment is now correct.

🎉 

> The bad news:
> 
> * Some of the unit tests are failing, as you can see from the Github checks in this PR.
> * I also encountered some failed tests from running `tox -e py`. 

Those logs seem to line up with what CI is generating for the macOS-latest/3.14 tests - so you're able to reproduce the problems that CI is generating (which is a good sign!)

I've taken a quick look, and the problem causing these test failures is *really* subtle (and is at least partially my fault for leading you astray). It turns out the way that the mock of `platform.machine()` and `platform.version()` is being created in a way that has a side effect *after* the test is created - you're actually attaching a mock to the *actual* platform module - so every test that is run *after* your new test thinks it is running on x86_64. 

You can fix this by adding:
```
dummy_command.tools.platform = MagicMock(spec_set=platform)
```
just before you set the mocks for `machine` and `version`. You'll also need to add an `import platform` at the top of the file. That makes the entire `tools.platform` reference a mock - but when the next test creates a new command, that mock will be overwritten. 

With that fix, the test suite should pass cleanly.

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -94,6 +94,21 @@ class macOSMixin(_MixinBase):
...
+        ):
+            raise BriefcaseCommandError(
+                "The Python interpreter that is being used to run Briefcase has been "
+                "compiled for x86_64, and is running in emulation mode on Apple "
+                "Silicon hardware. You must use a Python interpreter that has been "
+                "compiled for Apple Silicon, or is a Universal binary."
+            )
+
+        super().verify_tools()
+        return
```
freakboy3742: The return at the end of a function is implied; there's no need to provide it explicitly.
```suggestion
```

tests/platforms/macOS/test_macOSMixin_verify_tools.py
```diff
@@ -0,0 +1,28 @@
...
+    # Simulate that Mock platform is running on Apple Silicon with an x86_64 Python interpreter
+    dummy_command.tools.platform.machine = MagicMock(return_value="x86_64")
+    dummy_command.tools.platform.version = MagicMock(return_value="ARM64")
+
+    with pytest.raises(
+        BriefcaseCommandError,
+        match=r"The Python interpreter that is being used to run Briefcase has been "
+        "compiled for x86_64, and is running in emulation mode on Apple "
+        "Silicon hardware. You must use a Python interpreter that has been "
+        "compiled for Apple Silicon, or is a Universal binary.",
```
freakboy3742: While this is completely legal Python, it's visually ambiguous - it could easily be interpreted as "this raises() call has 5 arguments", because there are 5 lines at the same indentation level. As a style thing, if a string argument runs over multiple lines, and there are other arguments, we wrap that string in brackets and indent - that makes it more clear that there is only 1 argument. 
```suggestion
        match=(
            r"The Python interpreter that is being used to run Briefcase has been "
            r"compiled for x86_64, and is running in emulation mode on Apple "
            r"Silicon hardware. You must use a Python interpreter that has been "
            r"compiled for Apple Silicon, or is a Universal binary."
        ),
```

tests/platforms/macOS/test_macOSMixin_verify_tools.py
```diff
@@ -0,0 +1,28 @@
...
+import pytest
+
+from briefcase.exceptions import BriefcaseCommandError
+
+
+# test the verify_tools method of the macOSMixin class,
+# to ensure that it raises the expected error when running on Apple Silicon
+# with an x86_64 Python interpreter
+def test_verify_tools_on_macos(dummy_command):
+    """If you're on macOS, you can verify tools."""
```
freakboy3742: The explanatory comment should be the docstring of the test; it's also a good idea to make the name of the test reflect the specific purpose, in case we need to add a second macOS test later.
```suggestion
def test_verify_macos_cpu_arch(dummy_command):
    """Running through Rosetta on macOS will raise an error."""
```

changes/2736.bugfix.md
```diff
@@ -0,0 +1 @@
+Briefcase will warn user if using an x86 Python interpreter on an Apple silicon machine
```
freakboy3742:
```suggestion
Briefcase will now prevent the user from using an x86_64 Python interpreter on an Apple Silicon machine.
```

Issue #474, opened 2020-08-17
Add setuptools_scm support


Czaki:
In my project I use setuptools_scm for versioning (it calculate version from git tag and create proper file handling it). My project is distributed as python package and pyinstaler executable. I would like to switch from pyinstaller to briefcase (possibility of pip support) but I do not want to manually change version in file. 

**Describe the solution you'd like**
Add support to read version from setuptools_scm instead of manual update

**Describe alternatives you've considered**
Manual updating version. read version from file

freakboy3742:
Thanks for the suggestion. I agree that "version from SCM" is a use case worth considering.

There are a couple of complications to what you describe, though. Some packaging schemes have *very* strict version number requirements. Windows only allows strict numerical Triples. Android requires a build "sequence" number. IIRC, iOS only allows numbers. As a result, building an installer "for git hash xxxxx" isn't actually something you're going to be able to do in practice; even "for git tag xxxxx" will have some very serious constraints. We'll need to work out how to manage those constraints in any SCM-based version number solution.

There are also alternatives to `setuptools_scm` out there ([versioneer](https://github.com/warner/python-versioneer) being one that I am aware of); we'll need to make a decision whether to make an opinionated choice of one of those packages, or different version control systems (or even introduce a plugin system to support alternatives).

Czaki:
So maybe add option to provide version by function? 
Something like `get_version.py:get_version` Then user could provide simple wrapper function which provide version in proper format?

freakboy3742:
A user-provided function would be one way to provide compliant version numbers; however, that function would have to either (a) return a PEP508-compliant version number, which is then processed using the existing version number handling code, or (b) return a different version number for each platform.

What we really need at this point is a complete proposal for needs to change, and what the experience will be like as an end-user.

Czaki:
> A user-provided function would be one way to provide compliant version numbers; however, that function would have to either (a) return a PEP508-compliant version number, which is then processed using the existing version number handling code, or (b) return a different version number for each platform.

But is is enough to add description in documentation for this and fail if user function return version in wrong format.

mhsmith:
As mentioned in #1277, there is now a standard `dynamic` setting in pyproject.toml which may be useful for this.

andrewleech:
[...]

PR #2765, opened 2026-03-23
Translate TypeError from Version(version) into InvalidVersionError

akaIDIOT:
In trying to get briefcase to read my dynamic version, I noticed that setting `tool.briefcase.version` to something that would cause a `TypeError` during the coercion to a `Version` instance would trigger an except clause that assumes some configuration would be missing. This would cause briefcase to report a confusing error message:

> Briefcase configuration error: Global configuration is incomplete (missing )

By catching the `TypeError` close to coercing the configured value to a `Version`, the existing invalid version reporting triggers on an incompatible type too, showing up as a much more helpful output than "missing: nothing".

(no separate issue number)

## PR Checklist:

### DIFF
--- a/tests/config/test_GlobalConfig.py
+++ b/tests/config/test_GlobalConfig.py
@@ -135,7 +135,8 @@ def test_valid_app_version(input, expected):
 @pytest.mark.parametrize("input", INVALID_VERSIONS)
 def test_invalid_app_version(input):
     with pytest.raises(
-        BriefcaseConfigError, match=rf"Version number \({input}\) is not valid\."
+        BriefcaseConfigError,
+        match=rf"Version number \({input}\) is not valid\.",
     ):
         GlobalConfig(
             project_name="My Project",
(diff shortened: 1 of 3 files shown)


tests/config/test_GlobalConfig.py
```diff
@@ -53,6 +53,15 @@
...
     (Version("1.2.3"), "1.2.3"),  # Version object as input
 ]
+# Examples of invalid versions (input only)
+INVALID_VERSIONS = [
+    ("foobar",),
+    (42,),
+    ('{"file": "path/to/file"}',),
+    ({"file": "path/to/file"},),
+    ('{"attr": "myapp.module.__version__"}',),
+    ({"attr": "myapp.module.__version__"},),
```
akaIDIOT: These test values were taken from a [comment on #474](https://github.com/beeware/briefcase/issues/474#issuecomment-1616257388) and would also result in the confusing "missing" message mentioned in the PR body.

PR #2771, opened 2026-03-25
Introduce LinuxSystemAppConfig with 7 targeted casts.

filiplajszczak:
- `LinuxSystemAppConfig(FinalizedAppConfig)` in `linux/system.py` declares 7 platform attributes (`target_vendor`, `target_codename`, etc.).
- `finalize_app_config()` returns `LinuxSystemAppConfig` directly.
- 7 overrides that access platform attrs cast at the top of the method body.
- Platform-only methods take `LinuxSystemAppConfig` directly, no cast.

Reduces ty diagnostics from 366 ([on main](https://github.com/beeware/briefcase/actions/runs/23530011259/job/68491543155)) to  333 ([here](https://github.com/beeware/briefcase/actions/runs/23535318537/job/68508969774))

## PR Checklist:

### DIFF
--- a/src/briefcase/platforms/linux/system.py
+++ b/src/briefcase/platforms/linux/system.py
@@ -6,6 +6,7 @@ import subprocess
 import tarfile
 from collections.abc import Collection
 from pathlib import Path
+from typing import cast
 
 from briefcase.commands import (
     BuildCommand,
@@ -34,6 +35,21 @@ from briefcase.platforms.linux import (
 )
 
 
+class LinuxSystemAppConfig(FinalizedAppConfig):
+    """A FinalizedAppConfig with Linux system packaging attributes.
+
+    Set during ``finalize_app_config()`` on ``LinuxSystemMixin``.
+    """
+
+    target_vendor: str
+    target_codename: str
+    target_vendor_base: str
+    target_image: str
+    glibc_version: str
+    python_version_tag: str
+    packaging_format: str
+
+
 class LinuxSystemMixin(LinuxMixin):
     # The base mixin for system packages. It only supports native Linux usage, not usage
     # through Docker.
@@ -44,11 +60,13 @@ class LinuxSystemMixin(LinuxMixin):
     def build_path(self, app):
         # Override the default build path to use the vendor name,
         # rather than "linux"
+        app = cast(LinuxSystemAppConfig, app)
         return self.base_path / "build" / app.app_name / app.target_vendor
 
     def bundle_path(self, app):
         # Override the default bundle path to use the codename,
         # rather than "system"
+        app = cast(LinuxSystemAppConfig, app)
         return self.build_path(app) / app.target_codename
 
     def project_path(self, app):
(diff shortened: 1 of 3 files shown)


src/briefcase/platforms/linux/system.py
```diff
@@ -1032,6 +1055,7 @@ def _verify_packaging_tools(self, app: FinalizedAppConfig):
                 )
 
     def verify_app_tools(self, app):
```
```suggestion
    def verify_app_tools(self, app: FinalizedAppConfig):
```

src/briefcase/platforms/linux/system.py
```diff
@@ -175,7 +194,9 @@ def platform_freedesktop_info(self, app: AppConfig):
     def _finalize_target_image(self, app: AppConfig):
         app.target_image = f"{app.target_vendor}:{app.target_codename}"
 
-    def finalize_app_config(self, app: DraftAppConfig, **kwargs) -> FinalizedAppConfig:
+    def finalize_app_config(
+        self, app: DraftAppConfig, **kwargs
+    ) -> LinuxSystemAppConfig:
```
freakboy3742, line 201:
```suggestion
    def finalize_app_config(
        self,
        app: DraftAppConfig,
        **kwargs,
    ) -> LinuxSystemAppConfig:
```

src/briefcase/platforms/linux/system.py
```diff
@@ -241,7 +262,7 @@ def finalize_app_config(self, app: DraftAppConfig, **kwargs) -> FinalizedAppConf
 
         self.console.verbose(f"Targeting Python{app.python_version_tag}")
 
-        return super().finalize_app_config(app, **kwargs)
+        return LinuxSystemAppConfig(app, **kwargs)
```
freakboy3742: This doesn't seem right - it's assuming that the base class doesn't do any per-app finalisation. That's certainly true right now, but there's no reason to believe it will stay that way.

The most direct mapping here seems like it would be:
```suggestion
        return LinuxSystemAppConfig(super().finalize_app_config(app, **kwargs))
```

...but that doesn't feel right from a strict type check perspective. I guess it's falling into the same sort of  typing gap as we're exploiting with `**kwargs` handling, though?

filiplajszczak: Good point. What about widening `FinalizedAppConfig.__init__` to accept `AppConfig` instead of `DraftAppConfig`, then:

```python
return LinuxSystemAppConfig(super().finalize_app_config(app, **kwargs))
```

The narrowing to `DraftAppConfig` was meant to encode the lifecycle, but `finalize()` already prevents double finalization with `isinstance(app, FinalizedAppConfig)` — so the real guard is in runtime, not in type annotations.

freakboy3742: Yeah - I guess that makes sense. 

It could actually be *better*, as AppConfig has tighter type definitions for the base properties than DraftAppConfig - finalising a config that has (for example) a `license_files = None` definition would result in a FinalizedAppConfig that has a value that doesn't match the allowed type declaration, but is missed because of the `__dict__.update(**kwargs)` handling.

filiplajszczak: Committed, squashed and pushed.

mhsmith:
This may have broken test mode: see https://github.com/beeware/toga/issues/4278.

PR #2772, opened 2026-03-25
Support dynamic PEP621 metadata

akaIDIOT:
For projects that declare dynamic metadata (like automatically determining a version from source control or the current date), briefcase would complain that certain metadata fields (such as `version` in the example) were missing and refuse to continue. This pull requests adds the ability to resolve these dynamic fields using the build backend that the project would use to resolve these fields during the creation of a distribution. I think it should only trigger in these cases; dynamic metadata fields always require some configuration or hook to be able to resolve these fields at build time. Projects that solely use briefcase and no kind of build backend shouldn't see any negative effects from this.

The current documentation has a section on [PEP621-compatibility](https://briefcase.beeware.org/en/stable/reference/configuration/?h=configuration#pep621-compatibility), my intuition was to add something about what happens when some of these values are declared as dynamic there; would that be 

1. the right place?
2. sufficient?

Another point of contention is the translation 'table' in `briefcase.config._core_metadata_to_pep621`. I was hoping to find something along those lines in `packaging`, but that seems to have different use cases in mind. The fields that are included in that function are inspired by what `briefcase.config.merge_pep621_config` goes looking for (with the exception of `"license-files"`, which seems to not show up in the resulting metadata). The added unit test that's using `hatchling` has a silly amount of dynamic fields on purpose, showing the translation at work. If I've missed an existing implementation in a dependency for this, I'd be happy to take it back out, of course. If this way of going about it doesn't fit with briefcase, I'm open to suggestions.

References #474.

## PR Checklist:

### DIFF
--- a/src/briefcase/config.py
+++ b/src/briefcase/config.py
@@ -10,11 +10,10 @@ from email.utils import getaddresses
 from pathlib import Path
 from urllib.parse import urlparse
 
+from build import BuildBackendException
 from build.util import project_wheel_metadata
 from packaging.licenses import InvalidLicenseExpression, canonicalize_license_expression
 from packaging.version import InvalidVersion, Version
-
-from build import BuildBackendException
 
 if sys.version_info >= (3, 11):  # pragma: no-cover-if-lt-py311
     import tomllib
--- a/tests/config/test_parse_config.py
+++ b/tests/config/test_parse_config.py
@@ -3,11 +3,11 @@ from email.message import Message
 from unittest.mock import Mock
 
 import pytest
+from build import BuildBackendException
 
 from briefcase.config import parse_config
 from briefcase.console import Console
 from briefcase.exceptions import BriefcaseConfigError
-from build import BuildBackendException
 from tests.utils import create_file
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -291,11 +291,15 @@ ignore = [
     "SIM108",  # replace if-else with ternary operator (too opinionated)
 ]
 
+[tool.ruff.lint.isort]
+# If you've built a wheel locally, you'll have a build directory.
+# This confuses ruff into thinking build is a first-party Python package.
+known-third-party = ["build"]
+
 [tool.ruff.lint.per-file-ignores]
 # E501: line too long, to be fixed in future changes
 "tests/integrations/*" = ["E501"]
 "tests/platforms/*" = ["E501"]
-
 
 [tool.rumdl]
 flavor = "mkdocs"


freakboy3742 (changes requested):
Thanks for the PR!

I haven't done a *complete* review yet, but my initial impression is that this looks great. The implementation make sense; and while I'll probably tweak the documentation in the final version, what you've got here is (a) definitely in the right place, and (b) close enough for a first draft that I don't have any particular notes.

The only note I'd make at this point is about testing. The tests you've got here definitely exercise all the functionality - but they do so by live testing the actual build API, which will be slow, and will require network access to complete. This is a case where it would be preferable to mock the expected results from the `build` API, rather than actually invoke the build API to get those results.

akaIDIOT:
Thanks for the feedback, you're definitely right that those tests are a chunk slower than most, they take about a second each for me locally. I'll try to tweak them to run on mocks without actually calling build, hopefully in a way that makes it simple to run them for real to check they're still working without the mock 👍🏻 

I'll leave the documentation edits alone for now, I was expecting them to get tweaked or replaced anyways :)

akaIDIOT:
**Edit:** overtaken by subsequent changes

----

All the tests that would run the build backend for real are now parametrized into two parts:

- `id='real-build'` that will actually run the build backend;
- `id='mocked-build'` that mocks `project_wheel_metadata` to return the values that were already expected by the test.

The mocked variants take milliseconds, so that should remedy the speed concern. The `real-build` variant is currently always skipped, which might make it a bit useless. Running this for real could be more of an integration test than a unit test, so maybe it's better to remove the parametrization and just run the mocks (this would also reduce the diff a fair bit, I think). These *would* test the code that's been added to Briefcase, but might make it tough to generate the expected values in the future, as those don't need the `pyproject.toml` files that showcase how this would work for various build backends. Selecting the expensive tests with a pytest marker or something along those lines would make sense too; I'm open to suggestions there.

akaIDIOT:
On second thought, let's not go to integration test, 'tis a silly place.

The created files next to `pyproject.toml` in the happy-flow test that's using dynamic metadata (`test_pep621_dynamic_hatchling`) aren't strictly needed, they would be required to make the build work for hatch, but as the result gets patched, the test doesn't need the files to be there.

src/briefcase/config.py
```diff
@@ -1210,6 +1213,48 @@ def normalize_license_config(
...
+def resolve_dynamic_pep621_config(base_path, dynamic, console):
+    """Resolve dynamic PEP621 metadata using the project's configured build backend."""
+
+    try:
+        metadata = project_wheel_metadata(base_path, isolated=True)
+        # Provide fields declared as dynamic with corresponding metadata value
+        return {field: _core_metadata_to_pep621(field, metadata) for field in dynamic}
+    except BuildBackendException as e:
+        dynamic_fields = "    \n".join(dynamic)
+        console.warning(
```
freakboy3742: As this method will only be invoked if there is dynamic metadata, we can raise a hard error here.

tests/config/test_parse_config.py
```diff
@@ -1403,3 +1406,187 @@ def test_no_license_key(tmp_path):
...
+                    metadata["description"] = "The application is very awesome"
+                    metadata["license"] = "GPL-3.0"
+                    metadata["urls"] = {
+                        "Docs": "https://example.com/docs",
+                        "Homepage": "https://example.com/",
+                    }
+                    metadata["version"] = "1.2.3"
+            """
+        ),
+    )
```
freakboy3742: These files are no longer needed for testing

tests/config/test_parse_config.py
```diff
@@ -1403,3 +1406,187 @@ def test_no_license_key(tmp_path):
...
+    metadata["Name"] = "awesome"
+    metadata["Version"] = "1.2.3"
+    metadata["Summary"] = "The application is very awesome"
+    metadata["Project-URL"] = "Docs, https://example.com/docs"
+    metadata["Project-URL"] = "Homepage, https://example.com/"
+    metadata["Author-email"] = "Kim Park <kim@example.com>, John Doe <john@example.org>"
+    metadata["License-Expression"] = "GPL-3.0"
+    metadata["Requires-Dist"] = "toga>=0.5.3"
+
+    with patch("briefcase.config.project_wheel_metadata", return_value=metadata):
```
freakboy3742: For consistency - we use `monkeypatch` rather than `unittest.patch`. 

We can also use the mock to verify that the method *was* invoked with the expected parameters (or not invoked, as required).

tests/config/test_parse_config.py
```diff
@@ -1403,3 +1406,187 @@ def test_no_license_key(tmp_path):
             output_format="Xcode",
             console=Mock(),
         )
+
+
+def test_pep621_empty_dynamic(tmp_path):
```
freakboy3742: We try to include a docstring with every test explaining the purpose of the test.

tests/config/test_parse_config.py
```diff
@@ -1403,3 +1406,187 @@ def test_no_license_key(tmp_path):
...
+        [tool.briefcase.app.awesome]
+        formal_name = "Awesome Application"
+        """,
+    )
+
+    console = Mock()
+    with (
+        patch(
+            "briefcase.config.project_wheel_metadata",
+            side_effect=BuildBackendException(
```
freakboy3742: I found a second failure mode here - if the `requires` list is invalid, a `subprocess.CalledProcessError` is raised.

changes/2772.feature.md
```diff
@@ -0,0 +1 @@
+Dynamic PEP621 metadata in ``pyproject.toml`` will now be resolved by invoking the `[build-system]`.
```
freakboy3742:
```suggestion
Dynamic PEP 621 metadata in `pyproject.toml` will now be resolved by invoking the PEP 517 `[build-system]` interface.
```

Also - the filename should reference the original feature request (474), not the PR.

src/briefcase/config.py
```diff
@@ -1210,6 +1213,48 @@ def normalize_license_config(
...
+        case _:
+            # Let metadata's automatic normalization deal with selecting the right key
+            return metadata[pep621_key]
+
+
+def resolve_dynamic_pep621_config(base_path, dynamic, console):
+    """Resolve dynamic PEP621 metadata using the project's configured build backend."""
+
+    try:
+        metadata = project_wheel_metadata(base_path, isolated=True)
```
freakboy3742: This call might take a couple of seconds; it would be worth wrapping this in a `wait_bar`.

freakboy3742 (approved):
As promised, I've tweaked the docs; I've also done a bit of a minor cleanup of some of the tests cases (mostly to adhere to nebulous style consistency issues), and I caught an additional edge case in my own testing (the build backend returning subprocess.CalledProcessError). 

Otherwise, this is great - and resolves a long requested feature! Thanks for the PR!

PR #2788, opened 2026-04-07
Allow for creating a "skinless" android emulator

freakboy3742:
We've recently seen a lot of rate limiting (HTTP 429) when downloading a skin for [Android CI builds](https://github.com/beeware/toga/actions/runs/24030637432/job/70078652391). Since a skin is purely cosmetic, it makes sense to provide an option to build a "skinless" build option.

This modifies the Android emulator creation process so that if you specify a device type, but no skin, a skinless device is created. (i.e., `briefcase run -d '{"device_type":"pixel"}'`). This will avoid the attempt to download the skin.

This only impacts the "explicit device syntax" format; a skin is still selected when you go through the "create new device" menu option.

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1175,13 +1175,15 @@ In future, you can specify this device by running:
         """
         if device_type is None:
             device_type = self.DEFAULT_DEVICE_TYPE
-        if skin is None:
             skin = self.DEFAULT_DEVICE_SKIN
         if system_image is None:
             system_image = self.DEFAULT_SYSTEM_IMAGE
 
         # Ensure the required skin is available.
-        self.verify_emulator_skin(skin)
+        if skin:
+            self.verify_emulator_skin(skin)
+        else:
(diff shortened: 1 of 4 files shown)


rmartin16 (approved):
running through different permutations locally working  as expected 👍🏼

src/briefcase/integrations/android_sdk.py
```diff
@@ -1175,13 +1175,15 @@ def _create_emulator(
         """
         if device_type is None:
             device_type = self.DEFAULT_DEVICE_TYPE
-        if skin is None:
             skin = self.DEFAULT_DEVICE_SKIN
```
rmartin16, line 1178: just noticed this; this means that both `device_type` and `skin` must be specified by the user for `skin` to have any effect.

Previously, a command like this respected `skin` and defaulted the device type:
```console
briefcase run android -d '{"avd":"asdf", "skin": "pixel_3a"}'
```

freakboy3742: That is correct, and intentional; the documentation clarifies that usage.

PR #2797, opened 2026-04-13
Support Windows ARM64

### LINKED ISSUE #1887: Add support for Windows on ARM

Briefcase can currently only target Windows on x86. With the growing support for Windows on ARM, Briefcase should support targeting ARM.

The current stub should be built for ARM and the Visual Studio template should support targeting ARM.

The embedded Python should be the arm64 version. Although, this is only available back to Python 3.11.

Similar to macOS on Apple Silicon, Windows can run x86 binaries on ARM. I still need to actually try it but this may allow a Briefcase app built for x86 to run on ARM.

### Additional context

_No response_

abdnh:
This adds support for Windows ARM64.

- See related PR for the Visual Studio template: https://github.com/beeware/briefcase-windows-VisualStudio-template/pull/87
- Closes #1887
- More context: https://github.com/ankitects/anki/issues/4678

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,7 +235,7 @@ jobs:
       fail-fast: false
       matrix:
         framework: [ "toga", "pyside6", "pygame", "console" ]
-        runner-os: [ "macos-15-intel", "macos-latest", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest" ]
+        runner-os: [ "macos-15-intel", "macos-latest", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest", "windows-11-arm"]
 
   verify-apps:
     name: Build app
@@ -250,4 +250,4 @@ jobs:
       fail-fast: false
       matrix:
         framework: [ "toga", "pyside6", "pygame", "console" ]
-        runner-os: [ "macos-15-intel", "macos-15", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest" ]
+        runner-os: [ "macos-15-intel", "macos-15", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest", "windows-11-arm" ]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,3 +251,10 @@ jobs:
       matrix:
         framework: [ "toga", "pyside6", "pygame", "console" ]
         runner-os: [ "macos-15-intel", "macos-15", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest", "windows-11-arm" ]
+        exclude:
+          # Toga apps won't run on Windows ARM64 until toga#2782 is resolved
+          - framework: "toga"
+            runner-os: "windows-11-arm"
+          # Pygame hasn't published Windows ARM64 wheels
+          - framework: "pygame"
+            runner-os: "windows-11-arm"
--- a/changes/1887.feature.md
+++ b/changes/1887.feature.md
@@ -1,1 +1,1 @@
-Windows on ARM64 is now supported.
+Briefcase can now build Windows apps for ARM64 devices.
--- a/changes/1887.feature.md
+++ b/changes/1887.feature.md
@@ -1,0 +1,1 @@
+Windows on ARM64 is now supported.
--- a/docs/en/reference/platforms/windows/visualstudio.md
+++ b/docs/en/reference/platforms/windows/visualstudio.md
@@ -60,7 +60,7 @@ All Windows apps, regardless of output format, use the same icon formats, have t
 
 ## Pre-requisites
 
-Building the Visual Studio project requires that you install Visual Studio 2022 or later. Visual Studio 2022 Community Edition [can be downloaded for free from Microsoft](https://visualstudio.microsoft.com/vs/community/). You can also use the Professional or Enterprise versions if you have them.
+Building the Visual Studio project requires that you install Visual Studio. Visual Studio Community Edition [can be downloaded for free from Microsoft](https://visualstudio.microsoft.com/vs/community/). You can also use the Professional or Enterprise versions if you have them.
 
 Briefcase will auto-detect the location of your Visual Studio installation, provided one of the following three things are true:
 
@@ -75,6 +75,7 @@ When you install Visual Studio, there are many optional components. You should e
 - Desktop Development with C++
   - All default packages
   - C++/CLI support for v143 build tools
+  - MSVS v143 VS 2022 C++ ARM64/x64 build tools
 
 ## Application configuration
 
--- a/docs/en/reference/platforms/windows/visualstudio.md
+++ b/docs/en/reference/platforms/windows/visualstudio.md
@@ -75,7 +75,7 @@ When you install Visual Studio, there are many optional components. You should e
 - Desktop Development with C++
   - All default packages
   - C++/CLI support for v143 build tools
-  - MSVS v143 VS 2022 C++ ARM64/x64 build tools
+  - MSVC v143 VS 2022 C++ ARM64/x64 build tools
 
 ## Application configuration
 
--- a/src/briefcase/commands/create.py
+++ b/src/briefcase/commands/create.py
@@ -123,7 +123,15 @@ class CreateCommand(BaseCommand):
     def stub_binary_filename(self, support_revision: str, is_console_app: bool) -> str:
         """The filename for the stub binary."""
         stub_type = "Console" if is_console_app else "GUI"
-        return f"{stub_type}-Stub-{self.python_version_tag}-b{support_revision}.zip"
+        win_arm_suffix = (
+            "-arm"
+            if self.tools.host_os == "Windows" and self.tools.host_arch == "ARM64"
+            else ""
+        )
+        return (
+            f"{stub_type}-Stub-{self.python_version_tag}-b{support_revision}"
+            f"{win_arm_suffix}.zip"
+        )
 
     def stub_binary_url(self, support_revision: str, is_console_app: bool) -> str:
         """The URL of the stub binary to use for apps of this type."""
--- a/src/briefcase/commands/create.py
+++ b/src/briefcase/commands/create.py
@@ -129,8 +129,8 @@ class CreateCommand(BaseCommand):
             else ""
         )
         return (
-            f"{stub_type}-Stub-{self.python_version_tag}-b{support_revision}"
-            f"{win_suffix}.zip"
+            f"{stub_type}-Stub-{self.python_version_tag}{win_suffix}"
+            f"-b{support_revision}.zip"
         )
 
     def stub_binary_url(self, support_revision: str, is_console_app: bool) -> str:
--- a/src/briefcase/commands/create.py
+++ b/src/briefcase/commands/create.py
@@ -123,14 +123,14 @@ class CreateCommand(BaseCommand):
     def stub_binary_filename(self, support_revision: str, is_console_app: bool) -> str:
         """The filename for the stub binary."""
         stub_type = "Console" if is_console_app else "GUI"
-        win_arm_suffix = (
-            "-arm"
-            if self.tools.host_os == "Windows" and self.tools.host_arch == "ARM64"
+        win_suffix = (
+            f"-{self.tools.host_arch.lower()}"
+            if self.tools.host_os == "Windows"
             else ""
         )
         return (
             f"{stub_type}-Stub-{self.python_version_tag}-b{support_revision}"
-            f"{win_arm_suffix}.zip"
+            f"{win_suffix}.zip"
         )
 
     def stub_binary_url(self, support_revision: str, is_console_app: bool) -> str:
--- a/src/briefcase/integrations/base.py
+++ b/src/briefcase/integrations/base.py
@@ -195,7 +195,7 @@ class ToolCache(Mapping):
         self.base_path = Path(base_path)
         self.home_path = Path(os.path.expanduser(home_path or Path.home()))
 
-        self.host_arch = self.platform.machine()
+        self.host_arch = self._get_host_arch()
         self.host_os = self.platform.system()
         # Python is 32bit if its pointers can only address with 32 bits or fewer
         self.is_32bit_python = self.sys.maxsize <= 2**32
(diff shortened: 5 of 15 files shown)


abdnh:
(Windows tests failing because the stub binary URL has changed to include a platform suffix)

src/briefcase/platforms/windows/__init__.py
```diff
@@ -88,24 +88,24 @@ def distribution_path(self, app):
...
+    def architecture(self):
+        return "ARM64" if self.tools.host_arch == "ARM64" else "x64"
+
     def verify_host(self):
         super().verify_host()
-        # The stub app only supports x86-64 right now, and our VisualStudio and WiX code
-        # is the same (#1887). However, we can package an external x86-64 app on any
-        # build machine.
-        if self.tools.host_arch != "AMD64":
+        if self.tools.host_arch not in ("AMD64", "ARM64"):
```
freakboy3742, line 108: We recently found a class of errors on macOS where using an x86-64 interpreter on an ARM machine caused issues (see #2736); we should add a check similar to #2760 as a safety mechanism.

src/briefcase/platforms/windows/__init__.py
```diff
@@ -88,24 +88,24 @@ def distribution_path(self, app):
         suffix = "zip" if app.packaging_format == "zip" else "msi"
         return self.dist_path / f"{app.formal_name}-{app.version}.{suffix}"
 
+    def architecture(self):
```
freakboy3742: The naming for this one should (a) have consistent naming with the VSCode template, and (b) reflect that it is a VSCode (or at least *windows* specific) architecture name. Otherwise it won't be clear when to use this `architecture` vs `tools.host_arch`.

freakboy3742: Also - is there any reason this can't be a property, rather than a method?

src/briefcase/platforms/windows/app.py
```diff
@@ -23,9 +23,11 @@
 
 class WindowsAppMixin(WindowsMixin):
     output_format = "app"
-    packaging_root = Path("src")
     supports_external_packaging = True
 
+    def packaging_root(self):
```
freakboy3742, line 29: Is there any reason this needs to be a method, rather than a property?

freakboy3742 (changes requested):
Broadly, this looks great - a couple of minor details flagged inline, mostly around naming (similar naming issues to the template), and property vs method use.

src/briefcase/platforms/windows/__init__.py
abdnh: That's probably especially important for uv users as uv defaults to x86-64 interpreters: https://github.com/astral-sh/uv/issues/12906

freakboy3742: I worked out how this will bite us - A Toga app built using this PR fails if you use py install -3.13 because the py installer on Windows defaults to the AMD64 build; which leads to win-amd64 wheels being installed into the app.

freakboy3742:
I've done some initial testing of this, in conjunction with the updated Visual Studio template.

I got a VSStudio project to *build*, but I needed to:
* update the Visual Studio project to use the latest build tools (v145, rather than v143). That's probably a good idea regardless, so that a fresh install of Visual Studio 2026 is easier. 
* Re-install Python to make sure that I had an ARM64 build, rather than an AMD64 build. `py install -3.13` installs a AMD64 build by default, for the same reason that uv does.

Even with that, I get an error launching a Toga app because it can't load `_cffi_backend`. I haven't gotten to the bottom of this; it might be a Python.net issue.

abdnh:
Regarding build tools, I've not updated the version because the GitHub runner still uses Visual Studio 2022. This will apparently change [next month](https://github.blog/changelog/2026-02-05-github-actions-early-february-2026-updates/#windows-server-2025-with-visual-studio-2026-image-now-available-for-github-hosted-runners), but I'm not sure if the ARM runner will be updated at the same time: https://github.com/actions/partner-runner-images/issues/150

abdnh:
> Even with that, I get an error launching a Toga app because it can't load _cffi_backend. I haven't gotten to the bottom of this; it might be a Python.net issue.

Yes, looks like a Python.net issue. I had a look and was able to modify [pythonnet/clr-loader](https://github.com/pythonnet/clr-loader) to build ARM64 DLLs but the app then failed with this:
> AttributeError: function/symbol 'pyclr_initialize' not found in library '\app_packages\clr_loader\ffi\dlls\arm64\ClrLoader.dll': error 0x7f. Did you mean: 'pyclr_finalize'?

No luck figuring it out so far. Let me know if you think we should add an error if the Toga bootstrap is used for now.

freakboy3742:
I've found the source of the problem; the issue is that .NET Framework 4.x doesn't provide ARM64 builds, even though it's installed by default on Windows on ARM machine. You have to use .NET Core 8 (or newer) on an ARM64 machine, or use an x86-64 interpreter running in emulation mode. 

See https://github.com/beeware/toga/issues/2782#issuecomment-4257239304 for details; I've just pushed beeware/toga#4331 that adds ARM64 support to Toga's Winforms backend.

src/briefcase/platforms/windows/__init__.py
```diff
@@ -88,24 +88,36 @@ def distribution_path(self, app):
...
-        # is the same (#1887). However, we can package an external x86-64 app on any
-        # build machine.
-        if self.tools.host_arch != "AMD64":
+        if (
+            self.tools.host_arch == "ARM64"
+            and "AMD64" in self.tools.platform.python_compiler()
+        ):
+            raise UnsupportedHostError(
+                "The Python interpreter that is being used to run Briefcase has been "
+                "compiled for x86_64, and is running in emulation mode on ARM64"
```
freakboy3742:
```suggestion
                "compiled for x86_64, and is running in emulation mode on ARM64 "
```

changes/1887.feature.md
```diff
@@ -0,0 +1 @@
+Windows on ARM64 is now supported.
```
freakboy3742:
```suggestion
Briefcase can now build Windows apps for ARM64 devices.
```

src/briefcase/platforms/windows/__init__.py
```diff
@@ -88,24 +88,36 @@ def distribution_path(self, app):
...
+        return "ARM64" if self.tools.host_arch == "ARM64" else "x64"
+
     def verify_host(self):
         super().verify_host()
-        # The stub app only supports x86-64 right now, and our VisualStudio and WiX code
-        # is the same (#1887). However, we can package an external x86-64 app on any
-        # build machine.
-        if self.tools.host_arch != "AMD64":
+        if (
+            self.tools.host_arch == "ARM64"
```
freakboy3742, line 98: There's an edge case here that I (re)-discovered in testing: if you're on Python 3.10 or 3.11, `platform.machine()` returns the architecture of the executing binary, not the underlying machine architecture.

Since `host_arch` underpins a lot of this logic (and `host_arch` is populated by `platform.machine()`), we will need to add some extra logic to "correct" host_arch on Py < 3.12 on Windows.

freakboy3742 (changes requested):
Thanks for those updates. I've flagged some cosmetic details, plus one edge case with Python 3.10 and 3.11 that we'll need to handle.

We should also update the VSCode install instructions to remove the reference to VSCode 2022 (since 2026 is what you can download now), and include a reference to needing MSVS v143 VS 2022 C++ ARM64/x64 build tools.

The only other question left is exactly how to land the PR with the other dependencies. I've merged the Visual Studio template, but I can't automate tagging new binaries until we merge this PR; but this PR can't be merged until CI passes, and that won't happen until there's a Toga version that has ARM support. Obviously we'll need to test some pieces manually and add automated CI checks afterwards to break the stalemate; but I need to work out what sequence will be best for that.

src/briefcase/platforms/windows/__init__.py
abdnh: OK, I found a workaround.

src/briefcase/integrations/base.py
```diff
@@ -208,6 +208,36 @@ def __init__(
...
+            native_machine = ctypes.c_ushort(0)
+            if (
+                kernel32.IsWow64Process2(
+                    wintypes.HANDLE(kernel32.GetCurrentProcess()),
+                    ctypes.byref(process_machine),
+                    ctypes.byref(native_machine),
+                )
+                and native_machine.value == IMAGE_FILE_MACHINE_ARM64
+            ):
+                return "ARM64"
```
freakboy3742, line 238: This is clever - but I've found an even easier way that doesn't need system calls:
```suggestion
        ):
            return os.getenv("PROCESSOR_ARCHITECTURE", arch)
```

AFAICT, `PROCESSOR_ARCHITECTURE` is defined on all Win10 and 11 machines, and returns ARM64/AMD64 as appropriate.

docs/en/reference/platforms/windows/visualstudio.md
```diff
@@ -75,6 +75,7 @@ When you install Visual Studio, there are many optional components. You should e
 - Desktop Development with C++
   - All default packages
   - C++/CLI support for v143 build tools
+  - MSVC v143 VS 2022 C++ ARM64/x64 build tools
```
freakboy3742, line 78: There's also a mention of this in `integrations/visualstudio.py`, as part of the error message that is displayed if VSCode can't be found.

freakboy3742 (changes requested):
Thanks for those updates - two more small tweaks, and I think we're good to go.

src/briefcase/integrations/base.py
abdnh: Hmm, I tested that with an x86_64 interpreter (using `uv run --python cpython-3.11.15-windows-x86_64-none python`) and got `AMD64`, and it looks like this is what platform.machine() uses under the hood (at least in recent versions): https://github.com/python/cpython/blob/afde75664eb3ff3e147806f027c9da54c7eb77d4/Lib/platform.py#L816

freakboy3742: That's weird - the behavior shouldn't be impacted by the version of Python that runs it - it's an environment variable. Every Windows machine I have access to defines that variable (both x86 and ARM64) - does your machine set the `PROCESSOR_ARCHITECTURE` environment variable (or any similar variable)?

abdnh: It's not impacted by the Python version, but it does seem to return the architecture of the binary rather than the actual host machine: https://linaro.atlassian.net/wiki/spaces/WOAR/pages/28869623809/Identify+CPU+architecture

Yes, I have `PROCESSOR_ARCHITECTURE` set to `ARM64` in cmd/PowerShell.

I'm testing with a Windows 11 25H2 VM running under a Mac M4 chip by the way.

freakboy3742: Ah - I thought it was a property of the general runtime environment. I'm guessing the value that you see at a command shell is the value generated by the shell executable itself, which is why it looks like it reflects the native platform architecture. I guess the more complex "binary introspection" approach will be needed, then.

src/briefcase/commands/create.py
```diff
@@ -123,7 +123,15 @@ def support_package_url(self, support_revision: str) -> str:
...
         stub_type = "Console" if is_console_app else "GUI"
-        return f"{stub_type}-Stub-{self.python_version_tag}-b{support_revision}.zip"
+        win_suffix = (
+            f"-{self.tools.host_arch.lower()}"
+            if self.tools.host_os == "Windows"
+            else ""
+        )
+        return (
+            f"{stub_type}-Stub-{self.python_version_tag}-b{support_revision}"
+            f"{win_suffix}.zip"
```
freakboy3742, line 133:
```suggestion
            f"{stub_type}-Stub-{self.python_version_tag}{win_suffix}"
            f"-b{support_revision}.zip"
```

freakboy3742 (changes requested):
One last tweak to the archive name for the stub binary - this became apparent when I manually generated the stub binary so these tests would pass.

freakboy3742:
I've manually published stub binaries with the new names (using both your prefixing scheme, and the one that I've suggested inline), and re-run CI; and the ARM jobs are all passing, except for:
* pygame (pygame hasn't published Windows ARM wheels)
* Toga (that won't work until beeware/toga#4331 is in a published release).

We can put those two on the skip list for now.

.github/workflows/ci.yml
```diff
@@ -250,4 +250,4 @@ jobs:
       fail-fast: false
       matrix:
         framework: [ "toga", "pyside6", "pygame", "console" ]
-        runner-os: [ "macos-15-intel", "macos-15", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest" ]
+        runner-os: [ "macos-15-intel", "macos-15", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest", "windows-11-arm" ]
```
freakboy3742, line 253:
```suggestion
        runner-os: [ "macos-15-intel", "macos-15", "ubuntu-24.04", "ubuntu-24.04-arm", "windows-latest", "windows-11-arm" ]
        exclude:
          # Toga apps won't run on Windows ARM64 until toga#2782 is resolved
          - framework: "toga"
            runner-os: "windows-11-arm"
          # Pygame hasn't published Windows ARM64 wheels
          - framework: "pygame"
            runner-os: "windows-11-arm"
```

abdnh:
docs-lint apparently failed due to a temporary issue with an apple.com link.

freakboy3742 (approved):
I think we can call this done. I'm still working on Toga, and there's a couple of tweaks needed to the Visual Studio template before I can tag "official" binaries (including a minor walk-back of the stdout re-opening - see [this pr](https://github.com/beeware/briefcase-windows-VisualStudio-template/pull/89) for details) - but this does everything the Briefcase part needs to do. Thanks for the PR!

HalfWhitt:
It's great that we support ARM64 natively now, but do we really need to specifically *exclude* supporting x86_64 Python that's running in emulation mode? I just updated Briefcase in my Windows test VM and suddenly it's mad at me.

freakboy3742:
Do we *have* to? Possibly not. Is it significantly safer to? Yes. 

The problem with emulation mode is that it's really easy to end up on a situation where you *think* you're producing an ARM binary, but you're actually producing an x86-64 binary (especially given the incorrect "what is my platform" answers that Python will give on ARM64 prior to 3.13). The "only native" approach is the easiest way to guarantee platform confusion doesn't happen.

Is it the *only* way? Probably not. Patches welcome :-) 

I'd probably even consider a "yeah, I know what I'm doing" setting to disable that check if you can demonstrate that it works well enough for testing purposes.

HalfWhitt:
Ah, that makes sense. For my own needs, there's no reason I'd *prefer* to emulate, it's just what I happened to already have installed... which only reinforces your point, honestly. I hadn't even realized what architecture it was.

PR #2806, opened 2026-04-23
Include return codes in run failure messages.

freakboy3742:
While trying to diagnose intermittent Windows testing failures, I wasn't able to get any insight into why the run was failing, because the return code of the executed process  wasn't being reported.

This PR adds the return code to run failure messages.

## PR Checklist:

### DIFF
--- a/changes/2806.feature.md
+++ b/changes/2806.feature.md
@@ -1,0 +1,1 @@
+When an app fails to run or returns an error, the status code is now reported in the console.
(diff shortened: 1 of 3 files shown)


mhsmith (approved):

mhsmith:
["Return code 1"](https://github.com/beeware/briefcase/actions/runs/24887303893/job/72871021734) – so I guess that doesn't tell us much, but it might rule out some options.

freakboy3742:
Yeah - two theories I had that are now ruled out:

1. It's actually a failure of the log stream, not the app
2. `poll()` is retuning None for some reason.

It also occurs to me that with the recent console attachment changes, it *might* be possible to move back to a "direct stdout" rather than a log stream.

PR #2825, opened 2026-05-14
Validate full bundle identifier in config

### LINKED ISSUE #2824: Overly aggressive bundle identifier validator rejects valid bundle IDs
### Describe the bug

Similar to #1212 although pretty much the opposite example, Briefcase won't let me use `is.oja` as my bundle ID, despite it being valid (`https://oja.is` will be the website for the app).  There are plenty of examples of  allowing bundle ids along these lines.  I get:

```
Briefcase configuration error: 'is' is not a valid bundle identifier.

The bundle should be a reversed domain name. It must contain at least 2 dot-separated sections; each section may only include letters, numbers, and hyphens; and each section may not contain any reserved words (like 'switch', or 'while').
```

It's true that a bundle ID does have to have at least one dot, however Briefcase is actually asking for the bundle ID prefix here, since it forms the ID from this prefix plus a dot and the app name, thus supplying the required dot.  If the actual bundle ID was validated, instead of just the prefix, there would not be a problem.

### How to reproduce the bug

Create a new briefcase project, with the following in `pyproject.toml`

```pyproject.toml
[tool.briefcase]
project_name = "oja"
bundle = "is"
```

Attempt to build with briefcase.  Get the error:

```
Briefcase configuration error: 'is' is not a valid bundle identifier.

The bundle should be a reversed domain name. It must contain at least 2 dot-separated sections; each section may only include letters, numbers, and hyphens; and each section may not contain any reserved words (like 'switch', or 'while').
```
Expect the app to build.

### Minimum example code

```shell

```

### Screenshots

_No response_

### Environment details

- Operating system and version: 26.3.1 (a) (25D771280a)
- Python version: 3.13
- Software versions:
  - Briefcase: 0.4.2
  - Toga: 0.5.4

### Logs

```

```

### Additional context

_No response_

Rjvs:
Validate the full bundle id instead of just the bundle prefix.
This more correctly validates bundle ids, allowing single-dot ids.  Fixes #2824 

## PR Checklist:

Assisted-by: Claude Opus 4.7 or perhaps Codex 5.5, I can't remember

### DIFF
--- a/src/briefcase/config.py
+++ b/src/briefcase/config.py
@@ -646,11 +646,11 @@ class DraftAppConfig(AppConfig):
                 "and cannot end with '-' or '_')."
             )
 
-        if not is_valid_bundle_identifier(self.bundle):
+        if not is_valid_bundle_identifier(self.bundle_identifier):
             raise BriefcaseConfigError(
-                f"{self.bundle!r} is not a valid bundle identifier."
+                f"{self.bundle_identifier!r} is not a valid bundle identifier."
                 f"\n\n"
-                "The bundle should be a reversed domain name. It must contain at least "
+                "The bundle id should be a reversed domain name. It must contain at least "
                 "2 dot-separated sections; each section may only include letters, "
                 "numbers, and hyphens; and each section may not contain any reserved "
                 "words (like 'switch', or 'while')."
--- a/src/briefcase/config.py
+++ b/src/briefcase/config.py
@@ -648,12 +648,13 @@ class DraftAppConfig(AppConfig):
 
         if not is_valid_bundle_identifier(self.bundle_identifier):
             raise BriefcaseConfigError(
-                f"{self.bundle_identifier!r} is not a valid bundle identifier."
-                f"\n\n"
-                "The bundle id should be a reversed domain name. It must contain at "
-                "least 2 dot-separated sections; each section may only include "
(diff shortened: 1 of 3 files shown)


src/briefcase/config.py
```diff
@@ -646,14 +646,14 @@ def __init__(
...
+                f"{self.bundle_identifier!r} is not a valid bundle identifier."
                 f"\n\n"
-                "The bundle should be a reversed domain name. It must contain at least "
-                "2 dot-separated sections; each section may only include letters, "
-                "numbers, and hyphens; and each section may not contain any reserved "
-                "words (like 'switch', or 'while')."
+                "The bundle id should be a reversed domain name. It must contain at "
+                "least 2 dot-separated sections; each section may only include "
+                "letters, numbers, and hyphens; and each section may not contain any "
+                "reserved words (like 'switch', or 'while')."
```
freakboy3742: The validation approach here makes sense; but I'm a wary of the message being misleading. The thing that is user-modifiable is `bundle`; when that is combined with the app name should be at least 2-dot separated sections.

freakboy3742 (approved):
I've tweaked the error message and release note a bit, but the approach you've taken here (validating the full ID, not just the prefix) is clever. Thanks for the PR.

PR #2832, opened 2026-05-18
Clarify first-time bundled debug setup

### LINKED ISSUE #2621: Debug documentation issues
### Describe the bug

The debug instructions will not work as written, because the `briefcase build` and `briefcase run` commands don't include `-r`. So if pip has already been run, which happens during `briefcase create` on most if not all platforms, it won't be rerun, and the app will be built without the debug components. The app will still start, but it won't wait for the debugger to connect. When you hit the first breakpoint, it will print the current line of code and then immediately crash with a `BdbQuit` exception.

I guess `--debug` mode should work the same as `--test` mode in this respect, but I'm not familiar enough with the code to know whether this is a documentation-only issue, or whether something could be improved in the implementation short of full-blown dependency tracking (#807).

----

@freakboy3742: I don't know if this was mentioned somewhere in the gigantic discussion of #2173, but at what point are we going to stop calling this "experimental"? The design seems quite complete, even if it isn't implemented on every platform, so I'm not sure why we're calling it out in this way.

Once we remove those paragraphs from the `build` and `dev` pages, the `pdb` and `vscode` pages can become the single source of truth about which platforms they support.

### Environment

- Software versions:
  - Briefcase: 0.3.27

toby-bridges:
## Summary
- update the bundled PDB and VS Code debug guides to show the first debug build with `-r`
- explain the first-time bundled debug requirement in the `run --debug` reference page
- fix the PDB guide wording so it refers to the remote PDB flow instead of VS Code

## Testing
- `git diff --check`
- `python3 -m tox -e docs-lint` *(fails locally: `No module named tox` in this environment)*

Fixes #2621.

### DIFF
--- a/docs/en/how-to/debugging/pdb.md
+++ b/docs/en/how-to/debugging/pdb.md
@@ -16,16 +16,16 @@ This is currently an **experimental feature** that is only supported on Windows,
 
 To debug a bundled app, add `breakpoint()` somewhere in your code where the debugger should halt.
 
-Your app must then be modified to include a bootstrap that will connect to the VS Code debugger. This is done by passing the `--debug pdb` option to `briefcase build`:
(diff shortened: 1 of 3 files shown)


freakboy3742:
Closing on the basis that this is an AI-generated contribution and the contributor does not appear to be interested in responding to feedback.

PR #2843, opened 2026-05-19
Android emulator system image picker

### LINKED ISSUE #737: Improve flexibility of Android emulator system image

moondial-pal:
Added two step selection flow that fetches all available compatible images from `sdkmanager --list`.
Hard coded version is set if default is selected.

This enables the user to select an android version when creating a new emulator.

Fixes #737 

## PR Checklist:

Assisted-by: Claude Sonnet 4.6


moondial-pal:
The approach was revised during the PyCon sprint after a discussion with @mhsmith.
Am going to work on aligning tests and see if I can nail this down.

moondial-pal:
Are the 3 CI failures pre-existing or am I goofing something here?

-  docs-lint broken https://www.x.org/ link in x11passthrough.md, not touched by my PR
- Windows ARM PySide6 build, similar failure visible in recent main branch CI runs
- ReadTheDocs appears to be an infrastructure build failure? 

I checked recent CI runs on main and found some of these issues crop up there. 
I'm not sure how to get them to pass at this point 😅

mhsmith:
Yes, the failures aren't related to this PR, so you don't need to worry about them.

Thanks for the updates, I'll look at this in a couple of days.

src/briefcase/integrations/android_sdk.py
```diff
@@ -701,6 +702,51 @@ def verify_avd(self, avd: str):
         except KeyError:
             self.tools.console.debug(f"Device {avd!r} doesn't define a skin.")
 
+    def list_available_system_images(
+        self, min_version: int = ANDROID_MIN_OS_VERSION
+    ) -> list[str]:
```
mhsmith: There's no need to declare a default argument which is never used in production code; this only obscures the actual source of the default.

On the other hand, the test code *should* test the behavior of the argument; it looks like it currently doesn't.

src/briefcase/integrations/android_sdk.py
```diff
@@ -23,6 +23,7 @@
 from briefcase.integrations.subprocess import SubprocessArgT
 
 DEVICE_NOT_FOUND = re.compile(r"^error: device '[^']*' not found")
+ANDROID_MIN_OS_VERSION = 26
```
mhsmith: Why 26? The [template](https://github.com/beeware/briefcase-android-gradle-template/blob/main/cookiecutter.json) currently defaults to 24.

src/briefcase/integrations/android_sdk.py
```diff
@@ -701,6 +702,51 @@ def verify_avd(self, avd: str):
...
+        self, min_version: int = ANDROID_MIN_OS_VERSION
+    ) -> list[str]:
+        """Returns a sorted list of system image package identifiers available for the
+        current architecture and minimum Android version.
+
+        :param min_version: The minimum Android version to include. Defaults to
+        ``ANDROID_MIN_OS_VERSION``.
+        e.g., ``["system-images;android-31;default;x86_64",   "system-
+        images;android-34;default;x86_64",   "system-
+        images;android-34;google_apis;x86_64"]``
```
mhsmith: If the default argument is removed, it should be removed from the docstring as well.

Our docstrings now use Markdown, so code should be marked with single backquotes. The spacing and line breaking in the code probably won't render correctly either, but RTD obviously has a problem with the current commit, so we can't see for sure. If you merge from the main branch, that will probably fix the current error.

Also, if you intend the "e.g." paragraph to be part of the main docstring, then it should go before the `param` list, not after.

src/briefcase/integrations/android_sdk.py
```diff
@@ -701,6 +702,51 @@ def verify_avd(self, avd: str):
...
+        except subprocess.CalledProcessError as e:
+            raise BriefcaseCommandError(
+                "Unable to invoke the Android SDK manager"
+            ) from e
+
+        images = []
+        for line in output.splitlines():
+            package = line.split("|")[0].strip()
+            if not package.startswith("system-images"):
+                continue
```
mhsmith: The code above this point could be replaced with a call to `list_installed_system_images`.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1113,8 +1160,47 @@ def create_emulator(self) -> str:
         device_type = self.DEFAULT_DEVICE_TYPE
         skin = self.DEFAULT_DEVICE_SKIN
 
-        # TODO: Provide a list of options for system images.
-        system_image = self.DEFAULT_SYSTEM_IMAGE
+        # Get available images, raise an error if none found.
+        available_images = self.list_available_system_images(
+            min_version=getattr(app, "min_os_version", ANDROID_MIN_OS_VERSION)
```
mhsmith: Recommend factoring out this `getattr` expression into a function, and using it in `output_format_template_context`. That would eliminate any risk of the two locations going out of sync.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1113,8 +1160,47 @@ def create_emulator(self) -> str:
...
+No Android system images are available for your architecture
+({self.emulator_abi}).
+
+This may be caused by a network connectivity issue or an unsupported
+architecture. Check your network connection and re-run `briefcase run android`.
+"""
+            )
+
+        # Ask the user to select an Android version.
+        versions = sorted({img.split(";")[1].split("-")[1] for img in available_images})
```
mhsmith: The `split` expressions in this function are difficult to understand. Suggest creating a parsing function which converts a system image name into an object with fields for Android version, image type and ABI. Then we can either use that function both here and in `list_available_system_images`, or maybe even make `list_available_system_images` return the parsed objects.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1113,8 +1160,47 @@ def create_emulator(self) -> str:
...
+        version = self.tools.console.selection_question(
+            intro="Select the Android version for the emulator:",
+            description="Android version",
```
mhsmith: To avoid confusion, should say "API level" rather than "Android version".

src/briefcase/integrations/android_sdk.py
```diff
@@ -1113,8 +1160,47 @@ def create_emulator(self) -> str:
...
+            options=versions,
+            default="31",
+        )
+
+        # Ask the user to select an image type for the chosen version.
+        image_types = sorted(
+            {
+                f"{img.split(';')[1]};{img.split(';')[2]}"
+                for img in available_images
+                if img.split(";")[1].startswith(f"android-{version}")
```
mhsmith: Why is the version included in the type here, and why are we using `startswith`? I thought it might be because of decimal versions, but that doesn't quite seem right either. For example, if the version question has choices for 36 and 36.1, and the user chose 36, the second question will still include the images for 36.1. It's probably better not to do anything clever about that, and treat every version independently.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1113,8 +1160,47 @@ def create_emulator(self) -> str:
         device_type = self.DEFAULT_DEVICE_TYPE
         skin = self.DEFAULT_DEVICE_SKIN
 
-        # TODO: Provide a list of options for system images.
-        system_image = self.DEFAULT_SYSTEM_IMAGE
```
mhsmith, line 1136: `DEFAULT_SYSTEM_IMAGE` is still used in one other place, but it should be updated to make sure it's consistent with the default Android version and image type used in this function. This would probably involve creating `DEFAULT` properties for those two things, and using them in both locations.

tests/integrations/android_sdk/AndroidSDK/test__create_emulator.py
```diff
@@ -311,17 +318,25 @@ def test_default_name(mock_tools, android_sdk, tmp_path):
...
 
     # Mock the initial output of an AVD config file.
     avd_config_path = tmp_path / "home/.android/avd/beePhone.avd/config.ini"
     avd_config_path.parent.mkdir(parents=True)
     with avd_config_path.open("w", encoding="utf-8") as f:
         f.write("hw.device.name=pixel\n")
 
+    # Create a mock app
+    app = MagicMock()
+    del app.min_os_version  # ensure getattr fallback is used
```
mhsmith, line 336: There should be at least one test of the attribute being present.

src/briefcase/integrations/android_sdk.py
mhsmith: Sorry, not replaced exactly, but the duplicate code should be factored out.

moondial-pal:
Thanks for the detailed feedback. I'll work through each point. 
The suggestion to factor out the parsing logic and treat versions independently makes a lot of sense. I'll get back to you with updates.

src/briefcase/integrations/android_sdk.py
moondial-pal: Got it. I chose 26 because of a comment Russell made in issue #737: https://github.com/beeware/briefcase/issues/737#issuecomment-1165156168
Would you recommend I stick with the template default of 24?

freakboy3742: That's a typo on my part - we should definitely stick to the 24 minimum.

moondial-pal:
Digging into the API level sorting here is the current state of the API level selection wizard for feedback:

image

I had some questions before going any further.
1. `android-CANARY` and `android-CinnamonBun` have no numeric component, so currently they sort to the bottom of the list alphabetically rather than near a numeric position.
`CinnamonBun` is the codename for the 37 preview.
`CANARY` is a rolling pre-release with no fixed API level.
How would we want to sort those?

2. `android-37.0` and `android-CinnamonBun` both appear to refer to the same underlying release.
Should these be de-duplicated or is showing both acceptable?

3. The API level prompt currently defaults to 31.
Do we want to stick with that or have the wizard default to the latest?

That brings up another question that may be out of scope for this pr but how do we want to handle default API level selection as the years progress?

mhsmith:
> 1. `android-CANARY` and `android-CinnamonBun` have no numeric component, so currently they sort to the bottom of the list alphabetically rather than near a numeric position.
>    `CinnamonBun` is the codename for the 37 preview.
>    `CANARY` is a rolling pre-release with no fixed API level.
>    How would we want to sort those?

The bottom of the list looks fine to me, since they are the newest versions.

> 2. `android-37.0` and `android-CinnamonBun` both appear to refer to the same underlying release.
>    Should these be de-duplicated or is showing both acceptable?

If they have separate SDK packages, then let's show them separately. At some point, I think the package with the codename will disappear, but existing emulators should continue to work.

> 3. The API level prompt currently defaults to 31.
>    Do we want to stick with that or have the wizard default to the latest?
> 
> That brings up another question that may be out of scope for this pr but how do we want to handle default API level selection as the years progress?

The ideal solution is probably to make it default to `target_os_version`. But you don't have to do that in this PR unless you want to; sticking with the current default is fine.

src/briefcase/integrations/android_sdk.py
mhsmith: Also, although `min_os_version` is used as an integer in the Gradle template, Briefcase documents it as a string, for consistency with the other platforms. This will probably lead to a TypeError when it's compared to an integer in list_available_system_images. There should be a test to make sure we accept both types.

moondial-pal:
Sounds good on all three.
Keeping current sort order for numeric and named entries.
The default will remain `android-31` for now, will tackle `target_os_version` in a separate PR.

moondial-pal:
Still working on this. I have been on a camping/road trip for the past 9 days and won't return until 7/6/26 😅. I should be able to submit something that week.

moondial-pal:
After hitting an issue with a default selection for the tag, it comes to mind that a similar edge case could occur if  `android-31` isn't available for the user's architecture. Honestly not sure how likely that is, but the API level prompt will crash since `DEFAULT_API_LEVEL` won't be in the options list. Essentially, the tag prompt had the same issue and I fixed it with `self.DEFAULT_TAG if self.DEFAULT_TAG in tags else tags[0]`. Should I apply a similar pattern for the API level default, or would you prefer a different fallback?

mhsmith:
ARM64 and x86_64 are our only supported architectures for the build machine, and we know they are both available for the current hard-coded default API level. So I don't think we need to worry about that now.

mhsmith:
Hi, I saw your comment on the community call, so I'll look at this as soon as I can. In future, you can post a comment on a PR to let us know it's ready for another review.

moondial-pal:
> Hi, I saw your comment on the community call, so I'll look at this as soon as I can. In future, you can post a comment on a PR to let us know it's ready for another review.

I was on the fence on if that would be a nuisance or not. Only because I know that yourself and Russell have your work cut out for you as far as reviews go. Anytime I have submitted a PR both of you have always reviewed within a very reasonable time. 
I learn a little bit more about the contributing norms with each PR, so thanks for mentioning that.
Absolutely. Moving forward I'll make sure to communicate when I have something ready for review.

Honestly because I was on that long trip, I hadn't been able to work on submitting anything until 2 days before the meeting. I had nothing else to post or talk about so I thought I would at least update the progress of this PR 😅

moondial-pal:
I believe I have addressed all the feedback. Ready for review.

freakboy3742:
@moondial-pal Apologies for the delay in reviewing this - we should be able to get to this in the next couple of days.

src/briefcase/integrations/android_sdk.py
mhsmith: The duplication with `list_installed_system_images` is still unresolved.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1175,32 +1241,34 @@ def create_emulator(self, app: FinalizedAppConfig) -> str:
...
+            key=lambda x: (0 if x == "default" else 1, x),
         )
-        image_type = self.tools.console.selection_question(
-            intro="Select the system image type:",
-            description="Image type",
-            options=image_types,
-            default=image_types[0],
+        tag = self.tools.console.selection_question(
+            intro="Select the system image tag:",
+            description="Tag",
```
mhsmith, line 1266: Is there any precedent for calling this a "tag"? "Type" seems like a more obvious name, and has [previously been used in Android Studio](https://developer.android.com/studio/run/managing-avds#system-image) (see right side of screenshot).

src/briefcase/integrations/android_sdk.py
```diff
@@ -1132,8 +1226,49 @@ def create_emulator(self) -> str:
...
+architecture. Check your network connection and re-run `briefcase run android`.
+"""
+            )
+
+        # Parse available images once for use in both selection questions.
+        parsed_images = [_parse_system_image(img) for img in available_images]
+
+        # Ask the user to select an API level.
+        api_levels = sorted(
+            {api_level for api_level, _, abi in parsed_images},
```
mhsmith, line 1249:
```suggestion
            {api_level for api_level, _, _ in parsed_images},
```

src/briefcase/integrations/android_sdk.py
```diff
@@ -719,6 +774,44 @@ def verify_avd(self, avd: str):
         except KeyError:
             self.tools.console.debug(f"Device {avd!r} doesn't define a skin.")
 
+    def list_available_system_images(self, min_api_level: int) -> list[str]:
+        """Returns a sorted list of system image package identifiers available for the
+        current architecture and minimum Android version.
+
+        e.g., ``{"system-images;android-31;default;x86_64"}``
```
mhsmith, line 781: The sorting seems redundant, since they'll be sorted again in the one place that calls this function. In which case, maybe it could return a `set` instead, which would be more consistent with `list_installed_system_images`, and would make the docstring example syntax correct.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1132,8 +1226,49 @@ def create_emulator(self) -> str:
...
-        # TODO: Provide a list of options for system images.
-        system_image = self.DEFAULT_SYSTEM_IMAGE
+        # Get available images, raise an error if none found.
+        available_images = self.list_available_system_images(
+            min_api_level=min_api_level(app)
+        )
+        if not available_images:
+            raise BriefcaseCommandError(
+                f"""\
+No Android system images are available for your architecture
```
mhsmith, line 1236: This could use the same dedent function as your other PR.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1132,8 +1226,49 @@ def create_emulator(self) -> str:
...
+({self.emulator_abi}).
+
+This may be caused by a network connectivity issue or an unsupported
+architecture. Check your network connection and re-run `briefcase run android`.
```
mhsmith, line 1240: Network connectivity seems like an unlikely cause, because if the sdkmanager failed to contact the server, it should have caused a CalledProcessError. However, it would be worth mentioning `min_os_version` and its value, since that's the other thing used for filtering here. There's no guarantee that CANARY will continue to exist, so setting `min_os_version` to a very high number could return no results.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1175,32 +1241,34 @@ def create_emulator(self, app: FinalizedAppConfig) -> str:
...
+        # Ask the user to select an API level.
+        api_levels = sorted(
+            {api_level for api_level, _, abi in parsed_images},
+            key=_api_level_sort_key,
+        )
+        api_level = self.tools.console.selection_question(
+            intro="Select the API level for the emulator:",
+            description="API level",
+            options=api_levels,
+            default=self.DEFAULT_API_LEVEL,
```
mhsmith, line 1256: If `min_os_version` is set high enough, this may not be a valid default. There should be a fallback to default to the first item, like in the other question below.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1132,8 +1226,49 @@ def create_emulator(self) -> str:
...
+        # Ask the user to select a tag for the chosen API level.
+        tags = sorted(
+            {tag for level, tag, _ in parsed_images if level == api_level},
+            key=lambda x: (0 if x == "default" else 1, x),
+        )
+        tag = self.tools.console.selection_question(
+            intro="Select the system image tag:",
+            description="Tag",
+            options=tags,
+            default=self.DEFAULT_TAG if self.DEFAULT_TAG in tags else tags[0],
```
mhsmith, line 1268: It doesn't look like this `tags[0]` fallback is unit tested.

mhsmith (commented):
Sorry for taking so long. This is looking pretty good now, just a few more comments, plus one unresolved conversation from the previous round.

moondial-pal:
> Sorry for taking so long. This is looking pretty good now, just a few more comments, plus one unresolved conversation from the previous round.

Not an issue at all, I know this issue has been a bit drawn out. 
I'll work on the feedback, thanks.

src/briefcase/integrations/android_sdk.py
moondial-pal: Looking back, I had mistakenly associated "tag" with the version field from doing a bit of research here: [Code names, tags and build numbers](https://source.android.com/docs/setup/reference/build-numbers), but that refers to code tags, not system images. 
Changed the user facing label to "Type". I also updated the internal variable names where I used tag as well.

moondial-pal:
Apologies for the delayed submission.
@mhsmith this is ready for further review.

freakboy3742:
@moondial-pal Looks like you've hit a pre-commit problem...

mhsmith:
Since this PR was last merged from the main branch, we've started more strictly enforcing line length limits (#2979).

You'll also have to resolve some merge conflicts with #3033, which added the API level to the default emulator name.

PR #2845, opened 2026-05-19
Fix typing errors in console.py

Tridwoxi:
`src/briefcase/console.py` contained a few type errors according to `ty`. These have been fixed.

Behaviour changes:

- If `Console.close` is called more than once on the same instance, the method will do nothing instead of throwing an `AttributeError`.
- If the exception being captured by `Console.capture_stacktrace` has attribute `skip_logfile` yet is not a `BriefcaseError`, `skip_logfile` will be ignored. An alternative implementation more amiable to duck typing would be a `hasattr + isinstance` check.
- If `Console.capture_stacktrace` is called before an exception has been thrown, it will do nothing instead of throwing an `AttributeError` (due to implementation details of `rich` v15.0.0).
- If `sys.__stdout__ is None` then `Console.is_interactive` will return `False` instead of throwing a `AttributeError`.

I don't think these branches are taken under normal execution, so there should be no runtime impact. Corresponding tests have been added.

## PR Checklist:

Assisted-by: Claude Opus 4.7

### DIFF
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -514,17 +514,17 @@ class Console:
         :param label: An identifying label for the thread that has raised the
             stacktrace. Defaults to the main thread.
         """
-        _, exception, _ = sys.exc_info()
-        if exception is None:
+        exc_type, exc_value, exc_trace = sys.exc_info()
+        if exc_value is None:
             return
 
-        if isinstance(exception, BriefcaseError):
-            self.skip_log = exception.skip_logfile
+        if isinstance(exc_value, BriefcaseError):
+            self.skip_log = exc_value.skip_logfile
 
         trace = Traceback.extract(
-            type(exception),
-            exception,
-            exception.__traceback__,
+            exc_type,  # ty:ignore[invalid-argument-type]
+            exc_value,
+            exc_trace,
             show_locals=True,
         )
         self.stacktraces.append((label, trace))
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -9,12 +9,12 @@ import sys
 import textwrap
 import time
 import traceback
-from collections.abc import Callable, Collection, Iterable, Mapping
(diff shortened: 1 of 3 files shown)


src/briefcase/console.py
```diff
@@ -514,13 +514,20 @@ def capture_stacktrace(self, label="Main thread"):
         :param label: An identifying label for the thread that has raised the
             stacktrace. Defaults to the main thread.
         """
-        exc_info = sys.exc_info()
-        try:
-            self.skip_log = exc_info[1].skip_logfile
-        except AttributeError:
-            pass
+        _, exception, _ = sys.exc_info()
```
freakboy3742: Is there a typing-specific reason for throwing away the exception type and traceback here, and then re-constructing those a couple of lines later for the Traceback construction?

freakboy3742 (changes requested):
Thanks for this PR; these all makes sense, except for one flagged inline.

src/briefcase/console.py
Tridwoxi: The 3.10 docs for `sys.exc_info()` specify:

> If no exception is being handled anywhere on the stack, a tuple containing three `None` values is returned. Otherwise, the values returned are `(type, value, traceback)`.

In the type system, this shows up as `tuple[type[BaseException], BaseException, TracebackType] | tuple[None, None, None]`.

This means once we know the exception value is not `None`, the exception type is provably not `None`. Equally correct and more performant:

```py
        exc_type, exc_value, exc_trace = sys.exc_info()
        if exc_value is None:
            return

        if isinstance(exc_value, BriefcaseError):
            self.skip_log = exc_value.skip_logfile

        trace = Traceback.extract(
            exc_type,  # ty:ignore[invalid-argument-type]
            exc_value,
            exc_trace,
            show_locals=True,
        )
        self.stacktraces.append((label, trace))
```

I'll write this in a moment.

freakboy3742 (approved):
Thanks for those updates!

PR #2855, opened 2026-05-29
Support min_os_version on Windows

abdnh:
This adds support for specifying the minimum supported Windows build number using the `min_os_version` option.

## PR Checklist:

## Context

https://github.com/ankitects/anki/issues/4765

### DIFF
--- a/src/briefcase/platforms/windows/__init__.py
+++ b/src/briefcase/platforms/windows/__init__.py
@@ -263,7 +263,7 @@ class WindowsCreateCommand(CreateCommand):
         **kwargs,
     ):
         if template_min_version := self.target_windows_build(app):
-            min_version = getattr(app, "min_os_version", template_min_version)
+            min_version = int(getattr(app, "min_os_version", template_min_version))
             if min_version < template_min_version:
                 raise BriefcaseCommandError(
                     "Your Windows app specifies a minimum build number of "
--- a/src/briefcase/platforms/windows/__init__.py
+++ b/src/briefcase/platforms/windows/__init__.py
@@ -142,7 +142,7 @@ Install a 64bit version of Python and run Briefcase again.
         :return: version or None if one isn't specified
         """
         try:
-            return int(self.briefcase_toml(app)["briefcase"]["target_windows_build"])
+            return self.briefcase_toml(app)["briefcase"]["target_windows_build"]
         except KeyError:
             return None
 
@@ -264,7 +264,7 @@ class WindowsCreateCommand(CreateCommand):
     ):
         if template_min_version := self.target_windows_build(app):
             min_version = int(getattr(app, "min_os_version", template_min_version))
-            if min_version < template_min_version:
+            if min_version < int(template_min_version):
                 raise BriefcaseCommandError(
                     "Your Windows app specifies a minimum build number of "
                     f"{min_version}, but the app template only supports "
--- a/src/briefcase/platforms/windows/__init__.py
+++ b/src/briefcase/platforms/windows/__init__.py
@@ -243,6 +243,26 @@ class WindowsCreateCommand(CreateCommand):
             """,
         )
 
+    def _install_app_requirements(
+        self,
+        app: FinalizedAppConfig,
+        requires: list[str],
+        app_packages_path: Path,
+        **kwargs,
+    ):
+        support_min_version = 10240  # Windows 10
+        min_version = int(getattr(app, "min_os_version", support_min_version))
+        if min_version < support_min_version:
+            raise BriefcaseCommandError(
+                "Your Windows app specifies a minimum build number of "
+                f"{min_version}, but the support package only supports "
+                f"{support_min_version}"
+            )
+
+        return super()._install_app_requirements(
+            app, requires, app_packages_path, **kwargs
+        )
+
     def install_license(self, app: FinalizedAppConfig):
         """Install the license for the project as a single RTF document.
 
--- a/src/briefcase/platforms/windows/__init__.py
+++ b/src/briefcase/platforms/windows/__init__.py
@@ -264,7 +264,10 @@ class WindowsCreateCommand(CreateCommand):
     ):
         template_min_version = self.target_windows_build(app)
         if template_min_version:
-            min_version = getattr(app, "min_os_version", template_min_version)
+            min_version = getattr(app, "min_os_version", None)
+            if min_version is None:
+                app.min_os_version = template_min_version
+                min_version = template_min_version
             if min_version and int(min_version) < template_min_version:
                 raise BriefcaseCommandError(
                     "Your Windows app specifies a minimum build number of "
--- a/src/briefcase/platforms/windows/__init__.py
+++ b/src/briefcase/platforms/windows/__init__.py
@@ -262,13 +262,12 @@ class WindowsCreateCommand(CreateCommand):
         app_packages_path: Path,
         **kwargs,
     ):
-        template_min_version = self.target_windows_build(app)
-        if template_min_version:
+        if template_min_version := self.target_windows_build(app):
             min_version = getattr(app, "min_os_version", None)
             if min_version is None:
                 app.min_os_version = template_min_version
                 min_version = template_min_version
-            if min_version and int(min_version) < template_min_version:
+            elif min_version < template_min_version:
                 raise BriefcaseCommandError(
                     "Your Windows app specifies a minimum build number of "
                     f"{min_version}, but the app template only supports "
--- a/src/briefcase/platforms/windows/__init__.py
+++ b/src/briefcase/platforms/windows/__init__.py
@@ -134,6 +134,18 @@ Windows applications cannot be built using a 32bit version of Python.
 Install a 64bit version of Python and run Briefcase again.
 """)
 
+    def target_windows_build(self, app: FinalizedAppConfig) -> int | None:
+        """The minimum supported Windows build number for the app from
+        ``briefcase.toml``.
+
+        :param app: The config object for the app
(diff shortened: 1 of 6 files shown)


src/briefcase/platforms/windows/__init__.py
```diff
@@ -243,6 +243,26 @@ def _cleanup_app_support_package(self, support_path):
...
         )
 
+    def _install_app_requirements(
+        self,
+        app: FinalizedAppConfig,
+        requires: list[str],
+        app_packages_path: Path,
+        **kwargs,
+    ):
+        support_min_version = 10240  # Windows 10
```
abdnh: Hardcoded because there's no reliable way to detect the minimum version from the support package as far as I know.

freakboy3742: The support package won't include this information, but we could include it in the `briefcase.toml` in the app template - that is where the support package build number is defined, so it would make sense to tie the two together, rather than hard-coding the value here. If it isn't in the template, falling back to "not defined" would make sense to me.

src/briefcase/platforms/windows/__init__.py
```diff
@@ -243,6 +243,26 @@ def _cleanup_app_support_package(self, support_path):
...
+        min_version = int(getattr(app, "min_os_version", support_min_version))
+        if min_version < support_min_version:
+            raise BriefcaseCommandError(
+                "Your Windows app specifies a minimum build number of "
+                f"{min_version}, but the support package only supports "
+                f"{support_min_version}"
+            )
```
freakboy3742: This checks the value on the *installer* side, but it won't put the value into the template in the "default" case. i.e., it will catch the error if I try to build the app on Windows 8, but if I build the app on Windows 11, and then try to install on Windows 8, it won't pass that information down unless I explicitly put `min_os_version` in the app config.

freakboy3742 (changes requested):
Thanks for the update. This looks like a really useful improvement; a couple of suggestions inline regarding how this will be applied in practice.

tests/platforms/windows/app/create/test_create.py
```diff
@@ -259,3 +260,52 @@ def test_external(create_command, external_first_app, tmp_path):
...
+        (10240, "7601", False),
+        (10240, "10240", True),
+        (10240, "17763", True),
+        (None, 10240, True),
+        (10240, None, True),
+        (None, None, True),
+    ],
+)
+def test_min_os_version(
+    create_command, first_app_templated, template_version, app_version, compatible
```
freakboy3742: As a style thing, if a list of arguments long enough that it's triggering ruff's "put it on another line" rule, we go straight to the "one argument per line" formatting (which you get out of ruff automatically by adding a comma to the end of the last argument).

src/briefcase/platforms/windows/__init__.py
```diff
@@ -243,6 +255,30 @@ def _cleanup_app_support_package(self, support_path):
...
+        app: FinalizedAppConfig,
+        requires: list[str],
+        app_packages_path: Path,
+        **kwargs,
+    ):
+        template_min_version = self.target_windows_build(app)
+        if template_min_version:
+            min_version = getattr(app, "min_os_version", None)
+            if min_version is None:
+                app.min_os_version = template_min_version
```
freakboy3742: If this is going to propagate into the template, it can't be done here. 

`install_app_requirements` is invoked *after* the app template is generated, so with this approach, the `min_os_version` would only be annotated onto the app (a) after the template has been rolled out, and (b) only if requirements were being installed (which they should always be if you're creating the template, but it also means updating requirements would cause the app to be annotated).

If the app configuration is being updated, it should be done in a `finalize_app_config()` override so that *every* app configuration has a `min_os_version` under Windows.

src/briefcase/platforms/windows/__init__.py
```diff
@@ -243,6 +255,30 @@ def _cleanup_app_support_package(self, support_path):
...
 
+    def _install_app_requirements(
+        self,
+        app: FinalizedAppConfig,
+        requires: list[str],
+        app_packages_path: Path,
+        **kwargs,
+    ):
+        template_min_version = self.target_windows_build(app)
+        if template_min_version:
```
freakboy3742:
```suggestion
        if template_min_version := self.target_windows_build(app):
```

src/briefcase/platforms/windows/__init__.py
```diff
@@ -243,6 +255,30 @@ def _cleanup_app_support_package(self, support_path):
...
+            min_version = getattr(app, "min_os_version", None)
+            if min_version is None:
+                app.min_os_version = template_min_version
+                min_version = template_min_version
+            if min_version and int(min_version) < template_min_version:
```
freakboy3742: Isn't this an elif? Also - is the int conversion required?
```suggestion
            elif min_version < template_min_version:
```

freakboy3742 (changes requested):
Looking good; a couple of edge cases and style notes inline.

src/briefcase/platforms/windows/__init__.py
abdnh: But we can't read briefcase.toml in finalize_app_config() as it's not available at that stage?

Will setting a default min_os_version in cookiecutter.json work (same value as target_windows_build)?

src/briefcase/platforms/windows/__init__.py
abdnh: I was thinking it should a string (e.g `min_os_version = "17763"`) for consistency with other platforms, but that's probably unnecessary here.

src/briefcase/platforms/windows/__init__.py
freakboy3742: > But we can't read briefcase.toml in finalize_app_config() as it's not available at that stage?

Hrm - good point... 
 
> Will setting a default min_os_version in cookiecutter.json work (same value as target_windows_build)?

I guess that works - it's also the same approach that the iOS and macOS templates use, so there's at least *some* consistency there. 

The value in briefcase.toml is still needed because the support package can't report a minimum version - but the value can be templated from `{{ cookiecutter.min_os_version }}` to avoid duplication.

src/briefcase/platforms/windows/__init__.py
freakboy3742: Hrm - I guess that's fair. I guess it doesn't hurt to have the extra safety there.

src/briefcase/platforms/windows/__init__.py
abdnh: > but the value can be templated from {{ cookiecutter.min_os_version }} to avoid duplication.

This will break the build-time check though: target_windows_build will always match min_os_version.

freakboy3742: True... I think I've been getting myself tied in knots trying to form the analog with macOS/iOS, and that analog doesn't exist.

The core of the issue here is that we need the app's `min_os_version` to default to *at least* the support package's minimum version, but we can't get a default into the app's context until we've rolled out the template.

I think the fix here is to accept there will be some duplication in the template. 

If we modify the template so that `briefcase.toml` has hard coded value for `target_os_version`, and we set the default value for `min_os_version` in `cookiecutter.json` to be the same value, then we'll get the functionality that we need:

* If the user doesn't specify a `min_os_version`, the MSI will get the same value as the support package
* If the user specifies a `min_os_version`, it will be validated against the support package's `briefcase.toml` value, and the app won't build if it's older than the support package.

The only complication is that this requires duplicating the minimum Windows build version in 2 places in the template. That's less than ideal - but it's also something we can protect against with a CI check. If we add a unit test that validates that the value in `cookiecutter.json` matches the value in `briefcase.toml`, we get the functionality that we need, and we don't have to overthink the process of updating the default app value in the template.

abdnh: Added a CI check to the template.

tests/platforms/windows/app/create/test_create.py
```diff
@@ -259,3 +260,56 @@ def test_external(create_command, external_first_app, tmp_path):
...
+
+@pytest.mark.parametrize(
+    ("template_version", "app_version", "compatible"),
+    [
+        (10240, 7601, False),
+        (10240, 10240, True),
+        (10240, 17763, True),
+        (None, 10240, True),
+        (10240, None, True),
+        (None, None, True),
```
freakboy3742, line 273:
```suggestion
        (None, None, True),
        # Values provided as strings are converted to int
        ("10240", "7601", False),
        (10240, "7601", False),
        ("10240", 7601, False),
```

src/briefcase/platforms/windows/__init__.py
```diff
@@ -243,6 +255,26 @@ def _cleanup_app_support_package(self, support_path):
...
 
+    def _install_app_requirements(
+        self,
+        app: FinalizedAppConfig,
+        requires: list[str],
+        app_packages_path: Path,
+        **kwargs,
+    ):
+        if template_min_version := self.target_windows_build(app):
+            min_version = getattr(app, "min_os_version", template_min_version)
```
freakboy3742:
```suggestion
            min_version = int(getattr(app, "min_os_version", template_min_version))
```

freakboy3742 (approved):
I've restored the `int()` conversion (and added a handful of tests for that) - but otherwise, this looks great. Thanks for all your work on this - and apologies for the multiple back-and-forth passes while I got things straight in my head.

freakboy3742:
... and I just remembered I can't push the updates to the branch because it's a branch on an organisation.

If you merge those two suggestions, I'll merge the PR.

freakboy3742:
The docs CI issue has been fixed on main; so I'll merge with that CI failure.

PR #2863, opened 2026-06-02
Single option selection default

### LINKED ISSUE #2856: Single option default selection

In this use case, a user is creating an app for Android or iOS with no physical device plugged into their machine and no emulator(s) created.
In particular, I am thinking of someone using BeeWare for the first time. 

When they run either `briefcase run android` or `briefcase run iOS` the wizard presents them with a single option.

The single option being, create an emulator:

Image

If the user presses enter to continue, the wizard will show them a warning stating "Invalid Selection" and prompt them to select again:

Image

The solution I propose, would be to have the wizard select the single option as default.
When the user presses enter to create a new emulator, the wizard does so and proceeds to prompt the user for the emulator name:

Image

The original logic would play out if two or more options were available. 
Either multiple emulators:

Image

Or a plugged in device and one or more emulators:

Image

An alternative I considered proposing is to have the wizard always have a default regardless of how many options are available, but I believe that is too opinionated with multiple options and may lead to accidental selections with no option but to quit the wizard and restart the process.

The other alternative is to leave the wizard logic as is. The proposed solution is a micro experience enhancement. 

### Additional context

I noticed this quirk during the PyCon US time frame this year.
It occurred to me at a BeeWare open space that a new user would end up running into this and the proposed solution would smooth that experience out. 

I have PR ready for submission if this is an acceptable direction to go.

moondial-pal:
When only a single option is available in a selection prompt, it will be automatically selected as the default when the user presses enter.

Previously this would show an "Invalid Selection" warning and re-prompt the user.
From what I read, I believe this would be complementary to the user-defined defaults being worked on in #2279.

Fixes #2856 

## PR Checklist:

Assisted-by: Calude Sonnet 4.6

### DIFF
--- a/changes/2856.feature.md
+++ b/changes/2856.feature.md
@@ -1,0 +1,1 @@
+When only a single option is available in a selection prompt, it is automatically selected as the default.
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -1073,6 +1073,8 @@ class Console:
         self.prompt()
 
         choices = [str(index) for index in range(1, len(ordered) + 1)]
+        if len(ordered) == 1:
+            default = "1"
         index = self.input_selection(
             prompt=f"{description} [{default}]: " if default else f"{description}: ",
             choices=choices,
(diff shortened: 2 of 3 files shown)


mhsmith (approved):
Thanks!

Issue #2864, opened 2026-06-04
Long project descriptions cause icons to be blank on MSI installers


freakboy3742:
### Describe the bug

Originally reported by @initialcommit-io as #2465 

When an MSI installer creates a shortcut, if the description field is > 256 characters, the prefix of the icon link is shortened by the amount that the description field overflows the 256 character limit. 

The 

In this case, the description field was 272 characters, so the first 16 characters was being truncated.

This was only discovered by rolling back to Wix 3. Wix 3 raises a warning in this condition. Wix 5 is silent on the matter. I can't see any obvious "verbose" or "display warnings" mode that we've missed in wix.exe, so... thanks Wix. :-)

So - the immediate workaround is to truncate the description field.

Longer term, there are 2 (well... 3) fixes required:

A patch to the [briefcase-windows-app-template](https://github.com/beeware/briefcase-windows-app-template) and [briefcase-windows-visualstudio-template](https://github.com/beeware/briefcase-windows-visualstudio-template), ensuring that the usage of description in the shortcut link is limited to 256 characters. The actual change will be the same in both repositories.
A patch to Briefcase that does a check when a project configuration is parsed. The description field should be short; longer descriptions should be in the long_description field. So - if the description field is longer than 80 chars, we should output a warning to the user that their description is longer than it should be, may be truncated under some circumstances, and longer descriptions should be in the long_description field.

### How to reproduce the bug

1. Create a HelloWorld project with a project description that is more than 256 characters long (say, 260)
2. Run `briefcase package -p msi` to build an MSI
3. Install the MSI

Observe that the shortcut icon for the project is a blank page, not the grey Brutus icon.

If you press CTRL+R to open a run window, and run `shell:common programs` (for an all-users install) or `shell:programs` (for a per-user install), you can find the start menu shortcut that has been installed. Inspect the properties of the shortcut, and you should see that the path for the icon is malformed at the start, by as many characters as the description is longer than 256 characters. A 260 character description will mean the path is missing the first 4 characters (So it will start `sers\<username>...` instead of `C:\Users\<username>...`, or `DATA%\...` instead of `%APPDATA%\...`

### Environment details

- Operating system and version: Windows 11
- Python version: All (tested on 3.14)
- Software versions:
  - Briefcase: 0.4.2

### Logs

WiX produces no error or warning when this happens - it only manifests as a problem at runtime.

### Additional context

There are 3 fixes required:

[...]

PR #2869, opened 2026-06-04
Warn when an app description exceeds the recommended length

Labib-Bin-Salam:
## Changes

Partially addresses #2864.

That issue notes that a `description` longer than 256 characters silently corrupts the shortcut icon path in Windows MSI installers, and lists three required fixes. This PR implements the Briefcase-side fix:

> A patch to Briefcase that does a check when a project configuration is parsed. [...] if the `description` field is longer than 80 chars, we should output a warning to the user that their description is longer than it should be, may be truncated under some circumstances, and longer descriptions should be in the `long_description` field.

Specifically:

- Added a `MAX_DESCRIPTION_LENGTH` (80) constant to `briefcase.config`.
- `parse_config()` now emits a `console.warning(...)` when an app's `description` exceeds that length, pointing the user at `long_description`. The description value itself is preserved as-is, so this is purely advisory and non-breaking.
- Added tests covering both the over-length (warns) and at-limit (no warning) cases.
- Added a change note.

The two remaining fixes from the issue — capping the description at 256 characters in the [briefcase-windows-app-template](https://github.com/beeware/briefcase-windows-app-template) and [briefcase-windows-visualstudio-template](https://github.com/beeware/briefcase-windows-visualstudio-template) — live in separate repositories and are out of scope for this PR.

## PR Checklist:
Assisted-by: Claude Opus 4.8


src/briefcase/config.py
```diff
@@ -1463,6 +1469,20 @@ def parse_config(config_file: Path, platform, output_format, console):
         # Normalize license fields to PEP 639 representation.
         normalize_license_config(config, app_name, base_path, console)
 
+        # The description should be a short, single-line summary. Warn (but don't
+        # fail) if it's too long, as some packaging formats embed it in
+        # length-limited fields; for example, an over-long description corrupts
+        # the shortcut icon path in Windows MSI installers.
+        description = config.get("description")
+        if isinstance(description, str) and len(description) > MAX_DESCRIPTION_LENGTH:
+            console.warning(
```
freakboy3742: When a warning is long, we can now use the `console.warning_banner()` helper. That helper automatically wraps the the content, as well as standardising the presentation of config issues.

src/briefcase/config.py
```diff
@@ -41,6 +41,12 @@
     "and cannot end with '-' or '_')."
 )
 
+# The `description` is a short, single-line summary of the app. Longer values
+# belong in `long_description`. Some packaging formats embed the description in
+# length-limited fields (e.g. Windows MSI shortcuts truncate at 256 characters,
+# corrupting the icon path), so Briefcase warns when it exceeds this length.
+MAX_DESCRIPTION_LENGTH = 80
```
freakboy3742: This doesn't need to be factored out like this. It's used in exactly one place.

src/briefcase/config.py
```diff
@@ -1463,6 +1469,20 @@ def parse_config(config_file: Path, platform, output_format, console):
         # Normalize license fields to PEP 639 representation.
         normalize_license_config(config, app_name, base_path, console)
 
+        # The description should be a short, single-line summary. Warn (but don't
+        # fail) if it's too long, as some packaging formats embed it in
+        # length-limited fields; for example, an over-long description corrupts
+        # the shortcut icon path in Windows MSI installers.
```
freakboy3742: We don't need excessively long comments explaining code, especially when the warning message explains the underlying problem.
```suggestion
```

freakboy3742 (changes requested):
Thanks for the contribution. A couple of suggestions inline, mostly related to keeping comments terse. There's no need to provide multiple copies of a long description in the code.

The CI failure on documentation will be resolved after merging with main - we landed a change today that made a change to how documentation is structured. Don't worry about the Pyside6 failures - those tests have a tendency to be unstable for a reason that we haven't quite determined yet.

freakboy3742 (changes requested):
Please don't force-push to an existing PR - it makes it much harder to track what review comments have been addressed.

In this case - it's not even clear what has been updated as a result of your force push - it doesn't look like you've addressed any of the comments from my previous review.

freakboy3742:
@Labib-Bin-Salam Are you planning to continue work on this PR, or should we close it so someone else can pick up the issue?

Labib-Bin-Salam:
Yes, I'd like to continue — apologies for the delay, and for the earlier force-push.

I've pushed a new commit (cb8d8817) on top that addresses the review feedback, plus a merge of `main` to pick up the documentation restructure:

- The warning now goes through `console.warning_banner()`, so it's wrapped and presented consistently with the other configuration warnings.
- Removed the `MAX_DESCRIPTION_LENGTH` module-level constant and inlined the 80-character limit, since it's only used in one place.
- Dropped the long explanatory comments — the warning message itself already describes the problem.

I kept it as a separate commit (no force-push this time) so the changes since your review are easy to follow.

freakboy3742 (approved):
👍 Thanks for those updates!

PR #2873, opened 2026-06-05
Warn when project description exceeds 80 characters

### LINKED ISSUE #2864: Long project descriptions cause icons to be blank on MSI installers

wahajahmed010:
When an app's `description` exceeds 80 characters, MSI installers can experience icon corruption because the Windows shortcut description field is limited to 256 bytes — the overflow truncates the prefix of the icon path. This warning directs users to the `long_description` field for longer text, which is the intended use.

Changes:
- Added a console warning when `description` > 80 characters in `parse_config()`
- Warning includes the app name, character count, and a suggestion to use `long_description`
- The warning fires once per app during configuration parsing

Assisted-by: Claude Opus 4.5 (AI coding assistant - automated)

## PR Checklist:
Closes #2864

### DIFF
--- a/changes/2873.feature.md
+++ b/changes/2873.feature.md
@@ -1,0 +1,1 @@
+Added a warning when `description` exceeds 80 characters, prompting users to use `long_description` for longer text. Long descriptions can cause icon corruption in MSI installers.
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -1122,6 +1122,22 @@ any compatibility problems, and then add the compatibility declaration.
                     msg=f"Configuration for '{app_name}'",
                 )
 
+                # Warn if the app description is longer than 80 characters.
+                # Long descriptions can cause MSI shortcut icon corruption
+                # (the description field is limited to 256 chars in Windows
+                # shortcuts, and the description length overflows into the
+                # icon path prefix). Users should use the long_description
+                # field for longer text.
+                if len(self.apps[app_name].description) > 80:
+                    self.console.warning(
+                        f"The description for '{app_name}' is "
+                        f"{len(self.apps[app_name].description)} characters. "
+                        "Briefcase descriptions should be no more than 80 "
+                        "characters. Long descriptions may cause icon display "
+                        "issues on some platforms (e.g., MSI installers). "
+                        "Use the `long_description` field for longer text."
+                    )
+
         except OSError as e:
             raise BriefcaseConfigError(
                 f"""\
--- a/tests/commands/base/test_parse_config.py
+++ b/tests/commands/base/test_parse_config.py
@@ -3,6 +3,13 @@ from packaging.version import Version
 
 from briefcase.exceptions import BriefcaseConfigError
 
+from unittest.mock import MagicMock
+
+import pytest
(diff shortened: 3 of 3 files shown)


freakboy3742:
Thanks - but:
1. You've opened a pull request on an issue where there is an active PR in flight
2. You haven't followed our contribution process. 

Closing this PR on that basis.

PR #2877, opened 2026-06-12
Adopt os-log based logging.

freakboy3742:
macOS 26 changed the handling of default logging levels, so running a macOS app through Briefcase under macOS 26 results in the stdout content being displayed as `<private>`.

To work around this, we need to use an updated version of std-nslog which uses os_log rather than NSLog. However, this requires a binary module, which means the logger can't be loaded until app_packages has been added to the site path. This binary module requires a slightly different predicate pattern to extract log patterns.

However - as of Python 3.14, the functionality of std-nslog is baked into Python by default. As long as you're using 3.14.6+, you can get all of the functionality of std-nslog without importing the package. The predicate pattern here allows for native logging as well, and includes a cookiecutter extension to allow easy evaluation of whether the native shim can be used.

CI "Build app" tasks will fail on macOS until the templates are updated; the landing sequence will be to land the templates, then land this PR. See:
* beeware/briefcase-iOS-Xcode-template#75
* beeware/briefcase-macOS-Xcode-template#122

## PR Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,8 +50,10 @@ logs/
 .specify/templates
 .specify/extensions.yml
 .specify/extensions
+.specify/
 .specify/init-options.json
 .specify/integration.json
 .specify/integrations
 .specify/workflows
 .specify/workflow.yml
+specs
--- a/.gitignore
+++ b/.gitignore
@@ -50,7 +50,7 @@ logs/
 .specify/templates
 .specify/extensions.yml
 .specify/extensions
-.specify/
+.specify/feature.json
 .specify/init-options.json
 .specify/integration.json
 .specify/integrations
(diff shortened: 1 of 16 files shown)


.gitignore
```diff
@@ -50,8 +50,10 @@ logs/
 .specify/templates
 .specify/extensions.yml
 .specify/extensions
+.specify/
```
mhsmith: Does this mean all the other .specify entries can be removed?

freakboy3742: Not sure what happened here - the entry should be `./specify/feature.json`. I've pushed an update.

freakboy3742:
@mhsmith I'm going to merge this without a finalised review, because landing the iOS template means the current mainline Briefcase isn't able to view iOS logs, which is breaking Toga's CI (which will, in turn, cause problems with dependabot tomorrow). If you've got any deeper review comments, I'd still appreciate the review.

PR #2886, opened 2026-06-16
Add auto-resume feature

### LINKED ISSUE #2883: Simplify resume workflow for macOS notarisation

macOS apps must be notarized; however, the macOS notarisation workflow can take hours to complete. 

Briefcase already accommodates this, providing a `--resume` option to `briefcase package`. When notarisation starts, the command to resume notarisation is output to the log. 

This is very convenient for manual invocation, but not for invocation in automated environments, as it requires a different set of command line arguments to resume as it does to start the task - and it requires the user to store and remember the resume identifier.

Briefcase should be able to automatically resume a previous notarisation run without any additional command line handling.

When a notarisation run starts, Briefcase should output a metadata file into the `dist` directory, matching the name of the file being notarized (e.g., if you're notarising `Hello World-1.2.3.dmg`, then Briefcase should output `dist/Hello World-1.2.3.dmg.notarization-request`. This should be a TOML file that contains 2 keys: the signing identity, and the resume identifier.

On subsequent runs of `briefcase package`, Briefcase should look for the existence of a `dist/Hello World-1.2.3.dmg.notarization-request` file. If this file exists, then this is a resumed package run; check that the identity listed in the file matches the identity provided at the command line; if it does, run as if `--resume` had been provided at the command line. If the identity doesn't match, output an error indicating the problem, and suggest that deleting the `notarization-request` will be required if the user wants to start a new notarisation request.

When notarisation succeeds, the `.notarization-request` file should be deleted from the `dist` folder.

Provide other ways to pass in the `--resume` flag, such as an environment variable. Using a marker file allows for a completely hands-off resumption. 

### Additional context

With this change the `--resume` flag can be deprecated; the existing message suggesting the use of `--resume` can be updated to suggest re-running `briefcase package`.

Any existing documentation referencing the notarisation process and the `--resume` flag should also be updated.

kattni:
`briefcase package` now automatically resumes macOS notarisation when run after the notarisation process is interrupted.

Fixes #2883.

## PR Checklist:
 
 Assisted-by: Big Pickle, Claude Code


src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1346,6 +1426,42 @@ def package_app(
...
+
+            # Identity
+            if identity is not None and marker_data["identity"] != identity:
+                raise BriefcaseCommandError(
+                    f"Notarization request marker identity "
+                    f"{marker_data['identity']!r} does not match "
+                    f"the specified identity {identity!r}."
+                )
+            if identity is None:
+                identity = marker_data["identity"]
```
freakboy3742: A minor structural suggestion, so that it's easier to follow which branches will be followed when:
```suggestion
            if identity:
                if marker_data["identity"] != identity:
                    raise BriefcaseCommandError(
                        f"Notarization request marker identity "
                        f"{marker_data['identity']!r} does not match "
                        f"the specified identity {identity!r}."
                    )
            else:
                # User didn't specify an identity - use the identity defined
                # in the marker file.
                identity = marker_data["identity"]
```

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1346,6 +1426,42 @@ def package_app(
...
+                    )
+                if marker_data["installer_identity"] != installer_identity:
+                    raise BriefcaseCommandError(
+                        "Notarization request marker installer identity "
+                        f"{marker_data['installer_identity']!r} does not "
+                        f"match the specified installer identity "
+                        f"{installer_identity!r}."
+                    )
+            if installer_identity is None and "installer_identity" in marker_data:
+                installer_identity = marker_data["installer_identity"]
```
freakboy3742: Similar to the above - a minor organisational suggestion:
```suggestion
            if "installer_identity" in marker_data:
                if installer_identity:
                    if "marker_data["installer_identity"] != installer_identity:
                        raise BriefcaseCommandError(
                            "Notarization request marker installer identity "
                            f"{marker_data['installer_identity']!r} does not "
                            f"match the specified installer identity "
                            f"{installer_identity!r}."
                        )
                else:
                    # User didn't specify an installer identity - use the
                    # installer_identity from the marker file.
                    installer_identity = marker_data["installer_identity"]                
            elif installer_identity is not None:
                raise BriefcaseCommandError(
                    "Notarization request marker does not contain an "
                    "installer identity, but --installer-identity "
                    f"({installer_identity}) was specified."
                )
```

freakboy3742 (changes requested):
The core implementation looks great - a couple of minor suggestions about how the logic for resumption is ordered, but otherwise, this looks really solid.

The area that needs more work is tests. I can see that many of the existing tests have been refactored to make better use of parameterization - but that is at the cost of increased complexity *in* the test, which isn't always a good thing. I suspect we may want to revert those changes, and only add new tests (and test clauses) for this new functionality.

That leads into the other issue with the testing - as is common with a lot of AI-generated testing, when a test case already exists that exercise a given  execution path, it defaults to creating a new test rather than integrating new checks into the existing test. For example, there is an existing test that exercises a "successful notarization" path, there's no need to have a separate test that validates a marker file is created after a successful notarisation.

The extra tests for *failure* cases are generally going to be good - as those are cases where there wasn't previously a branch of code being exercised. 

My suggestion would be to revert the two test files to their original state, then get a coverage report showing which branches aren't executed. Add an assertion to each of the `test_notarize.py` and `test_resume_notarization.py` tests that are notarising that validates that a marker file does or doesn't exist, as required. Then add new tests for each of the failure modes, until you're back to 100% coverage.

tests/platforms/macOS/app/package/test_resume_notarization.py
```diff
@@ -424,6 +431,484 @@ def test_resume_notarize_pkg(
...
+
+    # The marker exists, but isn't valid TOML.
+    create_file(marker_path, "[not-valid-toml")
+    with pytest.raises(BriefcaseCommandError, match="is malformed"):
+        package_command.read_notarization_request(first_app_with_binaries)
+
+    # The marker is valid TOML, but is missing a required key.
+    create_file(marker_path, 'identity = "CAFEBEEF"\n')
+    with pytest.raises(BriefcaseCommandError, match="is missing required key"):
+        package_command.read_notarization_request(first_app_with_binaries)
```
freakboy3742: This is a good set of test assertions, but it should be 3 independent tests.

tests/platforms/macOS/app/package/test_resume_notarization.py
```diff
@@ -424,6 +431,484 @@ def test_resume_notarize_pkg(
...
+
+
+def test_resume_notarization_marker_conflict(
+    package_command,
+    first_app_with_binaries,
+    sekrit_identity,
+    sekrit_installer_identity,
+    tmp_path,
+):
```
freakboy3742, line 869: Similarly - this is a good set of test assertions, but it should be 3 independent tests - one for each problem case.

freakboy3742 (changes requested):
Looks good - one small cleanup needed to break some of the new tests up into smaller units.

freakboy3742 (approved):
👍 Awesome - nice work!

PR #2887, opened 2026-06-18
Add resume handling cross-platform

kattni:
`can_resume` returns `False` on all backends, except for the macOS override when applicable.

## PR Checklist:
 
 Assisted-by: Claude Code, Big Pickle

### DIFF
--- a/src/briefcase/commands/package.py
+++ b/src/briefcase/commands/package.py
@@ -57,6 +57,27 @@ class PackageCommand(BaseCommand):
             # Ensure the dist folder exists.
             self.dist_path.mkdir(exist_ok=True)
 
+    def can_resume(self, app, **options):
+        """Determine if we can resume a packaging operation.
+
+        The default backend does not support resume; subclasses that support
+        resumable operations (e.g., notarization) should override this method.
+
+        :param app: The app being packaged
+        :returns: ``True`` if packaging can be resumed; ``False`` otherwise.
+        """
+        return False
+
+    def verify_resume_app(self, app, **options):
+        """Verify the app for a resumed packaging operation.
+
+        Called instead of :meth:`verify_app` when ``can_resume()`` returns
+        ``True``. The default implementation delegates to :meth:`verify_app`.
+
+        :param app: app configuration
+        """
+        self.verify_app(app)
+
     def package_app(self, app: FinalizedAppConfig, **options):
         """Package an application.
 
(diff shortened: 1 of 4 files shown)


src/briefcase/commands/package.py
```diff
@@ -57,6 +57,27 @@ def clean_dist_folder(self, app, **options):
...
             self.dist_path.mkdir(exist_ok=True)
 
+    def can_resume(self, app, **options):
+        """Determine if we can resume a packaging operation.
+
+        The default backend does not support resume; subclasses that support
+        resumable operations (e.g., notarization) should override this method.
+
+        :param app: The app being packaged
+        :returns: ``True`` if packaging can be resumed; ``False`` otherwise.
```
freakboy3742: Docstrings need to be updated to Markdown format.

src/briefcase/commands/package.py
```diff
@@ -57,6 +57,27 @@ def clean_dist_folder(self, app, **options):
...
+
+    def verify_resume_app(self, app, **options):
+        """Verify the app for a resumed packaging operation.
+
+        Called instead of `verify_app` when `can_resume()` returns
+        `True`. The default implementation delegates to `verify_app`.
+
+        :param app: app configuration
+        """
+        self.verify_app(app)  # pragma: no cover
```
freakboy3742: Why is this a no-cover? That suggests we're not exercising this workflow in tests.

freakboy3742 (changes requested):
This looks great. One question inline about a `no cover`.

I also have some minor reservations about how `verify_app()` is split up; but I think that's more evidence of a deeper problem with what is being verified and how. In this case:

1. Verification for `package` is inheriting verification checks from the base command that aren't actually needed to verify packaging
2. The population of the packaging format isn't really a `verify` action.

However, those problems don't need to stand in the way of this PR; point (2) in particular will need to be addressed in the context of #2706

src/briefcase/commands/package.py
freakboy3742: From conversation in person - this is because there's an overridden implementation on the macOS backend, which is the only backend that allows for resumption - so there's no way to exercise this line of code.

I'd suggest the better approach would be to make the macOS implementation of `verify_resume_app` the implementation here; that way we'll get coverage *and* satisfy the functional requirement.

I also think we should be retaining the `verify_required_python` check as part of `verify_resume_app` - it's *mostly* redundant, but it's also a low effort check, so we might as well retain it.

freakboy3742 (approved):
Awesome - thanks for that fix!

PR #2890, opened 2026-06-19
Adds `--no-wait` to macOS notarization process

### LINKED ISSUE #2888: Allow opt-out of waiting for macOS notarisation

macOS notarization is a process that can take between minutes and hours. When notarization runs in CI, it ties up a runner until the task times out. It would be preferable to be able to opt-out of the waiting behavior so that CI resources (especially limited macOS runners) can be conserved, rather than be spent idling waiting for a notarization response from Apple.

Add a `--no-wait` option to the macOS `package` command, disabled by default (i.e., "wait=True" is the default value.)

When `--no-wait` is specified:
1. An initial notarization request will *submit* the notarization, and write the notarization request marker file, but will *not* attempt to finalise the notarization. The user should be shown an "info" level message that is very similar to the current warning, but indicates the next step is to re-run `briefcase package` to finalise.
2. A resumed notarization request will perform exactly *one* attempt to retrieve the notarization status. If the notarization attempt does not return a canonical response (either success or failure), the command should raise an error with a "Apple has not completed notarising the app; try again later" message, rather than sleeping and trying again. The wait bar "waiting for notarization acceptance..." message should also be updated to "Checking notarization status... " in the `--no-wait` case, to reflect the "one-shot" nature of the attempt.

* Status quo - let notarization soak up CI resources
* Raise an *error* on the initial notarization attempt. That would allow/require notarization to be a "re-run the same task until it succeeds" operation. Raising a success message means the *finalization* portion can be the only part that is re-run.

### Additional context

The argument and any related variables should be expressed as `wait`, `wait_for_finalization` etc, rather than the double-negative of `no_wait=True`

The `--no-wait` option should also be documented in the macOS guide as an option on the `package` command.

kattni:
The macOS `package` command now accepts `--no-wait`, which submits the app for notarisation, but does not wait for completion.

Fixes #2888

## PR Checklist:
 
 Assisted-by: Claude Code Sonnet 4.6, Opus 4.8

### DIFF
--- a/src/briefcase/platforms/macOS/__init__.py
+++ b/src/briefcase/platforms/macOS/__init__.py
@@ -1122,7 +1122,7 @@ class macOSPackageMixin(macOSSigningMixin):
         :param identity: The code signing used to notarize the app.
         :param installer_identity: The signing identity to use when signing the
             installer. Optional unless the packaging format is ``pkg``.
-        :param wait: If ``True``, wait for notarization to complete. If ``False``,
+        :param wait: If `True`, wait for notarization to complete. If `False`,
             submit the app and return without finalizing.
         """
         # Determine the arguments that would be needed to reproduce this notarization
@@ -1143,8 +1143,9 @@ class macOSPackageMixin(macOSSigningMixin):
 
         if not wait:
             self.console.info(
-                "The submission has been sent. You'll need to run briefcase "
-                "package at some point in the future to finalize the notarization."
+                f"{app.formal_name} as been submitted to Apple for notarization. "
+                "You'll need to run briefcase package at some point in the future "
+                "to finalize the notarization."
             )
         else:
             self.console.warning("""
@@ -1161,7 +1162,6 @@ resume it.
                 app,
                 identity=notarization_identity,
                 submission_id=submission_id,
-                wait=wait,
             )
 
     def submit_notarization(self, app, identity: SigningIdentity) -> str:
@@ -1338,7 +1338,7 @@ password:
         :param app: The app to notarize.
         :param identity: The code signing identity to use.
         :param submission_id: The submission ID of the notarization task to finalize.
-        :param wait: If ``True``, poll until notarization completes. If ``False``,
+        :param wait: If `True`, poll until notarization completes. If `False`,
             check the status once and raise an error if it is not yet complete.
         """
         try:
@@ -1395,8 +1395,9 @@ password:
                                     "Apple has not completed notarising the app; "
                                     "try again later"
                                 ) from e
-                            # Try again in 10 seconds.
-                            time.sleep(10)
+                            else:
+                                # Try again in 10 seconds.
+                                time.sleep(10)
                         else:
                             self.tools.subprocess.output_error(e)
                             raise BriefcaseCommandError(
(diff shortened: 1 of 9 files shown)


src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1113,6 +1122,8 @@ def notarize(
         :param identity: The code signing used to notarize the app.
         :param installer_identity: The signing identity to use when signing the
             installer. Optional unless the packaging format is ``pkg``.
+        :param wait: If ``True``, wait for notarization to complete. If ``False``,
```
freakboy3742:
```suggestion
        :param wait: If `True`, wait for notarization to complete. If `False`,
```

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1319,9 +1338,16 @@ def finalize_notarization(
         :param app: The app to notarize.
         :param identity: The code signing identity to use.
         :param submission_id: The submission ID of the notarization task to finalize.
+        :param wait: If ``True``, poll until notarization completes. If ``False``,
```
freakboy3742:
```suggestion
        :param wait: If `True`, poll until notarization completes. If `False`,
```

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1140,11 +1157,12 @@ def notarize(
...
-        self.finalize_notarization(
-            app,
-            identity=notarization_identity,
-            submission_id=submission_id,
-        )
+            self.finalize_notarization(
+                app,
+                identity=notarization_identity,
+                submission_id=submission_id,
+                wait=wait,
```
freakboy3742: This wait argument is redundant; we can fall back to the default.
```suggestion
```

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1130,7 +1141,13 @@ def notarize(
             installer_identity=installer_identity,
         )
 
-        self.console.warning("""
+        if not wait:
+            self.console.info(
+                "The submission has been sent. You'll need to run briefcase "
+                "package at some point in the future to finalize the notarization."
```
freakboy3742:
```suggestion
                f"{app.formal_name} as been submitted to Apple for notarization. "
                "You'll need to run briefcase package at some point in the future "
                "to finalize the notarization."
```

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1363,7 +1389,13 @@ def finalize_notarization(
...
                             # validated that it's a valid submission ID, so that means
-                            # notarization isn't complete yet. Try again in 10 seconds.
+                            # notarization isn't complete yet.
+                            if not wait:
+                                raise BriefcaseCommandError(
+                                    "Apple has not completed notarising the app; "
+                                    "try again later"
+                                ) from e
+                            # Try again in 10 seconds.
                             time.sleep(10)
```
freakboy3742: Even though it's functionally equivalent, using an else here makes it clear that there are two functional options.
```suggestion
                            else:
                                # Try again in 10 seconds.
                                time.sleep(10)
```

freakboy3742 (approved):
A couple of cosmetic docstring format fixes, and one rewording of an error message - but otherwise this is great! Thanks!

PR #2891, opened 2026-06-21
Reformat iOS `xcode.py` warning to use warning_banner

### LINKED ISSUE #2559: Add helper for formatting long warning messages

Yousuf24100286:
Fixes: #2559

**Description:** 

- `iOSXcodePassiveMixin.distribution_path()`: 
  - Replace hand formatted asterisk box with warning utility
  - raise error with empty message to prevent re-printing of the message

The hardcoded asterisk-box formatting was inconsistent with the rest of the codebase after #2563 introduced the `warning_banner()` helper.

## PR Checklist:

Assisted-by: Claude Opus 4.8

### DIFF
--- a/changes/+2559-xcode.misc.md
+++ b/changes/+2559-xcode.misc.md
@@ -1,0 +1,1 @@
+The warning banner helper was used in the iOS Xcode backend.
--- a/src/briefcase/console.py
+++ b/src/briefcase/console.py
@@ -310,13 +310,15 @@ class Console:
 
         return output_lines
 
-    def warning_banner(
-        self,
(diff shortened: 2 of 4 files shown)


mhsmith (commented):
The warning was previously attached to the exception, but now it's printed unconditionally. This prevents the warning from being suppressed by an exception handler.

Instead, the warning should be formatted and then attached to the exception. If there isn't already a public function to do the formatting without printing the warning, then one should be created.

mhsmith (approved):
Thanks!

PR #2901, opened 2026-06-22
Add man_page config option to override template-generated man pages on Linux

### LINKED ISSUE #2545: Actual man pages

Can't actually write man pages

Take in troff, ship man page

* Make a wrapping script that temporarily overrides the long description
* Run a side process that watches for briefcase to write out the man page and rewrite it
* Munge the resulting packages to insert the correct man page

### Additional context

This was discussed in discord

Yousuf24100286:
Fixes: #2545

## Description

added `man_page` option to `pyproject.toml` app configuration.

- added `man_page: str | None = None` to `AppConfig` and `DraftAppConfig`.
- modified `LinuxSystemBuildCommand.build_app()` to resolve `man_page` relative to the project root and copy it in place of the template default.
- Behavior
  - If `man_page` is set:
    - file doesn't exists, error is raised
    - else troff file replaces template generated man page
  - If `man_page` is not set
    - existing behavior is unchanged
- Example:
  ```toml
  [tool.briefcase.app.myapp]
  man_page = "docs/myapp.1"
  ```

## PR Checklist:

Assisted-by: Claude Opus 4.6

### DIFF
--- a/changes/2545.feature.md
+++ b/changes/2545.feature.md
@@ -1,0 +1,1 @@
+Linux system packages can now use a custom man page by setting ``man_page`` in ``pyproject.toml``.
--- a/changes/2545.feature.md
+++ b/changes/2545.feature.md
@@ -1,1 +1,1 @@
-Linux system packages can now use a custom man page by setting ``man_page`` in ``pyproject.toml``.
+Linux system packages can now use a custom man page by setting `man_page` in `pyproject.toml`.
--- a/src/briefcase/config.py
+++ b/src/briefcase/config.py
@@ -471,6 +471,7 @@ class AppConfig(BaseConfig):
     test_requires: list[str] | None
     supported: bool
     long_description: str | None
+    man_page: str | None
     license: dict | None
     license_files: list[str]
     console_app: bool
@@ -606,6 +607,7 @@ class DraftAppConfig(AppConfig):
         test_requires: list[str] | None = None,
         supported: bool = True,
         long_description: str | None = None,
+        man_page: str | None = None,
         console_app: bool = False,
         requirement_installer_args: list[str] | None = None,
         external_package_path: str | None = None,
@@ -635,6 +637,7 @@ class DraftAppConfig(AppConfig):
         self.test_requires = test_requires
         self.supported = supported
         self.long_description = long_description
+        self.man_page = man_page
         self.license = license
         self.license_files = [] if license_files is None else license_files
         self.console_app = console_app
--- a/src/briefcase/config.py
+++ b/src/briefcase/config.py
@@ -471,7 +471,6 @@ class AppConfig(BaseConfig):
     test_requires: list[str] | None
     supported: bool
     long_description: str | None
-    man_page: str | None
     license: dict | None
     license_files: list[str]
     console_app: bool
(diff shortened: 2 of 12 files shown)


src/briefcase/config.py
```diff
@@ -471,6 +471,7 @@ class AppConfig(BaseConfig):
     test_requires: list[str] | None
     supported: bool
     long_description: str | None
+    man_page: str | None
```
freakboy3742: As a heads up - `man_page` is a Linux-specific feature that doesn't require any special parsing. As a result, we wouldn't add that to the base AppConfig - we'd add it to the `LinuxSystemAppConfig` that is specific to the `system` platform handling.

freakboy3742:
The CI failure you're currently seeing hasn't been caused by anything you've done - there's been an update to one of our dependencies over the last hour, which is breaking tests everywhere. We're trying to get to the bottom of things; apologies for the inconvenience.

src/briefcase/config.py
Yousuf24100286: Thank You for the feedback.

Implementation updated. 
- `man_page` is no longer in the base `AppConfig/DraftAppConfig`.
-  now read as an optional platform-specific config key using `getattr(app, "man_page", None)` in `LinuxSystemBuildCommand.build_app()`

freakboy3742:
> The CI failure you're currently seeing hasn't been caused by anything you've done - there's been an update to one of our dependencies over the last hour, which is breaking tests everywhere. We're trying to get to the bottom of things; apologies for the inconvenience.

The fix for this problem is #2902; hopefully this will land in main very shortly. Once it does, you should be able to merge with main to resolve the issue you're seeing here.

changes/2545.feature.md
```diff
@@ -0,0 +1 @@
+Linux system packages can now use a custom man page by setting ``man_page`` in ``pyproject.toml``.
```
freakboy3742: We use Markdown format for docs, not ReStructured Text:
```suggestion
Linux system packages can now use a custom man page by setting `man_page` in `pyproject.toml`.
```

src/briefcase/platforms/linux/system.py
```diff
@@ -904,19 +904,27 @@ def build_app(self, app: FinalizedAppConfig, **kwargs):
         man_folder.mkdir(parents=True, exist_ok=True)
 
         with self.console.wait_bar("Installing man page..."):
-            manpage_source = self.bundle_path(app) / f"{app.app_name}.1"
-            if manpage_source.is_file():
-                with manpage_source.open(encoding="utf-8") as infile:
-                    outfile = gzip.GzipFile(
-                        man_folder / f"{app.app_name}.1.gz", mode="wb", mtime=0
+            man_page = getattr(app, "man_page", None)
+            if man_page:  # pragma: no-cover-if-is-windows
```
freakboy3742: I'm not clear why a no-cover is required on *only* this branch. There wasn't any no-cover required previously, and no-cover isn't required on the "default man file" path; why is the "non-default man file" path any different to that one?

freakboy3742 (changes requested):
The general structure of this looks good; a couple of minor tweaks and questions inline.

One additional detail that is required - this new feature needs to be documented. The new `man_page` option needs to be listed in the [configuration reference for the System packaging backend](https://briefcase.beeware.org/en/stable/reference/platforms/linux/system/#application-configuration).

src/briefcase/platforms/linux/system.py
Yousuf24100286: windows ci tests were failing, so added this because it was the first solution that came to mind.

removed it now.

freakboy3742 (approved):
Awesome - thanks for those updates!

Issue #2561, opened 2025-11-12
Automatical postinstall script for macos


hRtWzFe:
### What is the problem or limitation you are having?

I am now trying to use postinstall script in my pkg file on macos. It is present in cookiecutter template, but, nor pkgbuild or productbuild doesn't really do anything with it. Didn't find any workarounds or solutions for this in docs, and I don't want to spin xcode project in here.

### Describe the solution you'd like

I will like to have option to include custom payload for installer-only, inside resources, that will be packed alongside with .html file and license. 
My postinstall depends on pre-compiled executable (Yes, this is intended and fully needed for my usecase. Also, including some additional data in installer could be useful sometimes), which I would like to invoke from postinstall script. 
Is this even possible?

### Describe alternatives you've considered

I've checked the docs for some configuration options, checked older version of macos app template..

haikesan:
In briefcase/platforms/macOS/__init__.py there is macOSPackageMixin, which is writing this:
with self.console.wait_bar("Writing component manifest..."):
    with components_plist_path.open("wb") as components_plist:
        plistlib.dump(
            [
                {
                    "BundleHasStrictIdentifier": True,
                    "BundleIsRelocatable": False,
                    "BundleIsVersionChecked": True,
                    "BundleOverwriteAction": "upgrade",
                    "RootRelativeBundlePath": self.package_path(app).name,
                }
            ],
            components_plist,
        )
I could manually edit this after project generation to include BundlePostInstallScriptPath, but this solution doesn't look like pretty one for me..

haikesan:
Also, found interesting comment:
        # Console apps are installed in /Library/Formal Name, and include the
        # post-install scripts. Normal apps are installed in /Applications, and don't
        # include the scripts.

But, some "normal" apps also want to have postinstall.. 
For some reason, resources are also not updated, maybe reason is in some caches?
Also, maybe everything here is not intended and I should write plugin for myself?

freakboy3742:
Thanks for the request. 

Firstly - you're not missing anything. While some of the infrastructure for pre/post install scripts is present in the installer, it isn't something we currently support.

However, we've recently added support for pre/post install scripts on Windows MSI installers, and we're definitely interested in extending that to PKG installers for macOS. If you want to make a proposal (and/or turn that proposal into a pull request), that would be a very helpful way to contribute to Briefcase as a project.

haikesan:
My really DIRTY solution (later will think of how to clean it up and to contribute accordingly):

[...]

PR #2903, opened 2026-06-23
Add support for post-install scripts and installer resources on macOS PKG installers

### LINKED ISSUE #2865: Add support for post install scripts on macOS PKG installers

Originally reported by @haikesan as #2561.

MSI installers currently support the inclusion of post install scripts.

PKG installers also have the facility for post install scripts; however, Briefcase will not currently make any use of them.

It should be possible to add:
```
post_install_script = "scripts/post_install.sh"
```
to a macOS project configuration, and have that project configuration include the named script as part of a PKG installation.

n/a

### Additional context

There's no analog of a pre-uninstall script for PKG, so the user should be warned if they define a pre-uninstall script for a PKG installer. They should also be warned if they produce a DMG or ZIP installer and have pre-uninstall or post-install scripts defined.

haikesan:
PKG installers have always had the facility to run a post-install script, but Briefcase didn't expose it. This adds a `post_install_script` setting for macOS, matching the existing Windows MSI behaviour:

```toml
[tool.briefcase.app.myapp.macOS]
post_install_script = "scripts/post_install.sh"
```
  
The configured script is assembled into a dedicated scripts directory and passed to `pkgbuild --scripts`. The directory is rebuilt from scratch on every `package` run, so the template's own files are never mutated and re-packaging can't duplicate content.
  
As called out in the issue, scripts that an installer format can't use are now reported:
- a `pre_uninstall_script` on a PKG installer (there's no PKG equivalent) → warning;
- a `post_install_script` or `pre_uninstall_script` on a DMG/ZIP installer → warning;
- a `post_install_script` that doesn't exist on disk → error.

Fixes #2865, reported by me in #2561.

## PR Checklist:

Assisted-by: Claude Code (Opus 4.8)
Disclosure: Tests and documentation in this PR were AI-generated.

### DIFF
--- a/changes/2561.feature.md
+++ b/changes/2561.feature.md
@@ -1,0 +1,1 @@
+macOS PKG installers can now include additional installer resources, configured with the ``installer_resources`` setting.
--- a/changes/2561.feature.md
+++ b/changes/2561.feature.md
@@ -1,1 +1,1 @@
-macOS PKG installers can now include additional installer resources, configured with the ``installer_resources`` setting.
+macOS PKG installers can now include additional installer resources, configured with the `installer_resources` setting.
--- a/changes/2865.feature.md
+++ b/changes/2865.feature.md
@@ -1,1 +1,1 @@
-macOS PKG installers can now run a post-install script, configured with the ``post_install_script`` setting.
+macOS PKG installers can now run a post-install script, configured with the `post_install_script` setting.
--- a/changes/2865.feature.md
+++ b/changes/2865.feature.md
@@ -1,0 +1,1 @@
+macOS PKG installers can now run a post-install script, configured with the ``post_install_script`` setting.
--- a/docs/en/reference/platforms/macOS/index.md
+++ b/docs/en/reference/platforms/macOS/index.md
@@ -167,7 +167,7 @@ The minimum macOS version that the app will support. This controls the value of 
 /// note | Only used for PKG packaging
 ///
 
-A path, relative to the project root, to a shell script that will be executed during installation, after the installer content has been unpacked. The script must be a valid macOS installer `postinstall` script (i.e., a script with an appropriate shebang line such as `#!/bin/sh`).
+A path, relative to the project root, to a shell script that will be executed during installation, after the installer content has been unpacked. The script must be a valid macOS shell script file with an appropriate shebang line (e.g., `#!/bin/sh`).
 
 For console apps, Briefcase provides a post-install script that creates a symbolic link to the app on the `PATH`; your script will be run *after* that link has been created.
 
--- a/docs/en/reference/platforms/macOS/index.md
+++ b/docs/en/reference/platforms/macOS/index.md
@@ -167,7 +167,7 @@ A path, relative to the project root, to a directory of additional resources to 
 
 These resources are part of the installer itself, they are used while the installer is running, and are not installed onto the target machine.
 
-DMG and ZIP "installers" do not carry resources, if you define `installer_resources` while using one of those formats, the setting will be ignored.
+DMG and ZIP "installers" do not carry resources. If you define `installer_resources` while using one of those formats, the setting will be ignored with a warning.
 
 ### `min_os_version`
 
@@ -182,7 +182,7 @@ A path, relative to the project root, to a shell script that will be executed du
 
 For console apps, Briefcase provides a post-install script that creates a symbolic link to the app on the `PATH`; your script will be run *after* that link has been created.
 
-DMG and ZIP "installers" cannot run scripts; if you define a `post_install_script` while using one of those formats, the setting will be ignored. There is no PKG analog for a pre-uninstall script.
+DMG and ZIP "installers" cannot run scripts. if you define a `post_install_script` while using one of those formats, the setting will be ignored with a warning. There is no PKG analog for a pre-uninstall script.
 
 ### `universal_build`
 
--- a/docs/en/reference/platforms/macOS/index.md
+++ b/docs/en/reference/platforms/macOS/index.md
@@ -158,6 +158,17 @@ will result in an `Info.plist` declaration of:
 
 Any Boolean or string value can be used for an `Info.plist` value.
 
+### `installer_resources`
+
+/// note | Only used for PKG packaging
+///
+
+A path, relative to the project root, to a directory of additional resources to include in the PKG installer. The contents of this directory are added alongside Briefcase's own installer resources (such as the welcome screen and license), and can be referenced by the installer's distribution definition.
+
+These resources are part of the installer itself, they are used while the installer is running, and are not installed onto the target machine.
+
+DMG and ZIP "installers" do not carry resources, if you define `installer_resources` while using one of those formats, the setting will be ignored.
+
 ### `min_os_version`
 
 The minimum macOS version that the app will support. This controls the value of `MACOSX_DEPLOYMENT_TARGET` used when building the app.
--- a/docs/en/reference/platforms/macOS/index.md
+++ b/docs/en/reference/platforms/macOS/index.md
@@ -162,6 +162,17 @@ Any Boolean or string value can be used for an `Info.plist` value.
 
 The minimum macOS version that the app will support. This controls the value of `MACOSX_DEPLOYMENT_TARGET` used when building the app.
 
+### `post_install_script`
+
+/// note | Only used for PKG packaging
+///
+
+A path, relative to the project root, to a shell script that will be executed during installation, after the installer content has been unpacked. The script must be a valid macOS installer `postinstall` script (i.e., a script with an appropriate shebang line such as `#!/bin/sh`).
+
+For console apps, Briefcase provides a post-install script that creates a symbolic link to the app on the `PATH`; your script will be run *after* that link has been created.
+
+DMG and ZIP "installers" cannot run scripts; if you define a `post_install_script` while using one of those formats, the setting will be ignored. There is no PKG analog for a pre-uninstall script.
+
 ### `universal_build`
 
 A Boolean, indicating whether Briefcase should build a universal app (i.e, an app that can target both x86_64 and ARM64). Defaults to `true`; if `false`, the binary will only be executable on the host platform on which it was built - i.e., if you build on an x86_64 machine, you will produce an x86_65 binary; if you build on an ARM64 machine, you will produce an ARM64 binary.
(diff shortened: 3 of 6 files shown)


docs/en/reference/platforms/macOS/index.md
```diff
@@ -162,6 +162,17 @@ Any Boolean or string value can be used for an `Info.plist` value.
 
 The minimum macOS version that the app will support. This controls the value of `MACOSX_DEPLOYMENT_TARGET` used when building the app.
 
+### `post_install_script`
+
+/// note | Only used for PKG packaging
+///
+
+A path, relative to the project root, to a shell script that will be executed during installation, after the installer content has been unpacked. The script must be a valid macOS installer `postinstall` script (i.e., a script with an appropriate shebang line such as `#!/bin/sh`).
```
freakboy3742: It's not immediately clear what a "valid `postinstall` script" would be:
```suggestion
A path, relative to the project root, to a shell script that will be executed during installation, after the installer content has been unpacked. The script must be a valid macOS shell script file with an appropriate shebang line (e.g., `#!/bin/sh`).
```

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1646,6 +1681,41 @@ def finalize_package_zip(self, app: FinalizedAppConfig):
         ):
             self.ditto_archive(self.package_path(app), self.distribution_path(app))
 
+    def build_pkg_scripts(self, app: FinalizedAppConfig, installer_path: Path) -> Path:
+        """Assemble the scripts directory passed to ``pkgbuild --scripts``.
```
freakboy3742: We use Markdown format, not ReStructuredText
```suggestion
        """Assemble the scripts directory passed to `pkgbuild --scripts`.
```

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1646,6 +1681,41 @@ def finalize_package_zip(self, app: FinalizedAppConfig):
...
+
+        if scripts_path.exists():
+            self.tools.shutil.rmtree(scripts_path)
+        scripts_path.mkdir(parents=True)
+
+        post_install = (self.base_path / app.post_install_script).read_text(
+            encoding="utf-8"
+        )
+        if app.console_app and template_postinstall.is_file():
+            # Append the configured script to the template's postinstall.
```
freakboy3742: Appending isn't a good approach here, for multiple reasons:

Firstly, user content will be appended on every build - so if you re-run the PKG installer, the post-install script will be injected multiple times.

Secondly, the first line of the user-supplied script is documented as being a "valid script with a shebang" - which means you'll end up with a shebang in the middle of the post-install script file

Lastly, the existing `postinstall` script has a shebang of `#!/bin/sh`, which means the user's script will be run in that context. 

I can think of a couple of possible approaches:

1. Copy in the user-provided script under a *different*, but fixed name (`_postinstall`), and then include an `if [ -e "_postinstall" ]; then _postinstall; fi` clause at the bottom of the templated `postinstall` script. That will require changes to the macOS templates, but they should be fairly minimal.

2. Document that any user-provided post-install script overwrites the default one from the template, so if you have a console app, you are responsible for installing the symlink for the app. That would be simpler for *us* to implement; but at the cost of a more complex experience for end users, which I don't think is desirable.

3. *Always* treat the user-provided post-install content as "content that will be appended", but document that it will be appended to the existing script, and modify the template so that (a) the console app post-install logic is templated so that it's only included for a console app, and (b) with a clear marker in the file where user supplied content can be included (i.e., include "cut lines" so that all content between the cut lines is removed, and new content injected on every run).

Of those options, I think (1) is the one that is least likely to cause problems, but I'm open to discussion if you disagree.

freakboy3742 (changes requested):
This looks to be headed the right direction; a couple of comments about specifics inline.

src/briefcase/platforms/macOS/__init__.py
haikesan: Thanks for detailed feedback, on point 1 concern:
re-injection on every build is actually already avoided in the current implementation. build_pkg_scripts() doesn't mutate the template's postinstall, it wipes and rebuilds a separate pkg_scripts/ directory from scratch on every run.
but points 2 and 3 are valid, I'll proceed with cleanest first approach here.

changes/2561.feature.md
```diff
@@ -0,0 +1 @@
+macOS PKG installers can now include additional installer resources, configured with the ``installer_resources`` setting.
```
freakboy3742: We use Markdown format, not ReStructuredText
```suggestion
macOS PKG installers can now include additional installer resources, configured with the `installer_resources` setting.
```

changes/2865.feature.md
```diff
@@ -0,0 +1 @@
+macOS PKG installers can now run a post-install script, configured with the ``post_install_script`` setting.
```
freakboy3742:
```suggestion
macOS PKG installers can now run a post-install script, configured with the `post_install_script` setting.
```

docs/en/reference/platforms/macOS/index.md
```diff
@@ -158,10 +158,32 @@ will result in an `Info.plist` declaration of:
...
+### `installer_resources`
+
+/// note | Only used for PKG packaging
+///
+
+A path, relative to the project root, to a directory of additional resources to include in the PKG installer. The contents of this directory are added alongside Briefcase's own installer resources (such as the welcome screen and license), and can be referenced by the installer's distribution definition.
+
+These resources are part of the installer itself, they are used while the installer is running, and are not installed onto the target machine.
+
+DMG and ZIP "installers" do not carry resources, if you define `installer_resources` while using one of those formats, the setting will be ignored.
```
freakboy3742:
```suggestion
DMG and ZIP "installers" do not carry resources. If you define `installer_resources` while using one of those formats, the setting will be ignored with a warning.
```

docs/en/reference/platforms/macOS/index.md
```diff
@@ -158,10 +158,32 @@ will result in an `Info.plist` declaration of:
...
+### `post_install_script`
+
+/// note | Only used for PKG packaging
+///
+
+A path, relative to the project root, to a shell script that will be executed during installation, after the installer content has been unpacked. The script must be a valid macOS shell script file with an appropriate shebang line (e.g., `#!/bin/sh`).
+
+For console apps, Briefcase provides a post-install script that creates a symbolic link to the app on the `PATH`; your script will be run *after* that link has been created.
+
+DMG and ZIP "installers" cannot run scripts; if you define a `post_install_script` while using one of those formats, the setting will be ignored. There is no PKG analog for a pre-uninstall script.
```
freakboy3742:
```suggestion
DMG and ZIP "installers" cannot run scripts. if you define a `post_install_script` while using one of those formats, the setting will be ignored with a warning. There is no PKG analog for a pre-uninstall script.
```

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1707,19 +1816,31 @@ def package_pkg(
...
         else:
             install_args = ["--install-location", "/Applications"]
 
+        # Console apps always have a post-install script (it symlinks the binary onto
+        # the PATH); any app can also configure its own. If a custom script is
+        # configured, assemble a dedicated scripts directory for it.
+        scripts_path = installer_path / "scripts"
+        if post_install := getattr(app, "post_install_script", None):
+            with self.console.wait_bar("Installing post-install script..."):
+                scripts_path = self.build_pkg_scripts(app, installer_path)
```
freakboy3742: Suggestion to make this a little more explicit that there are two possible paths:
```suggestion
        if post_install := getattr(app, "post_install_script", None):
            with self.console.wait_bar("Installing post-install script..."):
                scripts_path = self.build_pkg_scripts(app, installer_path)
        else:
            scripts_path = installer_path / "scripts"
```

src/briefcase/platforms/macOS/__init__.py
```diff
@@ -1707,19 +1816,31 @@ def package_pkg(
...
+
+        if app.console_app or post_install:
+            install_args += ["--scripts", scripts_path]
+
+        # Assemble the installer resources. If the user has supplied additional
+        # resources, overlay them onto the templated resources in a clean directory.
+        resources_path = installer_path / "resources"
+        if getattr(app, "installer_resources", None):
+            with self.console.wait_bar("Installing additional installer resources..."):
+                resources_path = self.build_pkg_resources(app, installer_path)
```
freakboy3742: As above:
```suggestion
        if getattr(app, "installer_resources", None):
            with self.console.wait_bar("Installing additional installer resources..."):
                resources_path = self.build_pkg_resources(app, installer_path)
        else:
            resources_path = installer_path / "resources"
```

tests/platforms/macOS/app/package/test_package_pkg.py
```diff
@@ -410,6 +411,71 @@ def test_console_app_adhoc_signed(
...
+    first_app_with_binaries.packaging_format = "pkg"
+    first_app_with_binaries.post_install_script = "scripts/post_install.sh"
+
+    package_command.notarize = mock.Mock()
+    # Spy on chmod; the executable bit can't be checked via the file mode on Windows.
+    package_command.tools.os = mock.MagicMock(spec_set=os)
+    bundle_path = tmp_path / "base_path/build/first-app/macos/app"
+
+    # The templated scripts directory; its postinstall invokes `_postinstall`.
+    create_file(bundle_path / "installer/scripts/postinstall", "#!/bin/sh\nhook\n")
```
freakboy3742: To drive home the point of the test, it would be helpful to pull out this "shell script" content as a constant that is re-used in the assertion.

tests/platforms/macOS/app/package/test_package_pkg.py
```diff
@@ -410,6 +411,71 @@ def test_console_app_adhoc_signed(
...
+    # The user's script keeps its own interpreter.
+    create_file(
+        tmp_path / "base_path/scripts/post_install.sh",
+        "#!/usr/bin/env python3\nprint('custom')\n",
```
freakboy3742: As above - this would be better as a constant.

tests/platforms/macOS/app/package/test_package_pkg.py
```diff
@@ -410,6 +411,71 @@ def test_console_app_adhoc_signed(
...
+    # ...and the user's script is installed verbatim as an executable `_postinstall`.
+    assert (pkg_scripts / "_postinstall").read_text(encoding="utf-8") == (
+        "#!/usr/bin/env python3\nprint('custom')\n"
+    )
+    package_command.tools.os.chmod.assert_called_once_with(
+        pkg_scripts / "_postinstall", 0o755
+    )
+
+    # pkgbuild was given the generated scripts directory.
+    pkgbuild_args = package_command.tools.subprocess.run.mock_calls[0].args[0]
```
freakboy3742: This should verify the full list of args, not just a selection.

freakboy3742 (approved):
A couple of minor tweaks (I'm happy to make them manually), and this should be good to go - Thanks for the contribution!

PR #2904, opened 2026-06-24
Update AGENTS.md

kattni:
Add guidance on using Markdown syntax in docstrings and changenotes, and on following the PR template.

## PR Checklist:
 
 Assisted-by:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -10,6 +10,7 @@ Briefcase converts Python projects into standalone native applications for macOS
 ## Quick Reference
 
 - **Language**: Python >= 3.10 (3.10–3.14 supported)
+- **Docstrings**: Sphinx style with Markdown content
 - **Dev environment**: Python 3.13 virtualenv with `dev` dependency group
 - **License**: BSD-3-Clause
 - **Entry point**: `briefcase` via `src/briefcase/__main__.py:main()`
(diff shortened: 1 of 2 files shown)


freakboy3742 (approved):
No idea if it will work, but it makes sense to me.

mhsmith:
I don't think we should add things to AGENTS.md which duplicate instructions that already exist elsewhere; that's just giving ourselves a maintenance burden to keep them in sync. In particular, including a copy of the PR template here will likely cause an agent to ignore any changes we make to the live template in the future.

If the agents aren't reading the instructions elsewhere, then we should add language that will encourage them to do so. And if the instructions elsewhere are incomplete, then we should improve them, which will also help non-agent-assisted contributors.

freakboy3742:
> I don't think we should add things to AGENTS.md which duplicate instructions that already exist elsewhere; that's just giving ourselves a maintenance burden to keep them in sync. In particular, including a copy of the PR template here will likely cause an agent to ignore any changes we make to the live template in the future.
> 
> If the agents aren't reading the instructions elsewhere, then we should add language that will encourage them to do so. And if the instructions elsewhere are incomplete, then we should improve them, which will also help non-agent-assisted contributors.

Agreed there's something to discuss here; I've [opened a discussion rather than bury the conversation on a merged PR](https://github.com/beeware/beeware/discussions/630).

Issue #2231, opened 2025-04-02
Provide the option to use uv instead of pip for creating environments and installation


freakboy3742:
### What is the problem or limitation you are having?

uv is an alternative to pip and venv, with the advantage that it is especially fast at creating environments and installing packages. It should be possible to opt into the use of uv instead of pip.

### Describe the solution you'd like

It should be possible to configure Briefcase to use `uv` as the environment management tool. This could be a per-project setting (configured in the wizard); or it could be setting that is configured on a per *user* basis that a project might choose to override.

### Describe alternatives you've considered

Do nothing. `pip` and `venv` are perfectly viable solutions that work out of the box.

### Additional context

See also #596 - conda is another alternative to uv for environment creation.

Using uv for iOS and Android could be complicated, as those platforms have limited support for mobile platforms.

The speed of uv could make #1735 a lot more palatable. There's no reason we *couldn't* implement #1735 using pip and venv, but the user experience of taking 30 seconds to start the app every time would be poor. If uv can reduce that startup time to seconds, there's very little downside to creating a fresh virtual environment for every run.

rmartin16:
As a thought....we could use _only_ `uv`.... :)

freakboy3742:
That would be *an* option; however, supporting Conda is going to remain on the cards, so anything we do here will likely involve a pluggable interface. The question then becomes what value we use as a default.

At least in the short term, uv support for iOS and Android is likely to be the constraining factor here. uv *does* support `--platform` (or an analog - IIRC the spelling is different) - but unless we can install requirements for iOS and Android and/or create an "iOS/Android environment", making uv the default isn't a viable option.

alanbato:
I'd like to start the work on this as part of the EuroPython 2025 Sprints :)

timrid:
> At least in the short term, uv support for iOS and Android is likely to be the constraining factor here.

uv now supports [iOS](https://github.com/astral-sh/uv/pull/15640) and [Android](https://github.com/astral-sh/uv/pull/15646) since [v0.8.16](https://github.com/astral-sh/uv/releases/tag/0.8.16).

> uv does support --platform (or an analog - IIRC the spelling is different)

[...]

Issue #596, opened 2021-07-10
Support the use conda as a base environment for Briefcase apps


freakboy3742:
There is a significant part of the Python ecosystem (especially in the science/data space) that uses Conda to manage their Python environments. Some tools are easier to install via Conda; it therefore follows that some apps may benefit from being based on a Conda environment, rather than the default "vanilla" Python installation.

Briefcase should provide, as an option, the ability to build the app on a Conda base environment, rather than the "vanilla" Python install.

This will involve providing a configuration option in `pyproject.toml` to specify the base/foundation that an app wishes to use (with the current "vanilla Python" behavior being the default); then, depending on the configured foundation, Briefcase should change:
1. The default application template 
2. The default support package
3. The tool used to install app packages

Ideally, this would be done via a plugin interface, so that:
1. The conda backend can be developed independently without imposing a maintenance burden on the briefcase team
2. Other backends could be developed. There isn't a *huge* demand for this (pip and conda really are the only options in this space), but it might be useful to be able to configure different template + support package combinations as defaults.

**Implementation hints**

The affected code is all contained in `briefcase/commands/create.py`:
* The `support_package_url` property
* The `app_template_url` property
* The `install_app_dependencies()` method

If these three attributes/methods could be abstracted behind an interface, with Briefcase providing a default implementation for that interface that provides the existing behavior, an alternate Conda implementation of that interface could then be installed. 

The implementation to be used would be selected early in the lifecycle of the Create command, based upon a configuration option on the app looking up a key in a list of `pkg_resources` plugins (see `setup.cfg` and `briefcase/platforms/__init__.py` for details on how plugins are used at present to define the platforms). Ideally, the plugin interface would allow any command to have overridden entry points (so - Briefcase would define a `briefcase.foundation.default.create` plugin, and a Conda plugin for briefcase would define `briefcase.foundation.conda.create`; allowing for `briefcase.foundation.default.build` and `briefcase.foundation.conda.build` et al in future, if required).

**Questions/Considerations**

A (probably incomplete) list of questions/considerations for this project:

1. How do you create a standalone, relocatable Conda environment? Are binaries installed into a Conda site-packages folder location independent? Can the `.app` folder be moved once an installation has occurred? 
2. What configuration to that environment is needed to support having a `site-packages` folder in user space, *outside* the location of the packaged Conda environment.
[...]

PR #2917, opened 2026-06-30
Add support for alternative environment managers

freakboy3742:
**This is a work in progress; this should not be merged as is, and may be broken up into smaller contributions**.

Adds:
* A uv environment management backend (Refs #2231)
* A conda environment management backend (Refs #596)
* ~~A pixi environment management backend~~ Removed because Pixi integration needs more work
* Modifies the macOS, Windows, Linux and iOS platforms to use environment management for installing dependencies.

This is currently able to:
* generate development environments using all 3 environment backends; 
* generate macOS apps using the Xcode backend using beeware/briefcase-macOS-Xcode-template#125
* generate macOS apps using the App backend using beeware/briefcase-macOS-app-template#104. 
* generate Windows apps using the VisualStudio backend using beeware/briefcase-windows-VisualStudio-template#105

TODO:
* [x] Unit tests passing on macOS
* [ ] Unit tets passing on Linux
* [ ] Unit tests passing on Windows
* [x] 100% Coverage
* [ ] Validation apps passing
* [ ] Validation apps added for conda

---

## Key changes:

* App configurations gain an `env_manager` attribute, defaulting to `venv` (`uv` and `conda` also being legal values).
* There are new `uv` and `conda` VirtualEnvironment subclasses.
* The VirtualEnvironmentManager tool is now a mechanism for looking up the VirtualEnvironment *class*, rather than creating the venv.
* The VirtualEnvironment class (and subclasses) aren't used as a context manager. While shells have the concept of "activation" (which is a good match for context managers), that activation is really just setting some environment variables, which can be done on a per-invocation basis.
* Virtual environments identify if the environment is responsible for providing Python. This is to account for the fact that Conda/Pixi environments provide Python as a package, rather than needing a support package.
* There's a final `install_managed_python_env()` step when the environment provides Python; this is to copy the full `lib` folder into the app bundle. This is done instead of staging packages into an "app_packages" folder, and then "merging" them into a final location. 
* The conda backend uses `pip` to install editable source and local packages
* The conda backend doesn't support universal apps or x86-64 on macOS. Given x86-64 support is almost EOL, the extra effort didn't seem worth it.

## PR Checklist:


freakboy3742:
Closing as the last of this PR is now included in #2972.

Issue #1362, opened 2023-07-14
Better accomodate installing APK on physical device when existing app cannot be replaced


rmartin16:
### Describe the bug

The current APK install process can error if an existing version of the app cannot be replaced. For me, this happened because the existing app on the device was installed from a different computer.

```
[helloworld] Starting app on SM G960U (5759363737573398) (device ID 5759363737573398)

[helloworld] Installing app...

>>> Running Command:
>>>     /home/russell/.cache/briefcase/tools/android_sdk/platform-tools/adb -s 5759363737573398 shell am force-stop com.example.helloworld
>>> Working Directory:
>>>     /home/russell/tmp/beeware/helloworld
>>> Return code: 0
Stopping old versions of the app... done

>>> Running Command:
>>>     /home/russell/.cache/briefcase/tools/android_sdk/platform-tools/adb -s 5759363737573398 install -r /home/russell/tmp/beeware/helloworld/build/helloworld/android/gradle/app/build/outputs/apk/debug/app-debug.apk
>>> Working Directory:
>>>     /home/russell/tmp/beeware/helloworld
>>> Command Output:
>>>     adb: failed to install /home/russell/tmp/beeware/helloworld/build/helloworld/android/gradle/app/build/outputs/apk/debug/app-debug.apk: Failure [INSTALL_FAILED_UPDATE_INCOMPATIBLE: Package com.example.helloworld signatures do not match previously installed version; ignoring!]
>>>     Performing Streamed Install
>>> Return code: 1
Installing new app version...

Unable to install APK /home/russell/tmp/beeware/helloworld/build/helloworld/android/gradle/app/build/outputs/apk/debug/app-debug.apk on 5759363737573398
```

### Steps to reproduce

1. Install an APK on a physical device
2. Attempt to install APK from a different machine or using a different identity

### Expected behavior

Fallback to an uninstall and install workflow or at least provide better information to the user about what's wrong.

A simple remedy is `~/.cache/briefcase/tools/android_sdk/platform-tools/adb uninstall "com.example.helloworld"

### Environment

- Operating System: debian 11
- Python version: 3.10.6
- Software versions:
  - Briefcase: `0.3.16.dev9+gb6057a0d`

### Logs

[briefcase.2023_07_14-17_02_12.run.log](https://github.com/beeware/briefcase/files/12053969/briefcase.2023_07_14-17_02_12.run.log)

mhsmith:
[...]

PR #2949, opened 2026-07-20
Improve Android "install failed" messages

moondial-pal:
The two known failure markers `UPDATE_INCOMPATIBLE` and `VERSION_DOWNGRADE` now present the user with an informative error message indicating the probable cause and include a suggested route for resolution. 

Previously, both failures produced the same generic 'Unable to install APK' message, giving users no indication of what went wrong or how to fix it.

Refs #1362 

## PR Checklist:

Assisted-by: Claude Sonnet

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1481,18 +1481,45 @@ class ADB:
                 raise InvalidDeviceError("device id", self.device) from e
             raise
 
-    def install_apk(self, apk_path: str | Path):
+    def install_apk(self, apk_path: str | Path, package: str):
         """Install an APK file on an Android device.
 
         :param apk_path: The path of the Android APK file to install.
+        :param package: The package name that the Android APK will install.
         :returns: `None` on success; raises an exception on failure.
         """
         try:
             self.run("install", "-r", apk_path)
         except subprocess.CalledProcessError as e:
-            raise BriefcaseCommandError(
-                f"Unable to install APK {apk_path} on {self.device}"
-            ) from e
+            output = e.output or ""
+            if "INSTALL_FAILED_UPDATE_INCOMPATIBLE" in output:
+                raise BriefcaseCommandError(
+                    f"""\
+            Unable to install APK {apk_path} on {self.device}.
+
+            The app currently installed on the device was signed with a different key
+            than this APK (often because it was previously installed from a different
+            computer). To resolve this, uninstall the existing app, then try again:
+
+                $ adb -s {self.device} uninstall {package}
+            """
+                ) from e
+            elif "INSTALL_FAILED_VERSION_DOWNGRADE" in output:
+                raise BriefcaseCommandError(
+                    f"""\
+            Unable to install APK {apk_path} on {self.device}.
+
+            A newer version of this app is already installed, and Android won't install
+            an older version over it. To resolve this, uninstall the existing app, then
+            try again:
+
+                $ adb -s {self.device} uninstall {package}
+            """
+                ) from e
+            else:
+                raise BriefcaseCommandError(
+                    f"Unable to install APK {apk_path} on {self.device}"
+                ) from e
 
     def force_stop_app(self, package: str):
         """Force-stop an app, specified as a package name.
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -1495,14 +1495,16 @@ class ADB:
             if "INSTALL_FAILED_UPDATE_INCOMPATIBLE" in output:
                 raise BriefcaseCommandError(
                     f"""\
-            Unable to install APK {apk_path} on {self.device}.
+            Unable to install APK {apk_path} on {self.device}
 
             The app currently installed on the device was signed with a different key
             than this APK (often because it was previously installed from a different
-            computer). To resolve this, uninstall the existing app, then try again:
+            computer). To resolve this, run the following command to uninstall the
+            existing app, then try again:
 
-                $ adb -s {self.device} uninstall {package}
-            """
+                $ "{self.tools.android_sdk.adb_path}" -s {self.device} """
+                    f"""uninstall {package}
+"""
                 ) from e
             elif "INSTALL_FAILED_VERSION_DOWNGRADE" in output:
                 raise BriefcaseCommandError(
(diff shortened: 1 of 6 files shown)


moondial-pal:
`INSTALL_FAILED_VERSION_DOWNGRADE`

image

`INSTALL_FAILED_UPDATE_INCOMPATIBLE`

image

moondial-pal:
Looking for feedback on the error messages. Happy to change them if needed.

Also, since both messages suggest uninstalling as the remedy, should there be a note warning that this wipes the app's data? @mhsmith raised the same concern back in the original thread, just in the context of an automatic uninstall/reinstall flow (https://github.com/beeware/briefcase/issues/1362#issuecomment-1636756279), but it seems just as relevant here since someone could copy-paste the command without realizing the consequence.

moondial-pal:
Another note I'd like to mention is the adb command in these messages may not resolve if it's not on the user's PATH, still working on that fix.

moondial-pal:
Updated the error message to use Briefcase's managed adb path instead of assuming adb is already on the user's PATH. 
It is long, but can be copied then pasted. 
Running that to uninstall, then running `Briefcase run android` again will install cleanly.

@freakboy3742 or @mhsmith  whenever you get a chance, this is ready for review.

freakboy3742:
@moondial-pal Apologies for the delay in reviewing this - we should be able to get to this in the next couple of days.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1481,18 +1481,48 @@ def run(self, *arguments: SubprocessArgT, quiet: int = 0) -> str:
...
+                raise BriefcaseCommandError(
+                    f"""\
+            Unable to install APK {apk_path} on {self.device}
+
+            The app currently installed on the device was signed with a different key
+            than this APK (often because it was previously installed from a different
+            computer). To resolve this, run the following command to uninstall the
+            existing app, then try again:
+
+                $ "{self.tools.android_sdk.adb_path}" -s {self.device} """
```
mhsmith: Windows users might not recognize "$" as a prompt, and think it's part of the command line. Let's just have no prompt at all.

```suggestion
                "{self.tools.android_sdk.adb_path}" -s {self.device} """
```

src/briefcase/integrations/android_sdk.py
```diff
@@ -1481,18 +1481,48 @@ def run(self, *arguments: SubprocessArgT, quiet: int = 0) -> str:
...
-                f"Unable to install APK {apk_path} on {self.device}"
-            ) from e
+            output = e.output or ""
+            if "INSTALL_FAILED_UPDATE_INCOMPATIBLE" in output:
+                raise BriefcaseCommandError(
+                    f"""\
+            Unable to install APK {apk_path} on {self.device}
+
+            The app currently installed on the device was signed with a different key
+            than this APK (often because it was previously installed from a different
```
mhsmith: Better to focus on the user's current situation rather than making generalizations.

```suggestion
            than this APK (maybe because it was installed from a different
```

src/briefcase/integrations/android_sdk.py
```diff
@@ -1481,18 +1481,48 @@ def run(self, *arguments: SubprocessArgT, quiet: int = 0) -> str:
...
             self.run("install", "-r", apk_path)
         except subprocess.CalledProcessError as e:
-            raise BriefcaseCommandError(
-                f"Unable to install APK {apk_path} on {self.device}"
-            ) from e
+            output = e.output or ""
+            if "INSTALL_FAILED_UPDATE_INCOMPATIBLE" in output:
+                raise BriefcaseCommandError(
+                    f"""\
+            Unable to install APK {apk_path} on {self.device}
```
mhsmith: We recently added some helper methods to avoid this kind of confusing indentation (#2559). However, in this case we don't need the asterisk box to call attention to the message, because it's the last thing printed before Briefcase exits. 

So I think we should make `_dedent_and_wrap` a public function, and make its width argument default to 80 like the other functions. Then it can be used here to wrap the message, similarly to [this example](https://github.com/beeware/briefcase/pull/2891/files). It looks like it already avoids wrapping further-indented lines, such as the command line here.

src/briefcase/integrations/android_sdk.py
mhsmith: Also, this repeated command line should be factored out into a variable. The variable can also include the "To resolve this" sentence, along with the data loss warning you suggested.

src/briefcase/integrations/android_sdk.py
```diff
@@ -1481,18 +1481,48 @@ def run(self, *arguments: SubprocessArgT, quiet: int = 0) -> str:
...
+            an older version over it. To resolve this, run the following command to
+            uninstall the existing app, then try again:
+
+                $ "{self.tools.android_sdk.adb_path}" -s {self.device} """
+                    f"""uninstall {package}
+"""
+                ) from e
+            else:
+                raise BriefcaseCommandError(
+                    f"Unable to install APK {apk_path} on {self.device}"
```
mhsmith: The 3 copies of this message should be factored out into a variable.

tests/integrations/android_sdk/ADB/test_install_apk.py
```diff
@@ -31,20 +31,55 @@ def test_install_failure(adb, capsys):
...
+        side_effect=subprocess.CalledProcessError(
+            returncode=1,
+            cmd="install",
+            output="Failure [INSTALL_FAILED_UPDATE_INCOMPATIBLE: signatures do not match]",
+        )
+    )
+
+    with pytest.raises(BriefcaseCommandError) as exc_info:
+        adb.install_apk("example.apk", "com.example.helloworld")
+    assert "uninstall com.example.helloworld" in str(exc_info.value)
```
mhsmith, line 53: We don't need to include the full error text, but we should at least include enough words to check that the two different errors produce two different messages.

mhsmith:
> should there be a note warning that this wipes the app's data? @mhsmith raised the same concern back in the original thread, just in the context of an automatic uninstall/reinstall flow ([#1362 (comment)](https://github.com/beeware/briefcase/issues/1362#issuecomment-1636756279)), but it seems just as relevant here since someone could copy-paste the command without realizing the consequence.

Yes, that would be a good idea.

moondial-pal:
Thanks @mhsmith I'll get to work cleaning this up and addressing the feedback.

moondial-pal:
@mhsmith This is ready for review at your convenience.
Amended the output of the error messages dropping the prompt.
Factored out the repeated messages, the prefix and the remedy/warning text into shared variables and switched to the new public `dedent_and_wrap`.
Tests now check that messages are unique for each failure type.

mhsmith:
In future, please don't mark conversations as resolved: let the person who started the conversation do that. Automatic resolution from accepting a suggestion is OK.

mhsmith (approved):
Thanks, everything looks good to me.

moondial-pal:
> In future, please don't mark conversations as resolved: let the person who started the conversation do that. Automatic resolution from accepting a suggestion is OK.

Ah 😅, understood. Thanks for the patience @mhsmith .

PR #2958, opened 2026-07-23
Rework environment management interface.

freakboy3742:
Part of series of changes extracted from #2917.

Reworks the tooling around virtual environments.

Instead of the VirtualEnvironmentTool being a factory for creating an instance of a VirtualEnvironment  that is used as a context manager, the tool is now a factory for determining the appropriate *class* of VirtualEnvironment, and running verification on that class to confirm the tool exists. The command then constructs an instance of that class, and calls `prepare()` on that instance to ensure the instance exists. 

The class is no longer used as a context manager. While there is superficially a mapping to "entering and exiting" a virtual environment, that's not how they're used in practice - they're more akin to an application context, providing a set of wrappers around calls to `subprocess`. Each individual call to subprocess needs to "activate" the environment (by setting environment variables or other command prefixes); there isn't a global concept of "activation" that can be enabled. This also allows venv handling to be wrapped around other application contexts - most notably, Docker - so in principle, we could use different environment management inside Docker.

This also means that a command can create the instance of the venv early in the process, and then pass that venv around as context for requirements installation commands.

This also differentiates between the *name* of the environment, and the location of that environment; Tools like `pixi` have firm ideas about where environments should be stored, so we need to be able to name an environment, but leave it up to the backend to determine exactly where the environment will be stored.

Also adds an `env_manager` configuration item; this currently only accepts a value of `venv` , but we'll add `conda` and `uv` in the near future. Commands can then declare which environment managers they support. 

## PR Checklist:


src/briefcase/commands/dev.py
```diff
@@ -86,6 +83,11 @@ def add_options(self, parser):
             help="Run the app in test mode",
         )
 
+    def verify_app_tools(self, app: FinalizedAppConfig):
+        """Verify that tools needed to run the command for this app exist."""
+        super().verify_app_tools(app)
+        NativeAppContext.verify(tools=self.tools, app=app)
```
freakboy3742, line 89: This (and removing the NotImplemented for bundle_path) is needed to ensure that there is an app context; the venv commands are then run in that app context.

src/briefcase/integrations/virtual_environment/base.py
```diff
@@ -2,47 +2,77 @@
...
+        :param name: An identifying name for the environment. This will not be
+            the *full* name; the environment will add additional information to
+            disambiguate the app and environment type.
+        :param app: The app configuration
+        :param tools: The shared `ToolCache` instance.
+        :param base_path: The root of the project path.
+        :param platform: The platform for the environment. If this isn't the
+            host OS, the environment will be a cross build environment.
+        :param arch: The architecture for the environment. If this isn't the
+            host OS architeceture, the environment will be a cross build
```
freakboy3742:
```suggestion
            host OS architecture, the environment will be a cross build
```

mhsmith (approved):

PR #2959, opened 2026-07-23
Allow remote assembly loading in Windows ZIP packages

### LINKED ISSUE #1890: "Mark of the Web" prevents Windows app from running when distributed as a ZIP file
### Describe the bug

When a ZIP-packaged app is extracted and runs, a stacktrace is shown when loading DLLs fails (as seen in https://github.com/beeware/briefcase/discussions/1881).

This first happens when pythonnet tries to load its Python runtime DLL (Python.Runtime.dll) and then for the WebView DLL from toga-winforms.

The underlying issue is Windows' [Mark of the Web](https://learn.microsoft.com/en-us/deployoffice/security/internet-macros-blocked#additional-information-about-mark-of-the-web) (MotW) functionality that prevents running foreign code while the file is marked.

### Steps to reproduce

Package an app for Windows using ZIP. Extract and run that app on another machine.

### Expected behavior

The app should successfully run.

The issue can be worked around by manually removing the MotW from all DLLs in the app (via selecting "Unblock" in the DLL file's properties).

To bypass the MotW check automatically, we can tell the .NET runtime to [allow loading](https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/dd409252(v=vs.100)) remote sources.

This can be accomplished via a `<App Formal Name>.exe.config` file alongside the app's stub exe:
```xml
<configuration>
   <runtime>
      <loadFromRemoteSources enabled="true"/>
   </runtime>
</configuration>
```

Since this is not needed for MSI packages, it may be best to add this file only when packaging as a ZIP.

### Screenshots

_No response_

### Environment

- Operating System: Windows 11
- Python version: 3.11
- Software versions:
  - Briefcase: `0.3.20.dev15+g93ee7e9f`

### Logs

_No response_

### Additional context

_No response_

emerardd:
Fixes #1890.

Windows marks downloaded ZIP archives with Mark of the Web. This can prevent the
packaged app from loading remote assemblies such as `Python.Runtime.dll`.

ZIP packaging now injects an app-local `<executable>.exe.config` that enables
`loadFromRemoteSources`. The configuration is written directly into the archive,
so MSI packaging and the source tree are unchanged.

The Windows packaging tests cover the ZIP contents and verify the configuration
is not generated for MSI packages.

## PR Checklist:

Assisted-by: OpenAI Codex (GPT-5.6)

### DIFF
--- a/changes/1890.bugfix.md
+++ b/changes/1890.bugfix.md
@@ -1,1 +1,1 @@
-Windows ZIP packages now include an application configuration that permits .NET assemblies marked as originating from a remote source to load.
+Windows apps packaged as ZIP bundles are now safe from the "Mark of the Web".
--- a/changes/1890.bugfix.md
+++ b/changes/1890.bugfix.md
@@ -1,0 +1,1 @@
+Windows ZIP packages now include an application configuration that permits .NET assemblies marked as originating from a remote source to load.
--- a/tests/platforms/windows/app/test_package.py
+++ b/tests/platforms/windows/app/test_package.py
@@ -215,10 +215,8 @@ def test_package_msi(
 ):
     """A Windows app can be packaged as an MSI."""
 
-    package_command.package_app(
-        external_first_app if external else first_app_config,
-        **kwargs,
-    )
+    app = external_first_app if external else first_app_config
+    package_command.package_app(app, **kwargs)
 
     assert package_command.tools.subprocess.run.mock_calls == [
         # Compile MSI
@@ -246,6 +244,9 @@ def test_package_msi(
             cwd=tmp_path / "base_path/build/first-app/windows/app",
         ),
     ]
+    # The .NET remote source configuration is only included in ZIP packages.
+    executable_path = package_command.binary_path(app)
+    assert not executable_path.with_name(f"{executable_path.name}.config").exists()
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


emerardd:
Two example-build jobs failed for reasons that occur before the changed ZIP
packaging path runs:

- `pyside6 / ubuntu-24.04`: the openSUSE package mirror returned 404 while
  `zypper` installed `devel_basis`.
- `pyside6 / windows-11-arm`: the example app exited during `briefcase run`,
  before `briefcase package` was invoked.

The pre-commit, Towncrier, type, coverage, unit-test, documentation, and
remaining example-build checks passed. I attempted to rerun the failed jobs,
but contributors do not have permission to rerun workflows in the upstream
repository.

changes/1890.bugfix.md
```diff
@@ -0,0 +1 @@
+Windows ZIP packages now include an application configuration that permits .NET assemblies marked as originating from a remote source to load.
```
freakboy3742: This is functionally correct, but isn't a "user facing" representation of the change, because it doesn't explain the significance of the change.
```suggestion
Windows apps packaged as ZIP bundles are now safe from the "Mark of the Web".
```

tests/platforms/windows/app/test_package.py
```diff
@@ -282,20 +283,33 @@ def test_package_zip(package_command_with_files, first_app_config, kwargs, tmp_p
         "app_packages/clr.py",
         "app_packages/toga_winforms/command.py",
     )
+    executable_path = package_command_with_files.binary_path(first_app_config)
+    remote_sources_config = f"{executable_path.name}.config"
```
freakboy3742: This can be simplified a little if you include `remote_sources_config` directly in `source_folders_and_files`.

freakboy3742 (approved):
I've pushed a minor tweak to the changenote and the test case, but otherwise, this looks great! Thanks for the PR!

The two test failures you saw are a known problem; something about the Github Actions environment makes those two tests flaky.

PR #2970, opened 2026-07-30
Add uv support

### LINKED ISSUE #2231: Provide the option to use uv instead of pip for creating environments and installation

freakboy3742:
Add `uv` as an option for application environment managers.

Works on Windows, macOS and iOS. Works for Linux System apps when not building in Docker.

To test - add `env_manager = "uv"` to a project definition.

Fixes #2231 

## PR Checklist:


.pyspelling.yml
```diff
@@ -5,7 +5,7 @@ matrix:
     - docs/spelling_wordlist
     output: _build/dictionary/python.dic
   sources:
-  - 'docs/**/*.md'
+  - 'docs/en/**/*.md'
```
freakboy3742, line 8: This is to avoid trying to spell check `docs/superpowers` - I've been testing out obra/superpowers tooling, and it writes to this folder.

tests/commands/base/test_paths.py
```diff
@@ -31,15 +31,18 @@ def test_path_is_realpath(dummy_console, tmp_path):
...
+        monkeypatch.setattr(subprocess, "run", raise_calledprpcesserror)
+    else:
+
+        def raise_oserror(*a, **kw):
+            raise OSError()
 
-    # Patch run() since it's apparently quite difficult to find a
-    # location in Windows that is guaranteed to throw an error...
-    monkeypatch.setattr(subprocess, "run", raise_calledprpcesserror)
+        monkeypatch.setattr(os, "makedirs", raise_oserror)
```
freakboy3742, line 45: This change is unrelated; apparently Github has decided `/etc` is writable on *some* instances. I'll pull this out into a standalone PR.

freakboy3742: See #2971.

mhsmith (approved):

PR #2972, opened 2026-07-31
Add support for using conda as an environment manager.

freakboy3742:
Adds support for using Conda as an environment manager.

Adds one major new concept - the idea that the environment manager can be responsible for providing the Python install. When using Conda, there's no "support package" to install; Conda provides libpython, and the full Conda environment is copied into the application bundle as the last step of the install.

It's also assumed that if the environment is providing Python, Python packages are installed into that environment; so on update, updating the support package *or* the requirements requires rebuilding both.

Conda environments can only be used for macOS and Windows. A Conda doesn't have support for iOS or Android (yet... see [this draft CEP](https://github.com/conda/ceps/pull/183)); and a Conda environment can't be included as a Linux system package because it includes libraries (like libssl) that get flagged by DEB/RPM auditing tools. It might be possible to build a Flatpak that includes conda, but that is deferred from this PR.

On macOS, Conda cannot build universal apps; Conda binaries are single platform, and merging entire Conda environments would be complicated at best. As x86-64 macOS support is nearing EOL, I've decided to avoid the complication.

Add `env_manager = "conda"` to a project to test this configuration.

Building on macOS requires the use of custom templates:

* beeware/briefcase-macOS-Xcode-template#125
* beeware/briefcase-macOS-app-template#104

macOS requires a *different* stub binary for "framework" builds (which use the support package and Python.xcframework) vs "non-framework" builds (which use Conda and libpython3.X.so). 

Building on Windows requires the use of custom templates as well:

* beeware/briefcase-windows-VisualStudio-template#105
* beeware/briefcase-windows-app-template#112

I've preemptively published an experimental stub binaries (build x2917) to allow testing the macOS and Windows app templates. The app templates uses these stub binaries. The Xcode/VisualStudio templates can be merged today; once that happens and *this* PR is merged, the stub binary will need to be formally tagged, and the app template updated to use that new stub.

This PR also includes documentation for the general topic of environment management, and normalises the naming of the `venv` backend (renamed from `std_venv` - I realized the reason it was causing headaches was the .gitignore file).

## PR Checklist:


docs/en/reference/env_management/conda.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Prerequisites
+
+Briefcase requires the use of Conda 26.5 or higher.
+
+## Configuration
+
+To use Conda as the environment manager for your app, add `env_manager = "conda"` to your app configuration. This can be done as a global setting (in `[tool.briefcase]` section) or in a per-app setting (in the `[tool.briefcase.app.myapp]` section for `myapp`).
+
+The Conda environments that are created by Briefcase enable the use of `rattler` resolver.
```
mhsmith:
```suggestion
The Conda environments that are created by Briefcase are installed using the `rattler` resolver.
```

docs/en/reference/env_management/conda.md
```diff
@@ -0,0 +1,53 @@
...
+  ```python
+  requires = ["pillow=9.1"]
+  ```
+
+- Channel-qualified install
+  ```python
+  requires = ["conda-forge::pillow"]
+  ```
+
+Conda cannot install packages stored as local source references. If your requirements reference a local path, Briefcase will use `pip` to install that requirement (and it's dependencies) into your Conda environment. The following are examples of paths that would be installed with `pip`:
```
mhsmith:
```suggestion
Conda cannot install packages stored as local source references. If your requirements reference a local path, Briefcase will use `pip` to install that requirement (and its dependencies) into your Conda environment. The following are examples of paths that would be installed with `pip`:
```

docs/en/reference/env_management/uv.md
```diff
@@ -0,0 +1,54 @@
+# Environment management with uv
+
+Briefcase is able to use [uv](https://docs.astral.sh/uv/) to manage the installation of application requirements.
+
+## Prerequisites
+
+To use uv as the environment manager in your application, the `uv` binary must be available on your path. It can be installed using any of the [installation methods described in the uv documentation](https://docs.astral.sh/uv/getting-started/installation/).
```
mhsmith:
```suggestion
To use uv as the environment manager in your application, the `uv` binary must be available on your PATH. It can be installed using any of the [installation methods described in the uv documentation](https://docs.astral.sh/uv/getting-started/installation/).
```

mhsmith: For completeness, the other two files should also say how their tools are found:

* `conda` on the PATH.
* `pip` from the same environment as Briefcase itself.

docs/en/reference/env_management/conda.md
```diff
@@ -0,0 +1,53 @@
+# Environment management with Conda
+
+Briefcase is able to use [Conda](https://conda.io) to provide Python, and manage the installation of application requirements.
```
mhsmith: The fact that this mode gets Python from Conda and not from the Briefcase support package should be explicitly mentioned, both here and under the `support_package` / `support_revision` settings.

docs/en/reference/configuration.md
```diff
@@ -111,8 +111,9 @@ The environment manager to use when creating isolated Python environments and in
 
 * `venv` - The `venv` package provided by the Python standard library
 * `uv` - The [uv](https://docs.astral.sh/uv/) environment manager
+* `conda` - The [Conda](https://docs.conda.io/) environment manager
 
-Defaults to `venv`.
+Defaults to `venv`. For details on using different environment managers, see the `[environment management reference][environment-management]`.
```
mhsmith: Why is this option in the project section, which otherwise only includes high-level metadata? Could it be in the application section instead?

I remember we named `requirement_installer_args` on the assumption that when we later added this option, we would call it `requirement_installer`. I wonder if that might still be a good idea, because what the developer is really interested in is requirements, and environments are only a means to that end. 

The only exception is that Conda also provides Python itself, but I think that's a relatively minor detail compared to the clarity advantage of `requires`, `requirement_installer` and `requirement_installer_args` all having similar names, and appearing next to each other in the documentation. This would also remove any implication that `uv` is used to provide Python, which is a common use case but not one we currently support.

In that case, we would presumably rename `venv` to `pip`, at least in the the user-visible option.

(Remainder moved to https://github.com/beeware/beeware-docs-tools/issues/262)

src/briefcase/integrations/virtual_environment/conda.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+class CondaVirtualEnvironment(VirtualEnvironment):
+    """An environment manager using conda.
+
+    The environment is created as a prefix-based conda environment (`conda
+    create --prefix <path>`). The Python interpreter installed into the
+    environment matches the major/minor version of the interpreter running
+    Briefcase. Commands are executed against the environment by activating it
+    (prepending the environment's binary directory to ``PATH``).
```
mhsmith: This isn't applicable to Conda.

mhsmith:
> Conda environments can only be used for macOS and Windows.

What happens if you use conda or uv on a platform that doesn't support it, and is this documented and tested?

docs/en/reference/env_management/conda.md
freakboy3742: So... it *does* say that Conda provides Python... but I agree the implication isn't necessarily obvious, and the "negative" should also be mentioned. The cross-reference with support_package etc is also a good idea.

docs/en/reference/configuration.md
freakboy3742: I put it in the project section because it *can* be specified at the project level - which would seem to be the most likely usage as well - you're unlikely to use conda for one app, and uv in another app, inside a single project. 

As for `requirement_installer` - I see what you're driving at, but nobody thinks of uv or conda as a "requirements installer" - installing requirements is just one of the things you do *in* the environment. Granted, from the perspective of Briefcase, it's the most important (and almost the *only*) thing you do - but in the case of conda, its explicitly *not* just the requirements installer.

The overflow on names is something for the theme; I presume a different overflow behaviour is possible, but I'd need to dig into it.

freakboy3742:
Update:
* I've merged the Windows Visual Studio template, and tagged new binaries
* I've updated the Windows app template to use the new binary.
* I've merged the macOS Xcode template. I can't tag new binaries until this PR lands.
* I've updated the documentation to include the points flagged in reviews to date - mostly details of usage and edge cases
* I've merged with main, to pick up the fixes cases that caused issues during dependabot.

docs/en/reference/configuration.md
```diff
@@ -344,11 +344,13 @@ A file path or URL pointing at a tarball containing a Python support package. (i
...
 If this setting is not provided, Briefcase will use the default support package for the platform.
 
+The setting will be ignored if the app's environment manager is responsible for providing Python (e.g., Conda).
+
 #### `support_revision`
 
 The specific revision of a support package that should be used. By default, Briefcase will use the support package revision nominated by the application template. If you specify a support revision, that will override the revision nominated by the application template.
 
-If you specify an explicit support package (either as a URL or a file path), this argument is ignored.
+This argument will be ignored if you specify an explicit support package (either as a URL or a file path), the app's environment manager is responsible for providing Python (e.g., Conda).
```
mhsmith:
```suggestion
This argument will be ignored if you specify an explicit support package (either as a URL or a file path), or if the app's environment manager is responsible for providing Python (e.g., Conda).
```

docs/en/reference/configuration.md
```diff
@@ -396,7 +374,7 @@ A list of packages that are required for the test suite to run.
 
 Unlike most other keys in a configuration file, [`test_requires`][] is a *cumulative* setting. If an application defines requirements at the global level, application level, *and* platform level, the final set of requirements will be the *concatenation* of requirements from all levels, starting from least to most specific.
 
-See [`requires`][] for examples.
+The format for specifying requirements is determined by the [environment manager][env_manager] that is in use. For details on the format for specifying requirements, see the `[environment management reference][environment-management]`.
```
mhsmith: This link isn't rendering in any of the 3 places it appears on this page.

docs/en/reference/configuration.md
mhsmith: This needs to be updated.

mhsmith: Also, this section is out of alphabetical order.

docs/en/reference/configuration.md
freakboy3742: Not sure where those stray literal quotes came from... but fixed now.

docs/en/reference/configuration.md
freakboy3742: Fixed, and reordered. I've also clarified the docs for each environment manager to describe how requirements are installed.

mhsmith (approved):

Issue #2396, opened 2025-07-19
Add signing for Linux system packages


freakboy3742:
### What is the problem or limitation you are having?

Linux system packages do not currently have any signing or verification mechanism. It is my understanding that  signing of system packages is *possible* - if it is, Briefcase should support signing these packages.

### Describe the solution you'd like

`briefcase package linux system` should sign packages.

This will likely involve interacting with other tools to determine the available signing identities, and then applying that signing identity to the package.

### Describe alternatives you've considered

Status quo. Signing isn't *required*, it's a nice value add.

donkamileq:
Working on it EuroPython2025

jankoprowski:
We are currently working on signing .deb packages under Ubuntu 24.04. While there are many instructions available online about how to sign a Debian package, many of them are outdated. Several web pages recommend using tools like `debsign` or `dpkg-sig`, which are no longer readily available. These tools were used to sign Debian packages after they were built.

According to the [current instructions](https://www.debian.org/doc/manuals/maint-guide/build.en.html) for building Debian packages, it is recommended to use the `dpkg-buildpackage` command, which allows signing the package during the build process.
The older method using `dpkg-deb` should be replaced with:

`dpkg-buildpackage -k"gpg-identity@briefcase.org"`

You can also run this command without signing the package:

`dpkg-buildpackage -us -uc`

This process requires entering the password for your GPG key vault during signing. Additionally, the directory structure of the package has changed — instead of using a `DEBIAN` (uppercase) directory, you must now use a `debian` (lowercase) directory that contains numerous small metadata files.

After signing we will get files like:
`*.dsc`
`*.changes`
`*.deb`

We can verify the signature by following commands:

```
gpg --verify *.dsc
sha256sum *deb
grep *.deb *.changes
```
Sha256sum should be the same as entry inside `*.changes` file.

freakboy3742:
I'm going to close this as resolved by #2973, and open a new ticket for the one part that remains - adding code signing in a Docker context.

PR #2973, opened 2026-07-31
feat: sign linux system packages with GPG identities

siddhant-bayas:
Add GPG signing support to `briefcase package linux system`.

`briefcase package linux system` can now sign `.deb`, `.rpm` and `.pkg.tar.zst` packages with a GPG signing identity, using the `-i` / `--identity` option. The identity can be a GPG fingerprint (or any portion of it), a key ID, or the name/email address of an available secret key. If no identity is specified, Briefcase prompts you to select an identity from the available secret keys, or to opt out of signing; if only one secret key is available, it is offered as the default selection.

Signing is performed per format:

- `.deb` packages are signed in place with `debsigs --sign=origin --default-key=<identity>`, embedding an `origin` signature in the package.
- `.rpm` packages are signed in place with `rpmsign --define "_gpg_name <identity>" --addsign`, embedding a signature in the package header.
- `.pkg.tar.zst` packages are signed with `gpg --detach-sign`, producing a detached `.sig` file that must be distributed alongside the package.

If no identity is available, or `--adhoc-sign` is used, the package is produced unsigned with a warning. If the required signing tool (`debsigs`, `rpmsign` or `gpg`) is missing, an error suggests how to install it.

Signing is currently supported for local system builds only; it is not supported when building with Docker (i.e., using the `--target` option).

Refs #2396

## PR Checklist:
 Assisted-by: big-pickle (docs), claude-opus-4-8 (debugging)


docs/en/reference/platforms/linux/system.md
```diff
@@ -91,6 +91,20 @@ The Linux system app template includes a `LICENSE` and `CHANGELOG` file, with st
...
+## Code signing
+
+Linux system packages can be signed using a GPG signing identity. See the [how-to guide on Linux code signing](../../../how-to/code-signing/linux.md) for instructions on generating and obtaining a GPG signing identity.
+
+When signing a package, Briefcase uses the first secret key reported by `gpg --list-secret-keys`. The identity used to sign the package can be specified with the `-i` / `--identity` option; if no identity is specified and only one secret key is available, that key will be used. If multiple secret keys are available, Briefcase will prompt you to select one.
+
+Signing is performed as follows, depending on the packaging format:
+
+- `.deb` packages are signed with [`debsigs`](https://manpages.debian.org/bookworm/debsigs/debsigs.1.en.html), embedding an `origin` signature in the package. The `debsigs` tool must be installed.
+- `.rpm` packages are signed with [`rpmsign`](https://manpages.debian.org/bookworm/rpm/rpmsign.8.en.html), embedding a signature in the package header. The `rpm-sign` tool must be installed.
```
freakboy3742:
```suggestion
- `.rpm` packages are signed with [`rpmsign`](https://manpages.debian.org/bookworm/rpm/rpmsign.8.en.html), embedding a signature in the package header. The `rpm-sign` tool must be installed.
```

That URL doesn't seem right - pointing at Debian docs for an RPM tool?

docs/en/how-to/code-signing/linux.md
```diff
@@ -0,0 +1,57 @@
+# Linux { #code-signing-linux }
+
+## Overview
+
+Linux system packages (`.deb`, `.rpm` and `.pkg.tar.zst`) can be signed using a [GNU Privacy Guard](https://gnupg.org/) (GPG) key. Unlike macOS and Windows, there is no central authority that issues signing certificates for Linux. A GPG signature does not, by itself, make a package trusted; the person installing your package must have imported your public key into their GPG keyring, and must choose to trust it. However, a signature does allow a user to verify both the *integrity* and the *provenance* of the package, and it is a requirement for publishing packages to some official repositories.
```
freakboy3742: No need for the emphasis here.
```suggestion
Linux system packages (`.deb`, `.rpm` and `.pkg.tar.zst`) can be signed using a [GNU Privacy Guard](https://gnupg.org/) (GPG) key. Unlike macOS and Windows, there is no central authority that issues signing certificates for Linux. A GPG signature does not, by itself, make a package trusted; the person installing your package must have imported your public key into their GPG keyring, and must choose to trust it. However, a signature does allow a user to verify both the integrity and the provenance of the package, and it is a requirement for publishing packages to some official repositories.
```

docs/en/how-to/code-signing/linux.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Create a GPG key
+
+If you don't already have a GPG key, you can create one with:
+
+```console
+$ gpg --full-generate-key
+```
+
+You will be prompted to select a key type, key size and expiry date, and to provide a name and email address that will identify the key. The email address should be an address you control, as users will use it (along with your public key) to identify that the package really came from you.
```
freakboy3742: When I ran this, I was presented with a range of options for key type - can you provide any meaningful guidance about what type of key should be generated? On Ubuntu and Fedora, it suggests "ECC (sign and encrypt)" as a default - is that reasonable? What about the other options?

docs/en/how-to/code-signing/linux.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Distributing the signature
+
+When Briefcase signs a package, the signature is stored in a file alongside the package:
+
+- A `.deb` package is signed in place, and the signature is embedded in the package's `origin` signature field.
+- An `.rpm` package is signed in place; the signature is embedded in the package header.
+- An `.pkg.tar.zst` package is signed with a detached signature, producing a separate `<package>.sig` file that must be distributed alongside the package.
+
+Before distributing your packages, you should export your public key and make it available so that users can verify your packages:
```
freakboy3742, line 41: What does "make it available" mean? Where/how should that be done?

docs/en/how-to/code-signing/linux.md
```diff
@@ -0,0 +1,57 @@
...
+
+```console
+$ gpg --export --armor <fingerprint>
+```
+
+This will output your public key in ASCII armor format, which can be added to a user's keyring with `gpg --import`. A user can then verify the integrity and provenance of your package with `gpg --verify <signature file> <package file>` (for `.pkg.tar.zst` packages), or with the package manager's own verification tooling (for `.deb` and `.rpm` packages).
```
freakboy3742: Does "with the package managers own verification tooling" mean that the signature is automatically checked when the package is installed? Or that there's an option to `apt`/`dnf` that can verify a package?

docs/en/reference/platforms/linux/system.md
```diff
@@ -91,6 +91,20 @@ The Linux system app template includes a `LICENSE` and `CHANGELOG` file, with st
 
 The Linux system app template also includes an initial draft manfile for your app. This manfile will be populated with the [`description`][] and [`long_description`][] of your app. You may wish to add more details on app usage. If you need full control over the man page content, you can provide your own troff-formatted man page using the [`man_page`][] configuration option.
 
+## Code signing
+
+Linux system packages can be signed using a GPG signing identity. See the [how-to guide on Linux code signing](../../../how-to/code-signing/linux.md) for instructions on generating and obtaining a GPG signing identity.
+
+When signing a package, Briefcase uses the first secret key reported by `gpg --list-secret-keys`. The identity used to sign the package can be specified with the `-i` / `--identity` option; if no identity is specified and only one secret key is available, that key will be used. If multiple secret keys are available, Briefcase will prompt you to select one.
```
freakboy3742: This seems a little misleading as written - the first sentence suggests the first key will *always* be used, but AFAICT, that's only the case if there is one and exactly one secret key available.

Regardless, "Don't sign" should always be an available option. Auto-selecting an identity (even if there's only one of them) is inconsistent with other platforms; it would be OK for a single actual identity to be the default value in a menu, though.

pyproject.toml
```diff
@@ -360,6 +360,6 @@ unresolved-import = "ignore"
 [tool.codespell]
 skip = '.git,*.pdf,*.svg'
 # the way to make case-sensitive skips of words etc
-ignore-regex = '\bNd\b'
+ignore-regex = '\b(?:Nd|fpr)\b'
```
freakboy3742: What's this change for?

src/briefcase/platforms/linux/system.py
```diff
@@ -1043,7 +1045,256 @@ def debian_multiline_description(description):
...
+        by fingerprint, with the primary user ID as the value.
+    """
+    try:
+        output = tools.subprocess.check_output(
+            ["gpg", "--list-secret-keys", "--with-colons"],
+            quiet=1,
+        )
+    except (OSError, subprocess.CalledProcessError):
+        # No secret keys are available (gpg returns a non-zero exit code when
+        # there are no secret keys), or gpg isn't installed.
```
freakboy3742: These are two very different errors, and should be caught separately.

In particular, this feels like `gpg` should be wrapped as a tool integration, with a tool verification step to confirm availability of the gig binary.

src/briefcase/platforms/linux/system.py
```diff
@@ -1043,7 +1045,256 @@ def debian_multiline_description(description):
...
+            "pkg": ("gpg", "gpg", "gnupg"),
+        }[app.packaging_format]
+
+    def _verify_signing_tool(self, app: LinuxSystemAppConfig):
+        """Verify that the local environment contains the signing tool.
+
+        :param app: The app being signed
+        """
+        tool_name, executable_name, package_name = self._signing_tool(app)
+        if not self.tools.shutil.which(executable_name):
```
freakboy3742: This won't work for Docker builds - the tool needs to exist the app context, not the local build context.

src/briefcase/platforms/linux/system.py
```diff
@@ -1043,7 +1045,256 @@ def debian_multiline_description(description):
...
+                    return fingerprint
+            raise BriefcaseCommandError(f"Invalid signing identity {identity}")
+
+        if not identities:
+            # No signing identities are available.
+            return None
+
+        if len(identities) == 1:
+            # Only one identity is available; use it without prompting.
+            return next(iter(identities))
```
freakboy3742: As noted in the review of the docs - "Don't sign" should always be an option; if there's a single known identity, it's OK for *that* identity to be the default.

src/briefcase/platforms/linux/system.py
```diff
@@ -1043,7 +1045,256 @@ def debian_multiline_description(description):
...
+                "-u",
+                identity,
+                "--output",
+                str(self.signature_path(app)),
+                str(dist_path),
+            ],
+        }[app.packaging_format]
+
+        try:
+            self.tools.subprocess.run(sign_command, check=True)
```
freakboy3742: Again, this need to be run in the Docker context - `rpmsign` won't be available if I'm on macOS building over Docker.

That also means that for Docker builds, there's a need to export the signing key and add it to the docker container.

tests/platforms/linux/system/test_package__signing.py
```diff
@@ -0,0 +1,608 @@
...
+    )
+
+
+def test_identity_help(package_command):
+    """The identity help text is overridden for Linux."""
+    assert package_command.IDENTITY_HELP == (
+        "The GPG signing identity to use to sign the package. This can be the "
+        "fingerprint, key ID, or the name/email address of any GPG identity "
+        "available on the system."
+    )
```
freakboy3742: Asserting string values seems redundant - are these needed for coverage?

tests/platforms/linux/system/test_package__signing.py
```diff
@@ -0,0 +1,608 @@
...
+
+    assert (
+        package_command.signature_path(first_app)
+        == tmp_path / "base_path/dist/first-app-0.0.1-1-wonky.pkg.tar.zst.sig"
+    )
+
+
+# Tests for _verify_signing_tool().
+# #################################
```
freakboy3742: If there's clear "sub packages" to be tested, then the packaging tests should be broken into a subdirectory, with `test_LinuxSystemSigningMixin__verify_signing_tool.py` etc test modules.

freakboy3742 (changes requested):
Thanks for this PR - the core of this looks really solid. 

I've left some notes inline; but the broad themes of that feedback are: 

1. In your PR description, the AI tool usage declaration isn't looking for the name of the shell you used (e.g., "OpenCode") - we're looking for the model (Claude Opus 4.8, etc). The issue we're addressing is potential copyright exposure - if it turns out that one particular model is sanctioned because it contains copyrighted material, we need to be able to identify what pieces of our code are impacted.
2. As flagged inline, the "auto select identity" behavior isn't consistent with other platforms. If a user provides an identity at the command line, then that is the identity that should be used; but if there's no identity selected, a menu should *always* be shown, and "don't sign" should always be an option.
3. The docs all appear *correct*, but are missing some affordances for people not familiar with the packaging/signing process. I've flagged some areas of concern; in general, the docs need to be written from the position of assuming the user knows very little about their development environment, other than what they *have* to use on a daily basis for using their system (i.e., they will know how to install a package - possibly using a GUI tool, rather than command line - but they won't know how to drive `gpg` or `make`. Assume a non-expert user with Python programming experience, but not much more than that).
4. The implementation works great in my testing when used natively - but breaks on Docker. The signing tools need to verified inside the docker container; but the signature needs to be extracted from the *local* environment, and injected into the Docker context. That also means that the existence of GPG needs to be verified independent of the existence of `rpmsign` etc - because the execution environment could be macOS.

(4) is obviously a big piece of work. However, the core of what is needed will be the same - the only difference is the execution environment. In the interests of constraining scope creep, if you wanted to land a v1 of this feature that is "signing for local system builds only", I'd be OK with that, as long as there's clear error handling for the Docker build case, and the core features use the interface that *would* be Docker enabled (e.g., executing tools in the app context, not the system context).

docs/en/reference/platforms/linux/system.md
siddhant-bayas: fair point about the url. `manpages.debian.org` is a general man page archive that hosts the `rpmsign` page, but it does look odd for an rpm tool. Would you prefer i link to `rpm.org` or the arch man page instead?

docs/en/how-to/code-signing/linux.md
siddhant-bayas: thanks, agreed. i'll apply your suggested wording.

docs/en/how-to/code-signing/linux.md
siddhant-bayas: Yes, "ECC (sign and encrypt)" is the right default. Since GnuPG 2.4, ECC (Ed25519 for signing + Curve 25519 for encryption) is GnuPG's own recommended default,  it's fast, produces small signatures, and is what current GnuPG versions on Ubuntu/Fedora will suggest. I'll add a recommendation for that to the docs.

Of the other options: RSA (4096 bit) is the compatibility fallback — it works with much older GnuPG versions and older verification tooling (including `rpmsign`/`debsigs` on older distributions), at the cost of larger/slower keys. The other ECC curves (NIST P-384, Brainpool) offer no practical advantage for package signing, and DSA is long since deprecated, so I won't recommend those.

docs/en/how-to/code-signing/linux.md
siddhant-bayas: fair,  that was vague. I'll clarify that you should either publish the key to a keyserver (e.g.,` gpg --send-keys --keyserver keys.openpgp.org <fingerprint>`) or host the exported `.asc` file somewhere users can fetch it, and that users then import it into their keyring. I'll spell both out.

docs/en/how-to/code-signing/linux.md
siddhant-bayas: The signature is not automatically verified when the package is installed — I'll reword this to be explicit. The embedded signature on .deb/.rpm packages is checked with the dedicated tooling (`debsig-verify` for `.deb`, `rpmkeys --checksig/rpm --checksig` for` .rpm`), while `.pkg.tar.zst` signatures are verified with `gpg --verify`. I'll correct the doc so it doesn't imply automatic verification.

docs/en/reference/platforms/linux/system.md
siddhant-bayas: Agreed, and this is already on my list from the main review. I'll update both the behavior (menu always shown when no identity is given, with "Don't sign" as an option and a single identity as the default selection) and this doc to match.

docs/en/reference/platforms/linux/system.md
freakboy3742: rpm.org seems like a reasonably authoritative source - lets go with that.

src/briefcase/platforms/linux/system.py
siddhant-bayas: `OSError` (gpg not installed / not runnable) and `CalledProcessError` (gpg installed but no secret keys) should be handled separately. I'll split them, and I'll look at wrapping `gpg` as a managed tool integration with a proper verify step rather than the ad-hoc `shutil.which` check.

src/briefcase/platforms/linux/system.py
siddhant-bayas: this is the local-only constraint I outlined in the main review. I'll add a check so that when packaging targets a Docker context, signing is rejected with a clear "not supported for Docker builds" error for this v1, and the tool lookup goes through the app context so it can be swapped later.

src/briefcase/platforms/linux/system.py
siddhant-bayas: I'll make "Don't sign" a permanent menu entry and use the single identity only as the default selection, consistent with the other platforms.

src/briefcase/platforms/linux/system.py
siddhant-bayas: Same scope as the other Docker comments. For v1 (local builds only) I'll guard against the Docker case with a clear error. `The _signing_tool`/`sign_package` lookups will go through the app context so a future Docker version just needs key injection and in-container execution.

tests/platforms/linux/system/test_package__signing.py
siddhant-bayas: These guard the help-text overrides for Linux (ensuring `ADHOC_SIGN_HELP`/`IDENTITY_HELP `still override the platform-neutral base text, and stay in sync with the user-facing strings). If you'd rather, I can drop them, they're not needed for coverage.

tests/platforms/linux/system/test_package__signing.py
siddhant-bayas: I'll reorganize `test_package__signing.py` into a `test_package/signing/ `subdirectory with separate modules per mixin method (e.g., `test_verify_signing_tool.py`,` test_select_identity.py`, `test_sign_package.py`).

pyproject.toml
siddhant-bayas: This adds `fpr` to the codespell ignore regex. `gpg --list-secret-keys --with-colons` emits `fpr` as a record tag (for the key fingerprint), and codespell flags `fpr` as a misspelling. The `fpr` token appears in the source and tests, so it needs to be ignored.

freakboy3742: Ah - in which case, there are better ways to achieve that. 

There's a `ignore-words-list` setting two lines later in that file; but that's less appropriate for a word like `fpr` that *could* easily be a misspelling of `for`. 

The best option is an inline ignore - add `# codespell:ignore fpr` at the end of the line that uses the code and it will pass the code spell check- e.g. 
```
        elif record[0] == "fpr":  # codespell:ignore fpr
```
You might need to break the multi-line GPG output string into a concatenation of individual strings, rather than a multi-line string to make that pass:
```
GPG_OUTPUT = (
    "sec:u:255:22:F9FCBC4A7701B685:1785485940:::u:::scSC:::+::ed25519:::0:\n"
    "fpr:::::::::5F0B07E0E2D05DD5611BF7A3F9FCBC4A7701B685:\n"  # codespell:ignore fpr
    "grp:::::::::A93A9F4A002A4796BB0196F46C2ED4F603F95854:\n"
    "uid:u::::1785485940::E70814C3C8DC3AC3940E8BFBA0288F0CE55F2120::"
    "Jane Doe <jane@example.com>::::::::::0:\n"
    "ssb:u:255:22:5B0C17E0E2D05DD5611BF7A3F9FCBC4A7701B685:1785485940:::u:::e:::+::ed25519:::0:\n"
    "fpr:::::::::5B0C17E0E2D05DD5611BF7A3F9FCBC4A7701B685:\n"  # codespell:ignore fpr
)
```

tests/platforms/linux/system/test_package__signing.py
freakboy3742: I guess it doesn't hurt to retain them. It's a safety catch, doesn't cost much to keep it.

tests/platforms/linux/system/test_package__signing.py
freakboy3742: Please don't mark discussions as resolved; the existence of an outstanding discussion is part of how we (as reviewers) can track that a review item has been resolved.

pyproject.toml
siddhant-bayas: Thanks, I reverted the ignore-regex change in` pyproject.toml` and used inline `# codespell:ignore fpr` comments instead, on the source line and on each fpr record line in the test fixture. To keep the fixture's inline ignores on their own code lines I broke the multi-line `GPG_OUTPUT` string into a `"\n".join([...])` list (with the fingerprints pulled into `JANE`/`BOB` constants) rather than a parenthesized concatenation, but the effect is the same. `codespell` passes with the regex restored to `'\bNd\b'`.

changes/2396.feature.md
```diff
@@ -0,0 +1 @@
+Linux system packages (`.deb`, `.rpm` and `.pkg.tar.zst`) can now be signed with a GPG signing identity, using the `--identity` option to `briefcase package`. If no identity is specified, Briefcase will prompt you to select a signing identity from the secret keys available on your system, or to opt out of signing. Signing is not supported when building with Docker; packages built with the `--target` option must be produced without a signature.
```
freakboy3742: This is for the release notes - it doesn't need to be a full usage guide.
```suggestion
Linux system packages (`.deb`, `.rpm` and `.pkg.tar.zst`) can now be signed with a GPG signing identity, using the `--identity` option to `briefcase package`. Signing is not currently supported when building with Docker.
```

docs/en/how-to/code-signing/linux.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Create a GPG key
+
+If you don't already have a GPG key, you can create one with:
+
+```console
+$ gpg --full-generate-key
+```
+
+You will be prompted to select a key type, key size and expiry date, and to provide a name and email address that will identify the key. If possible, use an ECC key based on Curve 25519 (an `ed25519` signing key), which is the default in recent GnuPG versions and produces smaller, faster signatures; if you need to support older tools that don't understand ECC keys, generate an RSA key of at least 4096 bits instead. The email address should be an address you control, as users will use it (along with your public key) to identify that the package really came from you.
```
freakboy3742:
```suggestion
You will be prompted to select a key type, key size and expiry date, and to provide a name and email address that will identify the key. If possible, use an ECC key based on Curve 25519 (an `ed25519` signing key), which is the default in recent GnuPG versions and produces smaller, faster signatures. If you need to support older tools that don't understand ECC keys, generate an RSA key of at least 4096 bits instead. The email address should be an address you control, as users will use it (along with your public key) to identify that the package really came from you.
```

docs/en/how-to/code-signing/linux.md
```diff
@@ -0,0 +1,63 @@
...
+- An `.rpm` package is signed in place; the signature is embedded in the package header.
+- An `.pkg.tar.zst` package is signed with a detached signature, producing a separate `<package>.sig` file that must be distributed alongside the package.
+
+Before distributing your packages, you should export your public key and make it available so that users can verify your packages:
+
+```console
+$ gpg --export --armor <fingerprint>
+```
+
+This will output your public key in ASCII armor format. You can publish it on a public keyserver, for example with `gpg --send-keys --keyserver keys.openpgp.org <fingerprint>`, or you can host the exported `.asc` file on your own website. Users can then add your key to their keyring with `gpg --import <file>` (or `gpg --recv-keys <fingerprint>`).
```
freakboy3742:
```suggestion
This will output your public key in ASCII armor format. You can publish it on a public keyserver (e.g., with `gpg --send-keys --keyserver keys.openpgp.org <fingerprint>`), or you can host the exported `.asc` file on your own website. Users can then add your key to their keyring with `gpg --import <file>` (or `gpg --recv-keys <fingerprint>`).
```

docs/en/how-to/code-signing/linux.md
```diff
@@ -0,0 +1,63 @@
...
+
+A user can then verify the integrity and provenance of your package: with `gpg --verify <signature file> <package file>` for `.pkg.tar.zst` packages; with `debsig-verify` for `.deb` packages (using a policy that declares your key as trusted); or with `rpmkeys --checksig` (or `rpm --checksig`) for `.rpm` packages, once RPM is configured to trust your key.
```
freakboy3742:
```suggestion
A user can then verify the integrity and provenance of your package: 

- `debsig-verify` for `.deb` packages (using a policy that declares your key as trusted).
- `rpmkeys --checksig` (or `rpm --checksig`) for `.rpm` packages, once RPM is configured to trust your key.
- `gpg --verify <signature file> <package file>` for `.pkg.tar.zst` packages.
```

src/briefcase/integrations/gnupg.py
```diff
@@ -0,0 +1,73 @@
...
+                "operating system, and try again."
+            ) from e
+        except subprocess.CalledProcessError as e:
+            raise BriefcaseCommandError("Unable to invoke gpg.") from e
+
+        tools.gnupg = gpg
+        return gpg
+
+    @classmethod
+    def identities(cls, tools: ToolCache) -> dict[str, str]:
```
freakboy3742: This should be an instance method, not a class method; the tools class should be implied as part of tool verification.

src/briefcase/platforms/linux/system.py
```diff
@@ -1043,7 +1045,229 @@ def debian_multiline_description(description):
...
+        out of signing.
+
+        :param identity: A pre-specified identity (either the fingerprint, key
+            ID, or string name of the identity). If provided, it will be
+            validated against the list of available identities to confirm that
+            it is a valid signing identity.
+        :returns: The final identity to use, or `None` if no identity is
+            available and no identity was specified.
+        """
+        identities = GnuPG.identities(tools=self.tools)
```
freakboy3742: As noted in the review of the tool - this should be `self.tools.gnupg.identities()`

src/briefcase/platforms/linux/system.py
```diff
@@ -1043,7 +1045,229 @@ def debian_multiline_description(description):
...
+    def package_app(self, app, identity=None, adhoc_sign=False, **kwargs):
+        """Package an application.
+
+        :param app: The application to package
+        :param identity: The GPG signing identity to use to sign the package.
+            This can be a fingerprint, key ID, or the name/email address of an
+            identity. If unspecified, the identity will be determined from the
+            identities available on the system. Ignored if ``adhoc_sign`` is
+            True.
+        :param adhoc_sign: If ``True``, the package will not be signed.
```
```suggestion
            identities available on the system. Ignored if `adhoc_sign` is
            True.
        :param adhoc_sign: If `True`, the package will not be signed.
```

tests/integrations/gnupg/conftest.py
```diff
@@ -0,0 +1,39 @@
...
+        "sec:u:3072:1:4D68E1A93D2F47FB:1785485980:::u:::scESC:::::::",
+        f"fpr:::::::::{BOB}:",  # codespell:ignore fpr
+        "uid:u::::1785485980::89A2D4C6E5F8B3E7::Bob Builder <bob@example.com>::::::::::0:",
+    ]
+)
+BOB_OUTPUT += "\n"
+
+JANE_ALT_UID = (
+    "uid:u::::1785485940::E70814C3C8DC3AC3940E8BFBA0288F0CE55F2120::"
+    "Jane <jane@example.com>::::::::::0:\n"
```
freakboy3742: It took me a moment to spot the difference between the two "Jane" entries; it's worth putting a hat on it to make sure it stands out :-)
```suggestion
    "Jane's secret alternate identity <jane@example.com>::::::::::0:\n"
```

tests/platforms/linux/system/test_package.py
```diff
@@ -4,10 +4,19 @@
 import pytest
 
 from briefcase.exceptions import BriefcaseCommandError
+from briefcase.integrations.gnupg import GnuPG
 from briefcase.integrations.subprocess import Subprocess
 from briefcase.platforms.linux.system import LinuxSystemPackageCommand
 
 
+@pytest.fixture(autouse=True)
```
freakboy3742: This fixture is repeated - it should be rolled up into `system/conftest.py`. 

It also shouldn't need to use monkeypatch - mock_tools is a base tools fixture that can be extended with mocks for individual tools.

We also try to avoid autouse fixtures, as it's harder to debug their implicit behavior.

tests/platforms/linux/system/signing/test_package_app.py
```diff
@@ -0,0 +1,269 @@
...
+    ):
+        package_command.package_app(first_app)
+
+    package_command._package_deb.assert_not_called()
+    package_command._verify_signing_tool.assert_called_once_with(first_app)
+    package_command.sign_package.assert_not_called()
+
+
+def test_package_app_signs_raises_in_docker(
+    package_command, first_app, mock_identities
```
freakboy3742: A minor style nit - we try to avoid the "all on one line indented" format for arguments - if there's enough arguments that we need to break onto a separate line, (and there's no obvious shortening of the test name - like dropping "package_app", because it's redundant), then it's worth moving to 1-argument-per-line format.
```suggestion
    package_command,
    first_app,
    mock_identities,
```

freakboy3742 (changes requested):
Thanks for those updates - this is getting really close. A couple of fairly minor cleanups and style suggestions; the only one that is substantive is the use of `classmethod` on the GnuPG tool.

src/briefcase/integrations/gnupg.py
```diff
@@ -0,0 +1,71 @@
...
+            output = self.tools.subprocess.check_output(
+                ["gpg", "--list-secret-keys", "--with-colons"],
+                quiet=1,
+            )
+        except subprocess.CalledProcessError:
+            # gpg returns a non-zero exit code when no secret keys are available.
+            return {}
+        except OSError:
+            # gpg isn't installed.
+            return {}
```
freakboy3742: If gpg is missing, that should be caught by verification. If it's failing here, there's a much bigger problem going on, and we don't want it to fail silently.
```suggestion
```

src/briefcase/platforms/linux/system.py
```diff
@@ -1043,7 +1045,232 @@ def debian_multiline_description(description):
...
+        identity = self.console.selection_question(
+            intro="Select GPG signing identity to use:",
+            description="GPG Signing Identity",
+            options=options,
+            default=next(iter(identities)) if len(identities) == 1 else None,
+        )
+
+        if identity is None:
+            # The user opted not to sign the package.
+            return None
```
freakboy3742: For consistency with other platforms, we should provide the `--adhoc-sign` hint here as well.

src/briefcase/platforms/linux/system.py
```diff
@@ -1043,7 +1045,232 @@ def debian_multiline_description(description):
...
+                ):
+                    return fingerprint
+            raise BriefcaseCommandError(f"Invalid signing identity {identity}")
+
+        # Offer "Don't sign" as the first option, then the available identities.
+        # If only one identity is available, it becomes the default selection.
+        options = {None: "Don't sign", **identities}
+        identity = self.console.selection_question(
+            intro="Select GPG signing identity to use:",
+            description="GPG Signing Identity",
```
```suggestion
            description="GPG signing identity",
```

freakboy3742 (approved):
Thanks for those updates. I've pushed a couple of final really minor cosmetic tweaks; but otherwise, this is awesome! 

Thanks for the PR - this closes a major gap in Briefcase's feature set that I've wanted to close for a while. You've been a pleasure to work with through the whole code review process. I hope this is the first of many BeeWare PRs!

siddhant-bayas:
Thank you so much sir, this means a lot! This was actually my very first PR to a project of this size, so hearing that it closes a real gap and hearing that feedback is honestly really encouraging. I really enjoyed the review process too, appreciate you taking the time to work through it with me.

Looking forward to contributing more to Briefcase going forward. 🚀

PR #2978, opened 2026-08-03
Ensure requirement_install_args is passed to macOS and iOS installs.

freakboy3742:
#2967 refactored the use of app requirements to use virtual environments to install requirements. 

However, in doing this refactor, extra installer arguments were inadvertently omitted from the install process. This was observed as CI failures in Python-support-testbed (beeware/Python-support-testbed#192).

This ensures that the macOS and iOS invocations of requirements installs preserve the extra installer args.

Takes the opportunity to also switch back to rolling beeware/.github workflow pins, and to correct a permissions error with the dependabot changenote workflow.

## PR Checklist:

### DIFF
--- a/.github/workflows/check-pr-template.yml
+++ b/.github/workflows/check-pr-template.yml
@@ -10,4 +10,4 @@ permissions:
 jobs:
   check-pr-template:
     name: Check PR template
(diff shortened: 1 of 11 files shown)


freakboy3742:
@mhsmith Flagging for visibility - this was blocking dependabot updates.

PR #2985, opened 2026-08-06
Add hash verification to template usage.

### LINKED ISSUE #2981: Add hash verification for templates

Briefcase currently uses branch names to identify the template that is used to build the initial project template, and the base for apps.

However, tags and branches can be tampered with, without the awareness of the developer. Tools like `zizmor` encourage the use of full hashes, not just tags or branches, to identify releases of GitHub artefacts.

Briefcase should include the hash for all templates that it uses, not just the branch name. 

When the template is cloned/updated, the hash of the template should be verified; if the hash doesn't match, an error should be raised.

If a user provides a URL for a custom template (with `template`, or specifies a custom template branch (with `template_branch`), they should also be able to provide a `template_hash`. If a user specifies a custom template or branch, but doesn't specify a hash, a warning should be displayed.

Hashing should only be used for Briefcase-managed template clones, not for references to local directories.

The semantics of specifying hashes should match that in #2980 (i.e., an algorithm prefix, allowing for `unverified`).

Status quo - don't use template hashing.

### Additional context

It's worth noting that this will impose a maintenance burden - every template update will require a corresponding commit in Briefcase, and it won't be possible to "hotfix" a release by pushing an update to a template branch. This isn't a feature we use often, but we have done it (e.g., in the response to the security issue #2759); it's arguably a *good* thing, though.

freakboy3742:
Adds hash verification to any template that Briefcase clones.

Each Create command, plus New and Convert, set an expected template hash. The clone of the cookiecutter template is verified to ensure that the hash of the template branch matches that expected hash. If there's a mismatch, an error is raised.

An app configuration can also specify a template hash; this allows users to perform hash verification of their own template (and template branch) usage.

A utility script has been added to automatically update the code to new hashes; I've added mention of this script in the release process.

When the Briefcase version is an unreleased .dev version (i.e., during normal development), templates use the `main` branch; this usage is *not* hash verified, and a warning is raised. When we cut a release, there will be a tag, and release CI will fail unless the code has been updated to use the current hash on the template.

The one notable downside to this - it means we can't push an update to the v0.X.Y template branch to perform a hot-fix of any problems we find. That's both a good and bad thing; it's good security posture, but it also means if we find a security issue and want to backport that fix, we need to either need to cut a full release, or instruct users to add a `template_hash` value to their app configuration. That's inconvenient, but I think it's worth it from a security perspective.

Fixes #2981.

## PR Checklist:

Assisted-by: Claude Sonnet 5


mhsmith (approved):
Approved as long as CI passes once [GitHub Actions recovers](https://www.githubstatus.com/incidents/qcvjkzcs7j74).

Issue #2984, opened 2026-08-06
Add Linux code signing in Docker


freakboy3742:
### What is the problem or limitation you are having?

PR #2973 added support for code signing Linux system packages - but only when running natively. If you use `--target` to build a package in a Docker container, code signing fails.

### Describe the solution you'd like

It should be possible to build and sign a Linux package inside a Docker container. 

There's no changes needed to the general workflow - identity selection etc can continue as currently defined. The signing identity and gnupg must exist on the "host" machine (which may be a macOS machine). However, as part of container preparation, the Docker container must have the required tools installed, and the chosen signing identity must be injected into the Docker container (presumably by exporting the key from the local machine, and adding it to the keychain inside the Docker container). The signing step must also occur inside the Docker container.

### Describe alternatives you've considered

Status quo - only allow "native" signed package builds. This isn't ideal, as Docker builds are a major usability affordance.

freakboy3742:
Tagging @siddhant-bayas for visibility, as the author of #2973.

siddhant-bayas:
Hi @freakboy3742, I've finished implementing [Docker-based signing for Linux system packages](https://github.com/siddhant-bayas/briefcase/tree/feat/docker-system-signing) (the follow-up to #2973 / issue #2984). A few implementation details came up that I'd like to check with you before I put the PR up.

* **Signing tools are always installed in the image.** The Docker image gets built during `verify_app_tools`, which happens before the signing identity is chosen. Since identity selection is interactive, and existing images aren't rebuilt, there's no way to know at image build time whether signing will actually be used. Because of that, I add the signing tools (`debsigs` / `rpm-sign` / `gnupg`) to `SYSTEM_REQUIRES` for every image build. It means a few extra packages end up in every image, but it avoids needing a rebuild later if signing is enabled. Does that seem like a reasonable tradeoff?

* **Key injection uses a mounted file and a single container run.** The host exports the secret key to a file under `self.data_path`, which is already mounted at `/briefcase`. Then one container run does `gpg --batch --import <key> && <sign command>`, and the key file is cleaned up in a `finally` block. The key never ends up in an image layer or sticks around in the container. Is this what you had in mind when you mentioned exporting the key from the local environment and injecting it into the container, or would you prefer streaming it over stdin instead?

[...]

PR #2987, opened 2026-08-07
feat: sign Linux system packages when building with Docker

siddhant-bayas:
Add Docker support for code signing Linux system packages.

`briefcase package linux system` can now sign `.deb`, `.rpm` and `.pkg.tar.zst` packages when the package is built in a Docker container (i.e., using the `--target` option), using the same GPG signing identity as native builds.

When packaging inside Docker:

- The signing tool for the target format (`debsigs` for `.deb`, `rpm-sign` for `.rpm`, `gnupg` for `.pkg.tar.zst`) is installed in the build container. The tool is installed for every image build, since the image is built before the signing identity is selected. On SUSE, no additional package is installed, as `rpmsign` is provided by `rpm-build`, which is already part of the image.
- The selected secret key is exported from the host's GPG keyring into a file in the app's bundle folder (which is transient and app-specific, so the key is never at risk of being committed to version control or colliding with another project's key). The key is imported into the container and used to sign the package inside the container, and the exported key file is removed immediately after signing. The key is never stored in the Docker image.
- The `dist` folder is mounted into the container for the duration of the signing step only.

If the required signing tool is missing, the error now also suggests `--adhoc-sign` as an alternative to skipping signing, for consistency with other platforms.

As with native builds, a key that requires a passphrase cannot be used when signing a package built with Docker, since GnuPG cannot prompt for a passphrase inside a headless container. The code signing docs now describe the failure mode, and explain how to create a key without a passphrase for use with Docker builds.

Refs #2984

## PR Checklist:
 
 Assisted-by: big-pickle (docs), claude-opus-4-8 (debugging)

### DIFF
--- a/changes/2984.feature.md
+++ b/changes/2984.feature.md
@@ -1,0 +1,1 @@
+Linux system packages (`.deb`, `.rpm` and `.pkg.tar.zst`) can now be signed when building with Docker, using the same GPG signing identity as native builds. The signing identity is exported from the host and imported into the build container for the duration of the signing step.
--- a/changes/2984.feature.md
+++ b/changes/2984.feature.md
@@ -1,1 +1,0 @@
-Linux system packages (`.deb`, `.rpm` and `.pkg.tar.zst`) can now be signed when building with Docker, using the same GPG signing identity as native builds. The signing identity is exported from the host and imported into the build container for the duration of the signing step.
--- a/docs/en/how-to/code-signing/linux.md
+++ b/docs/en/how-to/code-signing/linux.md
@@ -64,4 +64,8 @@ As with other platforms, `--adhoc-sign` is useful during development and testing
 
 ## Docker builds
 
-Signing is not currently supported when building with Docker (i.e., when the `--target` option is used). When packaging with Docker, you must opt out of signing — either by selecting "Don't sign" when prompted, or by providing the `--adhoc-sign` option. Selecting a signing identity when building with Docker will cause an error.
+Linux system packages can be signed when building with Docker (i.e., when the `--target` option is used), using the same GPG signing identity as native builds.
+
+When signing a package built with Docker, Briefcase exports the selected secret key from the host machine's GPG keyring, and imports it into the build container so that the signing step can run inside the container. The exported key is removed immediately after signing, and is never stored in the Docker image.
+
+One caveat applies when building with Docker: because the signing step runs inside a headless container, GnuPG is not able to prompt for a passphrase. If your signing key requires a passphrase, the signing step will fail. To sign packages built with Docker, use a key (or subkey) that does not require a passphrase, or build the package without the `--target` option and sign it natively.
--- a/docs/en/how-to/code-signing/linux.md
+++ b/docs/en/how-to/code-signing/linux.md
@@ -14,6 +14,8 @@ $ gpg --full-generate-key
 
 You will be prompted to select a key type, key size and expiry date, and to provide a name and email address that will identify the key. If possible, use an ECC key based on Curve 25519 (an `ed25519` signing key), which is the default in recent GnuPG versions and produces smaller, faster signatures. If you need to support older tools that don't understand ECC keys, generate an RSA key of at least 4096 bits instead. The email address should be an address you control, as users will use it (along with your public key) to identify that the package really came from you.
 
+If you plan to sign packages built with Docker, create a key that does not require a passphrase: when GnuPG prompts you to enter a passphrase, leave the field blank and confirm. When a signing key is exported to a Docker container, GnuPG cannot prompt for the passphrase, so a key with a passphrase will fail the signing step. See [Docker builds](#docker-builds) for details.
+
 ## Obtain the identity of your key
 
 Briefcase uses the *fingerprint* of your key to identify the signing identity. To see the fingerprints of all the secret keys on your system, run:
@@ -68,4 +70,4 @@ Linux system packages can be signed when building with Docker (i.e., when the `-
 
 When signing a package built with Docker, Briefcase exports the selected secret key from the host machine's GPG keyring, and imports it into the build container so that the signing step can run inside the container. The exported key is removed immediately after signing, and is never stored in the Docker image.
 
-One caveat applies when building with Docker: because the signing step runs inside a headless container, GnuPG is not able to prompt for a passphrase. If your signing key requires a passphrase, the signing step will fail. To sign packages built with Docker, use a key (or subkey) that does not require a passphrase, or build the package without the `--target` option and sign it natively.
+One caveat applies when building with Docker: because the signing step runs inside a headless container, GnuPG is not able to prompt for a passphrase. If your signing key requires a passphrase, the signing step will fail. To sign packages built with Docker, use a key (or sub-key) that does not require a passphrase, or build the package without the `--target` option and sign it natively.
--- a/src/briefcase/platforms/linux/system.py
+++ b/src/briefcase/platforms/linux/system.py
@@ -2,11 +2,12 @@ from __future__ import annotations
 
 import gzip
 import re
+import shlex
 import subprocess
 import tarfile
 from collections.abc import Collection
 from pathlib import Path
-from typing import cast
+from typing import Any, cast
 
 from briefcase.commands import (
     BuildCommand,
@@ -1248,13 +1249,45 @@ or
             ],
         }[app.packaging_format]
 
+        subprocess_kwargs: dict[str, Any] = {}
+        key_file_path: Path | None = None
+        if isinstance(self.tools[app].app_context, DockerAppContext):
+            # When packaging inside Docker, the secret key must be made available
+            # to the container. Export the key to the data path (which is mounted
+            # into the container), then import it and sign the package in a single
+            # container run, so the key is not retained in the image or container.
+            key_file_path = self.data_path / f"{app.app_name}-signing-key.gpg"
+            subprocess_kwargs["mounts"] = [(self.dist_path, "/dist")]
+
         try:
-            self.tools[app].app_context.run(sign_command, check=True)
+            if key_file_path is not None:
+                self.tools.gnupg.export_secret_key(identity, key_file_path)
+                self.tools.os.chmod(key_file_path, 0o600)
+                sign_command = [
+                    "sh",
+                    "-c",
+                    " && ".join(
+                        " ".join(shlex.quote(arg) for arg in command)
+                        for command in [
+                            ["gpg", "--batch", "--import", str(key_file_path)],
+                            sign_command,
+                        ]
+                    ),
+                ]
+
+            self.tools[app].app_context.run(
+                sign_command,
+                check=True,
+                **subprocess_kwargs,
+            )
         except subprocess.CalledProcessError as e:
             raise BriefcaseCommandError(
                 f"Error while signing .{app.packaging_format} package for "
                 f"{app.app_name}."
             ) from e
+        finally:
+            if key_file_path is not None:
+                key_file_path.unlink(missing_ok=True)
 
     def clean_dist_folder(self, app, **options):
         super().clean_dist_folder(app, **options)
@@ -1283,12 +1316,6 @@ or
         else:
             identity = self.select_identity(identity=identity)
             if identity:
-                if self.use_docker:
-                    raise BriefcaseCommandError(
-                        "Signing system packages is not supported when using "
-                        "Docker. Re-run the package command without the "
-                        "`--target` option, or select `Don't sign`."
-                    )
                 # Signing is required; verify the signing tool is available.
                 self._verify_signing_tool(app)
             else:
--- a/src/briefcase/platforms/linux/system.py
+++ b/src/briefcase/platforms/linux/system.py
@@ -1258,27 +1258,19 @@ or
 
         subprocess_kwargs: dict[str, Any] = {}
         key_file_path: Path | None = None
-        if self.use_docker:
-            # When packaging inside Docker, the secret key must be made available
-            # to the container. Export the key to the bundle path (which is mounted
-            # in to the container), then import it and sign the package in a single
-            # container run, so the key is not retained in the image or container.
-            key_file_path = self.bundle_path(app) / "signing-key.gpg"
-            subprocess_kwargs["mounts"] = [(self.dist_path, "/dist")]
-            # The bundle folder is mounted at /app, and the dist folder at /dist;
-            # rewrite the sign command to use the container paths directly.
-            signature_path = self.signature_path(app)
-            container_dist_path = f"/dist/{dist_path.name}"
-            container_signature_path = f"/dist/{signature_path.name}"
-            sign_command = [
-                arg.replace(str(dist_path), container_dist_path).replace(
-                    str(signature_path), container_signature_path
-                )
-                for arg in sign_command
-            ]
 
         try:
-            if key_file_path is not None:
+            if self.use_docker:
+                # When packaging inside Docker, the secret key must be made available
+                # to the container. Export the key to the bundle path (which is mounted
+                # in to the container), then import it and sign the package in a single
+                # container run, so the key is not retained in the image or container.
+                #
+                # The bundle and dist folders are mounted in to the container, and the
+                # Docker layer rewrites the host paths in the commands to their
+                # container equivalents.
+                key_file_path = self.bundle_path(app) / "signing-key.gpg"
+                subprocess_kwargs["mounts"] = [(self.dist_path, "/dist")]
                 self.tools.gnupg.export_secret_key(identity, key_file_path)
                 self.tools.os.chmod(key_file_path, 0o600)
                 sign_command = [
@@ -1287,12 +1279,7 @@ or
                     " && ".join(
                         " ".join(shlex.quote(arg) for arg in command)
                         for command in [
-                            [
-                                "gpg",
-                                "--batch",
-                                "--import",
-                                f"/app/{key_file_path.name}",
-                            ],
+                            ["gpg", "--batch", "--import", str(key_file_path)],
                             sign_command,
                         ]
                     ),
@@ -1309,7 +1296,7 @@ or
                 f"{app.app_name}."
             ) from e
         finally:
-            if key_file_path is not None:
+            if self.use_docker:
                 key_file_path.unlink(missing_ok=True)
 
     def clean_dist_folder(self, app, **options):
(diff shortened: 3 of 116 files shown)


changes/2984.feature.md
freakboy3742, line 1: In this case, we don't need a second changenote. Ultimately, there's only one new feature - signing - the fact that it has been landed in 2 PRs is an implementation detail. 

So - in this case, the edit of the 2396.feature.md changenote is enough to satisfy towncrier, and we don't need this second changenote.

docs/en/how-to/code-signing/linux.md
```diff
@@ -64,4 +64,8 @@ As with other platforms, `--adhoc-sign` is useful during development and testing
...
-Signing is not currently supported when building with Docker (i.e., when the `--target` option is used). When packaging with Docker, you must opt out of signing — either by selecting "Don't sign" when prompted, or by providing the `--adhoc-sign` option. Selecting a signing identity when building with Docker will cause an error.
+Linux system packages can be signed when building with Docker (i.e., when the `--target` option is used), using the same GPG signing identity as native builds.
+
+When signing a package built with Docker, Briefcase exports the selected secret key from the host machine's GPG keyring, and imports it into the build container so that the signing step can run inside the container. The exported key is removed immediately after signing, and is never stored in the Docker image.
+
+One caveat applies when building with Docker: because the signing step runs inside a headless container, GnuPG is not able to prompt for a passphrase. If your signing key requires a passphrase, the signing step will fail. To sign packages built with Docker, use a key (or subkey) that does not require a passphrase, or build the package without the `--target` option and sign it natively.
```
freakboy3742: Two questions around this:

1. How do you create a key *without* a passphrase? In my testing on both Fedora and Ubuntu, passphrase creation was non-optional when using the `gpg --full-generate-key`
2. What is the failure mode if you *do* used a signed key for in-Docker signing?

src/briefcase/platforms/linux/system.py
```diff
@@ -727,6 +728,41 @@ def verify_system_python(self):
                 f"({system_version!r})."
             )
 
+    def _docker_signing_tool(self, app: LinuxSystemAppConfig) -> list[str]:
```
freakboy3742: How is this method any different to the existing `_signing_tool()`? There's clearly more logic required because of the differences related to SuSE - but we shouldn't have 2 independent places to define the package names for signing tools.

src/briefcase/platforms/linux/system.py
```diff
@@ -1202,13 +1251,45 @@ def sign_package(self, app: LinuxSystemAppConfig, identity: str):
             ],
         }[app.packaging_format]
 
+        subprocess_kwargs: dict[str, Any] = {}
+        key_file_path: Path | None = None
+        if isinstance(self.tools[app].app_context, DockerAppContext):
```
freakboy3742: I'm pretty sure this could be simplified to:
```suggestion
        if self.use_docker:
```

src/briefcase/platforms/linux/system.py
```diff
@@ -1202,13 +1251,45 @@ def sign_package(self, app: LinuxSystemAppConfig, identity: str):
...
+            # When packaging inside Docker, the secret key must be made available
+            # to the container. Export the key to the data path (which is mounted
+            # into the container), then import it and sign the package in a single
+            # container run, so the key is not retained in the image or container.
+            key_file_path = self.data_path / f"{app.app_name}-signing-key.gpg"
```
freakboy3742: The idea here is sound, but I don't think `self.data_path` is the best location. While it is a safe and mounted location, it's a little opaque (in that most users won't ever look in there), and shared between projects. 

I think it would be better to use a location in the bundle_path - that becomes an added level of safety because the bundle path is transient (so it won't get added to version control) and app specific (so there's no risk of name collision).
```suggestion
            key_file_path = self.bundle_path(app) / "signing-key.gpg"
```

tests/platforms/linux/system/test_package.py
```diff
@@ -165,6 +165,72 @@ def test_unknown_packaging_format(package_command, first_app):
...
+        ("debian", "system", "deb", ["debsigs"]),
+        ("rhel", "system", "rpm", ["rpm-sign"]),
+        ("arch", "system", "pkg", ["gnupg"]),
+        # Explicit output format is preserved, and adds the signing tool
+        ("debian", "deb", "deb", ["debsigs"]),
+        ("redhat", "rpm", "rpm", ["rpm-sign"]),
+        ("arch", "pkg", "pkg", ["gnupg"]),
+        # On SUSE, rpmsign is provided by rpm-build, which is already installed
+        # by the Docker image; no additional package is needed.
+        ("suse", "system", "rpm", []),
```
freakboy3742: Do we need a (`suse`, `rpm` ...) test case here?

tests/platforms/linux/system/test_package.py
```diff
@@ -165,6 +165,72 @@ def test_unknown_packaging_format(package_command, first_app):
...
+@pytest.mark.parametrize(
+    ("base_vendor", "input_format", "output_format", "expected_requires"),
+    [
+        # System packaging maps to known formats, and adds the signing tool
+        ("debian", "system", "deb", ["debsigs"]),
+        ("rhel", "system", "rpm", ["rpm-sign"]),
+        ("arch", "system", "pkg", ["gnupg"]),
+        # Explicit output format is preserved, and adds the signing tool
+        ("debian", "deb", "deb", ["debsigs"]),
+        ("redhat", "rpm", "rpm", ["rpm-sign"]),
```
freakboy3742: Is `redhat` a typo, or intentional?

docs/en/how-to/code-signing/linux.md
freakboy3742: Also - "subkey" is hitting the spell checker: 
```suggestion
One caveat applies when building with Docker: because the signing step runs inside a headless container, GnuPG is not able to prompt for a passphrase. If your signing key requires a passphrase, the signing step will fail. To sign packages built with Docker, use a key (or sub-key) that does not require a passphrase, or build the package without the `--target` option and sign it natively.
```

freakboy3742 (changes requested):
This looks pretty good, and works great in my testing. A couple of tweaks and questions inline. 

More generally - the process of exporting a key into the container still makes me a little nervous, because it's difficult to be 100% certain whether the key has been persisted into a cache somewhere. 

I know we've ruled out being able to mount the `.gnupg` folder directly because of version incompatibilities - but I've seen some suggestion online that it *might* be possible to use gpg in *agent* mode to export the gpg socket (via `gpgconf --list-dirs agent-extra-socket`) into the docker container, and do the signing using the *host* gpg. Could I ask you to investigate that option to see what might be possible?

docs/en/how-to/code-signing/linux.md
siddhant-bayas: 1. GnuPG does allow it, even though the UI hides that. In `gpg --full-generate-key`, press Enter at both the passphrase and repeat prompts, then answer `y` to the warning that an empty passphrase is not advisable. The prompts are mandatory, the passphrase is not. Your testing is right that this looks non-optional, and GnuPG documents it poorly, so the docs now spell it out in the "Create a GPG key" section.

2.  The container has no TTY, so gpg cannot start pinentry to ask for the passphrase, gpg refuses to sign, and Briefcase raises a `BriefcaseCommandError`. Honestly, the UX is the weak spot: the user gets a generic "Error while signing ... package for <app>" wrapping an opaque gpg pinentry failure. We do not detect or translate that failure, we only document the caveat. That is a real gap and it is worth being open about.

src/briefcase/platforms/linux/system.py
siddhant-bayas: They're now unified with no duplicate mappings. `_signing_tool()` (moved to `LinuxSystemMixin`, system.py:408) is the single source of truth: it resolves the format and returns the `(tool, executable, package)` triple from the shared `_SIGNING_TOOLS` map. `_docker_signing_tool()` (system.py:760) delegates to it and only adds the SUSE case (rpm-sign comes from `rpm-build`, already in the image) plus the list return type.

tests/platforms/linux/system/test_package.py
siddhant-bayas: Yes, it was a genuine gap. SUSE is the only known format where the docker tool list is intentionally empty, which is easy to break later. It is now covered at `test_package.py:181` (`("suse", "system", "rpm", [])`) and in `test_mixin__verify.py`.

tests/platforms/linux/system/test_package.py
siddhant-bayas: A typo. The constant is `RHEL = "rhel"`, and "redhat" is a value the os-release parser never produces. The test still passed because the format was explicit, so the vendor base was never consulted, but it was misleading. It is fixed to `rhel` in both parametrize blocks.

src/briefcase/platforms/linux/system.py
freakboy3742: This is an example of a place where you need to keep an AI agent on a short leash.

I'm guessing what you've done here is point your agent at this code, and my comment, and asked it to refactor. And that's exactly what it done... but it's gone too far, and somewhat in the wrong direction. `_SIGNING_TOOLS` is used in exactly one location, so there's no re-use value in having the constant. _SYSTEM_PACKAGING_FORMATS is used in 2 locations... but based on what I can see, one of those should be redundant - by the time you're invoking `_signing_tool()`, the packaging format should be a resolved property of the app.

And, my original question still stands - what is the difference between `_signing_tool()` and `_docker_signing_tool()`? Ultimately, both are returning details of the tools used to sign apps. `_docker_signing_tool()` does two things:
1. Drops all but the package name from the results
2. Performs a light transformation when we know the target platform is SuSE. 

The former doesn't need a separate method - it's "call the same method and ignore the first two results"; and AFAICT, the second is just as applicable to *local* SuSE installs - right now, if you run a local signing on a SuSE machine, and it can't find `rpmsign`, Briefcase will tell you to install `rpm-sign`... which is incorrect. 

I'm guessing this actually can't happen in practice because the packaging tools will be verified before the signing tools are checked - but that's not a reason to have 2 methods for one task.

src/briefcase/platforms/linux/system.py
```diff
@@ -1202,13 +1256,61 @@ def sign_package(self, app: LinuxSystemAppConfig, identity: str):
...
+            # rewrite the sign command to use the container paths directly.
+            signature_path = self.signature_path(app)
+            container_dist_path = f"/dist/{dist_path.name}"
+            container_signature_path = f"/dist/{signature_path.name}"
+            sign_command = [
+                arg.replace(str(dist_path), container_dist_path).replace(
+                    str(signature_path), container_signature_path
+                )
+                for arg in sign_command
+            ]
```
freakboy3742: This argument re-writing *shouldn't* be needed - one of the things the Docker layer does is path transformation. If that isn't working, it's possibly an indicator of a bigger bug.

src/briefcase/platforms/linux/system.py
```diff
@@ -1202,13 +1256,61 @@ def sign_package(self, app: LinuxSystemAppConfig, identity: str):
...
+
         try:
-            self.tools[app].app_context.run(sign_command, check=True)
+            if key_file_path is not None:
+                self.tools.gnupg.export_secret_key(identity, key_file_path)
```
freakboy3742, line 1272: The code layout here doesn't make sense. Why not move the `has_docker` branch inside the try, and then do all the checks based purely on whether docker is in use, instead of using `key_file_path` as a proxy?

freakboy3742 (changes requested):
A couple of refactoring/cleanup issues flagged. 

The bigger issue, however, is that you haven't addressed the question I asked in my last review - is the socket export approach workable? That would be much more preferable to a file export, as it *completely* avoids the "file in a layer" issue, and I *think* it might even resolve the key-with-password issue (since it will be the system GPG, *outside* the Docker container) that does the signing.

siddhant-bayas:
Thanks for the review! Both commits (`07a9a252` and `df023663`) address your feedback:

- `_SIGNING_TOOLS` is gone; the mapping lives inside `_signing_tool()` now
- The SUSE fix (rpmsign comes from rpm-build, there's no rpm-sign package) moved into `_signing_tool()`, so local SUSE installs get the correct hint too
- `_docker_signing_tool()` is collapsed into `_signing_tool()`, with the image prep step just ignoring the first two tuple elements
- `"system"` format resolution now happens in `finalize_app_config()`, so the packaging format is fully resolved before any signing logic runs, and `_SYSTEM_PACKAGING_FORMATS` only has one call site left
- `sign_package()` got restructured: the Docker branch sits inside the try block, everything keys off `self.use_docker`, and the manual path rewriting is gone since the Docker layer already handles that based on the mounts

On agent socket forwarding: I actually tried this instead of guessing. Bind-mounting the gpg-agent socket straight into the container only works on Linux-native Docker; Docker Desktop (macOS/Windows) can't forward arbitrary Unix sockets (see docker/for-mac #5297 and #7204). But briefcase already solves this kind of problem for X11 passthrough with a socat/TCP bridge, and the same trick works for gpg-agent:

1. Host side: `socat TCP-LISTEN:<port>,fork UNIX-CONNECT:$(gpgconf --list-dirs agent-extra-socket)`
2. Container side: another socat exposing `$GNUPGHOME/S.gpg-agent`, bridged to `host.docker.internal:<port>`
3. Import only the public key into the container

I got this working end to end: signed inside the container using the host's agent, then verified the signature back on the host. The secret key never leaves the host machine. This is literally what the extra-socket was designed for (it's in the gpg-agent docs), and the GnuPG wiki documents this exact socat pattern for SSH forwarding.

Your hunch about passphrases was right too. Passphrase prompts go to the host agent, so pinentry pops up on the user's desktop like normal. My test actually failed with "Operation cancelled" at first, which turned out to be because it was trying to launch pinentry on my headless host, so that confirms the routing works as expected.

A few caveats I hit:

- gpg-agent deliberately restricts remote sessions. Seeding the cache with `gpg-preset-passphrase` didn't carry over to the forwarded connection ("Bad passphrase"), and loopback pinentry from the container got refused ("Forbidden"), even with `allow-loopback-pinentry` set. So non-interactive CI signing would still need a passphrase-less dedicated key (which is basically what the current export/import approach gives you) or some more research.
- The TCP hop isn't authenticated, so it would need to bind to localhost or the docker bridge only.
- It adds socat as a dependency on both sides, plus relay lifecycle management.

Given all that, I'd lean toward keeping the current export/import approach for this PR. It works everywhere including CI, needs no extra tooling, and cleans up the key afterwards. Agent forwarding feels like a good follow-up for interactive users who want password-protected keys with Docker builds. Happy to open a tracking issue if you agree.

src/briefcase/platforms/linux/system.py
```diff
@@ -42,6 +44,33 @@
     parse_freedesktop_os_release,
 )
 
+# The packaging format implied by each system vendor base.
+_SYSTEM_PACKAGING_FORMATS = {
```
freakboy3742: As flagged in my previous review - this is an example of a constant that is used in one location. There's no reason to have it pulled out here.

src/briefcase/platforms/linux/system.py
```diff
@@ -1285,6 +1382,29 @@ class LinuxSystemPackageCommand(
...
+        :param app: The application to package
+        :param update: Should the application be updated (and rebuilt) first?
+        :param packaging_format: The format of the packaging artefact to create.
+        """
+        return super()._package_app(
+            app,
+            update,
+            _resolve_system_packaging_format(app, packaging_format),
+            **options,
+        )
```
freakboy3742: I can't see why this abstraction is needed. At the point `_package_app` is invoked the app config will be finalised. That means it has (or should be able to guarantee) a known-good `app.packaging_format` setting.

src/briefcase/platforms/linux/system.py
```diff
@@ -1697,6 +1802,32 @@ def _package_pkg(
...
+        :param packaging_format: The format of the packaging artefact to create.
+        :param channel: The resolved BasePublicationChannel instance
+        """
+        return super()._publish_app(
+            app,
+            update,
+            _resolve_system_packaging_format(app, packaging_format),
+            channel,
+            **options,
+        )
```
freakboy3742: As with `package` - this seems redundant.

src/briefcase/platforms/linux/system.py
```diff
@@ -273,6 +302,19 @@ def finalize_app_config(
...
+        # determine what that means. This must be done before the app tools are
+        # verified, so the Docker image can be built with the tools needed to
+        # package and sign the app.
+        if getattr(app, "packaging_format", None) in (None, "system"):
+            app.packaging_format = _SYSTEM_PACKAGING_FORMATS.get(app.target_vendor_base)
+            if app.packaging_format is None:
+                raise BriefcaseCommandError(
+                    "Briefcase doesn't know the system packaging format for "
+                    f"{app.target_vendor}. You may be able to build a package "
+                    "by manually specifying a format with -p/--packaging-format"
```
freakboy3742: Ok - this has exposed an interesting gap.

`finalize_app_config` is run on *every* command - including `create` and `run`. Those commands don't expose a `-p` option, so this help command is, at the very least, misleading. 

It also means that it's no longer possible to even *create* a Linux app for an unidentified Linux distribution without modifying the app configuration to add `packaging_format`. 

I think this might also be the cause of the roundabout logic you've added for `_package_app` - you're essentially adding code to work around the fact that finalise is setting the value for `-p`, which can then be overridden during packaging. Except that if the user has already had to bake in `packaging_format` to get past `create`, passing in *any* different value with `-p` will be an error.

So - there's a workflow issue we need to resolve here. At the very least, this needs to be only enforced for Docker builds; not supporting Docker builds for unknown linux distros would also be acceptable.

freakboy3742 (changes requested):
Thanks for the update; I've flagged a couple of issues inline, one of which (the availability of the `package_format` option) is going to require some thought.

The one remaining questions are around the export vs socket handling. 

Given the sensitive nature of signing keys, I'd much rather err on the side of caution. Although the window for exposure is *small* with exported files used as they are used in this PR, the window *exists*, and that means there's a potential problem. Therefore, I'd rather use the socket approach, even if that means there are limitations (in either the short or long term). 

Signing macOS packages on Linux is always going to be an edge case. I'm happy for it to be documented as such (or for macOS signing to require some extra config hurdles). If that means we defer macOS signing until a future iteration, that's fine as well.

As for CI environments: CI environments aren't ever going to allow password protected keys, so I don't think we need to go to extraordinary lengths to accommodate that use case (other than documenting the limitation).

freakboy3742:
Looks like you've broken something on the macOS builds - they're all failing with an error about `app.packaging_format`. 

As an aside - please don't force-push branches; it can cause historical comments to become orphaned. It's not a problem for commit history to be messy, as we use merge commits when we accept a PR.

PR #3003, opened 2026-08-13
fix: avoid shell=True in Windows cache dir creation

TrueFurina:
Replace `shell=True` in the Windows cache-directory creation (`src/briefcase/commands/base.py`) with an explicit `cmd.exe /c mkdir` argument list. The cmd.exe call (required to bypass the Windows Store Python sandbox) is preserved, but `data_path` is no longer interpreted by a shell — eliminating the shell-injection surface. Verified: `python -m py_compile` passes.

### DIFF
--- a/src/briefcase/commands/base.py
+++ b/src/briefcase/commands/base.py
@@ -260,8 +260,7 @@ a custom location for Briefcase's tools.
                 # allow normal interactions without attempting to sandbox them.
                 if platform.system() == "Windows":  # pragma: no-cover-if-not-windows
                     subprocess.run(
-                        ["mkdir", data_path],
-                        shell=True,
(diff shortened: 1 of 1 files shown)


freakboy3742:
To summarise:

1. There's no issue tied to this PR, so it's not clear what problem it is solving
2. There's no reproduction case in the PR (other than "it is required by Windows Store Python")
3. It's not intuitively obvious (to me at least) that this is a fix for anything
4. The PR doesn't follow the basic contribution guidelines laid out in our contribution guide.
5. The PR proposes a "testing" scheme which... I can't tell if it is intended as a joke, but it is hilarious.

If you believe you've found a problem, we're eager to accept a fix for that problem - but you need to work with us.

Closing on that basis.

PR #3022, opened 2026-08-27
Add install scope to Windows installer options

### LINKED ISSUE #2986: Make installer options scope optional

It is currently possible to define installer options for Windows MSI installers (with `install_option`)

Windows MSI installers can be installed with user scope, or with system scope - i.e., whether the app is being installed for one user, or for all users. 

There are some options (such as adding to the user path) that only make sense as a per-user option.

It should be possible to flag an install option as only being relevant to an install scope - for example:

```
[[tool.briefcase.app.helloworld.install_option]]
name = "foo"
title = "Foo the thing"
description = "Tell the app to foo the thing"
default = true
system = false
```

In this configuration:
* `system = true` would mean the option is only shown for all-user install scopes 
* `system = false` would mean the option is only shown for single-user scopes
* `system = "both"` (or undefined) would mean the option is used for both scopes
 
The list of installer options shown to the user should be filtered based on their chosen install scope. If an option isn't displayed, it should assume the default value for post-install script processing.

Pre-uninstall options should also honor this setting, if possible. The user doesn't choose the uninstall scope - that was determined at time of install - but if it's possible to preserve that value, that should be done.

This could be represented as 2 separate options (e.g., `system` and `user`) providing explicit definition for user and system installers. The downside is that `system=false, user=false` is a legal configuration.

### Additional context

_No response_

Xiaowen-Yang:
Add support for limiting Windows MSI install and uninstall options to a specific install scope.

The new `system` setting accepts:

- `true` for options that apply only to all-users installations;
- `false` for options that apply only to current-user installations; and
- `"both"` for options that apply to either scope.

When the setting is omitted, it defaults to `"both"` to preserve the existing behavior. Invalid values are rejected during configuration validation.

This change includes configuration validation, automated tests, Windows platform documentation, and a change note. The corresponding Windows app template change uses this setting to filter controls in the MSI install and uninstall dialogs.

Companion template branch:
https://github.com/beeware/briefcase-windows-app-template/pull/113

Fixes #2986

## PR Checklist:

Assisted-by: OpenAI Codex (GPT-5)

### DIFF
--- a/changes/2986.feature.md
+++ b/changes/2986.feature.md
@@ -1,0 +1,1 @@
+Windows MSI install and uninstall options can now be limited to current-user or all-users installs.
--- a/docs/en/reference/platforms/windows/index.md
+++ b/docs/en/reference/platforms/windows/index.md
@@ -219,7 +219,7 @@ If this setting is undefined, or set to an empty string, the installer will not 
 
 Windows MSI installers are able to present a panel of optional features to the user as part of the installation or uninstallation process. These features are binary flags which can then be used by a [post-install script][post_install_script] to perform additional installation behaviors, or by a [pre-uninstall script][pre_uninstall_script] to perform additional uninstallation behaviors.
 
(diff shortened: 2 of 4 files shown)


freakboy3742 (approved):
Awesome! This works great in my testing - and the extra typos you've picked up in the docs are also helpful. Thanks for the PR!

Issue #3031, opened 2026-09-01
Add an "auto" option for iOS simulator selection


freakboy3742:
### What is the problem or limitation you are having?

At present, `briefcase run iOS` requires either a `-d <simulator ID>` option, or a manual selection from a simulator list. This is inconvenient for CI purposes when the available simulators can change.

It should be able to "auto select" a simulator when the specific image doesn't matter.

### Describe the solution you'd like

`briefcase run iOS -d auto` should select a good candidate simulator - use the most recent iOS version, selecting an "SE-class" device (i.e., iPhone 17e). This matches the default behavior of the CPython iOS testbed runner.

### Describe alternatives you've considered

We could also make `-d auto` the default behavior, and make a bare `-d` an option to "show the full simulator list". 

### Additional context

This problem doesn't exist on Android because (a) Android requires explicit creation of a simulator, and (b) Briefcase already has a syntax to "auto drive" the creation of an emulator for test purposes.

PR #3033, opened 2026-09-02
Add auto device selection for Android.

freakboy3742:
A companion for #3031.

Adds automatic device selection (with creation, if needed) for Android. With this PR, `briefcase run android` will "just work", selecting an appropriate default emulator, creating that emulator if it doesn't already exist.

Also disables the use of skins if the code is running in a CI environment (if CI is set as an environment variable - which seems to be the defacto standard for identifying a CI environment).

## PR Checklist:

### DIFF
--- a/src/briefcase/integrations/android_sdk.py
+++ b/src/briefcase/integrations/android_sdk.py
@@ -197,11 +197,24 @@ class AndroidSDK(ManagedTool):
 
     @property
     def DEFAULT_DEVICE_SKIN(self) -> str:
+        # Skins are entirely cosmetic. In a CI environment, default to no skin.
+        if "CI" in os.environ:
+            return None
         return "pixel_7_pro"
 
     @property
+    def DEFAULT_ANDROID_SDK_LEVEL(self) -> int:
+        return 31
+
+    @property
     def DEFAULT_SYSTEM_IMAGE(self) -> str:
-        return f"system-images;android-31;default;{self.emulator_abi}"
+        sdk_level = self.DEFAULT_ANDROID_SDK_LEVEL
+        return f"system-images;android-{sdk_level};default;{self.emulator_abi}"
+
+    @property
+    def DEFAULT_AVD(self) -> str:
+        sdk_level = self.DEFAULT_ANDROID_SDK_LEVEL
+        return f"beePhone-{sdk_level}"
 
     @classmethod
     def sdk_path_from_env(cls, tools: ToolCache) -> tuple[str | None, str | None]:
@@ -722,7 +735,7 @@ connection.
     def list_installed_system_images(self) -> set[str]:
         """Returns a set of installed system image package identifiers.
 
-        e.g., ``{"system-images;android-31;default;x86_64"}``
+        e.g., `{"system-images;android-31;default;x86_64"}`
         """
         try:
             output = self.tools.subprocess.check_output(
@@ -744,7 +757,7 @@ connection.
         """Verify that the required system image is installed.
 
         :param system_image: The SDKManager identifier for the system image (e.g.,
-            ``"system-images;android-31;default;x86_64"``)
+            `"system-images;android-31;default;x86_64"`)
         """
         # Look for the directory named as a system image.
         # If it exists, we already have the system image.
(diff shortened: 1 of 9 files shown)


src/briefcase/integrations/android_sdk.py
```diff
@@ -197,11 +197,24 @@ def DEFAULT_DEVICE_TYPE(self) -> str:
...
     @property
     def DEFAULT_SYSTEM_IMAGE(self) -> str:
-        return f"system-images;android-31;default;{self.emulator_abi}"
+        sdk_level = self.DEFAULT_ANDROID_SDK_LEVEL
+        return f"system-images;android-{sdk_level};default;{self.emulator_abi}"
+
+    @property
+    def DEFAULT_AVD(self) -> str:
+        sdk_level = self.DEFAULT_ANDROID_SDK_LEVEL
+        return f"beePhone-{sdk_level}"
```
mhsmith, line 217: This needs to be reconciled with the code in `create_emulator`, which doesn't know about the SDK level suffix, but does sometimes add its own suffix to ensure uniqueness.

src/briefcase/platforms/android/gradle.py
```diff
@@ -397,9 +397,15 @@ def add_options(self, parser):
...
+            nargs="?",
+            default="auto",
+            const=None,
             help=(
                 "The device to target; either a device ID for a physical device, "
-                " or an AVD name ('@emulatorName') "
+                "or an AVD name ('@emulatorName'), or 'auto' (the default) to "
+                'automatically select a default "BeePhone" emulator (creating one '
+                "if necessary). Provide -d with no value to select from the full "
+                "list of available simulators."
```
mhsmith:
```suggestion
                "list of available emulators, or to create one interactively."
```

src/briefcase/platforms/android/gradle.py
```diff
@@ -397,9 +397,15 @@ def add_options(self, parser):
...
             "--device",
             dest="device_or_avd",
+            nargs="?",
+            default="auto",
+            const=None,
             help=(
                 "The device to target; either a device ID for a physical device, "
-                " or an AVD name ('@emulatorName') "
+                "or an AVD name ('@emulatorName'), or 'auto' (the default) to "
+                'automatically select a default "BeePhone" emulator (creating one '
```
mhsmith:
```suggestion
                'automatically select a default "beePhone" emulator (creating one '
```

src/briefcase/integrations/android_sdk.py
freakboy3742: Unless I'm missing something, I don't think there's a problem here.

`create_emulator()` prompts for an emulator name, providing a unique default name. It uses `beePhone` by default; if `beePhone` exists, it will suggest `beePhone1`, then `beePhone2` etc until a unique name is found.

The "default" emulator will be created as `beePhone-31`. If an AVD of that name already exists, it will be used. 

However, note that the "default" emulator has a hyphen; the `create_emulator` scheme does not. So the naming scheme for `create_emulator` names won't collide with the default emulator names. 

The only collision would be if a user manually created a `beePhone-31` emulator by specifying that name, in which case that emulator will be picked in future as the 'default'. That won't actually be a problem, other than the fact that we can't guarantee that it will be an SDK 31 emulator - but Briefcase isn't inherently guaranteeing that either. We're only guaranteeing and *an* emulator exists. If the user already has a beePhone-31 emulator, and selects that name, the name will be rejected as already being in use.

I included the SDK in the suffix to ensure that as a default behavior, when we need to bump the minimum required SDK, we can bump the default emulator SDK version, and we can guarantee the user will be provided with an emulator that user a compliant SDK. Otherwise, a user with a pre-existing Briefcase install and `beePhone` emulator would have a default emulator that isn't compatible with the new SDK minimum. This seemed an easier approach than adding parsing/error handling logic to catch the "emulator is too old" error case.

mhsmith: I agree with this general approach, but it seems confusing that we're generating emulator names in two different ways, both of which are `beePhone` followed by a number, but the number has different meanings depending on whether there's a hyphen or not.

This could be resolved by making `create_emulator` default to the same `beePhone-{sdk_level}` format (the SDK level will be configurable once #2843 lands), and to add a uniqueness suffix on top of that if necessary, e.g. `beePhone-31-1`. That way, there's only one naming scheme, and a manually-created emulator with the default name will also be found by the automatic selection.

freakboy3742: Fair point - I've modified the logic to use a consistent naming scheme between `create_emulator` and the default emulator creation scheme.

mhsmith (approved):

PR #3036, opened 2026-09-03
Log captured output when a non-streamed subprocess call fails

### LINKED ISSUE #1918: Log output for subprocess calls if it is not being printed
### Describe the bug

If you consider `Subprocess.check_output()`, it will [log](https://github.com/beeware/briefcase/blob/main/src/briefcase/integrations/subprocess.py#L655) both `stdout` and `stderr` in the event of an error.

However, `Subprocess.run()` does not attempt to log either `stdout` or `stderr` in the event of an error. The most likely reason for this is because `stdout` and `stderr` are being written to the console as the subprocess command runs. In the event that `stdout` and `stderr` are being piped, though, the content may be lost altogether.

### Steps to reproduce

One example is an error during macOS code signing since `stderr` is being piped for inspection is certain error cases but never written for fatal errors.

### Expected behavior

In the event that `subprocess.run()` is piping an output, ensure that output is captured in the log for an error.

### Screenshots

_No response_

### Environment

- Operating System: n/a
- Python version: n/a
- Software versions:
  - Briefcase: `0.3.20.dev47+g77cb18ec`

### Logs
In the log, you can see `stderr` has content, but it is not being fully captured in the log.

https://github.com/user-attachments/files/16044976/briefcase.2024_06_30-10_13_00.package.log

### Additional context

- https://github.com/beeware/briefcase/issues/1907

rawsun007:
Fixes #1918

## The problem

In non-streaming mode, `Subprocess.run()` handles a failure like this:

```python
    self._log_return_code(e.returncode)
    raise
```

`check_output()`, in the same situation, calls `self._log_output(e.output, e.stderr)` before logging the return code. So when a caller redirects stdout and/or stderr — which is exactly when the output is *not* on the console — the log ends up recording that a command failed with no indication of why. That is the case the issue describes, and the macOS signing example in #1910 is one instance of it.

## The change

One call added to that handler. When output was not redirected, `e.output` and `e.stderr` are `None` and `_log_output()` is a no-op, so unredirected calls log exactly as before.

## A test that pinned the bug

`test_calledprocesserror_exception_logging` built a `CalledProcessError` carrying both `output` and `stderr` and asserted that the log contained only the return code. It was encoding the behaviour this issue calls a bug, so it now asserts the output is present. I added `test_calledprocesserror_without_captured_output` to hold the other half — `output`/`stderr` unset, return code only — so the no-op path is pinned rather than assumed.

Flagging that explicitly because a reviewer seeing an existing test change should know it was deliberate.

## Verification

- `pytest tests/` — **4293 passed, 50 skipped**
- Coverage over the whole suite — **100.0%**, which the contributing guide requires with no exceptions
- Mutation check: reverting only the source change fails exactly one test, `test_calledprocesserror_exception_logging`, and the new unredirected test still passes
- `ruff check`, `ruff format --check`, `codespell` — clean on all three changed files
- Changenote added as `changes/1918.bugfix.md`

One thing I could not run: `pre-commit` fails to build the `docformatter` hook environment on Python 3.14 here (`untokenize` fails at `AttributeError: Constant object has no attribute s`). That is my interpreter, not this change; I ran ruff and codespell directly instead, and CI owns the full hook set.

## AI disclosure

Prepared with Claude Opus 5 in Claude Code: it located the asymmetry with `check_output()`, made the change, updated the tests, and drafted this description. I reviewed the diff and ran everything listed above locally. Per the contributing guide, the contribution is my responsibility.

## PR Checklist:

### DIFF
--- a/changes/1918.bugfix.md
+++ b/changes/1918.bugfix.md
@@ -1,0 +1,1 @@
+When a non-streamed subprocess call fails, any output captured on the exception is now written to the log. Previously only the return code was logged, so a failure whose output had been redirected left no explanation in the log file.
--- a/changes/1918.bugfix.md
+++ b/changes/1918.bugfix.md
@@ -1,1 +1,1 @@
-When a non-streamed subprocess call fails, any output captured on the exception is now written to the log. Previously only the return code was logged, so a failure whose output had been redirected left no explanation in the log file.
+When a non-streamed subprocess call fails, any output captured on the exception is now written to the log.
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -557,6 +557,12 @@ class Subprocess(Tool):
                 [str(arg) for arg in args], **self.final_kwargs(**kwargs)
             )
         except subprocess.CalledProcessError as e:
+            # If the caller redirected stdout and/or stderr, the output was never
+            # printed to the console; without logging it here, the log records a
+            # failure with no explanation. check_output() logs it in the same
+            # situation. When output wasn't redirected, these are None and
+            # _log_output() is a no-op.
+            self._log_output(e.output, e.stderr)
             self._log_return_code(e.returncode)
             raise
         self._log_return_code(command_result.returncode)
--- a/src/briefcase/integrations/subprocess.py
+++ b/src/briefcase/integrations/subprocess.py
@@ -557,11 +557,6 @@ class Subprocess(Tool):
                 [str(arg) for arg in args], **self.final_kwargs(**kwargs)
             )
         except subprocess.CalledProcessError as e:
-            # If the caller redirected stdout and/or stderr, the output was never
-            # printed to the console; without logging it here, the log records a
-            # failure with no explanation. check_output() logs it in the same
-            # situation. When output wasn't redirected, these are None and
-            # _log_output() is a no-op.
             self._log_output(e.output, e.stderr)
             self._log_return_code(e.returncode)
             raise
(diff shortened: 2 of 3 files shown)


rawsun007:
Heads-up on the one red check, so nobody has to dig: `Build app (pyside6, macos-26-intel)` is not related to this change.

It built the app fine, then hung in `briefcase run macOS app` for 24 minutes until the job was cancelled, leaving orphan `Verify App` and `Python` processes. The same job was also cancelled on the recent push to `main` ([run 33719116292](https://github.com/beeware/briefcase/actions/runs/33719116292)), and most recent runs on `drop-3.10` are cancelled or failing too, so it looks like that GUI-launch job is currently flaky repo-wide.

This diff only adds a logging call inside an existing `except subprocess.CalledProcessError` branch, so it cannot block or hang; the other 54 checks pass, including pre-commit, towncrier and the full unit matrix.

freakboy3742 (approved):
Thanks for the PR! I've pushed a small update - mostly trimming the verbosity of content that Claude has added. The extra test case wasn't testing anything new; and the 5 line comment explaining what "logs output" does also wasn't necessary.

As a general note, if you're planning to continue contributing to Briefcase and BeeWare - Claude's verbosity is something to keep an eye on. The PR body, for example, is *way* longer than it needs to be. This is, ultimately, a 1 line change. That does need *some* explanation, but it doesn't need 2 pages of explanation - and explaining your "testing regimen" when the regimen is "I ran the test suite" is entirely redundant.

PR #3044, opened 2026-09-08
Use the -embeddable Python download rather than -embed.

freakboy3742:
Python.org offers 2 "embeddeable" downloads for each version/architecture:

* python-3.X.Y-embed-amd64.zip 
* python-3.X.y-embeddable-amd64.zip

The public download page uses the former; but the `windows.json` release manifest only includes the latter. The release manifest includes release hashes; hashes for the -embed variants are on the website, but nowhere else.

The contents of the files are identical, except for a single `__install__.json` file in the `-embeddable` variant that includes some useful metadata. 

Since we need to be able to easily verify hashes, we should use the version that has readily available published hashes.

## PR Checklist:

### DIFF
--- a/changes/3044.misc.md
+++ b/changes/3044.misc.md
@@ -1,0 +1,1 @@
+Windows apps now use the `-embeddable` binary rather than the `-embed` binary.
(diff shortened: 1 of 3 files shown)


freakboy3742:
@mhsmith Tagging for visibility; needed to merge this to get a release out.

PR #3054, opened 2026-09-14
Reject a bundle identifier with a trailing newline

rawsun007:
`is_valid_bundle_identifier` applies `VALID_BUNDLE_RE` with `re.match`, and `$` matches before a final newline, so a bundle ending in one validates:

```pycon
>>> is_valid_bundle_identifier("com.example\n")
True
```

The value then reaches the places that use the identifier as a name — `flatpak.py` joins it onto the build path, and `appimage.py` puts it in a Docker image tag:

```pycon
>>> app = AppConfig(app_name="myapp", version="1.0", bundle="com.example\n", ...)
>>> app.bundle_identifier
'com.example\n.myapp'
```

`fullmatch` is what `is_valid_pep508_name` two hundred lines above already uses, and it makes the trailing `$` redundant, so that comes off too. Reverting only `config.py` fails exactly the one new case.

## PR Checklist:


mhsmith (approved):
Thanks!

PR #3057, opened 2026-09-15
Do not derive a Python keyword as the app class name

rawsun007:
`briefcase new` can generate an app that does not parse. The class name is derived from the formal name, and nothing stops that derivation landing on a Python keyword:

```
$ briefcase new --no-input -Q "formal_name=lambda" -Q "app_name=lambdaapp" ...
$ sed -n 9p lambdaapp/src/lambdaapp/app.py
class lambda(toga.App):
$ python -c "import ast; ast.parse(open('lambdaapp/src/lambdaapp/app.py').read())"
SyntaxError: invalid syntax
```

`briefcase dev` fails immediately on the app it just created. `None` and `class` do the same; `validate_formal_name` only checks that *some* identifier character survives.

The fix is the rule the function already applies one line above, where a name that cannot begin an identifier gets an underscore: a keyword gets one too, so `lambda` yields `_lambda`. Case still matters (`Lambda` is untouched) and soft keywords are left alone (`match` is a legal class name).

`is_reserved_keyword` already rejects keywords for *app* names, so this closes the same hole on the other name derived from user input.

Reverting `config.py` fails the three new cells. `pytest tests` is 4310 passed, 50 skipped.

## PR Checklist:


src/briefcase/config.py
```diff
@@ -107,6 +107,11 @@ def make_class_name(formal_name):
     ):
         class_name = f"_{class_name}"
 
+    # A keyword is a valid identifier everywhere except where this name is used:
+    # `class lambda(toga.App):` does not parse. Prepend an underscore, as above.
```
mhsmith:
```suggestion
    # A keyword is not a valid identifier, e.g. `class lambda(toga.App)`
    # does not parse. Prepend an underscore, as above.
```

mhsmith (approved):
Thanks!

PR #3060, opened 2026-09-17
doc: Note expired system certificate issue in how-to/code-signing/macOS

hyuri:
When an intermediate certificate is expired in the system — seems to be the **Developer ID** certificate —, instead of `This certificate is valid`, one will see `"Developer ID Application: [...]" certificate is not trusted` instead.

This adds a note on how to solve that issue — upgrading the system, or installing a fresh version of the intermediate certificate.

## PR Checklist:

Assisted-by:


docs/en/how-to/code-signing/macOS.md
```diff
@@ -105,6 +105,9 @@ Congratulations! You've just successfully installed the Developer ID Application
...
 
+/// note | Issue: "certificate is not trusted"
+If instead of `This certificate is valid` you see `"Developer ID Application: [...]" certificate is not trusted` in red, it most likely means the **Developer ID** intermediate certificate in your system, needed to validate your application certificate, is expired. To fix the issue, you need to either upgrade your system to the latest version (recommended), or download the latest [Developer ID - G2 certificate](https://www.apple.com/certificateauthority/DeveloperIDG2CA.cer). Other up-to-date certificates can also be downloaded from [Apple PKI](https://www.apple.com/certificateauthority) if needed.
```
freakboy3742, line 110: This isn't an issue right now, but for future safety of translations, we put a line break around `/// note` markup:
```suggestion
/// note | Issue: "certificate is not trusted"

If instead of `This certificate is valid` you see `"Developer ID Application: [...]" certificate is not trusted` in red, it most likely means the **Developer ID** intermediate certificate in your system, needed to validate your application certificate, is expired. To fix the issue, you need to either upgrade your system to the latest version (recommended), or download the latest [Developer ID - G2 certificate](https://www.apple.com/certificateauthority/DeveloperIDG2CA.cer). Other up-to-date certificates can also be downloaded from [Apple PKI](https://www.apple.com/certificateauthority) if needed.
```

freakboy3742 (changes requested):
Thanks for the PR! It's currently failing CI because of the need for a changenote, and a spelling issue (adding PKI to the spelling dictionary).

The only other thought (and I'm guessing you've fixed the problem, so you don't have the ability to test this any more) - can we catch this failure mode at runtime and provide advice? Do you still have logs from these failed runs?

hyuri:
> [...] I'm guessing you've fixed the problem, so you don't have the ability to test this any more) [...]

Yes, problem fixed on my end so I can't test this any further.

> The only other thought [...] can we catch this failure mode at runtime and provide advice? Do you still have logs from these failed runs?

[This part of the process](https://briefcase.beeware.org/en/stable/how-to/code-signing/macOS/) is done without running briefcase, so there are no failed runs.
The issue only manifests itself inside Keychain Access, after you install your Developer ID Application certificate.

hyuri:
> [...] and a spelling issue (adding PKI to the spelling dictionary).

I thought you meant you were adding it. Sorry. Done now.

freakboy3742 (approved):
I've pushed a couple of minor tweaks, but this is great! Thanks!

PR #3061, opened 2026-09-17
Fall back to a generic class name when none can be derived

rawsun007:
`make_class_name` drops every character that cannot appear in a Python identifier, so a formal name made only of punctuation, symbols or an emoji reduces to the empty string, and `AppConfig.class_name` returns `""`. Templates then emit `class (toga.App):`, which does not parse.

Reworked to the approach @freakboy3742 suggested: `!!!` is accepted, and the class name falls back to a generic one.

- `derive_class_name` is the stripping step, and still returns `""` when nothing survives.
- `make_class_name` returns `derive_class_name(formal_name) or "Briefcase"`, so the class name is always a usable identifier.

One knock-on that is worth a look, because it decides whether this is the whole change. `validate_formal_name` in `new.py` *is* the emptiness test:

```python
if not make_class_name(candidate):  # Check whether a class name may be derived
    raise ValueError(...)
```

With a fallback in `make_class_name`, that condition can never be true, and `briefcase new` would start accepting every formal name — `tests/commands/new/test_validate_formal_name.py` currently requires it to reject `""`, `" "`, `"\t"`, `"/"`, `"'"`, `"\\"` and `"/'\\"`. I have pointed the validator at `derive_class_name` so those rejections are unchanged, which keeps this PR to the bug you identified.

That does leave `briefcase new` refusing `"!!!"` while a hand-written `pyproject.toml` accepts it. If you would rather `new` accept it too — which is what "I can't see any reason it should be prohibited" suggests — that is a one-line change plus an update to those seven cases, and I am happy to make it; I did not want to loosen the interactive path without asking.

Also dropped the stale comment you spotted. Capitalisation is not applied and should not be: `make_class_name("match")` is `match` and `make_class_name("你好 世界")` is `你好世界`, both asserted in the existing tests, so adding `.capitalize()` would break them. The comment was the wrong half.

## Verification

`test_make_class_name` gains four cases (`!!!`, `...`, emoji, `""` → `Briefcase`), plus `test_derive_class_name` for the raw result and `test_make_class_name_is_always_an_identifier`. Removing only the `or "Briefcase"` fallback fails exactly 5 of those and leaves the rest green.

Full suite: 4325 passed, 50 skipped. `ruff check src tests` clean. `pre-commit` cannot build its environment here — the `docformatter` hook's `untokenize` dependency fails under Python 3.14 — so I ran ruff directly rather than claim a hook run I did not do.

The earlier CI failure on this branch was `test_rcedit_url`, which asserts a live `HEAD` against a GitHub release URL; that URL returns 200 from here and my diff does not reach it.

## PR Checklist:


freakboy3742 (changes requested):
Good catch, but I'm not sure the solution you've proposed is correct. 

`!!!` is definitely an *unconventional* formal name for an app, but I can't see any reason it should be *prohibited*. The issue is that the auto-generated class name, after stripping invalid characters, is empty.

I'd suggest that the better approach is that if the "cleaned" formal name generates an empty string, we set class name to something generic like "Briefcase". 

I also noticed that the *comment* in `make_class_name` says that after normalisation, the class name is capitalised... but there's capitalisation is never applied.

changes/3061.bugfix.md
```diff
@@ -0,0 +1 @@
+A formal name that contains no valid Python identifier characters, such as one made only of punctuation or an emoji, no longer produces an app with an empty class name; the class name falls back to ``Briefcase``.
```
freakboy3742, line 1: We use Markdown format, not RST.
```suggestion
A formal name that contains no valid Python identifier characters, such as one made only of punctuation or an emoji, no longer produces an app with an empty class name; the class name falls back to `Briefcase`.
```

tests/config/test_make_class_name.py
```diff
@@ -39,9 +41,39 @@
         # Soft keywords are legal class names, and case matters
         ("match", "match"),
         ("Lambda", "Lambda"),
+        # Nothing identifier-like survives, so a generic name is used
+        ("!!!", "Briefcase"),
+        ("...", "Briefcase"),
+        ("\U0001f389", "Briefcase"),
+        ("", "Briefcase"),
```
freakboy3742, line 48: This case in particular is problematic - it means a *literal* empty formal name will be accepted. We will need protection to ensure that "" isn't accepted (or, for that matter, "   ") isn't accepted as a formal name.

freakboy3742 (changes requested):
Thanks for those updates; this is definitely an improvement on the previous version, but it's got some fairly glaring problems that, as an experienced human reading the code, stand out to me.

You've declared that you're using Opus to develop this PR, and that's fine - but if your next interaction with Opus will be telling it "read the feedback from the reviewer and fix the problems" - providing that feedback is not a productive use of my time. Using GitHub as a proxy for my conversation with *your* Claude Code instance is an inefficient way to drive Claude; and won't result in *you* having a better understanding of the code either. 

If *you* are interested in receiving feedback, which *you* will fix, then I'm happy to provide that feedback. As a demonstration of that interest, I'd ask you to tell me what *you* think my criticisms might be.

In addition - please don't force push PR content. We don't care about a "messy" PR history, as we use merge commits when a PR is accepted; but when you force push, the context for previous PR reviews is lost.

